PR #85, opened 2016-09-02
Prototype flutter daemon mode

stevemessick:
@devoncarew @pq Here's a proof-of-concept for running flutter in daemon mode while debugging. Breakpoints work.

Now that I have a better understanding of how the pieces fit together I plan to rewrite most of this to make it more flexible. Currently, you can only debug one flutter app at a time. Eventually we will be able to have multiple devices connected and to debug multiple apps on a device.

Two files were copied from the Dart plugin then modified slightly to make connecting to the observatory possible in daemon mode. These changes could reasonably be back-ported to the Dart plugin with little risk, IMO.


src/io/flutter/run/DaemonManager.java
```diff
@@ -0,0 +1,348 @@
+package io.flutter.run;
```
devoncarew: Can you add the chromium copyright header to the top of this file?

devoncarew: (example from here: https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/FlutterBundle.java#L1)

src/io/flutter/run/DaemonManager.java
```diff
@@ -0,0 +1,348 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.lang.reflect.Type;
+import java.util.Map;
+
+/**
+ * Manage the flutter daemon.
+ * TODO Restructure process management to allow the daemon to run forever.
```
devoncarew: nit: `TODO(username):`

src/io/flutter/run/DaemonManager.java
```diff
@@ -0,0 +1,348 @@
...
+  private void sendCommand(@NotNull ProcessHandler handler, String command, Params params) {
+    OutputStream input = handler.getProcessInput();
+    if (input == null) {
+      LOG.error("No process input");
+      return;
+    }
+    int cmdId = myCommandId++;
+    Method method = new Method(command, params, cmdId);
+    try (FlutterStream str = new FlutterStream(input)) {
+      str.print("[");
```
devoncarew: You could also just issue one println() (`"[" + GSON.toJson(method) + "]"`).

src/io/flutter/run/DaemonManager.java
```diff
@@ -0,0 +1,348 @@
...
+        return;
+      }
+      String eventName = primEvent.getAsString();
+      JsonObject params = obj.getAsJsonObject("params");
+      if (eventName == null || params == null) {
+        LOG.error("Bad event from flutter: " + myCurrentJsonText);
+        return;
+      }
+      Event eventHandler = eventHandler(eventName);
+      if (eventHandler == null) return;
```
devoncarew: Does jetbrains style allow single line returns? Does it care?

src/io/flutter/run/DaemonManager.java
stevemessick: Ah yeah, forgot about that.

src/io/flutter/run/DaemonManager.java
stevemessick: I've never added a user name before. I'll try to remember that.

src/io/flutter/run/DaemonManager.java
stevemessick: I ran everything through the formatter. The single-line returns are according to style. In this sense, it is like Dart.

src/io/flutter/run/FlutterRunningState.java
```diff
@@ -31,6 +31,7 @@
 
   public FlutterRunningState(@NotNull ExecutionEnvironment environment) throws ExecutionException {
     super(environment);
+    //((TextConsoleBuilderImpl)getConsoleBuilder()).setUsePredefinedMessageFilter(true);
```
devoncarew: delete this commented out line?

src/io/flutter/run/DaemonManager.java
devoncarew: Cool, good to know :)

src/io/flutter/run/FlutterRunningState.java
stevemessick: I prefer to keep it for now. "Note to self" -- I do that a lot.

stevemessick:
I added copyrights. I'll deal with todo's in the next update.

src/io/flutter/run/DaemonManager.java
```diff
@@ -0,0 +1,353 @@
...
+
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.lang.reflect.Type;
+import java.util.Map;
+
+/**
+ * Manage the flutter daemon.
+ * TODO Restructure process management to allow the daemon to run forever.
+ * Currently it will be started when the debug session starts, and it ends
```
devoncarew: Can you open an issue to track the daemon process management refactor? Thanks!

devoncarew:
Re: the in-lined code from the Dart plugin, I think we'll want to:
- add a `third_party/intellij-plugins-dart` folder to the repo, and copy in the license file from here: https://github.com/JetBrains/intellij-plugins/blob/master/Dart/LICENSE.TXT
- add a readme.txt (or readme.md) file to that same repo, giving the url of the source repo (https://github.com/JetBrains/intellij-plugins/blob/master/Dart), and saying where the inlined files are (`src/com/jetbrains/lang/dart/`)

Also (as discussed in person) this is more motivation for finding a good solution for depending on the Dart plugin. For each dep, we'll either need to push some parameterizable service into the Dart plugin, or choose to in-line a graph of classes into the Flutter plugin (either put something into the Dart plugin that we can build on, or remove the dep entirely by in-lining).

devoncarew:
lgtm w/ the addition of a `third_party` directory. Thanks, and looking forward to trying this out!

stevemessick:
Merged. Thanks @devoncarew for keeping on top of the license issues. I did all you suggested, plus I moved the source files to third_party so we should not have to explain why we're not following established conventions. (Hope I got it right, please double-check my work.)

pq:
On update I get some new errors:

```
Error:(36, 24) java: cannot find symbol
  symbol:   class Gson
  location: class io.flutter.run.DaemonManager
...
```

I'm guessing I need to update module dependencies somewhere?  (I'm guessing the gradle build will need to be updated similarly?)

cc @stevemessick @devoncarew

pq:
I suppose I could do something like proposed here:

screen shot 2016-09-06 at 7 36 10 am

... but is that the right approach?

Do we need to update the README as well?

stevemessick:
Since I have no compilation errors it might be best to compare our projects
when you get in.

stevemessick:
Have you updated to the latest Dart plugin? (Source or distributed -- Alex
recently published a new one)

pq:
> Have you updated to the latest Dart plugin? (Source or distributed -- Alex
recently published a new one)

I haven't but the idea that I'd be depending on it to export the gson jar is worrisome.  Prone to break in all the obvious ways...

pq:
Updated Dart (and intellij community sources) and rebuilt but still no love.  Invalidating and restarting... Fingers crossed! :)

pq:
Still no `Gson`. 😢 

Let's take a look together in a bit.

pq:
OK, so after more fiddling around I took the bait and clicked "Add GSON to your classpath..."

The result was a new module dependency which doesn't seem right.

screen shot 2016-09-06 at 9 05 17 am

```    <orderEntry type="library" name="gson" level="project" />```

Anyway, I'm unblocked for now but we'll want to get this squared away for among other things the build.

UPDATE: I notice that the Dart Plugin has a similar GSON module dependency so maybe this *is* the right way to handle it?

PR #88, opened 2016-09-03
fold 'flutter run' in the flutter console

devoncarew:
- this folds the `flutter run` line in the console when running flutter apps

this:
screen shot 2016-09-02 at 5 22 16 pm

becomes this:
screen shot 2016-09-02 at 5 21 55 pm

and on hover:
screen shot 2016-09-02 at 5 22 10 pm

@pq @stevemessick 

@stevemessick, we can adjust this as the daemon run changes change the console output -

### DIFF
--- a/src/io/flutter/console/FlutterConsoleFolding.java
+++ b/src/io/flutter/console/FlutterConsoleFolding.java
@@ -21,11 +21,9 @@ public class FlutterConsoleFolding extends ConsoleFolding {
     if (!line.contains("flutter run")) return false;
 
     try {
-      final String homePath = FlutterSdk.getGlobalFlutterSdk().getHomePath();
-      if (homePath == null) {
-        return false;
-      }
-      final String flutterPath = FlutterSdkUtil.pathToFlutterTool(homePath);
+      FlutterSdk sdk = FlutterSdk.getGlobalFlutterSdk();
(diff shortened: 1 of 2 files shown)


src/io/flutter/console/FlutterConsoleFolding.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.List;
+
+public class FlutterConsoleFolding extends ConsoleFolding {
+  @Override
+  public boolean shouldFoldLine(String line) {
+    if (!line.contains("flutter run")) return false;
+
+    try {
+      final String homePath = FlutterSdk.getGlobalFlutterSdk().getHomePath();
```
pq: You might consider guarding against a null SDK here.  (Pretty sure an inspection would nag too since I bet `getGlobalFlutterSdk()` returns `@Nullable`.

pq:
LGTM! 👍

src/io/flutter/console/FlutterConsoleFolding.java
devoncarew: Thanks! Updated. I did fix an inspection for Nullable (and then later a related one for a no-op comparison, because of NonNullable).

stevemessick:
LGTM

Issue #94, opened 2016-09-06
Add `flutter doctor` hyperlink action to console.


pq:
`ConsoleFilter`s are the trick here.

pq:
In progress:

![screen shot 2016-09-06 at 9 47 32 am](https://cloud.githubusercontent.com/assets/67586/18282667/fcd0619a-7416-11e6-9e13-ca888283eba0.png)

where a click opens a fresh console w/ `doctor` results:

![screen shot 2016-09-06 at 9 47 40 am](https://cloud.githubusercontent.com/assets/67586/18282670/017a81da-7417-11e6-824f-3d2091ab1e6c.png)

PR #95, opened 2016-09-06
Hyperlink support for `flutter doctor` in flutter console (#94).

pq:
Fixes: https://github.com/flutter/flutter-intellij/issues/94

/cc @stevemessick @devoncarew

### DIFF
--- a/src/io/flutter/console/FlutterConsoleFilter.java
+++ b/src/io/flutter/console/FlutterConsoleFilter.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2000-2016 JetBrains s.r.o.
+ *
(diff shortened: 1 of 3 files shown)


stevemessick:
LGTM

pq:
Thanks!

src/io/flutter/console/FlutterConsoleFilter.java
```diff
@@ -0,0 +1,68 @@
+/*
```
devoncarew: This is a new file; we should use the chromium copyright.

pq: Whoops!

Hmmm, looks like my settings got borked...

PR #100, opened 2016-09-06
refactor location of the DartTestUtils class

devoncarew:
@stevemessick, refactor the location of the DartTestUtils class.


stevemessick:
LGTM

PR #104, opened 2016-09-06
fix the built plugin jar

devoncarew:
Fix an issue w/ the built plugin (`flutter-intellij-0.0.1.jar`) - it wasn't including its resources.

@pq @stevemessick


pq:
Awesome!

LGTM 👍

PR #111, opened 2016-09-07
Rename copy of debug service and make it a subclass of the real one

stevemessick:
@devoncarew See if this fixes your latest class loader problem with the built plugin.

### DIFF
--- a/third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/DartRunner.java
+++ b/third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/DartRunner.java
@@ -26,7 +26,7 @@ import com.jetbrains.lang.dart.ide.runner.base.DartRunConfigurationBase;
 import com.jetbrains.lang.dart.ide.runner.server.DartCommandLineRunConfiguration;
 import com.jetbrains.lang.dart.ide.runner.server.DartCommandLineRunningState;
 import com.jetbrains.lang.dart.ide.runner.server.DartRemoteDebugConfiguration;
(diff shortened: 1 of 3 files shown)


devoncarew:
(applying locally now)

stevemessick:
Thanks, cuz if it doesn't work I'd just have to revert this

devoncarew:
OK, this fixes the IllegalAccessExceptions. I do see:

```
Failed to connect to the VM observatory service: java.io.IOException: Failed to connect: ws://null:0/ws
Caused by: de.roderick.weberknecht.WebSocketException: unknown host: null
Caused by: java.net.UnknownHostException: null
/Users/devoncarew/projects/flutter/flutter/bin/flutter daemon

Process finished with exit code 130 (interrupted by signal 2: SIGINT)
Observatory connection never became ready
```

now, but I suspect it's a different issue.

lgtm; I'll file an issue for us to noodle ideas to avoid problems with the dual classloaders going forward

stevemessick:
I get that same "Failed to connect" message the first time I try to start
the debugger after firing up IntelliJ. Then it seems to be fine.

devoncarew:
> I get that same "Failed to connect" message the first time I try to start the debugger after firing up IntelliJ. Then it seems to be fine.

I did confirm, I was able to run and debug after restarting.

stevemessick:
Wait, what? You can do the demo from built plugins and all you say is "it
works"? Shouldn't we be celebrating more? 😂

stevemessick:
I see that emoticons may not convert the proper level of levity. I am
pretty happy right now!

devoncarew:
🎉 🎈 🍰 📈 🎆

PR #115, opened 2016-09-08
Pick up async stepping bug fix

stevemessick:
@devoncarew This brings in your fix for async stepping from last week.

I think this represents a worst-case scenario for our code copying model. A bug fix was applied to Dart plugin code that we had to copy into the Flutter plugin. I didn't notice that the bug fix modified copied code until today.

There really should be a null check on suspendContext. I didn't add it to keep the differences between the original and the copy to a minimum.


stevemessick:
@pq The header comment is probably of interest.

pq:
> @pq The header comment is probably of interest.

Missing some context (or coffee) --- what am I looking at?

_(Grabbing ☕ )_

devoncarew:
@stevemessick - there's a build failure here: https://travis-ci.org/flutter/flutter-intellij/builds/158465520#L159

It looks like this depends on API which is not yet in the stable Dart plugin. We'd need to cherry-pick in the async stepping fix, or in-line (much!) more of the debug code.

stevemessick:
Thankfully, we have CI in place. Is the async stepping change going to be
cherry-picked or will it wait until the next release?

devoncarew:
We should file a cherry-pick request (likely along with your other debugger API changes).

pq:
@stevemessick can we close this PR?

PR #122, opened 2016-09-16
don't run create with ansi drawing

devoncarew:
- pass `--no-color` into `flutter create`; this causes the pub get step to not try and draw using ansi codes


devoncarew:
@pq

PR #124, opened 2016-09-19
Tweak install notes.

pq:
/cc @devoncarew @stevemessick


stevemessick:
The instructions say either community or stable will work. However, if (like me) you use stable you need to be careful to name your sdK "Community Edition". If you don't you will end up with a modified .idea/ file.

PR #129, opened 2016-09-21
Add source path details to the CONTRIBUTING docs.

pq:
Instructions for getting platform sourcepaths wired up.

/cc @stevemessick @devoncarew

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ file.
 * In the project structure dialog, configure an IntelliJ platform SDK
   - point it to the `Contents` directory in your just downloaded copy of IntelliJ Community Edition (e.g, `IntelliJ IDEA CE.app/Contents`)
   - name it `IntelliJ IDEA Community Edition`
-* Optionally, If you want platform sources handy, clone the IntelliJ IDEA Community Edition repo (`git clone https://github.com/JetBrains/intellij-community`) and add a path to your local clone in the `Sourcepaths` tab of your `IntelliJ IDEA Community Edition` SDK and accept all the root folders found by the IDE after scanning.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -17,6 +17,7 @@ file.
 * In the project structure dialog, configure an IntelliJ platform SDK
   - point it to the `Contents` directory in your just downloaded copy of IntelliJ Community Edition (e.g, `IntelliJ IDEA CE.app/Contents`)
   - name it `IntelliJ IDEA Community Edition`
+* Optionally, If you want platform sources handy, clone the IntelliJ IDEA Community Edition repo (`git clone https://github.com/JetBrains/intellij-community`) and add a path to your local clone in the `Sourcepaths` tab of your `IntelliJ IDEA Community Edition` SDK and accept all the root folders found by the IDE after scanning.
```
stevemessick: Awesome! Thanks for figuring that out. Instead of making this optional I'd prefer to see a comment that you'd get decompiled code if you omit this step. Also, you have to do this for the Dart sources, too?

pq: I'll remove the optional bit.  Not doing this is really just not a good idea.

As for the Dart sources, we could do something similar with the plugins repo but I haven't tried.

pq: In that case it might make sense to also add instructions to checkout the stable branch.  (Technically we should do the same for the CE repo too.)

devoncarew: re: the Dart sources, we'll likely be in-lining parts of our major deps.

pq: In that case, I'll wait on those instructions.

In the meantime, a rubber stamp from someone? 😉

devoncarew:
lgtm 👍

PR #141, opened 2016-09-24
commit an older version of the Dart plugin

devoncarew:
- commit an older version of the Dart plugin

We now have two copies - the current stable version (`162.2017`), which we compile against, and the previous stable (`145.971.7`) - the last one compatible with the current version of Android Studio (`2.2`).

@stevemessick @pq


stevemessick:
LGTM

flutter-intellij.iml
```diff
@@ -13,15 +13,10 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
-    <orderEntry type="module" module-name="Dart" />
-    <orderEntry type="module" module-name="platform-api" />
```
stevemessick: I suspect removing these will cause problems for me, but I don't know that.

devoncarew: I can add them back - was was cleaning up some errors IntelliJ showed me, but they weren't affecting anything.

PR #143, opened 2016-09-24
fix a flutter sdk toast message

devoncarew:
Fix an issue where a notification would compare the flutter sdk path to the dart sdk path, instead of  flutter's dart sdk path to the dart sdk path (the warning would always show, even when using the dart sdk from the flutter sdk).

@pq @stevemessick


pq:
LGTM.

Hilarious.  I actually looked at this yesterday.  Not sure how it got this way, but it's awesome to see it fixed! 👍

PR #146, opened 2016-09-25
tweaks to the i18n file

devoncarew:
Minor tweaks to the i18n file:

- sort the entries
- change a raw string to a FlutterBundle.message

@pq @stevemessick


pq:
LGTM.

FWIW this may get un-ordered again if folks use the IDE's i18n injection support (e.g., have it extract strings to properties).  Anyway, we can cross that bridge if we need to!

devoncarew:
Cool - I didn't know about the auto-ide editing. Let's land; we can not worry about the sorting if editing via the IDE is common.

PR #149, opened 2016-09-26
Daemon service

stevemessick:
@devoncarew @pq 
Change debugger to use flutter daemon as a long-lived service.
I'd be happy to go over this in person if that would be faster.


src/io/flutter/run/FlutterAppState.java
```diff
@@ -0,0 +1,78 @@
...
+    final List<AnAction> actions = new ArrayList<>(Arrays.asList(super.createActions(console, processHandler, executor)));
+    addObservatoryActions(actions, processHandler);
+    return actions.toArray(new AnAction[actions.size()]);
+  }
+
+  protected void addObservatoryActions(List<AnAction> actions, final ProcessHandler processHandler) {
+    actions.add(new Separator());
+    //actions.add(new OpenDartObservatoryUrlAction(
+    //  "http://" + NetUtils.getLocalHostString() + ":" + myObservatoryPort,
+    //  () -> !processHandler.isProcessTerminated()));
```
pq: Is this commented-out bit related to the observatory action you were looking for but not seeing?

stevemessick: Probably. That's where the refresh button goes, too.

src/io/flutter/run/daemon/ConnectedDevice.java
```diff
@@ -0,0 +1,54 @@
...
+  @Override
+  public String deviceId() {
+    return myDeviceId;
+  }
+
+  @Override
+  public String platform() {
+    return myPlatform;
+  }
+}
```
pq: Nit: missing newline.

src/io/flutter/run/daemon/DaemonListener.java
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright 2016 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.run.daemon;
+
+public interface DaemonListener {
+
+  /**
+   * Process JSON input from the Flutter daemon running on the <code>controller.</code>
```
pq: Super-nit: `.` outside of code block?

src/io/flutter/run/daemon/DaemonListener.java
```diff
@@ -0,0 +1,24 @@
...
+   *
+   * @param json       The JSON string read from the Flutter daemon
+   * @param controller The FlutterDaemonController that controls the daemon
+   */
+  void daemonInput(String json, FlutterDaemonController controller);
+
+  /**
+   * Instruct the listener that it should enable device polling for the <code>controller.</code>
```
pq: And here...

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -0,0 +1,233 @@
...
+    return myController;
+  }
+
+  @Override
+  public boolean hasAppId() {
+    return myApp != null && myApp.appId != null;
+  }
+
+  @Override
+  public String appId() {
```
pq: Do we want a null check here?

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -0,0 +1,233 @@
...
+    return myApp.appId;
+  }
+
+  @Override
+  public String projectDirectory() {
+    return myApp.directory;
```
pq: ... And null checks for the following accessors?

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -0,0 +1,233 @@
...
+  @Override
+  public Object fetchWidgetHierarchy() {
+    throw new NotImplementedException();
+  }
+
+  @Override
+  public Object fetchRenderTree() {
+    throw new NotImplementedException();
+  }
+}
```
pq: Nit: missing newline.

src/io/flutter/run/daemon/FlutterDaemonController.java
```diff
@@ -0,0 +1,218 @@
...
+      addDeviceId(deviceId);
+    }
+  }
+
+  private boolean hasDeviceId(String id) {
+    return myDeviceIds.contains(id);
+  }
+
+  private void addDeviceId(String deviceId) {
+    myDeviceIds.add(deviceId);
```
pq: Do we need to worry about possible duplicates?

src/io/flutter/run/daemon/FlutterDaemonService.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  private FlutterDaemonService() {
+    Disposer.register(ApplicationManager.getApplication(), this::stopControllers);
+    schedulePolling();
+  }
+
+  /**
+   * Return the list of currently connected devices. The list is sorted by device name.
+   * TODO(pq,messick) Extend the debugger UI to allow selecting a device from this list.
```
pq: 👍

src/io/flutter/run/FlutterAppState.java
stevemessick: Actually, no this is something else.

src/io/flutter/run/FlutterAppState.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public boolean isConnectionReady() {
+    return myApp != null && myApp.port() > 0;
+  }
+
+  public int getObservatoryPort() {
+    return myApp.port();
```
stevemessick: What would be an appropriate alternative? If this method is called before isConnectionReady() returns true then I don't know what should happen.
Maybe adding a comment to say it can't be used too soon?

src/io/flutter/run/daemon/DaemonListener.java
stevemessick: This is intentional. I know looks unusual but it is consistent with the practice of using bold/italic/etc for trailing punctuation.

src/io/flutter/run/daemon/FlutterApp.java
stevemessick: I'm undecided about null checks in this class. If higher-level code (like UI) is trying to use this object before the async initialization has completed then you've got bigger problems. This should be fully initialized before it gets to UI-level code so I think NPE is an appropriate action here. I could be persuaded otherwise, but some sort of exception will be thrown and if it is checked it will complicate lots of things.

src/io/flutter/run/daemon/FlutterDaemonController.java
stevemessick: Yes, this is WIP (and currently broken).

PR #150, opened 2016-09-26
fix an npe when no devices are connected

devoncarew:
- fix an npe when no devices are connected

@pq @stevemessick

### DIFF
--- a/src/io/flutter/run/FlutterAppState.java
+++ b/src/io/flutter/run/FlutterAppState.java
@@ -23,6 +23,7 @@ import org.jetbrains.annotations.NotNull;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/FlutterAppState.java
```diff
@@ -48,7 +49,13 @@ protected ProcessHandler startProcess() throws ExecutionException {
     Project project = getEnvironment().getProject();
     String workingDir = project.getBasePath();
     assert workingDir != null;
-    ConnectedDevice device = service.getConnectedDevices().iterator().next();
+
+    Collection<ConnectedDevice> devices = service.getConnectedDevices();
+    if (devices.isEmpty()) {
+      throw new ExecutionException("No connected device");
```
stevemessick: I didn't think this condition could become true.

devoncarew: I don't think it will be going forward, but right now the run button can be enabled w/o any devices connected.

PR #155, opened 2016-09-26
Wire up restart commands

stevemessick:
@pq @devoncarew This adds Restart and Reload commands, as well as restoring the observatory connection. At least one, maybe both, need new icons. Reload does not seem to be working but restart does what's expected. It may be a problem in flutter. I notice that when I start the daemon in hot mode there are a lot more log messages printed than before, so I think hot mode is selected correctly.

### DIFF
--- a/src/io/flutter/actions/OpenComputedUrlAction.java
+++ b/src/io/flutter/actions/OpenComputedUrlAction.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.actions;
+
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.project.DumbAwareAction;
+import com.intellij.openapi.util.Computable;
+import com.jetbrains.lang.dart.DartBundle;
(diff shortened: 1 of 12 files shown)


src/io/flutter/actions/OpenComputedUrlAction.java
```diff
@@ -0,0 +1,35 @@
...
+import com.jetbrains.lang.dart.ide.runner.server.OpenDartObservatoryUrlAction;
+import icons.DartIcons;
+import org.jetbrains.annotations.NotNull;
+
+public class OpenComputedUrlAction extends DumbAwareAction {
+  private Computable<String> myUrl;
+  private Computable<Boolean> myIsApplicable;
+
+  public OpenComputedUrlAction(@NotNull final Computable<String> url, @NotNull final Computable<Boolean> isApplicable) {
+    super(DartBundle.message("open.observatory.action.text"), DartBundle.message("open.observatory.action.description"), DartIcons.Dart_16);
```
devoncarew: Note: I've tried to eliminate dependencies in `DartBundle` directly, and instead moved any necessary localized strings into the FlutterBundle resource, in order to reduce our API usage from the dart plugin.

devoncarew:
lgtm!

src/io/flutter/actions/OpenComputedUrlAction.java
pq: 👍 

I was actually thinking of doing a pass at extracting them all.  Happy to do that once this lands.

Issue #157, opened 2016-09-27
exception at startup; symptom is we lose the flutter sdk setting


devoncarew:
Perhaps a lifecycle issue with when we call `ApplicationLibraryTable.getApplicationTable()`? I'm not as familiar when it is or isn't available to call this method (called from io.flutter.sdk.FlutterSdk.getGlobalFlutterSdk(FlutterSdk.java:97).

/cc @pq @stevemessick 

```
[   4900]  ERROR - rationStore.ComponentStoreImpl - Cannot init libraryTable component state 
java.lang.IllegalArgumentException: Unknown macro $APP_CONFIG$
	at com.intellij.configurationStore.StateStorageManagerImpl.expandMacro(StateStorageManagerImpl.kt:353)
	at com.intellij.configurationStore.ApplicationStorageManager.expandMacros(ApplicationStoreImpl.kt:113)
	at com.intellij.configurationStore.StateStorageManagerImpl.createStateStorage(StateStorageManagerImpl.kt:212)
	at com.intellij.configurationStore.StateStorageManagerImpl.getOrCreateStorage(StateStorageManagerImpl.kt:165)
	at com.intellij.configurationStore.StateStorageManagerImpl.getStateStorage(StateStorageManagerImpl.kt:136)
	at com.intellij.configurationStore.ComponentStoreImpl.initPersistentComponent(ComponentStoreImpl.kt:275)
	at com.intellij.configurationStore.ComponentStoreImpl.initComponent(ComponentStoreImpl.kt:91)
	at com.intellij.openapi.components.impl.PlatformComponentManagerImpl.initializeComponent(PlatformComponentManagerImpl.java:54)
	at com.intellij.openapi.components.impl.ServiceManagerImpl$MyComponentAdapter.getComponentInstance(ServiceManagerImpl.java:237)
	at com.intellij.util.pico.DefaultPicoContainer.getLocalInstance(DefaultPicoContainer.java:246)
	at com.intellij.util.pico.DefaultPicoContainer.getComponentInstance(DefaultPicoContainer.java:212)
	at com.intellij.openapi.components.ServiceManager.doGetService(ServiceManager.java:46)
	at com.intellij.openapi.components.ServiceManager.getService(ServiceManager.java:37)
	at com.intellij.openapi.roots.impl.libraries.ApplicationLibraryTable.getApplicationTable(ApplicationLibraryTable.java:46)
[...]

PR #159, opened 2016-09-27
Fix Daemon initialization issue (#157).

pq:
Migrates initialization to a startup activity.  The `ApplicationTable` should be reliably available then.

Fixes: #157 

/cc @stevemessick @devoncarew


pq:
PS: sorry about the odd commit history...  I'll reset after this gets squashed.

PR #163, opened 2016-09-27
Eliminate device list duplication

stevemessick:
@pq @devoncarew Define equals() and hashCode() for FlutterDevice so the list stops growing. I had defined it as a set but didn't get around to defining these two until now.

### DIFF
--- a/src/io/flutter/run/daemon/ConnectedDevice.java
+++ b/src/io/flutter/run/daemon/ConnectedDevice.java
@@ -5,6 +5,8 @@
  */
 package io.flutter.run.daemon;
 
+import com.google.common.base.Objects;
+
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/daemon/ConnectedDevice.java
```diff
@@ -51,4 +53,20 @@ public String deviceId() {
...
+    } else {
+      return false;
+    }
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(myDeviceName, myDeviceId, myPlatform);
+  }
 }
```
pq: Nit: missing newline.

PR #170, opened 2016-09-28
add emulator info to devices

devoncarew:
This will let us know things like if the ios simulator is already running.

https://github.com/flutter/flutter/pull/6114 will need to land before this.

@stevemessick @pq


devoncarew:
Will land this after we push out a new build -

PR #172, opened 2016-09-28
Rev dart plugin

devoncarew:
- rev to 162.2233 of the dart plugin
- fix debugging for the latest Dart plugin

@stevemessick, this upgrades us to compiling against the latest published version of the Dart plugin, and fixes debugging against that latest version. We'll need to update this later - as part of debugging apps running under hot reload, we need to guess the uri prefix for the app. The Dart plugin currently only does this when remote debugging, so we spoof that here.

### DIFF
--- a/third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
+++ b/third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
@@ -284,6 +284,9 @@ public class DartVmServiceDebugProcessZ extends DartVmServiceDebugProcess {
   public boolean isRemoteDebug() {
     // TODO(devoncarew): This is a hack for now - we want to instead look for prefix matches
     // if remote debugging, or running a Flutter app.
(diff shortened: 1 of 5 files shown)


third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
```diff
@@ -282,7 +282,9 @@ public XDebuggerEditorsProvider getEditorsProvider() {
   }
 
   public boolean isRemoteDebug() {
-    return myRemoteDebug;
+    // TODO(devoncarew): This is a hack for now - we want to instead look for prefix matches
+    // if remote debugging, or running a Flutter app.
```
stevemessick: Could you explain this a bit more? If we will need to know about remote debugging then references to myRemoteDebug may be useful to preserve. As you know, this class will need to be reimplemented, and keeping internal notes like that around might make it easier.

devoncarew: Absolutely! Will add some docs. I think we'll iterate on this implementation a few times - would like to not have to hardcode this setting here.

PR #183, opened 2016-09-28
Filter out json formatting

stevemessick:
@pq @devoncarew Makes the console output more readable. I also added a note about configuring sources for Dart.

### DIFF
--- a/src/io/flutter/run/daemon/DaemonJsonInputFilterProvider.java
+++ b/src/io/flutter/run/daemon/DaemonJsonInputFilterProvider.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.run.daemon;
+
+import com.intellij.execution.filters.ConsoleInputFilterProvider;
+import com.intellij.execution.filters.InputFilter;
+import com.intellij.execution.ui.ConsoleViewContentType;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.Pair;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collections;
+import java.util.List;
+
+public class DaemonJsonInputFilterProvider implements ConsoleInputFilterProvider {
+
+  @NotNull
+  @Override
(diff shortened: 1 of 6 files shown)


src/io/flutter/run/daemon/DaemonJsonInputFilterProvider.java
```diff
@@ -0,0 +1,37 @@
...
+  @Override
+  public InputFilter[] getDefaultFilters(@NotNull Project project) {
+    return new InputFilter[]{new DaemonJsonInputFilter()};
+  }
+
+  private static class DaemonJsonInputFilter implements InputFilter {
+    @Nullable
+    @Override
+    public List<Pair<String, ConsoleViewContentType>> applyFilter(String text, ConsoleViewContentType contentType) {
+      if (text.startsWith("[{") && text.endsWith("}]\n")) {
```
devoncarew: @stevemessick, instead of filtering this from the console, what if we don't write the daemon / json output in the first place? The things we'd want to show there are likely just the log messages - direct output from the tool and the user's app (the `message.log` below).

stevemessick:
I couldn't find a way to do that. Do you know how?

devoncarew:
I don't, but we can land this and iterate towards a solution that doesn't use filtering - that only populates the console w/ user output. Sound good?

src/io/flutter/run/daemon/FlutterAppManager.java
```diff
@@ -275,11 +277,14 @@ private Event eventHandler(@NotNull String eventName, @Nullable String json) {
...
   private void eventLogMessage(@NotNull LogMessage message, @NotNull FlutterDaemonController controller) {
-    myLogger.info(message.message); // TODO(messick) Complete logging.
+    myLogger.info(message.message);
   }
 
   private void eventLogMessage(@NotNull AppLog message, @NotNull FlutterDaemonController controller) {
-    myLogger.info(message.log); // TODO(messick) Complete logging.
+    RunningFlutterApp app = findApp(controller, message.appId);
+    if (app != null && message.log != null) {
+      app.getConsole().print(message.log + "\n", ConsoleViewContentType.NORMAL_OUTPUT);
```
devoncarew: 👍

devoncarew:
lgtm!

stevemessick:
Yes, I like getting things working then making them better. I suspect we're already doing it the "IntelliJ way", though.

devoncarew:
(running this locally - looks great!)

stevemessick:
I agree. It is much better now.

PR #186, opened 2016-09-28
show progress from the daemon

devoncarew:
- when we get progress events from the daemon protocol, translate that into IntelliJ background tasks (the equivalent of Eclipse jobs); this displays the progress on the status line
- expose the baseUri field; we don't use this info yet

@pq @stevemessick: I'm not as familiar with how to access IntelliJ global state - like the current project - and when it's safe to do so...

### DIFF
--- a/src/io/flutter/run/daemon/FlutterAppManager.java
+++ b/src/io/flutter/run/daemon/FlutterAppManager.java
@@ -10,7 +10,11 @@ import com.intellij.execution.ExecutionException;
 import com.intellij.execution.ui.ConsoleViewContentType;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.progress.ProgressIndicator;
+import com.intellij.openapi.progress.ProgressManager;
+import com.intellij.openapi.progress.Task;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.project.ProjectManager;
 import com.intellij.util.TimeoutUtil;
 import gnu.trove.THashMap;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/daemon/FlutterAppManager.java
```diff
@@ -282,8 +291,52 @@ private void eventLogMessage(@NotNull LogMessage message, @NotNull FlutterDaemon
...
     RunningFlutterApp app = findApp(controller, message.appId);
-    if (app != null && message.log != null) {
-      app.getConsole().print(message.log + "\n", ConsoleViewContentType.NORMAL_OUTPUT);
+    if (app != null) {
+      if (message.progress) {
+        if (message.finished) {
+          clearLatch();
+        } else {
+          // TODO: Is this the correct API to use to get the current project?
+          Project project = ProjectManager.getInstance().getDefaultProject();
```
stevemessick: Nope. I don't know what the default project is but I recall it has null for its path.

stevemessick: I'd prefer not to have UI code this deep in the process management framework. Maybe FlutterAppState could host the progress updater. It probably has (or can easily have added) a reference to the Project. I admit to having been in a bit of a hurry when doing the console printing, and should have done it that way.

pq: Hmmmm...  As per our conversation I don't think this is quite right.  Better probably to have the project passed down through the `FlutterDaemonController`.  Alternatively it could pass in the `projectDir` and you could look for a match in `ProjectManager.getInstance().getOpenProjects()`.

devoncarew:
Happy to refactor before landing. I'll look at bumping the code up to somewhere closing to the UI, and where we can easily get the current project we're running in.

devoncarew:
will close and refactor in a new PR -

Issue #185, opened 2016-09-28
Add keybindings for app reload and restart.


pq:

pq:
cc @mit-mit

mit-mit:
`cmd-r` for hot reload
`cmd-shift-r` for restart

pq:
@stevemessick : I've got the incantation but there's a bit of rearchitecting to do.  Net-net, key-bound actions need to have a default constructor and we're using the constructor to inject the `ObservatoryConnector`.  To wit:

```java
  public ReloadFlutterApp(ObservatoryConnector connector) {
    super(connector, FlutterBundle.message("app.reload.action.text"), FlutterBundle.message("app.reload.action.description"),
          FlutterIcons.Flutter);
  }
```

Digging around but any thoughts on a clean way to do that appreciated!

devoncarew:
I wonder how they do some of the debugger actions, like for step-over/in/out?

pq:
OK, I've got a hack to find the connector and I think all's good.

The only issue now is around conflicting keybindings.

@mit-mit (or anyone) maybe take a look at the defined keybindings and come up with something less conflicting?

This [key](https://resources.jetbrains.com/assets/products/intellij-idea/IntelliJIDEA_ReferenceCard.pdf) may come in handy, though it differs a bit from mine which I based on eclipse. 😊

stevemessick:
Well, just to keep things moving, how about this:

> ctrl-f5 or (Mac) cmd-f5 for hot reload
> ctrl-shift-f5 or (Mac) cmd-shift-f5 for restart

IntelliJ likes to put process control stuff on function keys. Cmd-f2 stops the running job in the debugger. F5 is only defined for one other task so there shouldn't be any conflicts with standard keys. No way to predict what other plugins will do, though.

PR #188, opened 2016-09-28
Hot reload/restart keybindings (#185).

pq:
* keybindings for reload and restart (provisional)
* (temporary) connection-finding logic

Non-conflicting key-bindings are hard so I just punted.  
screen shot 2016-09-28 at 4 18 12 pm

They can easily be updated once we settle on better ones.  In the meantime, the actions can be invoked via the **Find Action** command (**⌘ + A**) and the keymappings updated locally for demo (**Preferences > Keymap**).

Console out from invocation:

screen shot 2016-09-28 at 4 29 14 pm

Fixes: https://github.com/flutter/flutter-intellij/issues/185

/cc @stevemessick @devoncarew

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -31,6 +31,14 @@
       <action id="Flutter.DeviceSelector" class="io.flutter.actions.DeviceSelectorAction"
               description="Flutter Device Selection"
               icon="FlutterIcons.Phone"/>
+      <action id="Flutter.ReloadFlutterAppKey" class="io.flutter.actions.ReloadFlutterAppKeyAction" text="Reload Flutter app">
+        <!-- TODO: settle on a less conflicting binding -->
+        <keyboard-shortcut first-keystroke="ctrl shift r" keymap="$default"/>
+      </action>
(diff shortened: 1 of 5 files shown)


pq:
Cool.  Thanks.  Anyone should feel free to fiddle with the bindings as they see fit.

resources/META-INF/plugin.xml
```diff
@@ -31,6 +31,14 @@
       <action id="Flutter.DeviceSelector" class="io.flutter.actions.DeviceSelectorAction"
               description="Flutter Device Selection"
               icon="FlutterIcons.Phone"/>
+      <action id="Flutter.ReloadFlutterAppKey" class="io.flutter.actions.ReloadFlutterAppKeyAction" text="Reload Flutter app">
```
devoncarew: I think the text for these actions should be the same as on the debugger toolbar. The names are similar, but do different things - I'd like to make it as easy as possible to distinguish them.

Right now we're using `Full Application Restart` and `Hot Reload`. Perhaps for these keybindings, we use something like `Flutter Restart Application` and `Flutter Hot Reload`?

And in general we should make hot reload easier to get to - slightly more promoted - than the full app restart.

pq: 👍 

And thanks for the follow-up PR!

PR #191, opened 2016-09-29
catch all daemon messages in the filter

devoncarew:
@stevemessick - catch all daemon messages in the filter; a few were slipping by before.


devoncarew:
closing - git issues

PR #193, opened 2016-09-29
rename action classes and labels

devoncarew:
- rename some action classes
- renames labels of actions and key bindings for uniformity
- show an error if the keybinding is invoked and there's no running flutter app

@pq


src/io/flutter/actions/RestartFlutterAppKeyAction.java
```diff
@@ -5,11 +5,15 @@
...
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationType;
+import com.intellij.notification.Notifications;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.jetbrains.lang.dart.ide.runner.ObservatoryConnector;
+import io.flutter.FlutterBundle;
 
 /**
- * A keystroke invoked {@link ReloadFlutterApp} action.
+ * A keystroke invoked {@link HotReloadFlutterApp} action.
```
pq: I think this should read `RestartFlutterApp`?  (My bad, sorry. 😛 )

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -84,12 +84,12 @@
   /**
    * Restart the app.
    */
-  void appRestart();
+  void performRestartApp();
 
   /**
    * Refresh the app.
```
pq: Maybe update to say "reload"?

src/io/flutter/actions/RestartFlutterAppKeyAction.java
devoncarew: done!

src/io/flutter/run/daemon/FlutterApp.java
devoncarew: done

stevemessick:
Renaming that class is causing me a lot of trouble.

PR #195, opened 2016-09-29
Finish enablng the new app buttons

stevemessick:
@pq @devoncarew The restart and reload buttons were not getting enabled correctly because I hadn't finished writing that code. Now it is complete. Please ignore the `// async` comments -- those are just notes to myself where synchronous access needs to be controlled.


devoncarew:
lgtm! I renamed some classes in a PR that just landed - this will need to be updated to in order to land; hopefully not too painful.

stevemessick:
Wasted too much time trying to salvage, so killing it instead.

PR #199, opened 2016-09-29
Stop apps cleanly

stevemessick:
@pq @devoncarew Cleaning up some concurrency issues and wiring up app.stop correctly. Now, in most cases the debugger is updated whether the app stops due to the Stop button being clicked or the app exiting on the device.

Actually disconnecting the device while the debugger is running causes a whole host of other problems that are not addressed here.

### DIFF
--- a/src/io/flutter/run/daemon/FlutterAppManager.java
+++ b/src/io/flutter/run/daemon/FlutterAppManager.java
@@ -59,7 +59,11 @@ public class FlutterAppManager {
     if (!waitForDevice(deviceId)) {
       return null;
     }
-    RunningFlutterApp app = new RunningFlutterApp(myService, controller, this, mode, isHot, target, route);
+    FlutterDaemonService service;
+    synchronized (myLock) {
+      service = myService;
+    }
+    RunningFlutterApp app = new RunningFlutterApp(service, controller, this, mode, isHot, target, route);
     AppStart appStart = new AppStart(deviceId, controller.getProjectDirectory(), isPaused, route, mode.mode(), target, isHot);
     Method cmd = makeMethod(CMD_APP_START, appStart);
     Runnable x = () -> {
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/daemon/FlutterAppManager.java
```diff
@@ -300,13 +312,19 @@ private void eventLogMessage(@NotNull AppLog message, @NotNull FlutterDaemonCont
...
   }
 
   private void eventDeviceRemoved(@NotNull DeviceRemoved removed, @NotNull FlutterDaemonController controller) {
-    for (ConnectedDevice device : myService.getConnectedDevices()) {
-      if (device.deviceName().equals(removed.name) && device.deviceId().equals(removed.id) && device.platform().equals(removed.platform)) {
-        myService.removeConnectedDevice(device);
+    synchronized (myLock) {
+      for (ConnectedDevice device : myService.getConnectedDevices()) {
+        if (device.deviceName().equals(removed.name) &&
+            device.deviceId().equals(removed.id) &&
```
devoncarew: devices should be unique by id - you won't have any on the same machine with the same ID

src/io/flutter/run/daemon/FlutterAppManager.java
```diff
@@ -346,9 +364,11 @@ private void error(JsonObject json) {
   }
 
   private RunningFlutterApp findApp(FlutterDaemonController controller, String appId) {
-    for (FlutterApp app : myApps) {
-      if (app.getController() == controller && app.hasAppId() && app.appId().equals(appId)) {
-        return (RunningFlutterApp)app;
+    synchronized (myLock) {
+      for (FlutterApp app : myApps) {
+        if (app.getController() == controller && app.hasAppId() && app.appId().equals(appId)) {
```
devoncarew: The very very lastest flutter codebase returns unique app ids as well, even across daemon sessions. Not sure if you want to update the code to reflect that yet though, since it depends on everybody being in the most up to date flutter.

devoncarew:
👍 lgtm

PR #202, opened 2016-09-29
Better enablement for toolbar buttons

stevemessick:
@pq @devoncarew Here's a rewrite of today's earlier branch that I gave up trying to merge.

I didn't preserve everything, but the user-facing bits are all there.


devoncarew:
lgtm

PR #210, opened 2016-09-29
faster stops

devoncarew:
Speed up the stop implementation; we don't wait for a response from the daemon.

@stevemessick, I wasn't able to find a way to notify that the debug session was completed, when we receive an `app.stop` event. The session already completes when the app stops - the service protocol connection closes, and the debug session ends because of that. Still, it would be nice to also be able to close it on receipt of `app.stop`.

### DIFF
--- a/src/io/flutter/run/daemon/FlutterAppManager.java
+++ b/src/io/flutter/run/daemon/FlutterAppManager.java
@@ -200,7 +200,8 @@ public class FlutterAppManager {
     AppStop appStop = new AppStop(app.appId());
     Method cmd = makeMethod(CMD_APP_STOP, appStop);
     sendCommand(app.getController(), cmd);
-    FlutterJsonObject obj = waitForResponse(cmd);
+    // We don't wait for a response here; we'll get an app.stop event later as a notification.
+    //waitForResponse(cmd);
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/daemon/FlutterAppManager.java
```diff
@@ -200,7 +200,8 @@ void stopApp(@NotNull FlutterApp app) {
     AppStop appStop = new AppStop(app.appId());
     Method cmd = makeMethod(CMD_APP_STOP, appStop);
     sendCommand(app.getController(), cmd);
-    FlutterJsonObject obj = waitForResponse(cmd);
+    // We don't wait for a response here; we'll get an app.stop event later as a notification.
+    //waitForResponse(cmd);
```
stevemessick: I tried that. It is faster but the app does not terminate.

devoncarew: In my (one) test it did - we do sent the `stop` request, we just don't wait here for the response.

devoncarew:
> I'm going to rewrite this class tomorrow.

How do you feel about landing this change in the meantime? I'll confirm again that this does terminated the app.

Issue #179, opened 2016-09-28
Device selector does not update when simulator is closed.


pq:
After killing my emulator, this remained:

![screen shot 2016-09-28 at 10 11 19 am](https://cloud.githubusercontent.com/assets/67586/18924171/fdde60d8-8563-11e6-9014-987181445540.png)

Also note that I killed it since observatory timed-out and I wanted to retry.

![screen shot 2016-09-28 at 10 12 23 am](https://cloud.githubusercontent.com/assets/67586/18924213/276601d6-8564-11e6-9ba4-24c1e054ab44.png)

Issue #182, opened 2016-09-28
Support to launch simulator from device selector menu


pq:
Idea: have an item at the end of the devices pulldown to start the iOS Simulator, if we’re on a mac and there’s no simulator running.

PR #212, opened 2016-09-30
device selector listens to device notifications

devoncarew:
- have the device selector listen to device notifications
- auto-select a device if we find one and there is no current selection
- on the mac, show a menu item for opening the ios simulator

screen shot 2016-09-29 at 11 21 14 pm

screen shot 2016-09-29 at 11 21 27 pm

@pq @stevemessick

### DIFF
--- a/src/io/flutter/actions/DeviceSelectorAction.java
+++ b/src/io/flutter/actions/DeviceSelectorAction.java
@@ -134,7 +134,7 @@ public class DeviceSelectorAction extends ComboBoxAction implements DumbAware {
     final boolean enabled;
 
     OpenSimulatorAction(boolean enabled) {
-      super("Open Simulator…");
+      super("Open iOS Simulator");
 
       this.enabled = enabled;
     }
--- a/src/io/flutter/actions/DeviceSelectorAction.java
+++ b/src/io/flutter/actions/DeviceSelectorAction.java
@@ -5,12 +5,24 @@
  */
 package io.flutter.actions;
 
+import com.intellij.execution.ExecutionException;
+import com.intellij.execution.configurations.GeneralCommandLine;
+import com.intellij.execution.process.OSProcessHandler;
+import com.intellij.execution.process.ProcessAdapter;
+import com.intellij.execution.process.ProcessEvent;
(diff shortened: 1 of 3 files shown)


devoncarew:
fix https://github.com/flutter/flutter-intellij/issues/179, https://github.com/flutter/flutter-intellij/issues/182

src/io/flutter/run/daemon/FlutterDaemonService.java
```diff
@@ -94,7 +104,9 @@ private FlutterDaemonService() {
   /**
    * @return the currently selected device
    */
-  public @Nullable ConnectedDevice getSelectedDevice() {
+  public
+  @Nullable
+  ConnectedDevice getSelectedDevice() {
```
stevemessick: We usually put access on the declaration line.

src/io/flutter/actions/DeviceSelectorAction.java
```diff
@@ -20,27 +32,36 @@
...
+
+    if (actions.isEmpty()) {
+      group.add(new NoDevicesAction());
+    }
+    else {
+      group.addAll(actions);
+    }
+
+    if (SystemInfo.isMac) {
+      boolean simulatorOpen = false;
```
stevemessick: Not used?

eseidelGoogle:
"..." used to imply "opens a dialog" in Apple's HUG, no clue if that's
still the case, but my instinct is you don't want the ... here.

src/io/flutter/run/daemon/FlutterDaemonService.java
devoncarew: done

src/io/flutter/actions/DeviceSelectorAction.java
devoncarew: We pass it into the `OpenSimulatorAction` ctor, int order to toggle enablement of the action.

devoncarew:
Re: the ellipses my initial impl had show a toast on success. I'll experiment with showing progress as the simulator opens, and remove the ellipses if that doesn't work out.

pq:
\o/

PR #213, opened 2016-09-30
use the path and cwd info from the launch config

devoncarew:
- use the path and cwd info from the launch config (so we can launch things other than `lib/main.dart`)

@stevemessick @pq


src/io/flutter/run/FlutterAppState.java
```diff
@@ -58,8 +61,6 @@ protected ProcessHandler startProcess() throws ExecutionException {
     FlutterDaemonService service = FlutterDaemonService.getInstance();
     assert service != null;
     Project project = getEnvironment().getProject();
-    String workingDir = project.getBasePath();
-    assert workingDir != null;
```
stevemessick: Shouldn't we ensure the Project is not the default project before starting the daemon?

devoncarew: Cool, added this back in.

PR #223, opened 2016-10-03
Check for minum required Dart Plugin version (#168).

### LINKED ISSUE #168: at runtime, check for and complain if the dart plugin version is less than our minimum
We should show a toast if the Dart plugin version doesn't meet our minimum; currently `162.2233`.

pq:
Adds an inspection to check for a minimum Dart Plugin version.

Also adds a new class `DartPlugin` to centralize our access to the Dart Plugin.

If this approach looks good, I'd like to move many if not all of our Dart Plugin API access to this class.  A step in the direction of better managing our dependency...

fixes #168

/cc @devoncarew @stevemessick

### DIFF
--- a/src/io/flutter/dart/DartPlugin.java
+++ b/src/io/flutter/dart/DartPlugin.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.dart;
+
+import com.intellij.ide.plugins.IdeaPluginDescriptor;
+import com.intellij.ide.plugins.PluginManager;
+import com.intellij.openapi.util.Version;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+/**
+ * Provides access to the Dart Plugin.
+ */
+public class DartPlugin {
+
+  /**
+   * Tracks the minimum required Dart Plugin version.
+   */
+  private static final String MINIMUM_REQUIRED_PLUGIN_VERSION = "162.2233";
+  private static final Version MINIMUM_VERSION = Version.parseVersion(MINIMUM_REQUIRED_PLUGIN_VERSION);
+
+  private static final DartPlugin INSTANCE = new DartPlugin();
+
+  private Version myVersion;
+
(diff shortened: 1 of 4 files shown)


src/io/flutter/dart/DartPlugin.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Provides access to the Dart Plugin.
+ */
+public class DartPlugin {
+
+  /**
+   * Tracks the minimum required Dart Plugin version.
+   */
+  private static final String MINIMUM_REQUIRED_PLUGIN_VERSION = "162.2233";
```
stevemessick: Just wondering if this should be in a file that's part of the Flutter plugin?

src/io/flutter/dart/DartPlugin.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private static final DartPlugin INSTANCE = new DartPlugin();
+
+  private Version myVersion;
+
+  public static DartPlugin getInstance() {
+    return INSTANCE;
+  }
+
+  private IdeaPluginDescriptor getIdeaPluginDescriptor() {
```
stevemessick: I like to define private constructors for singleton classes.

src/io/flutter/dart/DartPlugin.java
pq: You mean it should be in the Dart Plugin?

My reasoning is that it captures the Flutter Plugin's expectations about the Dart Plugin so it makes more sense for it to be there.  In an ideal world this would be in the `plugin.xml` but IDEA doesn't support version constraints.

Or maybe I'm missing your thrust?

src/io/flutter/dart/DartPlugin.java
pq: It's funny, the reason I even made this a singleton rather than just having a bunch of static methods was with an eye towards testing and I don't think a private constructor plays nicely with that but now you've got me wondering...

pq:
Thanks!

stevemessick:
On Mon, Oct 3, 2016 at 2:13 PM, Phil Quitslund <notifications@github.com>
wrote:

> You mean it should be in the Dart Plugin?


No. I meant not having it in a string constant. It doesn't really make any
practical difference, I suppose.

PR #227, opened 2016-10-03
Console file path hyperlinks for `flutter create`.

pq:
Adds path hyperlinks to `flutter create` output.

screen shot 2016-10-03 at 4 00 41 pm

/cc @devoncarew @stevemessick


devoncarew:
`flutter create` is really very noise. We started printing out all files when we were only creating 4 or so files. That moved to ~30 with the ios and android templates. Perhaps we should summarize all the files in subdirs? Or in `ios/` and `android/`? /cc @danrubel

danrubel:
Good point. I'll pull something together tomorrow.

Issue #221, opened 2016-10-03
Don't hang if debugging is started while already debugging


mit-mit:
Current if you hit the Debug button while already debugging the session hangs, and you have to restart IntellJ. If the longer term fix to this -- which I believe involves supporting multiple debugging sessions -- is too hard, then short term perhaps we can just 'do nothing' when we are already debugging?

devoncarew:
🎉

PR #234, opened 2016-10-04
App management improvements

stevemessick:
A few improvements and bug fixes for Flutter app management.

* De-bounce the Debug button. This still needs work but now if it is clicked when an app is running it won't break the world. It just does nothing. #221 
* Stop trying to stop an app if it isn't running.
* After the process has terminated, re-launch the polling service if needed.

@devoncarew @pq


pq:
LGTM.

Awesome!

Issue #226, opened 2016-10-03
Pause button of debugger throws NPE


stevemessick:
Start a flutter app, hit a breakpoint, click Continue, then click the Pause button (two blue vertical bars) in the left tool bar.

java.lang.NullPointerException
	at com.jetbrains.lang.dart.ide.runner.server.vmService.DartVmServiceDebugProcess.startPausing(DartVmServiceDebugProcess.java:344)
	at com.intellij.xdebugger.impl.XDebugSessionImpl.pause(XDebugSessionImpl.java:523)
	at com.intellij.xdebugger.impl.actions.handlers.XDebuggerPauseActionHandler.perform(XDebuggerPauseActionHandler.java:31)
	at com.intellij.xdebugger.impl.actions.handlers.XDebuggerActionHandler.perform(XDebuggerActionHandler.java:34)
	at com.intellij.xdebugger.impl.actions.XDebuggerActionBase.perform(XDebuggerActionBase.java:100)
	at com.intellij.xdebugger.impl.actions.XDebuggerActionBase.performWithHandler(XDebuggerActionBase.java:92)
	at com.intellij.xdebugger.impl.actions.XDebuggerActionBase.actionPerformed(XDebuggerActionBase.java:81)
	at com.intellij.openapi.actionSystem.ex.ActionUtil$1.run(ActionUtil.java:199)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAware(ActionUtil.java:215)
	at com.intellij.openapi.actionSystem.impl.ActionButton.actionPerformed(ActionButton.java:188)
	at com.intellij.openapi.actionSystem.impl.ActionButton.performAction(ActionButton.java:154)
	at com.intellij.openapi.actionSystem.impl.ActionButton.processMouseEvent(ActionButton.java:325)

PR #236, opened 2016-10-04
Make Run button work

stevemessick:
We just needed to pass the proper mode argument through to the daemon.

Apps launched via the Run button will not stop properly. I reported https://github.com/flutter/flutter/issues/6208

@pq @devoncarew


stevemessick:
Also fixes the NPE reported in #226 That was supposed to be a separate branch but I typed the wrong git command and it ended up here.

Issue #44, opened 2016-08-17
Create run configurations during project creation


pq:

mit-mit:
I'd really like to see this; it was confusing having to create this manually.

pq:
Questions on a few defaults:

* Config name. (Same as project?)
* Dart file?  (`lib/main.dart` if found, or?)
* Working dir?  (Project root?)

devoncarew:
Should we solve this by having `flutter create` emit the right intellij metadata? So IntelliJ doesn't create the launch config when opening an existing flutter project, but flutter projects (created by `flutter create`) just have all the bootstrapping info from the get go?

mit-mit:
That is a good point Devon! We do that today with the .atom folder, and could just as well create a .idea folder with the config! CC @danrubel WDYT?

It looks like .idea folder is preferred over the .ipf file format:
https://intellij-support.jetbrains.com/hc/en-us/community/posts/206915235-IPR-vs-IDEA

mit-mit:
As for config, I think we roughly need:

1. To enable Dart support
2. To have lib/main.dart open in an editor tab
3. To expand the lib folder in the project list
4. To create a config
  - Name: `Flutter`
  - Dart file: `<project dir>/lib.main.dart`
  - Checked mode: `disabled`

mit-mit:
I went through an experiment making the changes I described above, and between each change made a commit to git. The resulting [commit log](https://github.com/flutter/flutter-intellij/files/511148/IdeaFileChanges.pdf) shows a pretty staggering set of changes :-|

mit-mit:
@pq I was thinking that this is not blocking for fishfood (we can provide config instructions), but that it blocks the [external alpha](https://github.com/flutter/flutter-intellij/milestone/7) instead. WDYT?

pq:
> @pq I was thinking that this is not blocking for fishfood 

👍  

That said, I got it mostly working yesterday so I should have a PR later this AM. 

> 1. To enable Dart support
> 2. To have lib/main.dart open in an editor tab
> 3. To expand the lib folder in the project list
> 4. To create a config
>   - Name: `Flutter`
>   - Dart file: `<project dir>/lib.main.dart`
>   - Checked mode: `disabled`

Hits all these high points. 😄

mit-mit:
Excellent!

pq:
Re-framing this issue back to being about run config generation.  We can open another to track other template generation ideas.

pq:
[...]

PR #243, opened 2016-10-05
Generate a run config during project creation (#44).

pq:
Generates a default run configuration for new projects.

Also fixes a race condition with opening `main.dart` by forcing a filesystem refresh before doing the existence check.

Fixes: https://github.com/flutter/flutter-intellij/issues/44

/cc @stevemessick @devoncarew

### DIFF
--- a/src/io/flutter/sdk/FlutterSdk.java
+++ b/src/io/flutter/sdk/FlutterSdk.java
@@ -6,6 +6,9 @@
 package io.flutter.sdk;
 
 import com.intellij.execution.ExecutionException;
+import com.intellij.execution.RunManager;
+import com.intellij.execution.RunnerAndConfigurationSettings;
+import com.intellij.execution.configurations.ConfigurationFactory;
 import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.process.OSProcessHandler;
 import com.intellij.execution.process.ProcessAdapter;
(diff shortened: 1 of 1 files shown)


src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -211,11 +219,48 @@ void onTerminate(@Nullable Module module,
...
           if (workingDir != null && module != null && !module.isDisposed()) {
+            final Project project = module.getProject();
+
             // Enable Dart.
             FlutterSdkUtil.enableDartSupport(module);
-            final FileEditorManager manager = FileEditorManager.getInstance(module.getProject());
-            // Open main.
-            final VirtualFile main = workingDir.findFileByRelativePath("lib/main.dart");
+            final FileEditorManager manager = FileEditorManager.getInstance(project);
+
```
stevemessick: Do you really want two blank lines here?

pq: Nope!

stevemessick: Still LGTM

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -211,11 +219,47 @@ void onTerminate(@Nullable Module module,
...
+            final Project project = module.getProject();
+
             // Enable Dart.
             FlutterSdkUtil.enableDartSupport(module);
-            final FileEditorManager manager = FileEditorManager.getInstance(module.getProject());
-            // Open main.
-            final VirtualFile main = workingDir.findFileByRelativePath("lib/main.dart");
+            final FileEditorManager manager = FileEditorManager.getInstance(project);
+            
+            // Create a basic run configuration.
```
devoncarew: Should we check to see if one already exists? Or, asked another way, how would this react if `flutter create` is changed to create an intellij launch configuration as part of its template?

pq: I don't think `flutter create` is the way to go here since I think run configs live in project metadata somewhere?

As for a pre-existing run config, I guess I assumed that with a fresh project there couldn't be one but since this could be called during module creation maybe that's not safe.

https://github.com/flutter/flutter-intellij/issues/245

Issue #231, opened 2016-10-04
Dart not always enabled for new Flutter Projects created from source.


stevemessick:
I created a new Flutter project on existing sources and found that it did not have Dart support enabled. I had to open the pubspec.yaml file and enable it there. It should have been done as part of project creation.

pq:
Strange!  What steps did you use to create the project?

Using [these steps](https://github.com/flutter/flutter-intellij/wiki/User-Journeys#create-a-flutter-project-based-on-existing-source) that should _just work_.

stevemessick:
That's what I did.

pq:
Did `flutter create` get called?

stevemessick:
How can I tell? Is there a log file somewhere?

BTW this was a big Flutter project, sync'd to pristine state. It had not had flutter doctor or pub get done on it previously. It took a long time to do pub get. I think I restarted the IDE at one point, but I'm currently single-stepping through (some of) the code.

pq:
There should be a message pane at the bottom chrome:

<img width="440" alt="screen shot 2016-10-04 at 9 13 48 am" src="https://cloud.githubusercontent.com/assets/67586/19082417/ebfc5c40-8a12-11e6-85ea-23b48608b317.png">

pq:
Updated title and add a bug label to indicate that this is expected to work.

stevemessick:
That message is long gone. I believe I restarted IntelliJ at one point.

stevemessick:
This is straightforward to reproduce. 
1. Get a pristine version of flutter/examples/flutter_gallery. `git status` should report no modifications.
2. Create a new Flutter project in IntelliJ pointing to that directory.
3. Open lib/main.dart
4. See the prompt to enable Dart support, as in the screen shot.

<img width="1096" alt="screen shot 2016-10-05 at 3 52 03 pm" src="https://cloud.githubusercontent.com/assets/8518285/19134646/194fe750-8b14-11e6-9a40-658481e0c92f.png">

I don't know if you're supposed to be opening the flutter_gallery directory like that. If not, I bet I won't be the only one who does. The message about obtaining a lock might be a clue.

pq:
This is helpful thanks.  That message is definitely interesting!

pq:
A few quick thoughts.

[...]

PR #246, opened 2016-10-06
Hoist Dart module config out of `flutter create` process wait (#231).

pq:
Fixes: https://github.com/flutter/flutter-intellij/issues/231

/cc @stevemessick @devoncarew


stevemessick:
LGTM but not able to use GitHub at this location.

Issue #256, opened 2016-10-08
exception on debug after a launch timeout


devoncarew:
When launching an app (on a very slow laptop), I often see a timeout. On the 2nd debug launch, I see the exception:

```
null
java.lang.NullPointerException
	at io.flutter.run.daemon.RunningFlutterApp.deviceId(FlutterApp.java:198)
	at io.flutter.run.daemon.FlutterAppManager.lambda$isAppRunning$2(FlutterAppManager.java:95)
	at java.util.stream.MatchOps$1MatchSink.accept(MatchOps.java:90)
	at java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1351)
...
```

PR #257, opened 2016-10-08
fix an exception on launch

devoncarew:
- fix https://github.com/flutter/flutter-intellij/issues/256
- fix an issue w/ slow launches; after a timeout, the 2nd launch would NPE

@stevemessick


stevemessick:
LGTM

PR #259, opened 2016-10-08
group dart and flutter wizards

devoncarew:
- group the dart and flutter wizards in the new project dialog

screen shot 2016-10-08 at 6 07 10 am

I'm of two minds about this; I like the Dart and Flutter wizards being grouped, but in a more web based intelliJ product, this would but the Flutter wizard in with the web focused wizards. @pq @stevemessick


stevemessick:
LGTM
I agree with your concern but don't have any thing to add.

devoncarew:
Yeah, we can change it if we find that it's not working.

Issue #254, opened 2016-10-07
Remove Dart nature to get back to a trim build.


pq:
(No need to bundle the Dart SDK.)

![screen shot 2016-10-07 at 1 22 19 pm](https://cloud.githubusercontent.com/assets/67586/19204325/16084028-8c91-11e6-8d63-3069550a6524.png)

PR #260, opened 2016-10-08
remove the dart nature for smaller build artifacts

devoncarew:
- remove the dart nature for smaller build artifacts
- fix https://github.com/flutter/flutter-intellij/issues/254

We're back down to around 205k for the `flutter-intellij.jar` file; @pq.


pq:
Lean and mean.  LGTM!

Issue #178, opened 2016-09-28
Restart after plugin install throws Fatal IDE Exception.


pq:
![screen shot 2016-09-28 at 10 04 33 am](https://cloud.githubusercontent.com/assets/67586/18923939/19b38c12-8563-11e6-81fc-b638509b7861.png)

![screen shot 2016-09-28 at 10 04 43 am](https://cloud.githubusercontent.com/assets/67586/18923943/1bb7a07a-8563-11e6-8527-eba515830a26.png)

pq:
The work-around is to set the SDK and restart.  (Restart is required to get the daemon rebooted.)

devoncarew:
Ran into this again (on a fresh install):

```
The Flutter SDK is not configured
com.intellij.execution.ExecutionException: The Flutter SDK is not configured
	at io.flutter.run.daemon.FlutterDaemonController.createCommandLine(FlutterDaemonController.java:210)
	at io.flutter.run.daemon.FlutterDaemonController.forkProcess(FlutterDaemonController.java:118)
	at io.flutter.run.daemon.FlutterAppManager.startDevicePoller(FlutterAppManager.java:153)
	at io.flutter.run.daemon.FlutterDaemonService.lambda$schedulePolling$0(FlutterDaemonService.java:197)
	at com.intellij.openapi.application.impl.ApplicationImpl$2.run(ApplicationImpl.java:309)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

pq:
@stevemessick before I go mucking around in here, I wonder if this is on your radar...  If so, please re-assign.  In any event, we should coordinate.

stevemessick:
I'm not looking at this. Today is vacation anyway.

pq:
Did a bit of plumbing in this weekend and have a few questions.  Maybe we can chat today.

PR #277, opened 2016-10-11
Watch for SDK config changes and trigger daemon starts/stops (#178).

pq:
Makes the device daemon more “self-healing”.

With this, the “Clean Install” user story becomes:

1. Install Flutter plugin
2. Restart IDE
3. Create Flutter project (setting SDK in the process)
4. Profit

:)

No more extra restart and fatal exceptions.

Fixes: https://github.com/flutter/flutter-intellij/issues/178


pq:
> So maybe it took a little more than 10 minutes :)

Ha!  Yeah.  There was a little more to it in the end...  Thanks a million for your help!

PR #278, opened 2016-10-11
Quietly recover from OS process exit

stevemessick:
@pq @devoncarew Gracefully recover from the flutter daemon process being killed (eg: kill -9 ...), or any other form of unplanned exit.


pq:
LGTM!

PR #296, opened 2016-10-13
Version `.idea/` contents.

pq:
As per Alex's suggestion this morning, add `.idea` to versioning so that we have consistency across developers.  (This is part of _the Jetbrains way_...)

To be fair, these files may benefit from some tweaking but once we get it right we'll all have the same settings which I agree is a good thing.

/cc @stevemessick @devoncarew

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 out/
-.idea/
 .gradle/
 build/
 flutter-intellij.jar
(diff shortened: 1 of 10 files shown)


stevemessick:
LGTM

devoncarew:
If this is the jetbrains way, I'm all for it :)

I think some of these files are naturally per-developer however? And others are per team?

pq:
> I think some of these files are naturally per-developer however? And others are per team?

It sure seems that way.  Looking at the intellij-plugins repo though, all they ignore is `.DS_Store`!

Issue #293, opened 2016-10-13
Bump Flutter `VERSION` so that we can detect stale SDKs.


pq:
See #194.

PR #299, opened 2016-10-13
Track and check minimum supported Flutter SDK (#194).

### LINKED ISSUE #194: Add a notification if the currently configured Flutter SDK is too old.
The rub is that out of date Flutter SDKs can produce unpredictable results (notably debugger issues).  To that end, it would be great to show a toast and probably an error in the preference and project creation pages if this case is detected.

Open question: what from our version info is right to key off?

screen shot 2016-09-29 at 5 39 06 am

cc @abarth @sethladd since I think you guys were thinking about this?

pq:
Adds an editor notification that pops up when Flutter is less than `0.0.2`, the earliest tagged reload-friendly SDK. (See: #293.)

Fixes #194 

/cc @stevemessick @devoncarew

### DIFF
--- a/src/io/flutter/inspections/SdkConfigurationNotificationProvider.java
+++ b/src/io/flutter/inspections/SdkConfigurationNotificationProvider.java
@@ -27,19 +27,24 @@ import io.flutter.FlutterBundle;
 import io.flutter.module.FlutterModuleType;
 import io.flutter.sdk.FlutterSdk;
 import io.flutter.sdk.FlutterSdkService;
+import io.flutter.sdk.FlutterSdkVersion;
 import io.flutter.settings.FlutterSettings;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public class WrongDartSdkConfigurationNotificationProvider extends EditorNotifications.Provider<EditorNotificationPanel>
+public class SdkConfigurationNotificationProvider extends EditorNotifications.Provider<EditorNotificationPanel>
   implements DumbAware {
+
+  // Minimum SDK known to support hot reload.
+  private static final FlutterSdkVersion MIN_SUPPORTED_SDK = FlutterSdkVersion.forVersionString("0.0.2");
+
   private static final Key<EditorNotificationPanel> KEY = Key.create("FlutterWrongDartSdkNotification");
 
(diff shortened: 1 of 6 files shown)


src/io/flutter/settings/FlutterSettings.java
```diff
@@ -49,4 +50,12 @@ public boolean ignoreMismatchedDartSdks() {
...
     this.ignoreMismatchedDartSdks = ignoreMismatchedDartSdks;
   }
-}
\ No newline at end of file
+
+  public void setIgnoreOutOfDateSdks(boolean ignoreOutOfDateSdks) {
+    this.ignoreOutOfDateSdks = ignoreOutOfDateSdks;
+  }
+
+  public boolean ignoreOutOfDateSdks() {
```
stevemessick: Convention is to use some boolean-sounding name, for example isIgnoringOutOfDateSdks()

src/io/flutter/inspections/SdkConfigurationNotificationProvider.java
```diff
@@ -108,6 +117,21 @@ public EditorNotificationPanel createNotificationPanel(@NotNull VirtualFile file
...
 
+  private EditorNotificationPanel createOutOfDateFlutterSdkPanel() {
+
+    final FlutterSettings settings = FlutterSettings.getInstance(project);
+    if (settings == null || settings.ignoreOutOfDateSdks()) return null;
+
+    EditorNotificationPanel panel = new EditorNotificationPanel();
+    panel.setText(FlutterBundle.message("flutter.old.sdk.warning"));
+    panel.createActionLabel("Dismiss", () -> {
+      settings.setIgnoreOutOfDateSdks(true);
```
stevemessick: How does this ever get reset to false?

pq: It doesn't yet.  My thinking is that we'll want to introduce some kind of event that blanks the setting.  Not sure what that should be just yet though.  Maybe the next time the min version bumps?

I admit to have thought about it just not settled yet so I punted.  Thoughts welcome!

stevemessick: You could add something to the Flutter preferences page until the automagical event is devised.

PR #301, opened 2016-10-13
Checkpoint multi-debugging

stevemessick:
@pq @devoncarew Checkpoint. I'm probably not going to commit this as-is. But if you get a chance to take a look I'd like to know if you think the approach makes sense.

The new class, `DaemonProcessWrapper`, is a bit of trickery. It is a wrapper around the OSProcessHandler that controls the daemon process. All it does is change object identity so that the debugger connects to the correct process. The reflective bits are due to `ProcessHandler` code that cannot be simply redirected. I'll add a class comment to that effect before the final commit.

### DIFF
--- a/src/io/flutter/run/daemon/DaemonProcessWrapper.java
+++ b/src/io/flutter/run/daemon/DaemonProcessWrapper.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.run.daemon;
+
+import com.intellij.execution.process.ProcessHandler;
+import com.intellij.execution.process.ProcessListener;
+import com.intellij.openapi.util.Key;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.OutputStream;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class DaemonProcessWrapper extends ProcessHandler {
+  private ProcessHandler myHandler;
+
+  DaemonProcessWrapper(ProcessHandler handler) {
+    super();
+    myHandler = handler;
+  }
+
+  public void startNotify() {
+    myHandler.startNotify();
+  }
+
+  @Override
+  public void destroyProcessImpl() {
+    reflectivelyInvoke("destroyProcessImpl");
+  }
+
+  @Override
+  public void detachProcessImpl() {
+    reflectivelyInvoke("detachProcessImpl");
+  }
+
+  @Override
+  public boolean detachIsDefault() {
+    return myHandler.detachIsDefault();
+  }
+
+  public boolean waitFor() {
+    return myHandler.waitFor();
+  }
+
+  public boolean waitFor(long timeoutInMilliseconds) {
+    return myHandler.waitFor(timeoutInMilliseconds);
+  }
+
+  public void destroyProcess() {
+    myHandler.destroyProcess();
+  }
+
+  public void detachProcess() {
+    myHandler.detachProcess();
+  }
+
+  public boolean isProcessTerminated() {
+    return myHandler.isProcessTerminated();
+  }
+
+  public boolean isProcessTerminating() {
+    return myHandler.isProcessTerminating();
+  }
+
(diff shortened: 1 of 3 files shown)


pq:
My gut wants me to suggest some other approach (wanting to avoid reflection tricks for all the obvious reasons) but I'm not seeing one.  I'll defer to @devoncarew who knows this space better but otherwise LGTM.

src/io/flutter/run/daemon/DaemonProcessWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+  public void removeProcessListener(final ProcessListener listener) {
+    myHandler.removeProcessListener(listener);
+  }
+
+  public void notifyProcessDetached() {
+    reflectivelyInvoke("notifyProcessDetached");
+  }
+
+  public void notifyProcessTerminated(final int exitCode) {
+    reflectivelyInvoke("notifyProcessTerminated");
```
devoncarew: Should this pass along the exitCode?

src/io/flutter/run/daemon/DaemonProcessWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+    return myHandler.isSilentlyDestroyOnClose();
+  }
+
+  private void reflectivelyInvoke(String methodName) {
+    try {
+      Method method = myHandler.getClass().getDeclaredMethod(methodName);
+      method.setAccessible(true);
+      method.invoke(myHandler);
+    }
+    catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException ex) {
```
devoncarew: I did not know you could use `|` here -

src/io/flutter/run/daemon/DaemonProcessWrapper.java
stevemessick: Yes. Thanks for catching that. However, I think I'm going to discard this entire class.

src/io/flutter/run/daemon/FlutterDaemonController.java
```diff
@@ -116,7 +114,7 @@ void removeDeviceId(String deviceId) {
   public void forkProcess(Project project) throws ExecutionException {
     try {
       GeneralCommandLine commandLine = createCommandLine(project);
-      myHandler = new OSProcessHandler(commandLine);
+      myHandler = new DaemonProcessWrapper(new OSProcessHandler(commandLine));
```
devoncarew: It looks like both startDeviceController and startApp call forkProcess, and then the result is stored in an instance variable. Is a `FlutterDaemonController` shared between multiple running apps?

I'm woefully unfamiliar with the architecture here, and why we're wrapping a process here, instead of creating a new DaemonProcess style subclass of ProcessHandler.

src/io/flutter/run/daemon/DaemonProcessWrapper.java
stevemessick: I didn't either but when I saw IntelliJ give the separate catch clauses that special yellow highlighting I was glad to see it. This is something I've long wanted.

devoncarew:
I don't have enough familiarity here to comment reasonably; lgtm if you think this is the right thing to do. Please add more docs to help the next person through!

> All it does is change object identity so that the debugger connects to the correct process.

That does help explain things a bit.

src/io/flutter/run/daemon/FlutterDaemonController.java
stevemessick: With this approach we'd have a single `flutter daemon` process for each app (i.e. flutter project directory). There's a single `FlutterDaemonController` in charge of each external daemon process. So yes, the controller can be shared between running apps provided they are started from the same directory but run on different devices. And the device polling job can be handled by the same process. That was part of the original design.

I didn't create a new subclass because Window and Linux/Mac have different subclasses of ProcessHandler already, so we would have needed to do it twice. The wrapper is intended to fix the cases where the debugger uses object identity to identify debugger tabs.

However, there are enough other odd things going on, and maintenance is likely to be so confusing, that I'm experimenting with a non-shared daemon approach. We'll end up with a lot of external process (potentially) but that's not really a big deal. Having readable code is probably more important than an elegant system architecture.

stevemessick:
Closing in favor of #306

PR #303, opened 2016-10-14
Fix ignores.

pq:
This should take care of the churn.

/cc @devoncarew @stevemessick


alexander-doroshko:
👍

pq:
Thanks @alexander-doroshko for the tips!

For the curious, it turns out these files are defaulted out in the IDE (**Preferences | Version Control | Ignored Files**) so if we were using it's VCS we wouldn't have this problem.

PR #306, opened 2016-10-14
Stop sharing daemons

stevemessick:
@pq @devoncarew This makes #301 obsolete.

Instead of using the minimal number of daemon process the approach is to spawn a separate daemon for each application. I've only tested this with two devices but it seems solid, except for the bug I filed about breakpoints.

I chose to leave the old code in place as a comment until we're sure it won't be needed.


pq:
LGTM!

PR #313, opened 2016-10-17
Only show device selector for Flutter projects (#267).

### LINKED ISSUE #267: Only display device list for Flutter projects
If the Flutter plugin is installed the device list pull-down is displayed in the tool bar unconditionally. It should be possible to adapt that behavior so the pull-down is only displayed when the current project is a Flutter project.

pq:
Fixes #267

/cc @stevemessick @devoncarew

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkUtil.java
+++ b/src/io/flutter/sdk/FlutterSdkUtil.java
@@ -179,7 +179,7 @@ public class FlutterSdkUtil {
    *
    * @return true if an open Flutter project is found
    */
-  public static boolean isFlutterProjectOpen() {
+  public static boolean isFluttering() {
     return Arrays.stream(ProjectManager.getInstance().getOpenProjects()).anyMatch((Project p) -> FlutterSdkUtil.hasFlutterModule(p));
   }
 
--- a/src/io/flutter/sdk/FlutterSdkUtil.java
+++ b/src/io/flutter/sdk/FlutterSdkUtil.java
@@ -176,6 +176,7 @@ public class FlutterSdkUtil {
 
   /**
    * Checks the workspace for any open Flutter projects.
+   *
    * @return true if an open Flutter project is found
    */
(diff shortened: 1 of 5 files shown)


devoncarew:
lgtm!

pq:
> isFlutterProjectOpen() ==> isFluttering()?

Cute.

And done!

src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -163,6 +166,23 @@ private static String readVersionFile(String sdkHomePath) {
...
+    return ModuleUtil.hasModulesOfType(project, FlutterModuleType.getInstance());
+  }
+
+  /**
+   * Checks the workspace for any open Flutter projects.
+   *
+   * @return true if an open Flutter project is found
+   */
+  public static boolean isFluttering() {
+    return Arrays.stream(ProjectManager.getInstance().getOpenProjects()).anyMatch((Project p) -> FlutterSdkUtil.hasFlutterModule(p));
```
alexander-doroshko: `DeviceSelectorAction` is a project-specific action. I suggest to add `Project` parameter to `isFluttering()` and pass `e.getProject()` as an argument.

alexander-doroshko: Or, better, delete `isFluttering()` method and call `hasFlutterModule(p)` right from the `DeviceSelectorAction.update()`.
It is very unusual when the presence of some other open project affects UI or behavior of the currently focused one.

pq: Interesting.  I shied away from `e.getProject()` because sometimes it returns null but then I now notice that it's getting called more than once...  So maybe it's not null when it needs to be?

pq: OK, I went with your suggestion.  Thanks!

https://github.com/flutter/flutter-intellij/pull/314

Issue #230, opened 2016-10-04
Emulate AS Instant Run button for debugger & hot reload


stevemessick:
From @mit-mit 
Try to change the debug icon appearance and behavior to match Android Studio:

![android-studio-intant-run](https://cloud.githubusercontent.com/assets/8518285/19076423/91331200-89fd-11e6-8f50-526e7a1a0082.gif)

Clicking Debug starts the debugging session and changes the Debug icon to a "Debug Refresh" icon. Hitting this would do a hot reload for Flutter. Clicking stop ends the session.

Search for "Instant Run" to get started in AS sources.

stevemessick:
Let's use standard abbreviations.

mit-mit:
Tentatively tagging this for the parity milestone.

devoncarew:
@stevemessick - you've looked at this - do you have a sense for approx. how much work this is? I'm wondering if it's small/medium/large. Thanks!

stevemessick:
Actually, I haven't had a chance to do any more than the quick search I described. I don't know what's really involved in implementing it. It sounds like UI work, and not something standard in IntelliJ, so it could be a big job even though it seems small-ish. I would de-prioritize it until we were actually trying to integrate with AS. Is that happening now?

stevemessick:
I made some progress on this, since it would also solve #332 and #334. See https://github.com/flutter/flutter-intellij/tree/enablement

It is blocked on flutter #6400.

mit-mit:
@stevemessick and @johnmccutchan can you please sync on this today, Friday? Even if we cannot resolve this by early next week for 0.1, I think we really need this fixed in 0.2.

devoncarew:
@mit-mit - would you say this is covered by https://github.com/flutter/flutter-intellij/issues/407?

mit-mit:
Covered by #407

PR #318, opened 2016-10-18
Fix inspections in ProgressHandler

stevemessick:
@pq @devoncarew @alexander-doroshko This is the first step towards fixing #230 

Tested manually, like most of the debug stuff.


pq:
LGTM.

PR #291, opened 2016-10-12
Getting started tweaks (#287, #288).

pq:
Fixes #287 
Fixes #288

/cc @devoncarew @stevemessick

PR #319, opened 2016-10-18
Update getting_started.md

mit-mit:
A few minor edits: 

* Change 'OS X' to the new name 'macOS'
* Remove mentioning of Windows as we don't support it yet (could mislead some to think we do)


mit-mit:
@pq just minor tweaks to your update from last week in #291

mit-mit:
Thanks!

PR #321, opened 2016-10-18
Add a single Observatory button

stevemessick:
@pq @devoncarew @alexander-doroshko 

Due to refactoring, we were adding two identical buttons to the run console that would open the observatory.


devoncarew:
the more the merrier?

lgtm

pq:
> the more the merrier?

ha!

LGTM2

Issue #328, opened 2016-10-19
Exception while activating newly installed flutter plugin.


mravn-google:
IntelliJ seems to indicate that we are accessing UI components from a background thread.

Steps to reproduce: I was following installation instructions:

> Flutter plugin installation
> 
> During alpha development, the Flutter plugin is hosted in its own plugin repository. Before it can be installed, the repository needs to be added to the IDE.
> 
> Open plugin preferences (Preferences>Plugins on macOS, File>Settings>Plugins on Linux)
> Select "Browse repositories…"
> In the Browse Repositories dialog, click "Manage repositories…"
> In the resulting Custom Plugin Repositories pop-up, click the + button and paste https://flutter.github.io/flutter-intellij/alpha/updatePlugins.xml into the Repository URL: field and click OK
> In the Browse Repositories dialog, select flutter-intellij from the list of plugins (be sure it is not being filtered out of your results view by specifying Category: All or limiting it to the Flutter repository in the pull-down by the search box) and click the green Install button
> Once downloaded, click the Restart IntelliJ IDEA button
> Flutter plugin configuration
> 
> Open preferences (Preferences on macOS, File>Settings on Linux)
> Select Languages & Frameworks>Flutter
> Set the path to the root of your flutter repo
> Click OK

Exception occurred after the final OK click. Happens again after restart of IntelliJ and repetition of the last four steps above.

```
"AWT-EventQueue-0 2016.1.3#IC-145.1617.8 IDEA, eap:false, os:Linux 3.13.0-96-generic, java-version:JetBrains s.r.o 1.8.0_76-release-b198" prio=0 tid=0x0 nid=0x0 waiting on condition
     java.lang.Thread.State: WAITING
 on java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject@280eada7
	at sun.misc.Unsafe.park(Native Method)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
	at java.awt.EventQueue.getNextEvent(EventQueue.java:554)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:170)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
[...]

PR #331, opened 2016-10-19
Add IDEA version constraints to plugin.xml (#328).

pq:
Fixes: #328

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -8,6 +8,8 @@
 
   <version>0.1.1</version>
 
+  <idea-version since-build="162.1" until-build="162.*"/>
+  
 <change-notes>
 <![CDATA[
 0.1.0:


alexander-doroshko:
Note that in JetBrains' repos `master` branch currently targets 2017.1, so `since-version` would be `171.1`. For 2016.3 releases there's branch `163` and for 2016.2 - branch `162`. If you decide to follow our branch structure make sure to have appropriate `since-build` and `until-build` in each flutter-plugin branch.

pq:
@alexander-doroshko : what would you recommend we do here?  I'm not familiar with the pros and cons...  We really just want to ensure that a people don't get into a state where they're using the plugin with an incompatible earlier version.

alexander-doroshko:
I recommend that in flutter-plugin repository you have the same branches as in Dart plugin and IntelliJ Community and build compatible plugin build from appropriate branch. In each branch you'll have different Dart.jar that you compile against. And also each branch will have different version of IntelliJ Plugin SDK.

alexander-doroshko:
Pros: you develop and compile against JARs that Flutter-plugin will be executed against in runtime (well, almost)
Cons: everyone will need to understand what these branches are for, and also to have several installations of IntelliJ IDEA (if he/she develops in different branches).

pq:
Good points.

Assuming we *do* want to go this road, are these constraints right for the current 162 Dart plugin?

alexander-doroshko:
One more con: you'll need to carefully make sure that all commits are correctly cherry-picked to each branch where it is expected to be.
So all cons are about discipline and better understanding of the processes.
But the pro is that "code compiles => plugin won't fail", and it is possible with similar branches only.

Still you may have conflicts and incompatibilities when IntelliJ API or Dart plugin API changes within a branch. You may need to have subbranches and/or more granular since/until. The only way to fix ALL incompatibility issue is to run IntelliJ IDEA Community from sources + Dart plugin from sources + Flutter plugin from sources. But with simplified single-module Flutter-plugin project setup the best you can do is to follow our branches structure.

alexander-doroshko:
yes, for 2016.2 these constraints are correct.

devoncarew:
lgtm

We'll need to carefully evaluate our dev process when we decide to support android studio - they use an older version of the IntelliJ APIs.

PR #343, opened 2016-10-19
Cleanup FlutterSettingsConfigurable

alexander-doroshko:
A bunch of fixes for minor bugs, mostly related to the Flutter page in Preferences + some code simplification.


src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -69,33 +61,9 @@ private FlutterSdk(@NotNull final String homePath) {
     this(homePath, FlutterSdkUtil.getSdkVersion(homePath));
   }
 
-  /**
-   * Returns the same as {@link #getGlobalFlutterSdk()} but much faster
-   */
   @Nullable
   public static FlutterSdk getFlutterSdk(@NotNull final Project project) {
```
pq: I believe we borrowed this pattern from the Dart plugin.  Is it unnecessary there too?

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -108,27 +83,23 @@ public String getId() {
 
   @Nullable
   @Override
-  public Runnable enableSearch(String option) {
-    //TODO(pq): enable search
```
pq: I'd like to be sure we implement search and would rather leave a TODO than delete it.  Unless you think we shouldn't?

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -66,34 +64,11 @@ protected void textChanged(final DocumentEvent e) {
...
 
-    final TextComponentAccessor<JComboBox> textComponentAccessor = new TextComponentAccessor<JComboBox>() {
-      @Override
-      public String getText(final JComboBox component) {
-        return component.getEditor().getItem().toString();
-      }
-
-      @Override
-      public void setText(@NotNull final JComboBox component, @NotNull final String text) {
-        if (!text.isEmpty() && !FlutterSdkUtil.isFlutterSdkHome(text)) {
```
pq: Is this validation made unnecessary by the logic in `apply`?  (Is that why you're deleting it?)

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -66,34 +64,11 @@ protected void textChanged(final DocumentEvent e) {
...
-
+    sdkCombo.addBrowseFolderListener("Select Flutter SDK Path", null, null,
+                                     FileChooserDescriptorFactory.createSingleFolderDescriptor(),
+                                     TextComponentAccessor.STRING_COMBOBOX_WHOLE_TEXT);
 
-    final ComponentWithBrowseButton.BrowseFolderActionListener<JComboBox> browseFolderListener =
-      new ComponentWithBrowseButton.BrowseFolderActionListener<>("Select Flutter SDK Path", null, sdkCombo,
-                                                                 null,
-                                                                 FileChooserDescriptorFactory.createSingleFolderDescriptor(),
-                                                                 textComponentAccessor);
```
pq: Again, I thought having a validating combo accessor made sense.  Is that not the preferred pattern?

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -66,34 +64,11 @@ protected void textChanged(final DocumentEvent e) {
...
-    sdkCombo.addBrowseFolderListener(null, browseFolderListener);
+    versionDetails.setBackground(UIUtil.getPanelBackground());
```
pq: Ah.  Good catch!  I'm guessing this would look bad otherwise with custom color themes?

Does this change the current appearance?

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -142,24 +113,22 @@ public void apply() throws ConfigurationException {
...
 
   private void updateVersionText() {
-
     final FlutterSdk sdk = FlutterSdk.forPath(getSdkPathText());
     if (sdk == null) {
       versionDetails.setVisible(false);
     }
     else {
       try {
+        final ModalityState modalityState = ModalityState.current();
```
pq: Curious why you're hoisting this out of the runnable.  Don't we want the modality state when it gets invoked?

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -209,6 +178,6 @@ private String getErrorMessage() {
   }
 
   private String getSdkPathText() {
-    return sdkCombo.getComboBox().getEditor().getItem().toString().trim();
+    return FileUtilRt.toSystemIndependentName(sdkCombo.getComboBox().getEditor().getItem().toString().trim());
```
pq: Cool!

src/io/flutter/sdk/FlutterSdk.java
alexander-doroshko: It was done in the Dart plugin for 2 reasons:
- at that time reference resolution was done by IDE (now by server) and getGlobalDartSdk() was called thousands times a second
- bottleneck of getGlobalDartSdk() is java.io.File.exists(). It takes much less than 1ms, but when called thousands times a second it was visible in CPU snapshot

Neither is the case for Flutter. It is neither called frequently nor has File.exists() in its implementation.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
alexander-doroshko: This logic was written for the following use case:
- assume I have `dart` folder with `dart-sdk` and `dartium` subfolders
- in file chooser I select not the `dart-sdk` folder, but its parent - `dart` folder
- ide notices that selected path is not a dart sdk folder and looks for dart-sdk subfolder
- if dart-sdk subfolder contains Dart SDK then IDE behaves as if user has chosen correct folder himself

This logic is not applicable for Dart because it is unlikely that Flutter SDK is in `dart-sdk` folder and user selects parent folder instead.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
alexander-doroshko: See previous comment. Accessor was made not for validation and it is not intended to do any validation. Default implementation only converts slashes - native in UI, forward in model.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
alexander-doroshko: 'Flutter' word is successfully searched for automatically. Do you think you'll want to look for anything else? Ready to return TODO back.

src/io/flutter/sdk/FlutterSdk.java
pq: Ok.  Thanks for the explanation!

src/io/flutter/sdk/FlutterSettingsConfigurable.java
pq: Great.  Thanks for the context.  I'm convinced.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
pq: OK.  Thanks.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
pq: I'll defer to @devoncarew but I think we actually *want* a contrasting background for the version details.  That said, I think the way we had it originally was probably not right.

To be clear, @devoncarew : do we intend the version background to be different than the rest of the panel?  (I think it looks good when it pops a little...)

src/io/flutter/sdk/FlutterSettingsConfigurable.java
pq: No.  I think that's fine.  I didn't know about the automatic behavior and just didn't want to forget.  Let's remove it. 👍

src/io/flutter/sdk/FlutterSettingsConfigurable.java
pq: Ah, OK.  I guess I thought `updateVersionText()` was getting called on the AWT thread but it's hard to keep straight. 😄 

If I'm understanding you right, isn't it odd that this would ever have worked?


screen shot 2016-10-20 at 5 38 24 am

Why am I not seeing an exception on HEAD?

(Incidentally, these version details are something we don't want to lose from the UI when we make changes to Flutter configuration.  They're very useful for debugging.)

src/io/flutter/sdk/FlutterSettingsConfigurable.java
alexander-doroshko: How about showing version details as a standard multiline JLabel? It will look the same as 'Flutter SDK path' or 'Version details' label an will take several lines according to the `flutter -version` output. Happy to update the PR.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
alexander-doroshko: Looks like it never worked before.

> Why am I not seeing an exception on HEAD?

I'm sure you should see it. If you are running from sources see `idea.log` tab of the Debug tool window.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
pq: Sounds good to me!  We're obviously not that familiar with IntelliJ UI bits.  Any improvements are appreciated.

Could you add a screenshot of the "new look"?  (Assuming it's any different.)

src/io/flutter/sdk/FlutterSettingsConfigurable.java
pq: Sorry.  Not seeing that tab but I'm probably not looking in the right place...

screen shot 2016-10-20 at 6 02 38 am

Time for more coffee! ☕️

alexander-doroshko: Got it. You probably use IntelliJ IDEA 2016.2 and assertion was added later, see [this commit](https://github.com/JetBrains/intellij-community/commit/abbffc933ff652660146eeb4731818eeeb340ddb).
`idea.log` tab is probably new in 2016.3 as well (but in 2016.2 exceptions appear in Console tab). So you do not see exception, but it doesn't mean that you use ModalityState correctly :).

src/io/flutter/sdk/FlutterSettingsConfigurable.java
devoncarew: looks great!

pq:
Thanks for all the back and forth Alex!  As always, learned a bunch. 😄

PR #353, opened 2016-10-20
implement a missing method

devoncarew:
I see a compile error locally for a missing abstract method - adding one here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ pubspec.lock
 .pub/
 .idea/workspace.xml
 .idea/shelf
+.idea/inspectionProfiles/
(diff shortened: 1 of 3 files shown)


devoncarew:
(TBR)

alexander-doroshko:
@devoncarew This must be some misconfiguration on your end. The code is compilable with Java 1.8. Make sure your Plugin SDK is based on Java 1.8 and that you do not have local changes in project files from `.idea` folder.

alexander-doroshko:
@devoncarew Sorry, my bad. You are right.
Default implementation of SearchableConfigurable#enableSearch is available in IntelliJ IDAE 2016.3 only. Since master branch here has to be 2016.2-compatible, empty FlutterSettingsConfigurable.enableSearch() is required.

devoncarew:
OK, makes sense.

This does point out that we need to re-enable our build system on the bot to get earlier warnings of these issues :)

alexander-doroshko:
@devoncarew I see that you also added `.idea/inspectionProfiles/` to `.gitignore`. This means that every developer of the flutter-intellij plugin will have his/her own set of enabled inspections and will see yellow code in different places.

I suggest that you remove this folder from `.gitignore` and commit `.idea/inspectionProfiles/Project_Default.xml` instead. For example you may switch off inspection 'Declaration access can be weaker' that generates a lot of yellow code here (if you think that you do not need this inspection) and keeping `Project_Default.xml` in git will guarantee that inspection is disabled for everyone working on the project.

devoncarew:
@alexander-doroshko, makes sense

PR #354, opened 2016-10-20
include inspectionProfiles in committed files

devoncarew:
- update our .gitignores and committed files to include `.idea/inspectionProfiles` (see https://github.com/flutter/flutter-intellij/pull/353)

@alexander-doroshko


alexander-doroshko:
lgtm

Issue #329, opened 2016-10-19
Revisit/refactor Flutter project structure.


pq:
There's an ongoing conversation around Flutter project structure and in particular our choice to use an application library to track the Flutter SDK location.  This is a placeholder for a revisit of that decision.

/cc @alexander-doroshko

alexander-doroshko:
https://github.com/flutter/flutter-intellij/pull/355

pq:
Closing as work is under way in https://github.com/flutter/flutter-intellij/pull/361

PR #355, opened 2016-10-20
Change how path to Flutter SDK is tracked in a project model (#329)

alexander-doroshko:
Addresses #329.
With this change 'Flutter SDK' global library is not created any more to store path to the Flutter SDK. Instead getFlutterSdkRoot() / setFlutterSdkRoot()  delegate to getDartSdkRoot() / setDartSdkRoot() with path adjustment ("/bin/cache/dart-sdk").

This commit doesn't change anything in UI. The only thing changed in UX is that Flutter SDK field validation is a bit more strict now. IDE considers that path to the Flutter SDK is valid if both:
- [sdk root]/bin/flutter file exists
- [sdk root]/bin/cache/dart-sdk/lib folder exists

Previously there was no check for the presence of a Dart SDK inside Flutter SDK installation. Now you won't be able to configure Flutter SDK in the IDE until 'flutter doctor' is done.

@pq @devoncarew


alexander-doroshko:
See updated PR with required changes in WrongModuleTypeNotificationProvider: #361

Issue #341, opened 2016-10-19
Support for adding Flutter Module Type to existing projects.


pq:
From #233, @stevemessick notes:

> With the latest version you cannot continue working on a project that was created with an earlier version of the Flutter plugin. The problem is that we only recently started marking Flutter projects with a module type and old projects do not have that. The solution is to create a new project pointing to the same source directory, and overwrite existing files (.idea, *.iml). The symptom is that there is no device selector in the tool bar.

FWIW, the Dart plugin supports this in it's preference page.

<img width="789" alt="screen shot 2016-10-19 at 9 32 29 am" src="https://cloud.githubusercontent.com/assets/67586/19527949/f40d56aa-95de-11e6-9f06-6e7fcae7c463.png">

Depending on how we settle #329, we might consider doing something similar.

/cc @alexander-doroshko

alexander-doroshko:
Dart support doesn't care about module type, so nothing in common here with Flutter unless you decide to get rid of Flutter module type for some reason.
Flutter is closer to Go plugin in this regard. See [WrongModuleTypeNotificationProvider](https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/671943ad70ec7f1c4e2b59662cdb06d84f81cf1c/src/com/goide/inspections/WrongModuleTypeNotificationProvider.java).

pq:
> Dart support doesn't care about module type, so nothing in common here with Flutter unless you decide to get rid of Flutter module type for some reason.

Sorry, wasn't clear...  What I mean is that the Dart plugin, in Preferences, provides a means to add Dart support to existing projects.

Having said that, I like Go's solution.  An inspection would be perfect here actually.

Thanks! 👍

alexander-doroshko:
> Sorry, wasn't clear... What I mean is that the Dart plugin, in Preferences, provides a means to add Dart support to existing projects.

Sorry for not being clear, I meant that there can be nothing in common in the implementation of 'Enable support' feature (regardless of #329), because it is not possible to change module type on the fly, you need to reload project.

pq:
Great clarification.  Thanks!  Unfortunate that a restart is required but I still think it's very much the right approach.  In particular, I'd like to take your cue from Go and add an inspection that asks the user if they want to "Enable Flutter support", when a project is detected that

  * does not already have the Flutter module type, and
  * looks Fluttery which I take to be:
    * having a `flutter.yaml` file in the root directory

[...]

PR #356, opened 2016-10-20
Notification to detect Flutter projects w/o Flutter Module Types (#341).

pq:
Fixes: #341 

Taking inspiration from the Go plugin, this check tests for the presence of a `flutter.yaml` file and the absence of the Flutter module type and gives users the opportunity to enable Flutter support.

screen shot 2016-10-20 at 12 44 07 pm

/cc @alexander-doroshko @devoncarew @stevemessick

### DIFF
--- a/src/io/flutter/inspections/WrongModuleTypeNotificationProvider.java
+++ b/src/io/flutter/inspections/WrongModuleTypeNotificationProvider.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.inspections;
+
+
+import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.fileEditor.FileEditor;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleUtilCore;
+import com.intellij.openapi.project.DumbAware;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.project.ProjectManager;
+import com.intellij.openapi.ui.Messages;
+import com.intellij.openapi.util.Key;
+import com.intellij.openapi.util.text.StringUtil;
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.ui.EditorNotificationPanel;
+import com.intellij.ui.EditorNotifications;
+import com.intellij.util.containers.ContainerUtil;
+import com.jetbrains.lang.dart.DartFileType;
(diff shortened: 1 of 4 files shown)


src/io/flutter/inspections/WrongModuleTypeNotificationProvider.java
```diff
@@ -0,0 +1,108 @@
...
+      VirtualFile flutterYaml = baseDir.findFileByRelativePath(FLUTTER_YAML_FILE);
+      return flutterYaml != null && flutterYaml.exists();
+    }
+    return false;
+  }
+
+  @NotNull
+  private static EditorNotificationPanel createPanel(@NotNull Project project, @NotNull Module module) {
+    EditorNotificationPanel panel = new EditorNotificationPanel();
+    panel.setText("'" + module.getName() + "' is not a Flutter Module, device debugging is not fully supported.");
```
stevemessick: Shouldn't these strings be in the Flutter bundle?

pq: I started to and then saw that we're only half doing it.  I'm tempted to not bother since I doubt we'll ever internationalize.

Looking at the Dart plugin it's pretty fastidious.  Go, on the other hand, is not.

Happy to go with whatever you all prefer.

stevemessick: I agree that we probably won't internationalize. OTOH it is standard practice. And IntelliJ does all the work for you. That's why I do it. That, and @alexander-doroshko wouldn't accept a PR unless I did. :)

pq:
Updated Strings and added `HardCodedStringLiteral` inspection to our default profile to ease future externalizations. 👍

src/io/flutter/inspections/WrongModuleTypeNotificationProvider.java
```diff
@@ -0,0 +1,111 @@
...
+    panel.createActionLabel(FlutterBundle.message("change.module.type.to.flutter.and.reload.project"), () -> {
+      int message = Messages.showOkCancelDialog(project, FlutterBundle.message("updating.module.type.requires.project.reload.proceed"),
+                                                FlutterBundle.message("update.module.type"),
+                                                FlutterBundle.message("reload.project"), CommonBundle.getCancelButtonText(), null);
+      if (message == Messages.YES) {
+        module.setOption(Module.ELEMENT_TYPE, FlutterModuleType.getInstance().getId());
+        ApplicationManager.getApplication().runWriteAction(() -> {
+          FlutterSdkUtil.enableDartSupport(module);
+          FlutterSdkGlobalLibUtil.enableFlutterSdk(module);
+          project.save();
```
pq: @alexander-doroshko : I feel like this could be better; thoughts welcome!

pq:
Landing but happy to integrate changes from @alexander-doroshko (when he wakes up 😉).

PR #357, opened 2016-10-20
`findConnector` cleanup.

pq:
Use project associated with action when invoking via key events.

Alex pointed this out in a side-band conversation and this is just follow-up.

/cc @stevemessick @alexander-doroshko


pq:
😄

Issue #114, opened 2016-09-08
Create a manual testing doc.


pq:
While trying to debug the deployed plugin to track down the NPE in #109, I was hit with the fact that the plugin is not compatible with my installed EAP.  The rub is I need to test against the latest "stable" which from looking at the gradle build is: "IntelliJ IDEA 2016.2.3 / Build #IC-162.1812.17".  It would be good to add this to a testing doc.

@devoncarew : is this how you're testing?  DO you have a (now old) IDEA 2016.2.3 around that you test against?  I'm probably being thick but do you have a download link?

devoncarew:
+1 to a testing script (here's a nice sample one: https://github.com/Microsoft/vscode/wiki/Smoke-Test)

I am using IntelliJ CE to test - d/l'ed from here: https://www.jetbrains.com/idea/download/.

pq:
As per our discussion, the user journey's doc will serve double-duty as our test script.

devoncarew:
Done?

PR #361, opened 2016-10-21
Change how path to Flutter SDK is tracked in a project model (#329)

alexander-doroshko:
Addresses #329.
With this change 'Flutter SDK' global library is not created any more to store path to the Flutter SDK. Instead getFlutterSdkRoot() / setFlutterSdkRoot()  delegate to getDartSdkRoot() / setDartSdkRoot() with path adjustment ("/bin/cache/dart-sdk").

This commit doesn't change anything in UI. The only thing changed in UX is that Flutter SDK field validation is a bit more strict now. IDE considers that path to the Flutter SDK is valid if both:
- [sdk root]/bin/flutter file exists
- [sdk root]/bin/cache/dart-sdk/lib folder exists

Previously there was no check for the presence of a Dart SDK inside Flutter SDK installation. Now you won't be able to configure Flutter SDK in the IDE until 'flutter doctor' is done.

@pq @devoncarew


devoncarew:
> [sdk root]/bin/cache/dart-sdk/lib folder exists

drive by comment: this might be a problem - the flutter sdk may exist at that location, but the dart sdk might not have been downloaded yet. That happens as part of the shell script bootstrapping.

@pq for review

src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -94,37 +91,10 @@ private static String sdkRelativePathTo(@NotNull String sdkPath, @NotNull String
...
-    File bin = new File(flutterSdk, "bin");
-    if (!bin.isDirectory()) {
-      throw new ExecutionException(FlutterBundle.message("flutter.sdk.is.not.configured"));
-    }
-    if (!(new File(bin, "flutter").exists())) {
-      throw new ExecutionException(FlutterBundle.message("flutter.sdk.is.not.configured"));
-    }
-    return path;
+    final File flutterToolFile = new File(path + "/bin/flutter");
+    final File dartLibFolder = new File(path + "/bin/cache/dart-sdk/lib");
```
devoncarew: Because the dart sdk may not yet exist in a valid flutter sdk, perhaps check for `/VERSION` here?

devoncarew: (in addition to the existing `/lib/flutter` check)

alexander-doroshko: Valid dart SDK is crucial for the suggested model. Under the hood we store 'Dart SDK' global library. Without existing `/bin/cache/dart-sdk/lib` we'll fail to save Dart SDK and thus will loose knowledge about Flutter SDK path.
I think this requirement (to run `flutter doctor` in advance and have valid Dart SDK) is acceptable, as Flutter SDK is unusable without Dart SDK: no analysis, no pub, etc.

devoncarew:
@alexander-doroshko, lgtm

I'd like to find a way to allow the user to set valid flutter sdk locations that don't yet have their dart-sdk provisioned. Perhaps something to iterate towards.

alexander-doroshko:
> I'd like to find a way to allow the user to set valid flutter sdk locations that don't yet have their dart-sdk provisioned

Thanks, I'll keep it in mind when working on SDK per module configuration. There will be a lot of work and  this may appear to be be an easy-to-get side effect.

devoncarew:
Makes sense. Perhaps we can suggest some user actions to remedy a partially set up flutter sdk. Something like:

- recognize that it is a valid flutter sdk, but one that does not yet have its dart-sdk provisioned
- respond w/ an error when the user tries to set it as the flutter sdk, but say something like 'Please run the flutter command-line tool to finish set up of the flutter sdk'.

pq:
> say something like 'Please run the flutter command-line tool to finish set up of the flutter sdk'.

Could we reasonably initiate this for them?  Or is it to fiddly to interact with this external process?

@alexander-doroshko : sorry for the slow response on this.  I'm just getting back online post summit.  TBH the biggest hold-up for me has been putting together a test plan (#114 -- *sigh*) so that I can better understand how this might impact existing installations and scenarios like Devon described.  Shooting to do that today though! 👍

pq:
A few follow-up questions:

* what happens in existing installations if Dart and Flutter SDKs are different?  (It's not currently recommended but it is possible.)
* how will this story evolve when we *do* want to support projects that have Dart and Flutter Dart SDKs (for Dart Web and Flutter Dart modules)?  I know we did not reach a conclusion at the summit but it seems like if we're going to change this maybe we should fix it once?
* @devoncarew makes a good point about partially installed Flutter SDKs and messaging we should provide.  How do you feel about that?

Thanks for your patience!  _Slowly_ coming back online... 😉

alexander-doroshko:
- If user has not matching Dart and Flutter SDKs on plugin update Dart's one will win.
- That's the benefit of the suggested model. When I add support of Dart SDK per module configuration you'll get Flutter SDK per module configuration, almost for free. So users will be able to set different Dart and Flutter SDK for different modules.
- Ready to add a message in a follow-up PR.

pq:
Cool.  Thanks!

alexander-doroshko:
> partially installed Flutter SDKs and messaging we should provide

Can it wait until the next week? I'm out of time today and tomorrow is a state holiday.

pq:
> Can it wait until the next week? I'm out of time today and tomorrow is a state holiday.

Of course!  Enjoy your holiday!

pq:
Tracking bug opened: https://github.com/flutter/flutter-intellij/issues/415

PR #367, opened 2016-10-21
Widen version constraints to play nice with EAP.

pq:
Technically we’re overreaching but this is necessary for the plugin to load in the latest 2016.3 EAP.

FYI @stevemessick I think this address your loading issue of earlier today.

/cc @devoncarew


devoncarew:
lgtm. we should plan to have a matrix of bots building against the supported platform versions and products

Issue #271, opened 2016-10-11
Failure when opening flutter examples


mit-mit:
Tested with plugin version 0.0.6:

1. Open IJ, and do File>Open
2. Navigate to `/flutter/examples/hello_world` and hit OK
3. It prompted that Dart support is not enabled, enable it
4. When prompted that "'Pub get' has not been run" hit 'Get dependencies"

Fails with:

```
Working dir: /Users/mit/dev/github/flutter/examples/hello_world
/Users/mit/dev/github/flutter/bin/cache/dart-sdk/bin/pub get
Resolving dependencies...
The Flutter SDK is not available.
Depended on by:
- hello_world 
Process finished with exit code 69
``` 

Pub should be available there:
```
$ /Users/mit/dev/github/flutter/bin/cache/dart-sdk/bin/pub --version
Pub 1.20.0-dev.10.1
$
```

mit-mit:
@pq known issue in 0.0.6?

pq:
cc @devoncarew wonder if this has anything to do with our checked-in project metadata?

devoncarew:
This is something we'll need to handle. The way those pubspecs are written (https://github.com/flutter/flutter/blob/master/examples/hello_world/pubspec.yaml#L5), you can't use `pub get` or `pub upgrade` with them (which IntelliJ is doing). You can to run `flutter packages get` and `flutter packages upgrade`.

pq:
Hmmmm....   A few thoughts:

1. Should we train the Dart plugin to ignore pubspecs with `sdk: flutter` set or maybe better still, 
1. should we make the Dart Plugin pub support retargetable so that we can have it call `flutter packages get`? Or, 
1. is there a plan in place to make `pub get` "just work" (a la Dan's shim idea)?

devoncarew:
flutter_tools impl of `flutter packages get` here: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/dart/pub.dart#L54

devoncarew:
Can we replace the `pubspec.yaml` contributions from the Dart plugin with ones from the Flutter plugin?

pq:
That may be tricky as there are actions _and_ inspections we'd want to hijack/retarget.

What about the idea of making `pub get|upgrade` run "the right way" if `sdk: flutter` is specified?  That is, by setting the right environment variables?

pq:
[...]

PR #377, opened 2016-10-26
Bump minimume Dart to `162.2485`.

pq:
Ensures that we have a Dart plugin where `pub` works as expected.

See: https://github.com/flutter/flutter-intellij/issues/271

/cc @alexander-doroshko @devoncarew @stevemessick


devoncarew:
lgtm

PR #382, opened 2016-10-26
Delete user_journeys.md

mit-mit:
Deleting this, as this is now covered on flutter.io, and we won't want duplication


mit-mit:
@devoncarew please review

mit-mit:
@devoncarew please review

pq:
LGTM!

devoncarew:
This was never really meant to be user facing, but more to help us guide our developer efforts. We should reinstate this - perhaps moving it into a wiki.

PR #393, opened 2016-11-01
Start and stop the daemon based on project open and close

### LINKED ISSUE #372: Plugin active in non-flutter project
Great plugin.

I experienced a problem with this plugin though. I was not working on a flutter project, just some other dart project, but this plugin was freezing my instance somehow.

It kept moaning about an unauthorized device (which is my phone I am charging). I noticed it every 4 seconds or so in the logs;

```
2016-10-24 22:31:54,673 [3179484]   INFO - r.run.daemon.FlutterAppManager - Device 55a2e1b9 is not authorized.
You might need to check your device for an authorization dialog. 
```

I don't think the plugin needs to be active when there's no flutter enabled project?

Idea community edition 2016.2.4 and flutter intellij 0.1.0. 
Will test 0.1.3 later (I just authorized my device, then I noticed the update....)

stevemessick:
@pq @devoncarew 
Make the daemon lifecycle mirror its projects.

The daemon is running now if any Flutter project is open. It doesn't have to have focus, since it takes a few seconds to spin up the device list. The alternative is to start and stop the daemon based on the project that has focus.

Also reformatted everything in flutter.io.run and picked up a few little things.

Fixes #372

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkManager.java
+++ b/src/io/flutter/sdk/FlutterSdkManager.java
@@ -5,7 +5,6 @@
  */
 package io.flutter.sdk;
 
-
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.project.ProjectManager;
 import com.intellij.openapi.project.ProjectManagerAdapter;
@@ -30,7 +29,7 @@ public class FlutterSdkManager {
   private FlutterSdkManager() {
     listenForSdkChanges();
     // Cache initial state.
-    isFlutterConfigured = isGlobalFlutterSdkSet();
+    isFlutterConfigured = isGlobalFlutterSdkSetAndNeeded();
   }
 
   public static FlutterSdkManager getInstance() {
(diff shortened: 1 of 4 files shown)


src/io/flutter/sdk/FlutterSdkManager.java
```diff
@@ -50,7 +79,7 @@ public void removeListener(@NotNull Listener listener) {
   }
 
   private static boolean isGlobalFlutterSdkSet() {
-    return FlutterSdk.getGlobalFlutterSdk() != null;
+    return FlutterSdk.getGlobalFlutterSdk() != null && FlutterSdkUtil.isFluttering();
```
pq: Probably doesn't matter but I'd be inclined to rename the method to make it clear another test is happening.

pq: Also, at a quick glance it looks like `isFluttering` is often (if not always) checked before calling this method?  So maybe it's not necessary?  (But then you put it here for some reason! ;) )

src/io/flutter/sdk/FlutterSdkManager.java
```diff
@@ -39,6 +42,32 @@ public static FlutterSdkManager getInstance() {
...
+      }
+      @Override
+      public void projectClosed(@NotNull Project project) {
+        if (FlutterSdkUtil.isFluttering()) return;
+        checkForFlutterSdkRemoval();
+      }
+    });
+  }
+
+  private void checkForFlutterSdkAddition() {
```
pq: I wonder if this could take advantage of `FlutterSdkManager` which does a similar check?

stevemessick: I think you're asking whether I meant to replace the code I copied from the listener to the main class. I did, and forgot to.

src/io/flutter/sdk/FlutterSdkManager.java
stevemessick: Good catch. I was able to remove the prior checks but it is necessary to have that call here.

I changed the name but I'm not sure it is better.

stevemessick:
PTAL

PR #380, opened 2016-10-26
Relax version constraint to work with current Dart plugin sources

stevemessick:
@pq @devoncarew

PR #395, opened 2016-11-01
Reduce upper bound of version range

stevemessick:
@pq @devoncarew @alexander-doroshko 

Following up from Alex's comments about version ranges.

I looked for a way to revert the fix to #380 but realized typing the number and doing a normal commit would be faster.


devoncarew:
lgtm!

Issue #406, opened 2016-11-02
stdout output is duplicated when debugging


Hixie:
In the P5 usability study, all exceptions from the framework got garbled in the console output in IntelliJ. This made them borderline unusable.

The garbled output is also visible in: https://github.com/flutter/flutter/issues/6647

pq:
cc @stevemessick for thoughts

A snippet from  flutter/flutter#6647 that illustrates the log issue:

```
When the exception was thrown, this was the stack:
#2      Text.Text (package:flutter/src/widgets/text.dart:139:12)
I/flutter : #2      Text.Text (package:flutter/src/widgets/text.dart:139:12)
I/flutter : #3      _MyHomePageState.build (/data/user/0/com.yourcompany.flutterSample2/cache/flutter-sample2mfnkhu/lib/main.dart:70:20)
I/flutter : #4      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:2766:23)
I/flutter : #5      BuildableElement.rebuild (package:flutter/src/widgets/framework.dart:2667:5)
I/flutter : #6      StatefulElement.update (package:flutter/src/widgets/framework.dart:2897:5)
I/flutter : #7      Element.updateChild (package:flutter/src/widgets/framework.dart:1984:15)
I/flutter : #8      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:2778:16)
#3      _MyHomePageState.build (/data/user/0/com.yourcompany.flutterSample2/cache/flutter-sample2mfnkhu/lib/main.dart:70:20)
#4      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:2766:23)
I/flutter : #9      BuildableElement.rebuild (package:flutter/src/widgets/framework.dart:2667:5)
I/flutter : #10     ProxyElement.update (package:flutter/src/widgets/framework.dart:3015:5)
I/flutter : #11     Element.updateChild (package:flutter/src/widgets/framework.dart:1984:15)
I/flutter : #12     ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:2778:16)
I/flutter : #13     BuildableElement.rebuild (package:flutter/src/widgets/framework.dart:2667:5)
I/flutter : #14     StatelessElement.update (package:flutter/src/widgets/framework.dart:2816:5)
#5      BuildableElement.rebuild (package:flutter/src/widgets/framework.dart:2667:5)
[...]

PR #409, opened 2016-11-03
Remove stdout duplicate

devoncarew:
Remove a hard-coded setting for `isRemoteDebug()`. This was used to invoke the logic to guess path prefixes when setting breakpoints. We now get the path prefix from flutter_tools when launching apps, so we no longer need this workaround.

The workaround was the cause of the duplicated stdout text. We were getting stdout both from the flutter_tools logging code, and from the service protocol itself. We really only want to print the stdout info from the service protocol when we are remote debugging, and have no other access to stdout from the app.

fix https://github.com/flutter/flutter-intellij/issues/406; @stevemessick


third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
```diff
@@ -621,4 +607,61 @@ private static String threeSlashize(@NotNull final String uri) {
     if (uri.startsWith("file:")) return "file:///" + uri.substring("file:".length());
     return uri;
   }
+
+  private boolean baseUriInited = false;
```
stevemessick: Fields should be defined at the beginning of a class declaration, if we're being consistent with Dart. What do you think about renaming it to something that clearly indicates it is solving a problem via reflection that should be dealt with by refactoring the superclass?

devoncarew: Makes sense! I moved the field definition, renamed it a bit, and will look at opening the access level in the parent class.

Issue #403, opened 2016-11-02
Unclear how to config Debug|Profile|Release for a run config


sethladd:
Intent: create a "run config" for "debug mode" for a Flutter app.

I don't have run configs, so I tried to create one. I'm on this screen, but I'm not certain how to say "use debug mode" or "use profile mode".

![screen shot 2016-11-02 at 2 16 07 pm](https://cloud.githubusercontent.com/assets/5479/19947815/0b5150f4-a107-11e6-9f9c-d5946a1b4144.png)

I see a "checked mode", but not sure what that would do in the context of Flutter.

Should we have radio buttons for the different modes on this dialog?

pq:
I agree that checked mode is odd here.  Likely we should remove it.

As for creating a debug config, the model in IntelliJ is to have a single run configuration be runnable in debug or run modes.  A little background [here](https://www.jetbrains.com/help/idea/2016.2/run-debug-configuration.html).

pq:
Closing in favor of these follow-ups:

* @mit-mit opened #410 to track support for more run modes.
* https://github.com/flutter/flutter-intellij/pull/412 removes the `checked` option.

PR #412, opened 2016-11-03
Remove checked mode Flutter Run option (#403).

pq:
See: #403 


pq:
Or @stevemessick 👍

alexander-doroshko:
I wanted to say that you should remove it from `FlutterRunnerParameters` as well, but it appears that `FlutterRunnerParameters` is an empty subclass of `DartCommandLineRunnerParameters`. Not sure that you anywhere need `FlutterRunnerParameters` to be `DartCommandLineRunnerParameters`. Since the set of parameters is going to be completely different you may remove inheritance and add all needed stuff to `FlutterRunnerParameters` directly.

alexander-doroshko:
Not sure if `FlutterRunConfiguration` needs to extend `DartRunConfigurationBase` as well. Do you use this fact anyhow?

alexander-doroshko:
Well, looks like you do use it as `FlutterRunner` extends `DartRunner`. Revisiting flutter run/debug code requires more time than I can spend now. Never mind.

pq:
Thanks for the comments guys!

PR #420, opened 2016-11-06
Bump to `0.1.5`.

pq:
/cc @devoncarew

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -16,7 +16,7 @@ javaVersion = 1.6
 # To test against a different IDEA product.
 #intellij.alternativeIdePath =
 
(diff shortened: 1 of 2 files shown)


pq:
Just after pushing this my IDEA caught up with `0.1.4.1`.  Happy to put this on ice until our scheduled release.

devoncarew:
Let's sync tomorrow re a release -

pq:
Yeah.  Given the patch, I think we can defer in the interest of sticking to the schedule but I'm easy!

pq:
Updated `CHANGELOG`.  PTAL.

@devoncarew

PR #423, opened 2016-11-08
add a hot reload checkbox

devoncarew:
- add a hot reload checkbox to the launch config

@pq 

screen shot 2016-11-07 at 11 03 13 pm

### DIFF
--- a/src/io/flutter/run/FlutterAppState.java
+++ b/src/io/flutter/run/FlutterAppState.java
@@ -86,7 +86,7 @@ public class FlutterAppState extends DartCommandLineRunningState {
       }
     }
 
-    myApp = service.startApp(project, cwd, device.deviceId(), myMode, relativePath);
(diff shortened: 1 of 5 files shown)


eseidelGoogle:
Do we want to allow non-hot reload support?  When would I not want this option?

devoncarew:
I imagine for the same reason we want it from the command-line? Are there plans to remove `--no-hot`?

eseidelGoogle:
I'm all for simplifying our product whenever possible.  I don't know any usecase for --no-hot other than that it might be more stable than --hot at this time.

Unless there is a reason to add --no-hot to IntelliJ, I'd vote for keeping things simple as possible and just making --hot the only mode in IntelliJ.

pq:
Code LGTM.  Neutral about the value of the checkbox for the reasons that Eric mentioned.  Don't fee strongly either way though.

devoncarew:
I'll put this on hold and we can determine how it fits in with the run/debug UI refactor.

Issue #415, opened 2016-11-03
Provide feedback when a user configures a partially setup Flutter SDK


pq:
Follow-up from #361:

> Perhaps we can suggest some user actions to remedy a partially set up flutter sdk. Something like:
> * recognize that it is a valid flutter sdk, but one that does not yet have its dart-sdk provisioned
> * respond w/ an error when the user tries to set it as the flutter sdk, but say something like 'Please run the flutter command-line tool to finish set up of the flutter sdk'.

alexander-doroshko:
#428

PR #428, opened 2016-11-09
Suggest user to run 'flutter doctor' if Flutter SDK installation is not complete (#415)

alexander-doroshko:
Address #415.
@pq @devoncarew 
Ideas about better wording are welcome.
image

### DIFF
--- a/src/io/flutter/FlutterBundle.properties
+++ b/src/io/flutter/FlutterBundle.properties
@@ -18,6 +18,7 @@ debugger.trying.to.connect.vm.at.0=Connecting to Dart VM at {0}
 
 error.folder.specified.as.sdk.not.exists=Error: the folder specified as the Flutter SDK home does not exist.
 error.path.to.sdk.not.specified=Error: path to the Flutter SDK is not specified.
+error.flutter.sdk.without.dart.sdk=Error: Flutter SDK installation is not complete, please run 'flutter doctor'.
 error.sdk.not.found.in.specified.location=Error: Flutter SDK is not found in the specified location.
 
 flutter.command.exception=Flutter Command Exception: {0}
--- a/src/io/flutter/FlutterBundle.properties
+++ b/src/io/flutter/FlutterBundle.properties
@@ -18,7 +18,7 @@ debugger.trying.to.connect.vm.at.0=Connecting to Dart VM at {0}
 
 error.folder.specified.as.sdk.not.exists=Error: the folder specified as the Flutter SDK home does not exist.
 error.path.to.sdk.not.specified=Error: path to the Flutter SDK is not specified.
-error.flutter.sdk.without.dart.sdk=Error: Flutter SDK installation is not complete, please run 'flutter doctor'.
+error.flutter.sdk.without.dart.sdk=Error: Flutter SDK installation is incomplete, see <a href='https://flutter.io/setup/#get-the-flutter-sdk'>setup instructions</a>.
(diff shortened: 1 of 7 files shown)


lukechurch:
Is there some way we can launch it from within IJ?

If not, then I'd suggest at least advising that they do this from the terminal. Is there a predictable way of know where they should run it, in case the installation hasn't set the paths?

alexander-doroshko:
> Is there some way we can launch it from within IJ?

Technically possible, but a lot of work and not sure about the best UI for 'flutter doctor' output. One more modal dialog with Terminal-like functionality..? I think we are good with an error message at this point.

> If not, then I'd suggest at least advising that they do this from the terminal. Is there a predictable way of know where they should run it, in case the installation hasn't set the paths?

If paths are not set then [Flutter SDK]/bin/flutter is the correct script to run. Suggestions on better text are welcome (the shorter the better :)).

pq:
> Technically possible, but a lot of work and not sure about the best UI for 'flutter doctor' output.

As it happens we do support running `flutter doctor` (from a console hyperlink in `flutter create` output) but it's only minimal.  It prints it's output to the console and is not actionable beyond that.  In the medium run we'll want to add better support.


Alternatively, we could link out to external docs on installing flutter?

Something like:

> Error: This Flutter SDK installation is incomplete; setup instructions [here](https://flutter.io/setup/#get-the-flutter-sdk).

but word-smithed? -- paging @mit-mit !

alexander-doroshko:
How about this? Should some part of the error text be red as it was before?
image

eseidelGoogle:
I may be missing something obvious, but how would one run `flutter doctor`
if one doesn't know where the flutter sdk is (which contains `flutter
doctor`)?

pq:
Better integration of `flutter doctor` is high on my list of things to address down the road.  TBH I'm not sure what it will look like but we've clearly got a ways to go.

In the meantime,

> I may be missing something obvious, but how would one run `flutter doctor`
if one doesn't know where the flutter sdk is (which contains `flutter
doctor`)?

I'm hoping the setup instructions hyperlinked in the message will be sufficient to get folks moving.  This is the link:

https://flutter.io/setup/#get-the-flutter-sdk

Maybe there's a better one?

Or if you have ideas for a better message or action, please do chime in!

alexander-doroshko:
>  how would one run `flutter doctor` if one doesn't know where the flutter sdk is

Hyperlink or advice to run 'flutter doctor' is shown if flutter sdk path is correct but it doesn't contain dart sdk inside. If selected path is not a flutter sdk at all we still show as before: 'Error: Flutter SDK is not found in the specified location.'

pq:
LGTM!

alexander-doroshko:
I haven't added a code with the link yet, only a screenshot. :)
So, is black error text ok (was red)?
For errors without link - black text too?
Text - ok?
```
Error: path to the Flutter SDK is not specified.
Error: the folder specified as the Flutter SDK home does not exist.
Error: Flutter SDK is not found in the specified location.
Error: Flutter SDK installation is incomplete, see <a href='https://flutter.io/setup/#get-the-flutter-sdk'>setup instructions</a>.
```

pq:
> So, is black error text ok (was red)?
> For errors without link - black text too?
> Text - ok?

I'm neutral about black vs. red text.  Whatever is most consistent with how errors are presented elsewhere in IntelliJ?  If that's not easy to answer, I say pick one and we can always revisit.

Thanks!

Issue #424, opened 2016-11-08
remove uses of Project.getBaseDir() and Project.getBasePath()


devoncarew:
There are 3 usages of `Project.getBaseDir()` and 2 more usages of `Project.getBasePath()` methods in the flutter-intellij plugin source code.

JetBrains has discussed removing these methods as nobody reads javadoc and those who read still hope that these methods would work for them. The problem is that in most cases these methods do what you expect, but it should be treated just as a coincidence. Good plugin should not rely on some specific project structure.

Long story short:
- There's no such thing as 'Project root' in IntelliJ world.
- You never need Project.getBaseDir() and Project.getBasePath(), they do not do what you expect, see javadoc
- In most cases what you may need instead is:
  - Flutter project root. Effectively it is a parent folder of the flutter.yaml file
  - Module content root. Note that module may contain 0, 1, 2 or more content roots.

Make sure that you do manual or automated testing with more complex scenarios than single-module project with one content root in a module that is at the same time a Flutter project root and at the same time it is a parent of .idea folder. Always keep in mind the most complicated scenarios:
- There may be 0, 1, 2 and more modules in the project
- There may be 0, 1, 2 and more content roots in each module
- There may be 0, 1, 2 or more Flutter project in each content root, at any level in the content root folder structure
- Location of .idea folder is not anyhow connected with the location of module content roots. `Project.getBaseDir()` may be on a different physical drive than the content roots and project files

pq:
Grabbing this since I started looking at it last night. 👍

PR #430, opened 2016-11-09
Fix project basepath calculation (#424).

pq:
Fixes: #424.


src/io/flutter/util/ModuleUtil.java
```diff
@@ -0,0 +1,36 @@
...
+import com.intellij.openapi.vfs.VirtualFile;
+import io.flutter.module.FlutterModuleType;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Collection;
+
+public class ModuleUtil {
+
+  public static  @Nullable VirtualFile getProjectContentRoot(@NotNull Module module) {
```
alexander-doroshko: IntelliJ's formatter places annotations on a separate line, so we prefer to write it before `public static`, otherwise formatting looks not ideal.

src/io/flutter/run/daemon/FlutterDaemonController.java
```diff
@@ -205,7 +206,7 @@ private static GeneralCommandLine createCommandLine(Project project) throws Exec
       }
 
       flutterSdkPath = flutterSdk.getHomePath();
-      workingDir = project.getBasePath();
+      workingDir = ModuleUtil.getProjectBasePath(project);
```
alexander-doroshko: Flutter daemon is not project-specific, is it? Looks like it doesn't need any special working dir. 
Don't miss a chance to completely get rid of `ModuleUtil.getProjectBasePath` method. :) 

The presence of this method may provoke others to use it, whereas methods like this are kind of 'harmful'. It is impossible to rationally explain why it takes first content root of first Flutter module. And it's name doesn't reflect its behavior.

If I ever have to write a harmful method, I prefer to name it as such, for example `mayBeGetSomeContentRootOfSomeFlutterModule()`. Do you agree it is much better than `getProjectBasePath()`? Because it won't be used by other developers occasionally! :) And because there's no such thing as project base path.

Another example of 'harmful' method is `isFluttering()`. It was unused when it was introduced, but developers noticed and now it has several usages. Well, may be `isFluttering()` is not that bad. Especially taking into consideration that its implementation uses cached results. But every call to `ProjectManager.getOpenProjects()` and 'ModuleManager.getModules()` rings my bells.

But methods called like `ModuleUtil.getProjectBasePath()` worry me much more.

src/io/flutter/util/ModuleUtil.java
alexander-doroshko: Also, `project content root' is something unnatural. There's no such entity in IntelliJ world. If you need to keep this method, name like `getSomeContentRoot(Module module)` would better reflect its behavior and less likely be used incorrectly by other devs.
Looks like the whole `ModuleUtil` class can be removed, see other comments.

src/io/flutter/inspections/WrongModuleTypeNotificationProvider.java
```diff
@@ -45,11 +46,13 @@ public WrongModuleTypeNotificationProvider(@NotNull Project project) {
...
 
-  private static boolean hasFlutterYaml(Module module) {
-    VirtualFile baseDir = module.getProject().getBaseDir();
-    if (baseDir != null) {
+  private static boolean hasFlutterYaml(@NotNull Module module) {
+    final VirtualFile[] roots = ModuleRootManager.getInstance(module).getContentRoots();
+    for (VirtualFile baseDir : roots) {
       VirtualFile flutterYaml = baseDir.findFileByRelativePath(FLUTTER_YAML_FILE);
-      return flutterYaml != null && flutterYaml.exists();
+      if (flutterYaml != null && flutterYaml.exists()) {
```
alexander-doroshko: IMHO `findChild()` would look better here than `findFileByRelativePath()`.

src/io/flutter/console/FlutterConsoleFilter.java
```diff
@@ -61,7 +64,9 @@ public void navigate(final Project project) {
       final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
       if (sdk != null) {
         try {
-          sdk.run(FlutterSdk.Command.DOCTOR, module, project.getBaseDir(), null);
+          final VirtualFile contentRoot = ModuleUtil.getProjectContentRoot(module);
+          assert contentRoot != null;
+          sdk.run(FlutterSdk.Command.DOCTOR, module, contentRoot, null);
```
alexander-doroshko: `flutter doctor` doesn't seem to need a project-specific base dir, does it? I suggest to get rid of 'harmful' `ModuleUtil` class at all! :)

src/io/flutter/run/daemon/FlutterDaemonController.java
pq: Thanks for all the useful context and for sharing your naming conventions.

You're totally correct that the CWD is not required by the daemon.  For consistency, @devoncarew  and I agreed that we'll just set it to flutter root.  (Another step closer to removing `ModuleUtil` 😉 )

src/io/flutter/inspections/WrongModuleTypeNotificationProvider.java
pq: Sure.  Done.

src/io/flutter/console/FlutterConsoleFilter.java
pq: In this case `flutter doctor` can actually take advantage of directory context so it would be most ideal to provide it.   Agreed though that we can remove `ModuleUtil`.  Stay tuned...

pq:
Major issues addressed.  PTAL @alexander-doroshko .

pq:
Thanks!

> Curious, how flutter doctor behavior depends on CWD?

Chatting with @devoncarew  this AM he suggested that it adds a few more checks.

@danrubel maybe you can elaborate?

danrubel:
`flutter doctor` shows information about the flutter installation and related tools and should not depend upon the current working directory. Please open [an issue](https://github.com/flutter/flutter/issues/new) and assign to me if you find out that it does.

pq:
> flutter doctor shows information about the flutter installation and related tools and should not depend upon the current working directory. Please open an issue and assign to me if you find out that it does.

@devoncarew : maybe I misunderstood or things have changed...  thoughts?

devoncarew:
I _had_ thought we'd added code to do some lightweight verification if run from the root of a project.

PR #431, opened 2016-11-10
Cache version for dismissed SDK checks (#429).

### LINKED ISSUE #429: Ensure Flutter SDK bounds check is done on plugin updates.
Currently, once you dismiss the nag to update the Flutter SDK it is gone forever.  It would be safer if this dismissal state was flushed on new plugin installs.  (To catch, for example, #427.)

pq:
Fixes #429.

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -87,8 +87,6 @@
 
     <moduleType id="FLUTTER_MODULE_TYPE" implementationClass="io.flutter.module.FlutterModuleType"/>
 
-    <projectService serviceInterface="io.flutter.settings.FlutterSettings" serviceImplementation="io.flutter.settings.FlutterSettings"/>
-
     <!-- Plugin service with SmallIDE default, optionally overridden by product-specific implementations -->
     <projectService serviceInterface="io.flutter.sdk.FlutterSdkService" serviceImplementation="io.flutter.sdk.FlutterSmallIDESdkService"
                     overrides="false"/>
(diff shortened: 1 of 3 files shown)


devoncarew:
lgtm!

I think it would also be ok to warn once per session - i.e., to cache the dismissal in memory.

alexander-doroshko:
Some general comments:

1. `FlutterSettings` is a `PersistentStateComponent` managed by the IntelliJ Platform. Public fields are serialized. Private fields with public getter/setter are serialized as well. Now you have mixed style. It may lead to confusion for those who'll maintain this class in future.

2. Agree with Devon, remembering dismission per session might be ok as well. You won't have to have  `PersistentStateComponent` at all.

3. If you decide to keep `PersistentStateComponent` for dismission, consider storing it in `.idea/workspace.xml` instead of a dedicated `.idea/FlutterSettings.xml` file. All files in `.idea` folder except `.idea/workspace.xml`  should be kept in version control. Are you sure you want all team members to get dismission that one of the developers clicked? This looks more like personal settings (like recently opened files, recent searches, etc.) and `.idea/workspace.xml` is exactly for such personal things. If/when you add more project-specific shareable settings, storing them in `.idea/FlutterSettings.xml` will be the correct solution.

pq:
Thanks for the comments.  After some thinking, I'm leaning towards @devoncarew 's suggestion of caching per session.  A nice simplification!

pq:
Updated settings to be scoped by session.

PTAL?

/cc @devoncarew @alexander-doroshko

alexander-doroshko:
`FlutterSettings` class remains registered as a project service in the `plugin.xml`, but with the new implementation of its `getInstance()` you never let Platform to instantiate it as a component. Also with the new `getInstance()` implementation you've lost per-project settings: user's dismiss action in one project now affects other projects in this session. May be it's what you want, just decide:
- either `FlutterSettings` is app-level , then remove it from `plugin.xml`
- or it is project-level, then restore previous implementation of `getInstance(Project)`

pq:
Good catch!

> either FlutterSettings is app-level , then remove it from plugin.xml

This is what we want.

Updated `plugin.xml` 👍

PR #432, opened 2016-11-10
refactor the launch config code to split out from Dart base

devoncarew:
- refactor the launch config code to split out from Dart base

A larger PR, but this sets the stage for future work, including creating a second Flutter launch config to support the Bazel/internal use case. This fully separates out the Flutter and Dart launch config code.

@pq

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -23,5 +23,17 @@
       <option name="processLiterals" value="true" />
       <option name="processComments" value="true" />
     </inspection_tool>
+    <inspection_tool class="unused" enabled="false" level="WARNING" enabled_by_default="false">
+      <option name="LOCAL_VARIABLE" value="true" />
+      <option name="FIELD" value="true" />
+      <option name="METHOD" value="true" />
+      <option name="CLASS" value="true" />
+      <option name="PARAMETER" value="true" />
+      <option name="REPORT_PARAMETER_FOR_PUBLIC_METHODS" value="true" />
+      <option name="ADD_MAINS_TO_ENTRIES" value="true" />
+      <option name="ADD_APPLET_TO_ENTRIES" value="true" />
+      <option name="ADD_SERVLET_TO_ENTRIES" value="true" />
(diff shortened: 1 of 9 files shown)


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -23,5 +23,17 @@
...
+      <option name="FIELD" value="true" />
+      <option name="METHOD" value="true" />
+      <option name="CLASS" value="true" />
+      <option name="PARAMETER" value="true" />
+      <option name="REPORT_PARAMETER_FOR_PUBLIC_METHODS" value="true" />
+      <option name="ADD_MAINS_TO_ENTRIES" value="true" />
+      <option name="ADD_APPLET_TO_ENTRIES" value="true" />
+      <option name="ADD_SERVLET_TO_ENTRIES" value="true" />
+      <option name="ADD_NONJAVA_TO_ENTRIES" value="true" />
+    </inspection_tool>
```
pq: I'm guessing adding this is intentional?  (Neutral either way.)

src/io/flutter/FlutterBundle.properties
```diff
@@ -48,3 +48,8 @@ update.module.type=Update Module Type
 updating.module.type.requires.project.reload.proceed=Updating module type requires project reload. Proceed?
 
 waiting.for.flutter=Waiting for debug connection...
+
+path.to.dart.file.not.set=path to Dart source not specified
+dart.file.not.found=Dart file not found: {0}
+not.a.dart.file.or.directory=Specified file is not a Dart file or directory: {0}
+work.dir.does.not.exist=Working directory doesn''t exist: {0}
```
pq: > doesn''t

Seems like a typo?

Also, weren't we keeping these sorted alphabetically?

devoncarew: I think this is to escape the quote? This is what it looks like in the dart properties file.

.idea/inspectionProfiles/Project_Default.xml
devoncarew: Yup, intentional, to reduce the warnings in some files.

Issue #442, opened 2016-11-15
Hot reload cannot be mapped to a key


mit-mit:
This repros in 0.1.5 in a Mac: I can no longer bind hot reload to a key. I can add the key in preferences, but nothing happens when I hit the key.

devoncarew:
I can repro this -

pq:
OK, this traces back to the enablement fixes we introduced.  Looking at a fix...

PR #444, opened 2016-11-15
Fix keymap assignments (#442).

pq:
Fixes: #442.

/cc @devoncarew

### DIFF
--- a/src/io/flutter/actions/FlutterAppAction.java
+++ b/src/io/flutter/actions/FlutterAppAction.java
@@ -5,6 +5,7 @@
  */
 package io.flutter.actions;
 
+import com.intellij.openapi.actionSystem.ActionManager;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.project.DumbAwareAction;
 import com.intellij.openapi.util.Computable;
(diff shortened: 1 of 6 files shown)


src/io/flutter/actions/FlutterAppAction.java
```diff
@@ -17,21 +18,22 @@
...
       getTemplatePresentation().setEnabled(myIsApplicable.compute() && isRunning());
     }
   };
 
-  public FlutterAppAction(ObservatoryConnector connector, String text, String description, Icon icon, Computable<Boolean> isApplicable) {
+  public FlutterAppAction(ObservatoryConnector connector, String text, String description, Icon icon, Computable<Boolean> isApplicable, @NotNull String actionId) {
     super(text, description, icon);
     myConnector = connector;
     myIsApplicable = isApplicable;
+    ActionManager.getInstance().registerAction(actionId, this);
```
devoncarew: This object is registering itself globally in its ctor; I assume this ctor is only called once?

devoncarew:
lgtm! less code and sharing the same action instance

src/io/flutter/actions/FlutterAppAction.java
pq: Yep.  It should only get called once (when contributed via plugin registration).  I'll look a little deeper to verify though (and maybe add a guard to be safe).

pq: Guard added.

PR #447, opened 2016-11-15
Restart/reload action mapping fixes.

pq:
* actions may be null if not debugging
* actions get created fresh for new debugging sessions (so need to be re-registerable)

/cc @devoncarew


pq:
FWIW: what identified these issues was alternating between run and debug sessions.  Something to consider for the test script...

devoncarew:
lgtm

PR #449, opened 2016-11-15
Action registry fixes.

pq:
TL;DR; the action manager doesn't permit over-writing action mappings so we need to remove first.

/cc @devoncarew


devoncarew:
At the risk of making the smoke test script too long, is exercising this bug a candidate for the smoke test?

lgtm

pq:
> At the risk of making the smoke test script too long, is exercising this bug a candidate for the smoke test?

Adding something now.  There's actually a lot going on besides this when you run a sequence of sessions so I think it's worth capturing.

PR #451, opened 2016-11-16
Update plugin.xml

devoncarew:
Retroactively update the changelog for the last patch release; @pq.


devoncarew:
No, this was from the 0.14.0 branch - just wanted to make sure we didn't lose it.

Issue #305, opened 2016-10-14
Restore breakpoints after hot reload


stevemessick:
When the Hot Reload button is clicked the code on the selected device is updated but the breakpoints are lost for that device. If multiple devices are connected only the breakpoints on the reloaded device are lost. The breakpoint locations are not changing in this case (I just changed one character in a string to test that). This needs to be solved by the IDE since the daemon has no information about breakpoints.

stevemessick:
Oddly, doing a full app restart seems to preserve breakpoints.

devoncarew:
Yeah, we'll need some API changes to enable us to preserve (or re-set) breakpoints during a hot reload.

stevemessick:
Can't we just use the observatory, as we already do to set and unset them? I'm actually wondering why the breakpoints do not get lost on a restart. Maybe the daemon should be doing something different, and this isn't even an IDE issue. @danrubel could that be the case?

devoncarew:
The resolution from this issue: https://github.com/flutter/flutter-intellij/issues/273 will help us out. I think the steps will be:

- we send in `reloadSources(pause: true)` via the daemon protocol
- the VM reloads; we get a new event type back (`kPausePostRequest`)
- on receipt of that event, we should manually restore all previous breakpoints

eseidelGoogle:
FYI @johnmccutchan

devoncarew:
This is currently blocked on https://github.com/flutter/flutter/issues/6890 (a crash in the runtime).

/cc @johnmccutchan, @turnidge I see this crash on both the iOS simulator and with Android devices.

johnmccutchan:
https://github.com/flutter/flutter/issues/6890 is fixed.

devoncarew:
PR for the Dart plugin out here: https://github.com/JetBrains/intellij-plugins/pull/457

devoncarew:
This is in progress in the Dart plugin; moving it out of the 0.1.8 milestone.

devoncarew:
This has landed in both the Flutter and Dart plugins.

PR #456, opened 2016-11-18
restore breakpoints across a reload

devoncarew:
- some work to restore breakpoints across a hot reload; this tests for the capability in the dart plugin and if it's found, it requests that an isolate be paused after a reload
- also, inline a `StopWatch` to remove a dep on apache.commons

@pq


devoncarew:
in support of #305

devoncarew:
I'm not sure; I saw that it came from a velocity jar in CE; not sure if this is generally available in the platform or not. In any case, this avoids the dep entirely.

Issue #468, opened 2016-11-23
validate against the latest 2016.3 version of the Dart plugin


devoncarew:
- validate we compile and run with the latest version of the Dart plugin compatible with the recently released `2016.3` IntelliJ

PR #469, opened 2016-11-24
remove old jars; add the latest Dart plugin jar

devoncarew:
- remove old jars; add the latest Dart plugin jar
- fix https://github.com/flutter/flutter-intellij/issues/468

@pq @alexander-doroshko

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,10 +35,8 @@ allprojects {
   intellij {
     pluginName = 'Flutter'
 
-    // IntelliJ IDEA 2016.2.3 / IC-162.1812.17
-    // Dart 162.2233
-    version = '2016.2.3'
-    plugins = ['Dart:162.2485']
+    version = '2016.3'
+    plugins = ['Dart:163.8651']
 
     // IntelliJ IDEA 2016.1.4 (~Android Studio 2.2) / IC-145.2070.6
     // Dart 145.971.7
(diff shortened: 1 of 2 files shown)


alexander-doroshko:
- There are 3 more Dart.jar files that probably can be removed as well
- Good time to update since-build to 163.7743

alexander-doroshko:
Also need to make sure somehow that all developers update "IntelliJ IDEA Community Edition" SDK to point to 2016.3. It doesn't happen automatically even if you update your working IntelliJ IDEA, SDK may be  still pointing to another version.
The soft way is just to inform colleagues somehow.
The hard way is to rename SDK to version-specific, like 'IntelliJ Community 2016.3' and commit misc.xml. Everyone who updates gets non-compilable project with errors and is asked to setup SDK 'IntelliJ Community 2016.3' manually.

devoncarew:
> There are 2 more Dart.jar files that probably can be removed as well

I left some older ones in there deliberately. Longer term, I want to be able to build and test against our version ranges - have our CI build and test against the oldest Dart plugin version we support, and against the newest one we support (for each supported IntelliJ produect - CE/Android Studio/AppCode).

> Good time to update since-build to 163.7743

Thanks! I'll open a bug about this - we do want to be able to deploy to 2016.2 for perhaps ~4-6 more weeks.

> The soft way is just to inform colleagues somehow. The hard way is to rename SDK to version-specific, like 'IntelliJ Community 2016.3' and commit misc.xml.

Great point. It looks like we're already committing misc.xml, so perhaps we should just change the SDK name?

alexander-doroshko:
>  It looks like we're already committing misc.xml, so perhaps we should just change the SDK name?

Yup

Issue #371, opened 2016-10-24
Disable update prompts from Dart plugin


mit-mit:
As a Flutter developer, getting this is pretty confusing:
![screen shot 2016-10-24 at 11 29 57](https://cloud.githubusercontent.com/assets/13644170/19640487/38468a10-99dd-11e6-9e4f-decfd5fdccb6.png)

Given they get the Dart SDK via the Flutter SDK, we should see if we can configure the Dart plugin to not show this.

pq:
cc @alexander-doroshko 

I've been thinking about this too.  In fact, we probably want to remove the update UI from the Dart plugin preferences when configured to use Flutter as well.

pq:
Currently we call  `DartSdkUpdateOption.setDartSdkUpdateOption(DartSdkUpdateOption.DoNotCheck);` when setting the Dart SDK via Flutter SDK configuration but since that's not the only path it's not sufficient.

* we could add more calls to cover more cases, or
* we could push some (more) awareness of Flutter into the Dart plugin (added bonus we could suppress the update UI)

/cc @alexander-doroshko

alexander-doroshko:
[Done](https://github.com/JetBrains/intellij-plugins/commit/c64cf7815d4e6b1dbfa58859f4a115ceef916027)

pq:
Fantastic!

How do you feel about suppressing or disabling the update UI in `DartConfigurable`?

alexander-doroshko:
[Done](https://github.com/JetBrains/intellij-plugins/commit/8573cd83bb6826e339d60be7b9cdb97add1af2a2), thanks!

pq:
Thank you!

@mit-mit : do you want to update our blip of known issues to say this is fixed up-stream?

mit-mit:
> @mit-mit : do you want to update our blip of known issues to say this is fixed up-stream?

Just leave the label on there, then when we push an update I can query for all closed bugs with the known issues label and see what needs to be updated.

pq:
Great.  Thanks!

PR #476, opened 2016-12-01
Remove fixed issues from known issues list

mit-mit:
Removed two fixed issues

Not sure if I can remove https://github.com/flutter/flutter-intellij/issues/371 -- that was fixed but was it pushed by JetBrains?


pq:
LGTM.

As for the fix for #371, @alexander-doroshko : can you confirm this is in the latest published plugin?  (I believe it is.)

alexander-doroshko:
> As for the fix for #371, @alexander-doroshko : can you confirm this is in the latest published plugin?

Confirm

PR #484, opened 2016-12-02
Flip the bit on `final` inspections (#483).

### LINKED ISSUE #483: Reduce review churn w/ a decision on `final`.
The code-base is schizophrenic and the `final` modifier is used inconsistently.  This has come up a few times in reviews and in the interest of streamlining things moving forward I think we should make a decision.

My gut says we should just re-enable the inspection that enforces (gratuitous) use of `final` and encourage the habit of inspecting as a presubmit.  

(Aside: it'd be great to automate this in the build.)

pq:
Fixes #483 .

(This assumes we decide that yes we do want to go this way.  FWIW I'm not hugely invested in this style in particular but I lean towards thinking that biting the bullet and enforcing via an inspection presubmit may be easiest in the long run.)

/cc @devoncarew

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,6 +1,7 @@
 <component name="InspectionProjectProfileManager">
   <profile version="1.0">
     <option name="myName" value="Project Default" />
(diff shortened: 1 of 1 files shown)


pq:
FWIW this *would* trigger a bunch of warnings in the existing source base.  If we decide to land, I'll follow-up with a bulk cleanup PR.

devoncarew:
+1, lgtm

PR #485, opened 2016-12-02
Make locals and fields `final` where possible.

pq:
Follow-up from: #483, #484.

/cc @devoncarew


devoncarew:
That's a lot of final! lgtm

pq:
Yes. Finally!

PR #487, opened 2016-12-02
Suppress “CanBeFinal” for reflected fields.

pq:
These data classes get their fields filled in by reflection so the “CanBeFinal” inspection is inappropriate.

Last bit of follow-up from #483 (I promise!)

/cc @devoncarew


devoncarew:
lgtm

Issue #232, opened 2016-10-04
Device selection should be per-window, not global


stevemessick:
Open two Flutter projects (on separate code bases). Attach two devices. Select one device in one window, select the other device in the other window. Return to the first window and see that the device selection has counter-intuitively changed to what is current in the second window.

devoncarew:
Makes sense. Looking at the code, the action surfacing the pulldown is setting and getting its info from the `FlutterDaemonService`. That's a singleton for the app:

```
FlutterDaemonService service = FlutterDaemonService.getInstance();
```

We'd need to either make one daemon instance per window (probably a good idea) or have a device selection service, and have one instance of that per window.

PR #490, opened 2016-12-04
change FlutterDaemonService to a project (not app) service

devoncarew:
- change FlutterDaemonService to be a project service (from an application service)
- fix https://github.com/flutter/flutter-intellij/issues/232
- address an issue where the device selector would sometimes not show all available devices

@pq, @stevemessick


devoncarew:
> for possible gotchas in switching service scopes

This made me think; I updated the PR to deallocate the FlutterDaemonService on project disposal (not on the app one).

devoncarew:
(landing...)

PR #493, opened 2016-12-04
respond to an app stopped event

devoncarew:
- respond to an app stopped event (terminate the debug process). Previously we only knew about an app stopping if we'd requested it ourselves; this handles the case were the phone goes away or the app stops for some other reason
- only show the run button as active if we can run something (it behaves similarly to the debug button now)
- some refactoring to remove unused code and to simplify things a bit

@pq

### DIFF
--- a/src/io/flutter/run/daemon/FlutterAppManager.java
+++ b/src/io/flutter/run/daemon/FlutterAppManager.java
@@ -41,7 +41,7 @@ public class FlutterAppManager {
 
   private static final Gson GSON = new Gson();
 
-  private final Logger myLogger = Logger.getInstance("#io.flutter.run.daemon.FlutterAppManager");
+  private static final Logger LOG = Logger.getInstance(FlutterAppManager.class.getName());
   private final FlutterDaemonService myService;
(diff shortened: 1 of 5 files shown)


src/io/flutter/run/daemon/FlutterAppManager.java
```diff
@@ -41,7 +41,7 @@
 
   private static final Gson GSON = new Gson();
 
-  private final Logger myLogger = Logger.getInstance("#io.flutter.run.daemon.FlutterAppManager");
+  private static final Logger LOG = Logger.getInstance(FlutterAppManager.class.getName());
```
pq: If I'm understanding this right, we're doing away with unique loggers per instance.  I *think* that's OK but it may be confusing in the cases where we're debugging multiple apps.

(Or wait, is there an app manager per app, or?)

EDIT: probably OK either way.  Feel free to ignore!

devoncarew: I'm just changing the logging to match the pattern used in other files. If we do want to keep the logger as an instance logger (and not static), I'm happy to do that...

Issue #499, opened 2016-12-05
when debugging, the we're not able to resolve paths correctly


devoncarew:
IntelliJ opens `/main.dart` (from the VM), instead of `~/lib/main.dart`. I see this in the log:

```
2016-12-05 11:33:30,863 [ 268649]   WARN - yzer.DartAnalysisServerService - execution_mapUri(1, null, /Users/devoncarew/IdeaProjects/smoke_test_12_5/lib/main.dart) returned error INVALID_PARAMETER: Invalid parameter 'uri'. Invalid URI. 
```

devoncarew:
After a full reload, debugging resolution works as expected. I don't think this is an analysis server regression, I think this is due to switching the initial loader over to using a snapshot.

PR #500, opened 2016-12-05
fix an issue with debugging snapshot based apps

devoncarew:
- fix an issue with debugging snapshot based apps
- fix https://github.com/flutter/flutter-intellij/issues/499

@pq


third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
```diff
@@ -527,6 +527,10 @@ public XSourcePosition getSourcePosition(@NotNull final String isolateId, @NotNu
 
       if (myDASExecutionContextId != null && !isDartPatchUri(uri)) {
         if (getRemoteProjectRootUri() == null || !uri.contains(getRemoteProjectRootUri())) {
+          if (uri.startsWith("/")) {
+            // Convert a file path to a file: uri.
+            uri = new File(uri).toURI().toString();
+          }
```
pq: Oh man.  We had stuff like this in server too.  Good catch.

Issue #401, opened 2016-11-02
Update IntelliJ to use flutter run rather than flutter daemon


danrubel:
Once [flutter run --machine](https://github.com/flutter/flutter/pull/6645) lands, we should
* [support launching bazel targets](https://github.com/flutter/flutter-intellij/issues/392)
* update other launch targets to use `flutter run --machine` rather than `flutter daemon`

pq:
When we do this should we bump the version of flutter tools and up our minimum requirement?  Or should we support the old and the new worlds?

@danrubel @devoncarew

danrubel:
I recommend bumping the flutter tools version and requiring that version in IJ.

pq:
Sounds good.  Can we bump it when you have the (unrelated) new events work landed?  Or should we do it sooner?

danrubel:
I'll address https://github.com/flutter/flutter/issues/6665 first, determine if there is anything else we need for this, then bump the number.

PR #505, opened 2016-12-07
Flutter run machine

devoncarew:
- update IntelliJ to use `flutter run --machine` (instead of flutter daemon `app.startApp`) when starting an app (fix https://github.com/flutter/flutter-intellij/issues/401). We still use the daemon protocol, but use it in the one-shot manner provided by flutter run --machine

@pq, @danrubel

### DIFF
--- a/src/io/flutter/run/daemon/DaemonJsonInputFilterProvider.java
+++ b/src/io/flutter/run/daemon/DaemonJsonInputFilterProvider.java
@@ -31,7 +31,12 @@ public class DaemonJsonInputFilterProvider implements ConsoleInputFilterProvider
     public List<Pair<String, ConsoleViewContentType>> applyFilter(String text, ConsoleViewContentType contentType) {
       if (!VERBOSE) {
         final String trimmed = text.trim();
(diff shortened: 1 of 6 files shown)


src/io/flutter/run/daemon/DaemonJsonInputFilterProvider.java
```diff
@@ -31,9 +31,14 @@
     public List<Pair<String, ConsoleViewContentType>> applyFilter(String text, ConsoleViewContentType contentType) {
       if (!VERBOSE) {
```
danrubel: Is the VERBOSE final ever set true? Should it be removed and this code simplified?

devoncarew: It's just for use during development - when you're working on the launch code itself.

devoncarew:
Re: the version check, we did bump to needing `0.0.3` (as part of the --machine change).

devoncarew:
Thanks for the reviews!

Issue #466, opened 2016-11-22
IntelliJ opens example project without Flutter module


jsiedentop:
When I open one of the flutter examples in IntelliJ (tested with hello_world, hello_service and gallery), the IDE loads the Dart module but no Flutter module. 

In the _Modul Settings_ (_context menu on the root node in the project tree -> Open Module Settings_) Is one problem listed: "_Project hello_world: project SDK is not defined_". I did't find a way to activate the Flutter Module in the IDE. 

1. There should be a a way, to activate Flutter Support in the IDE
2. The IDE should recognize, that the Flutter module needs to be loaded. Maybe update the Flutter plugin or add a _flutter.yaml_ file in the example projects.

**Informations to reproduce the issue and a workaround to activate the Flutter support:**
1. Get the hello_world example:
> https://github.com/flutter/flutter/tree/b9bff6a912c62c9dae5eddbaef80927655b769d7/examples/hello_world

2. Open the example project in IntelliJ

3. Click on "_Get dependencies_" in the "_'Pub get' has not been run_" message

4. The following error appear:
```
Working dir: /Users/jonas/dart/hello_world
/Users/jonas/Library/flutter/bin/cache/dart-sdk/bin/pub get
Resolving dependencies...
The Flutter SDK is not available.
Depended on by:
- hello_services 
Process finished with exit code 69
```

5. Close IntelliJ and create the file _flutter.yaml_ with this content:
```
name: hello_world
uses-material-design: true
```

6. Restart IntelliJ with the hello_world example.

7. The message "_'hello_world' is not a Flutter module, devide debugging is not fully supported._" appear. 

8. Click on _Change module type to Flutter and reload project_

9. Run the pub action _Get dependencies_ (Should work fine without error).

**Flutter doctor output:**

```
$ flutter doctor

[✓] Flutter (on M.ac OS, channel master)

    • Flutter at /Users/jonas/Library/flutter

    • Framework revision a0c567f751 (5 hours ago), 2016-11-21 20:17:41

    • Engine revision cea5ed2b9b

    • Tools Dart version 1.21.0-dev.6.0



[✓] Android toolchain - develop for Android devices (Android SDK 25.0.0)

    • Android SDK at /Users/jonas/Library/Android/sdk

    • Platform android-25, build-tools 25.0.0

    • Java(TM) SE Runtime Environment (build 1.8.0_111-b14)



[✓] iOS toolchain - develop for iOS devices (Xcode 8.1)

[...]

PR #511, opened 2016-12-08
Improve test for Flutteriness in module type check (#466).

pq:
Partially address: #466

With this fix Flutter projects that lack a `flutter.yaml` file will get detected by their flutter pubpspec dependencies.

/cc @devoncarew 

Note the hacky regexp.  Happy to refine.

### DIFF
--- a/src/io/flutter/inspections/WrongModuleTypeNotificationProvider.java
+++ b/src/io/flutter/inspections/WrongModuleTypeNotificationProvider.java
@@ -44,7 +44,7 @@ public class WrongModuleTypeNotificationProvider extends EditorNotifications.Pro
   private static final String FLUTTER_YAML_FILE = "flutter.yaml"; //NON-NLS
   private static final String PUBSPEC_YAML_FILE = "pubspec.yaml"; //NON-NLS
 
-  private static final Pattern FLUTTER_SDK_DEP = Pattern.compile(".*dependencies:\\s*flutter:\\s*sdk:\\s*flutter"); //NON-NLS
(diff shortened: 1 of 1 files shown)


devoncarew:
Let's iterate on the regex a bit, and makes sure it doesn't throw out valid flutter pubspec files.

src/io/flutter/inspections/WrongModuleTypeNotificationProvider.java
```diff
@@ -33,27 +33,52 @@
...
+import java.util.regex.Pattern;
 
 
 public class WrongModuleTypeNotificationProvider extends EditorNotifications.Provider<EditorNotificationPanel> implements DumbAware {
   private static final Key<EditorNotificationPanel> KEY = Key.create("Wrong module type");
   private static final String DONT_ASK_TO_CHANGE_MODULE_TYPE_KEY = "do.not.ask.to.change.module.type"; //NON-NLS
   private static final String FLUTTER_YAML_FILE = "flutter.yaml"; //NON-NLS
+  private static final String PUBSPEC_YAML_FILE = "pubspec.yaml"; //NON-NLS
+
+  private static final Pattern FLUTTER_SDK_DEP = Pattern.compile(".*dependencies:\\s*flutter:\\s*sdk:\\s*flutter"); //NON-NLS
```
pq: I was thinking about this on my way in actually.  I'm tempted to keep it super simple and just look for `\\s*sdk:\\s*flutter` since that's the salient bit and I can't imagine many false positives.

devoncarew: 👍

devoncarew:
lgtm w/ the regex update

PR #512, opened 2016-12-09
refresh the icons for hot reload and restart

devoncarew:
- redo the icon for app restart, so that it borrows the lightening bolt from the hot reload button
- create a debug version of the hot reload button
- switch our current hot reload action to use the combined run+debug variant of the reload icon. This should make it less confusing when you hit debug to start an app, but get the lightening bolt run icon

In the future - when we have reload and restart icons in the toolbar - the reload icon can use either reload_debug.png or reload_run.png depending on how the app was started.

screen shot 2016-12-08 at 9 53 55 pm

screen shot 2016-12-08 at 9 54 05 pm

screen shot 2016-12-08 at 9 54 10 pm

@pq, @mit-mit


mit-mit:
These look great!

PR #522, opened 2016-12-12
add a new 'Profile' launch type

devoncarew:
This adds a new `Profile` launch type (IntelliJ `Executor`) - this shows up as a new top level button on the IntelliJ toolbar, next to the Run and Debug buttons. This launches a flutter app in `--profile` mode - useful for getting a sense of the release mode runtime performance of an app, but with some observatory profiling info still available.

@pq, @mit-mit we should talk a bit about this, to make sure that this is how we want to expose this functionality and esp. whether we do want to add another top-level launch type button.

screen shot 2016-12-11 at 12 40 43 pm

In that screenshot, `flutter run --profile` is the third button in.


mit-mit:
Is seems a little specialised to have in the toolbar. Is there no way to add it so that it's in the Run menu but not in the toolbar?

devoncarew:
resolution: won't land today; useful info for later

PR #523, opened 2016-12-12
have the flutter upgrade action better locate pubspec files

devoncarew:
- have the flutter upgrade action better locate pubspec files; previously, it would only work when invoked on `flutter.yaml` directly. Now it looks in the Module for the most likely pubspec to act on.
- some refactoring of our use of dart+flutter constants

@pq


pq:
LGTM!

Issue #527, opened 2016-12-12
Make it obvious when Flutter commands finish.


pq:
Following up from #525, we should make sure all commands display something on finish.

pq:
(Got this.)

devoncarew:
FYI, the Dart plugin suffixes processes with `Process finished with exit code 0`.

pq:
Cool.  I'm tempted to add this for all commands.  (Unless you think otherwise?)

devoncarew:
Yeah, making it uniform sounds great.

PR #529, opened 2016-12-12
Add console output on Flutter command completion (#527).

pq:
Fixes: #527.

* adds process termination console output
* fixes pubspec finding (when `CommonDataKeys.PSI_FILE` or `LangDataKeys.MODULE` cache-miss)

screen shot 2016-12-12 at 12 25 57 pm

/cc @devoncarew

### DIFF
--- a/src/io/flutter/actions/FlutterSdkAction.java
+++ b/src/io/flutter/actions/FlutterSdkAction.java
@@ -26,6 +26,9 @@ import io.flutter.sdk.FlutterSdk;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Base class for Flutter commands.
  */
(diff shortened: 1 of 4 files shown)


src/io/flutter/actions/FlutterSdkAction.java
```diff
@@ -49,15 +50,25 @@
...
+      return findPubspecFrom(ModuleManager.getInstance(project).getModules());
     }
     final VirtualFile file = psiFile.getVirtualFile();
     final VirtualFile contentRoot = ProjectRootManager.getInstance(project).getFileIndex().getContentRootForFile(file);
     return contentRoot == null ? null : contentRoot.findChild(FlutterConstants.PUBSPEC_YAML);
   }
 
+  private static VirtualFile findPubspecFrom(@NotNull Module[] modules) {
+    for (Module module : modules) {
+      final VirtualFile file = findPubspecFrom(module);
```
devoncarew: This works well if there's one module in a project (or even, one dart module in a project). What should be the behavior if there's more than one matching module? Return the first, or fail (return null)?

devoncarew:
lgtm; one question about how best to support multi-module projects

src/io/flutter/actions/FlutterSdkAction.java
pq: My thinking was to return the first pubspec we find on the assumption that there should not likely be more than one (or that if there were it wouldn't matter).  Not sure if that's the right solution though TBH.

devoncarew: If there's only one match, returning that makes sense. If there's more than one, we may be better off failing here than running the command for a different module than expected.

pq: Works for me.  Added a check and failure.

pq:
(Check and error added.)

Issue #516, opened 2016-12-09
consider adding a flutter context menu on flutter.yaml files


devoncarew:
We have some Pub actions on the context menu for pubspec.yaml files. This would be similar, and expose things that acted on the project, like packages get, upgrade, ...

pq:
(Looking at this now.)

pq:
AIs:

* [ ] move actions to butter bar
* [ ] add "Flutter" context menu to `flutter.yaml` files

pq:
Closing in favor of tracking the new file context menu in #537.

PR #532, opened 2016-12-12
Context menu for `flutter.yaml` files (#516)

pq:
Fixes: #516.

screen shot 2016-12-12 at 3 37 45 pm

/cc @devoncarew

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -126,6 +126,15 @@
               text="Submit Feedback…"
               description="Provie feedback for the Flutter plugin"/>
     </group>
+
+    <!-- editor actions -->
+    <group>
+      <add-to-group group-id="EditorPopupMenu" anchor="first"/>
+      <reference ref="flutter.packages.get"/>
+      <reference ref="flutter.packages.upgrade"/>
+      <separator/>
+    </group>
+
   </actions>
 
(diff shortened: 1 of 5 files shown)


mit-mit:
Why a context menu as opposed to adding two more actions next to upgrade and doctor?

pq:
> Why a context menu as opposed to adding two more actions next to upgrade and doctor?

My thinking is that the toolbar would be too crowded and that these actions are the less frequently needed (making them good candidates for a context menu).  I'm open to discussion though.

mit-mit:
I think they should just as common, if not more common? At least once things stabilize a bit, I would expected packages to rev much more frequently than the core framework.

pq:
Updated butterbar:

screen shot 2016-12-13 at 10 30 35 am

Label-smithing welcome.

@devoncarew @mit-mit

mit-mit:
Looks great, labels good as they are!

devoncarew:
Can you update to only have a JSeparator between the "project" related actions and the "global" related ones? Flutter upgrade acts on both, so it's an odd duck, but I'd put the splitter after it.

pq:
> Can you update to only have a JSeparator between the "project" related actions and the "global" related ones? Flutter upgrade acts on both, so it's an odd duck, but I'd put the splitter after it.

👍

devoncarew:
👍 lgtm

Issue #392, opened 2016-10-31
create an IntelliJ launch configuration that supports launching bazel run targets


devoncarew:
This will support the internal workflow.

devoncarew:
This will depend on https://github.com/flutter/flutter/issues/6593.

devoncarew:
Moving out of 0.1.8 - work will continue on this, but it doesn't gate the release.

devoncarew:
Marking as closed; we can re-ope more specific issues as they arise.

PR #534, opened 2016-12-13
add a preference to show and hide the device selector

### LINKED ISSUE #513: device selector sometimes doesn't show devices until you click on it
I've seen this twice; you start IntelliJ with a device already connected; the device selector shows but w/o any devices listed. When you first click on the selector, it then populates with devices.

### LINKED ISSUE #533: the stop button can take a while to return after hitting it

devoncarew:
- add a preference to show and hide the device selector
- when launching, only pass the device ID in if the selector is showing
- launches can be enabled w/ no devices selected if the selector is hidden
- added a project section in the flutter preference page
- project settings get written to the `FlutterSettings` class (an existing class with that name renamed to `FlutterUIConfig`)
- fix an issue with the stop button hanging the UI for several seconds (fix #533)
- fix an issue with the device selector not populating until the user clicks on it (fix #513)

screen shot 2016-12-09 at 12 45 32 pm

@pq 

(partial work towards https://github.com/flutter/flutter-intellij/issues/392)


src/io/flutter/run/daemon/FlutterAppManager.java
```diff
@@ -178,7 +182,7 @@ void stopApp(@NotNull FlutterApp app) {
       // streams are closed which immediately terminates the process.
       CompletableFuture
         .completedFuture(sendCommand(app.getController(), cmd))
-        .thenApply(this::waitForResponse)
+        .thenApply((Method method) -> waitForResponse(method, 300))
```
pq: Out of curiosity  how'd you settle on 300 here?

devoncarew: 300ms - it seems like a good max time to wait, in terms of the stop button being responsive.

PR #550, opened 2016-12-15
Bazel run

devoncarew:
- introduce a new bazel launch configuration
- refactor the existing launch code so that the bazel launch classes either subclass it, or share a common (`Base`) parent class
- remove unused code from the older Dart launch config implementation
- having a device selected is optional for a bazel launch config (though if one is selected, it will use it)
- devices are still necessary for regular flutter launches

screen shot 2016-12-14 at 8 49 22 pm

screen shot 2016-12-14 at 8 49 50 pm

(not yet ready for review)


devoncarew:
@pq, this is ready for review. The implementation is behind a flag (the launch config is commented out in the plugin.xml file) pending some supporting work on the bazel / shell script side.

src/io/flutter/run/FlutterAppState.java
```diff
@@ -79,16 +83,15 @@ protected ProcessHandler startProcess() throws ExecutionException {
...
+    if (service.isActive()) {
       final Collection<ConnectedDevice> devices = service.getConnectedDevices();
       if (devices.isEmpty()) {
         throw new ExecutionException("No connected device");
       }
       device = service.getSelectedDevice();
     }
 
-    final FlutterRunnerParameters parameters = ((FlutterRunConfiguration)getEnvironment().getRunProfile()).getRunnerParameters().clone();
+    final FlutterRunnerParameters parameters = ((FlutterRunConfigurationBase)getEnvironment().getRunProfile()).getRunnerParameters().clone();
```
pq: Actually, looking at `FlutterAppStateBase` this looks safe.  Does it make sense to override `getEnvironment()` and add a `@NotNull` annotation to make this obvious?

src/io/flutter/run/FlutterRunner.java
```diff
@@ -52,15 +52,17 @@ public boolean canRun(final @NotNull String executorId, final @NotNull RunProfil
...
-      if (service == null || !service.hasSelectedDevice()) {
-        return false;
-      }
+    if (FlutterSdk.getFlutterSdk(project) == null) {
+      return false;
+    }
+
+    final FlutterDaemonService service = FlutterDaemonService.getInstance(project);
+    //noinspection SimplifiableIfStatement
+    if (!service.isActive() || !service.hasSelectedDevice()) {
```
pq: I think we need a null check here?:

    if (service == null || !service.isActive() || !service.hasSelectedDevice())

src/io/flutter/run/FlutterRunnerParameters.java
```diff
@@ -119,6 +97,25 @@ public VirtualFile getDartFile() throws RuntimeConfigurationError {
   }
 
   @NotNull
+  VirtualFile getBestContextFile(Project project) throws RuntimeConfigurationError {
```
pq: Do we need this project param?

src/io/flutter/run/bazel/FlutterBazelAppState.java
```diff
@@ -0,0 +1,68 @@
...
+        device = service.getSelectedDevice();
+      }
+    }
+
+    final FlutterRunnerParameters parameters =
+      ((FlutterRunConfigurationBase)getEnvironment().getRunProfile()).getRunnerParameters().clone();
+    final String cwd = parameters.computeProcessWorkingDirectory(project);
+
+    final String launchingScript = parameters.getLaunchingScript();
+    assert (launchingScript != null);
```
pq: Nit: lose the parens?  (Or add them to the working dir check above.)

src/io/flutter/run/bazel/FlutterBazelRunConfigurationType.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    @NotNull
+    public RunConfiguration createConfiguration(String name, RunConfiguration template) {
+      // Override the default name which is always "Unnamed".
+      return super.createConfiguration(template.getProject().getName(), template);
+    }
+
+    @Override
+    public boolean isApplicable(@NotNull Project project) {
+      return FileTypeIndex.containsFileOfType(DartFileType.INSTANCE, GlobalSearchScope.projectScope(project));
```
pq: Do we also want to check for Flutter module type?  (Looks like we'd want to update `FlutterRunConfigurationType` similarly.)

src/io/flutter/run/daemon/FlutterDaemonService.java
```diff
@@ -68,14 +69,16 @@ public void processTerminated(ProcessHandler handler, FlutterDaemonController co
     void deviceRemoved(ConnectedDevice device);
   }
 
-  @Nullable
+  @NotNull
   public static FlutterDaemonService getInstance(@NotNull Project project) {
```
pq: Ah!  I see why you weren't checking this before.

It looks like you're you treating the case where `com.intellij.openapi.components.ServiceManager#doGetService` returns `null` as a kind of assert which works for me!

src/io/flutter/sdk/FlutterSdkManager.java
```diff
@@ -41,7 +41,9 @@ public static FlutterSdkManager getInstance() {
   }
 
   private void listenForSdkChanges() {
+    // TODO(devoncarew): I don't believe this always fires when it should.
```
pq: Let's chat about this offline.  I'm curious what makes you think this.

src/io/flutter/run/FlutterAppState.java
```diff
@@ -79,16 +83,15 @@ protected ProcessHandler startProcess() throws ExecutionException {
     ConnectedDevice device = null;
 
     // Only pass the current device in if we are showing the device selector.
-    final FlutterSettings settings = FlutterSettings.getInstance(project);
-    if (settings.isShowDevices()) {
+    if (service.isActive()) {
```
devoncarew: I changed it to not be null (and added a `@NotNull` annotation to its definition).

src/io/flutter/run/FlutterAppState.java
devoncarew: Overridden and added the annotation. IntelliJ didn't seem to complain w/o `@NotNull` from the superclass however...

src/io/flutter/run/FlutterRunner.java
devoncarew: `FlutterDaemonService.getInstance(project)` has a non-null guarantee.

src/io/flutter/run/FlutterRunnerParameters.java
devoncarew: Hmm, no, looks like it's not used.

src/io/flutter/run/bazel/FlutterBazelAppState.java
devoncarew: Lost! A habit from Dart I guess -

src/io/flutter/run/bazel/FlutterBazelRunConfigurationType.java
devoncarew: Sounds good for the regular Flutter launch type. For this one, I'm inclined to leave it slightly looser. This launch can work w/o a flutter sdk configured; perhaps also it's good for it to work in a project that isn't strictly a flutter one. Thoughts?

pq: SGTM!

src/io/flutter/run/daemon/FlutterDaemonService.java
devoncarew: Yeah, I think that case is more of a runtime exception - we couldn't construct the class for some reason.

PR #557, opened 2016-12-15
Bump to `0.1.7`.

pq:
@devoncarew

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -18,6 +18,18 @@
 
   <change-notes>
     <![CDATA[
+
+0.1.7:
+<ul>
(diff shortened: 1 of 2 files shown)


resources/META-INF/plugin.xml
```diff
@@ -18,6 +18,18 @@
 
   <change-notes>
     <![CDATA[
+
+0.1.7:
```
devoncarew: should we bump the versions in the various metadata files as well?

devoncarew:
lgtm

resources/META-INF/plugin.xml
pq: Yep.  Done. 👍

Issue #556, opened 2016-12-15
Device selector disappears after project close and open


pq:
Haven't tried to debug, but to repro:

1. open a project `foo`
2. close the project
3. open a project `bar`

No device selector.

I confirmed that the preference to see it is set:

<img width="855" alt="screen shot 2016-12-15 at 11 27 57 am" src="https://cloud.githubusercontent.com/assets/67586/21238973/eaca77d2-c2b9-11e6-9210-d3a9237f8041.png">

cc @devoncarew

pq:
I then created another project and still no selector.

devoncarew:
Yeah, likely a bug; let me look.

pq:
So I think I have an idea what's happening here.  In `DeviceSelectorAction.update` we register a listener that updates state and are careful not to register it multiple times.   In this case though I think we need to re-register.  

As a quick test,

remove this check:

    if (!isListening) { 

and life is good again (albeit with gratuitous updates).

pq:
One more thought is that it might be good for us to consolidate our project-scoped listeners in the `FlutterDaemonService` as it's doing some disposing based on project lifecycle.  Specifically, see the constructor where this registration happens:

    Disposer.register(project, this::stopControllers);
    Disposer.register(project, this::stopListeningForSdkChanges);

devoncarew:
Cool, sounds like good cleanup.

devoncarew:
OK, it looks like the action has a lifetime of the application, not the project.

PR #560, opened 2016-12-16
actions have a application lifetime

devoncarew:
- update the DeviceSelectorAction class - actions live for the life of the app
- fix https://github.com/flutter/flutter-intellij/issues/556

@pq


src/io/flutter/actions/DeviceSelectorAction.java
```diff
@@ -71,8 +69,9 @@ public void update(final AnActionEvent e) {
 
     final FlutterDaemonService service = FlutterDaemonService.getInstance(project);
 
-    if (!isListening) {
-      isListening = true;
+    if (!knownProjects.contains(project)) {
+      knownProjects.add(project);
+      Disposer.register(project, () -> knownProjects.remove(project));
```
pq: Down the road it'd be great to migrate this logic into `FlutterDaemonService` where we've already got disposers registered on projects but that makes sense probably for a bigger cleanup.

PR #561, opened 2016-12-16
fix an npe on project creation

devoncarew:
Fix an NPE found during smoke testing; an NPE can be thrown on project creation. This is racy, so likely wasn't a regression.

@pq


devoncarew:
Yeah, the NPE was down inside IntelliJ code - the window hadn't been created yet.

PR #567, opened 2016-12-16
add a module config editor page

### LINKED ISSUE #317: implement ModuleConfigurationEditorProvider on our custom module type
Custom module type gives great control on module level configuration in the Project Structure dialog (IntelliJ IDEA only). I mean that you can add any tabs (like for Java you have Sources, Paths and Dependencies). Currently you have none, so user even can't see/configure content roots, and excluded folders. Consider implementing ModuleConfigurationEditorProvider that would return something inherited from CommonContentEntriesEditor.

devoncarew:
- add a module config editor page (fix #317)

Instead of being blank, the module page in the structure dialog now shows:

screen shot 2016-12-16 at 3 17 14 pm

@pq, @alexander-doroshko


alexander-doroshko:
Super!

PR #568, opened 2016-12-19
handle another ios crash format in console folding

devoncarew:


devoncarew:
@pq

src/io/flutter/console/FlutterConsoleFolding.java
```diff
@@ -28,7 +28,14 @@
   //     [x86_64] libnetcore-856.20.4
   // 0   libsystem_network.dylib             0x0000000111918682 __nw_create_backtrace_string + 123
   // 1   libnetwork.dylib                    0x0000000111ab2932 nw_socket_add_input_handler + 3100
-  private static final String iosCrashPrefix = "\t        [";
+  private static final String iosCrashFormat1 = "\t        [";
```
pq: Should we use a regexp here?  The space count seems brittle?

PR #569, opened 2016-12-19
Debugger code refactor

devoncarew:
- code motion and refactoring to simplify the debugger codebase
- rename FlutterAppManager to FlutterDaemonControllerHelper

This is a largely semantic free refactoring of the debugger codebase to improve maintainability. There's now:

- one daemon service per IntelliJ window
- one controller active for device detection
- one controller per launch, with a lifetime of that launch
- the FlutterDaemonControllerHelper, which is wholly owned by and assists the daemon controller

@pq

### DIFF
--- a/src/io/flutter/utils/StopWatch.java
+++ b/src/io/flutter/utils/StopWatch.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.utils;
+
+public class StopWatch {
+  private long startTime;
+  private long stopTime;
+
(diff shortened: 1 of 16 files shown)


src/io/flutter/utils/StopWatch.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.utils;
+
+public class StopWatch {
```
pq: Aside: I can't remember but is this essentially a pared down version of the apache one?  (I'm guessing we want to avoid an additional dep?)

devoncarew: Yup; I can make a note of that.

PR #570, opened 2016-12-19
some work towards support of small IDEs

devoncarew:
- some work towards support of small IDEs

This adds a new project wizard contribution for small IDEs. It also does some cleanup where we display a combobox of SDKs - we now populate and maintain the SDK history (list of previously selected SDKs). This PR does not change our supported IntelliJ platforms - we want to set up some testing on the bots per platform when we do add support.

@pq


src/io/flutter/actions/DeviceSelectorAction.java
```diff
@@ -109,7 +109,7 @@ public void deviceRemoved(ConnectedDevice device) {
   }
 
   private void updateVisibility(final Project project, final Presentation presentation) {
-    ApplicationManager.getApplication().invokeAndWait(() -> {
+    FlutterUtils.invokeAndWait(() -> {
```
pq: Good idea! 👍

resources/META-INF/plugin.xml
```diff
@@ -172,8 +174,14 @@
     <!-- configurationType implementation="io.flutter.run.bazel.FlutterBazelRunConfigurationType"/>
     <programRunner implementation="io.flutter.run.bazel.FlutterBazelRunner"/ -->
 
+    <!-- IDEA only -->
     <moduleType id="FLUTTER_MODULE_TYPE" implementationClass="io.flutter.module.FlutterModuleType"/>
```
pq: Maybe this should move to `idea-contribs.xml`?

src/io/flutter/module/FlutterSmallIDEProjectGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+import io.flutter.FlutterErrors;
+import org.jetbrains.annotations.NotNull;
+
+import javax.swing.*;
+
+public class FlutterSmallIDEProjectGenerator extends WebProjectTemplate<String> {
+  @NotNull
+  @Override
+  public String getName() {
+    return "Flutter";
```
pq: Should this String be externalized?

src/io/flutter/sdk/FlutterSdkManager.java
```diff
@@ -20,26 +21,20 @@
  * Monitors the application library table to notify clients when Flutter SDK configuration changes.
  */
 public class FlutterSdkManager {
-
-  private static FlutterSdkManager INSTANCE;
   private final EventDispatcher<Listener> myDispatcher = EventDispatcher.create(Listener.class);
   private final LibraryTableListener myLibraryTableListener = new LibraryTableListener();
   private boolean isFlutterConfigured;
 
+  public static FlutterSdkManager getInstance() {
```
pq: Are we expecting this to be non-null at call sites (or do we want to be able to?)  If so, maybe add `@NotNull`?

src/io/flutter/actions/DeviceSelectorAction.java
devoncarew: This change lets us run on the current Android Studio (the dev channel build).

src/io/flutter/sdk/FlutterSdkManager.java
devoncarew: Yup; added.

resources/META-INF/plugin.xml
devoncarew: sgtm, will move the `FlutterModuleConfigurationEditorProvider`. Off-hand, do you know why the four `editorNotificationProvider` are in the IDEA-only file?

PR #573, opened 2016-12-22
add a test configuration for the plugin

devoncarew:
- add a test configuration for the plugin
- commit some additional jars for the Dart plugin (necessary for running unit tests)
- remove an older version of the Dart plugin (used for older Android Studios)

@pq


flutter-intellij.iml
```diff
@@ -28,5 +28,23 @@
...
+          <root url="jar://$MODULE_DIR$/lib/dart-plugin/163.10154.19/resources_en.jar!/" />
+        </CLASSES>
+        <JAVADOC />
+        <SOURCES />
+      </library>
+    </orderEntry>
+    <orderEntry type="module-library" scope="PROVIDED">
+      <library>
+        <CLASSES>
+          <root url="jar://$MODULE_DIR$/lib/dart-plugin/163.10154.19/snakeyaml-1.13.jar!/" />
```
pq: I wonder if we couldn't move these dependent jars our of the specific plugin directories so that we don't need to copy or re-up them when we add new plugin versions.

For example,

    lib/dart-plugin/
      deps/
        snakeyaml-1.13.jar
      163.10154.19/
        Dart.jar  

or something similar?

.idea/runConfigurations/Flutter_Plugin__tests_.xml
```diff
@@ -0,0 +1,29 @@
+<component name="ProjectRunConfigurationManager">
```
pq: Curious about the double underscore in the filename.  Is this a convention?  Cool either way! 👍

eseidelGoogle:
You might want to look into an alternative storage mechanism for your jars.  Git is particularly bad at large binary files.  They all end up unique, and all get pulled into all clones for all time.  So every time you add a 10mb binary, you're making every checkout for all time 10mb bigger iirc. :/

flutter-intellij.iml
devoncarew: Sounds good! Will reduce he jar file churn.

.idea/runConfigurations/Flutter_Plugin__tests_.xml
devoncarew: I think IntelliJ does that to escape away any chars that might be dodgy on the file system. In this case, the name of the launch config in the UI is `'Flutter Plugin (tests'`.

flutter-intellij.iml
devoncarew: done!

devoncarew:
Re: the committed jars - two are small; the Dart plugin one is ~4MB, and gets updated perhaps monthly. I don't think it's an immediate issue in terms of bloating the repo, but is good to know. It would definitely be possible to provision as part of a build process. We have some build system and CI plans in Q1, so redoing some of the jar provisioning could be factored into that.

PR #579, opened 2016-12-27
switch our CI build to use ant

devoncarew:
- switch to using ant (remove the gradle files)
- the build downloads the necessary artifacts, compiles the plugin (with the jetbrains forms and non-null aware version of javac), and packages the artifacts
- we can now specify the platform to build against (IntelliJ CE / Android Studio / Webstorm) and the version
- shard the travis build to run two builds per commit, one testing against the earliest supported IntelliJ CE, and one testing against the latest supported one

@pq 

Some follow-up work for this will be to move to downloading the larger files we need to compile and build against to google storage.


building.md
```diff
@@ -1,30 +1,22 @@
 ## Building the plugin
 
-We use gradle to build the plugin, and require version 2.1 or greater (`gradle -version`).
+We use ant to build the plugin; to check your ant install, run `ant -version`.
```
pq: Do we have a minimum ant version?

building.md
```diff
@@ -1,30 +1,22 @@
...
+build/flutter-intellij.jar
 ```
 
-## More information
-
-Gradle can be installed via `brew` on the Mac (`brew install gradle`), or via `SDKMAN` as
-described on the gradle [web site](http://www.gradle.org/downloads).
-
-Documentation for the options in `gradle.properties` is available at
-[github.com/JetBrains/gradle-intellij-plugin](https://github.com/JetBrains/gradle-intellij-plugin).
```
pq: Maybe we can add instructions for installing ant here?  Is `brew` the way to go?

tool/travis.sh
```diff
@@ -11,10 +11,12 @@ set -e
...
-
 # Print a report for the API used from the Dart plugin
 pub get
 dart tool/grind.dart api
+
+# Run the ant build.
+ant build \
+  -Didea.product=$IDEA_PRODUCT \
+  -Didea.version=$IDEA_VERSION \
+  -Ddart.plugin.version=$DART_PLUGIN_VERSION
```
pq: Where does `DART_PLUGIN_VERSION` get set?

building.md
devoncarew: The latest available is `1.9.7`; the current version on travis is `1.8.2`. That seems to meet our needs - and we're not using anything 1.9 specific - so we can take that as our current min.

building.md
devoncarew: Sure, added. `brew install ant` will get the the latest bits.

tool/travis.sh
devoncarew: This gets set from the travis entry-point script, `//.travis.yml`. We get specific versions of these env vars for each bot that we configure (for each line in the `env:` section).

devoncarew:
> Besides the general hilarity of going back to ant, LGTM!

Yeah, I weighed the pros and cons of 4-5 different options. Ant really did come out on top; it works great with Java projects, and the build script is very straightforward to tinker with.

PR #580, opened 2016-12-27
Bazel run config mnemonics (and cleanup).

pq:
* adds missing form UI mnemonics.

mnemonics

* misc. cleanup

@devoncarew

### DIFF
--- a/third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
+++ b/third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
@@ -198,9 +198,6 @@ public class DartVmServiceDebugProcessZ extends DartVmServiceDebugProcess {
     });
   }
 
-  public void scheduleConnect() {
-  }
-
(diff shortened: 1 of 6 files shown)


third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
```diff
@@ -198,9 +198,6 @@ public void logInformation(final String message, final Throwable exception) {
     });
   }
 
-  public void scheduleConnect() {
```
devoncarew: What does the parent implementation do? Does this method (even an empty implementation) effect behavior?

devoncarew:
lgtm

third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
pq: I wondered the same thing but there's no parent implementation --- the closest is `scheduleConnect(String)` which has a different signature obviously and is private.

Issue #584, opened 2016-12-30
build updates


devoncarew:
Some updates to our build system:

- [x] download artifacts from google storage (to improve reliability)
- [x] test against an additional IntelliJ platform (to ensure the build system is robust wrt other IntelliJ platforms)

PR #585, opened 2016-12-30
Build improvements

devoncarew:
- move to downloading large build prereq files from google storage
- add the ability to build against other IntelliJ platforms
- break the tests out into 'unit' and 'integration' source directories. unit tests don't need access to the IntelliJ APIs, and integration tests do. I don't love this - I'd rather that unit tests were small, self-contained tests (whether they reference the IntelliJ APIs or not), and integration tests were larger tests of things like use cases and user journeys. Until we can bring up IntelliJ on travis however, this split helps us
- run the unit tests as part of the idea builds
- shard travis into 4 bots, testing our earliest and latest supported versions of IntelliJ CE, the current webstorm, and 2.3 beta 1 of android studio
- fix https://github.com/flutter/flutter-intellij/issues/584

@pq


building.md
```diff
@@ -21,3 +21,21 @@ build/flutter-intellij.jar
...
+## The build pre-reqs
+
+Several large files required for building the plugin are downloaded from Google Storage
+and cached in the `artifacts/` directory. We use timestamps to know if the local cached
+copies are up-to-date.
+
+To see a list of all current IntelliJ build pre-reqs:
+
+```shell
+$ gsutil ls gs://flutter_infra/flutter/intellij/
```
pq: Maybe for later, but perhaps a link to `gsutil` install instructions?

devoncarew: Will add in a follow-up

Issue #336, opened 2016-10-19
Newly created flutter project shows broken pub banner


eseidelGoogle:
I suspect this will be fixed by https://github.com/flutter/flutter-intellij/issues/271, but in case not I'm filing here.

I'm not even sure the banner is needed, but it shows up and then the buttons don't actually do what you want.
<img width="732" alt="screen shot 2016-10-19 at 8 24 31 am" src="https://cloud.githubusercontent.com/assets/11857803/19525450/cfb2d234-95d5-11e6-9d30-b05063a3fdda.png">

pq:
Right.  There are (at least) two issues:

1. the pub banner shows even though we've run pub behind the scenes in `flutter create` --- maybe @alexander-doroshko can chime in on what we might do here...
1. `pub get` doesn't do the right thing for Flutter w/o some massaging --- as you note, this is fixed with #271 and will work as soon as a new Dart plugin is pushed

Thanks!

alexander-doroshko:
1. Banner is shown if there's no `.packages` file near `pubspec.yaml` or if `.packages` file is older than pubspec.
1. I guess you want me to cherry-pick that fix to 162 branch and update the Dart plugin?

pq:
> Banner is shown if there's no .packages file near pubspec.yaml or if .packages file is older than pubspec.

Ah.  That makes sense.  What's odd is that I can repro this with a project that looks like this:

```
[~] $ ls -alT
total 56
drwxr-xr-x  14 pquitslund  eng   476 Oct 19 09:18:53 2016 .
drwxr-xr-x  12 pquitslund  eng   408 Oct 19 09:18:52 2016 ..
drwxr-xr-x   3 pquitslund  eng   102 Oct 19 09:17:55 2016 .atom
-rw-r--r--   1 pquitslund  eng    84 Oct 19 09:17:55 2016 .gitignore
drwxr-xr-x   8 pquitslund  eng   272 Oct 19 09:18:53 2016 .idea
-rw-r--r--   1 pquitslund  eng  1599 Oct 19 09:17:59 2016 .packages
-rw-r--r--   1 pquitslund  eng   154 Oct 19 09:17:55 2016 README.md
drwxr-xr-x   4 pquitslund  eng   136 Oct 19 09:17:55 2016 android
-rw-r--r--   1 pquitslund  eng    48 Oct 19 09:17:55 2016 flutter.yaml
[...]

PR #586, opened 2017-01-03
Suppress pub warnings during `flutter create` (#336).

pq:
* adds new Dart plugin and bumps dependency lower-bound
* uses new `DartPubActionBase.setIsInProgress()` API
* fixes incompatible Dart version message

Fixes: #336.

@devoncarew


src/io/flutter/dart/DartPlugin.java
```diff
@@ -18,7 +18,7 @@
   /**
    * Tracks the minimum required Dart Plugin version.
    */
-  private static final String MINIMUM_REQUIRED_PLUGIN_VERSION = "162.2485";
+  private static final String MINIMUM_REQUIRED_PLUGIN_VERSION = "163.10154";
```
pq: Note that this is the bound @alexander-doroshko suggested.  It could probably just as well be `163.11306`.

build.xml
```diff
@@ -11,8 +11,8 @@
 
   <!-- defaults -->
   <property name="idea.product" value="ideaIC"/>
-  <property name="idea.version" value="2016.3.1"/>
-  <property name="dart.plugin.version" value="163.10154.19"/>
+  <property name="idea.version" value="2016.3.2"/>
```
pq: This is required by the `163.10154.19` Dart plugin.  Is it OK to bump our IDEA version?

pq: As per our conversation, I'll change this back and opt for a reflective solution.

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -104,6 +105,26 @@ private static void printExitMessage(@Nullable Project project, @Nullable Module
     });
   }
 
+  private static void setPubInProgress(boolean inProgress) {
+    try {
+      DartPubActionBase.class.getMethod("setIsInProgress", boolean.class).invoke(null, inProgress);
+    }
+    catch (Exception e) {
```
devoncarew: Perhaps catch `Throwable`? This would catch Errors as well, which ones that would otherwise be compile time errors probably are.

devoncarew:
lgtm w/ the change in the catch handler

src/io/flutter/sdk/FlutterSdk.java
pq: Done!

PR #595, opened 2017-01-05
Form cleanup.

pq:
* adds missing mnemomic to Analytics checkbox

screen shot 2017-01-05 at 6 54 37 am

* wraps settings error text in a scrollpane (as per recommended IDEA style)

@devoncarew

### DIFF
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.form
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.form
@@ -14,6 +14,9 @@
           <grid row="0" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="0" fill="3" indent="0" use-parent-layout="false"/>
         </constraints>
         <properties/>
+        <clientProperties>
+          <BorderFactoryClass class="java.lang.String" value="com.intellij.ui.IdeBorderFactory$PlainSmallWithIndent"/>
+        </clientProperties>
         <border type="etched" title="SDK"/>
         <children>
           <component id="a59e7" class="javax.swing.JLabel">
@@ -66,6 +69,9 @@
           <grid row="1" column="0" row-span="1" col-span="1" vsize-policy="3" hsize-policy="3" anchor="0" fill="3" indent="0" use-parent-layout="false"/>
         </constraints>
         <properties/>
+        <clientProperties>
+          <BorderFactoryClass class="java.lang.String" value="com.intellij.ui.IdeBorderFactory$PlainSmallWithIndent"/>
+        </clientProperties>
         <border type="etched" title="General"/>
         <children>
           <component id="6304a" class="javax.swing.JCheckBox" binding="myReportUsageInformationCheckBox">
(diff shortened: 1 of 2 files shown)


src/io/flutter/sdk/FlutterSettingsConfigurable.form
```diff
@@ -14,9 +14,6 @@
           <grid row="0" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="0" fill="3" indent="0" use-parent-layout="false"/>
         </constraints>
         <properties/>
-        <clientProperties>
```
devoncarew: Please add this (and the section below) back in. The UI editor always wants to remove it, but it gives us the nice (single line on top) border between the preference section.

pq: Oh!  Sure.  Sorry about that.  It'd be great to prevent the UI editor from doing this.

@alexander-doroshko : thoughts?

pq:
Annoying to have to post-edit but I don't see an obvious alternative in the properties.

screen shot 2017-01-05 at 8 41 12 am

Maybe @alexander-doroshko can advise when he's back.

alexander-doroshko:
Strange. I never edit form files manually and never get client properties lost.
I confirm that BorderFactoryClass set as a client property for titled panels is the preferred style. Here's where I set it in UI:
image

alexander-doroshko:
@devoncarew @pq
I've got an idea. Looks like **BorderFactoryClass** is shown as **Client Property** in UI only in internal mode. To enable internal mode you need to add `idea.is.internal=true` to [idea.properties](https://www.jetbrains.com/help/idea/2016.3/file-idea-properties.html) file. This should allow you to forget about manual forms editing.

pq:
Just getting to try this but not seeing the option.

screen shot 2017-01-12 at 5 18 04 am

screen shot 2017-01-12 at 5 18 28 am

@alexander-doroshko : am I missing any other settings?  Is it possible that this only works in Ultimate?

alexander-doroshko:
Do you have Tools | Internal Actions in the main menu? I guess no. That might be a wrong idea.properties file. 
I've found a better way: click Help | Edit Custom VM Options, agree to create a file, add 
```
-Didea.is.internal=true
```
to it, restart the IDE.

pq:
Hmmm.  Actually I do have that menu!

screen shot 2017-01-12 at 6 02 51 am

Anything else I might need to enable?

And while we're on the topic, what should I explore in here?  Lots of interesting looking stuff!

alexander-doroshko:
Ok, next attempt: edit `.idea/misc.xml` and add `ClientPropertiesManager` to the beginning:
```
<project version="4">
  <component name="ClientPropertiesManager">
    <properties class="javax.swing.JPanel">
      <property name="BorderFactoryClass" class="java.lang.String" />
    </properties>
  </component>
...
```
Safe to commit if it works.

As for internal actions - no special recommendations. It appeared to be not needed.

alexander-doroshko:
@pq Does it work?

pq:
> @pq Does it work?

Sorry!  Was distracted by the Dart SDK roll into Flutter...  Picking this up again now.

pq:
Sorry for the delay.  Works a charm.

https://github.com/flutter/flutter-intellij/pull/635.

Thanks!

PR #598, opened 2017-01-06
Revert scrollpane wrapping.

pq:
Inspection be damned, wrapping this in a scrollpane produces an ugly UI artifact.

screen shot 2017-01-06 at 10 14 11 am

Reverting to remove.

@devoncarew 

FYI: @alexander-doroshko : advice on addressing the inspection appreciated.

screen shot 2017-01-06 at 11 22 26 am

### DIFF
--- a/src/io/flutter/module/FlutterGeneratorPeer.form
+++ b/src/io/flutter/module/FlutterGeneratorPeer.form
@@ -52,6 +52,14 @@
         <properties/>
         <border type="none"/>
         <children>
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
If `errorText` needs to be a `JTextPane` then inspection is right, and the correct fix would be to control the visibility of the wrapping panel (may be JScroppPane itself) instead of the inner `errorText` component (io.flutter.module.FlutterGeneratorPeer#validate).

pq:
Thanks!

https://github.com/flutter/flutter-intellij/pull/600

PR #600, opened 2017-01-09
Error Text scrollpane fix.

pq:
Follow-up from: https://github.com/flutter/flutter-intellij/pull/598.

* scrollpane as per @alexander-doroshko's recommendation
* no border to keep presentation clean (and the same)

screen shot 2017-01-09 at 5 48 04 am

@alexander-doroshko @devoncarew


alexander-doroshko:
If it can't be simplified down to a simple label then lgtm.

devoncarew:
lgtm

Issue #611, opened 2017-01-10
IntelliJ - bring the simulator window forward


Sfshaza:
## Steps to Reproduce

I had launched the simulator (but had hidden the app on my desktop), and opened my example in the IntelliJ IDE (Community Edition). I clicked the "run" button and saw some alarming text in the message window. (I tried copying the text, but IntelliJ doesn't let me select it, so I'm including a screenshot.)

![screen shot 2017-01-10 at 12 43 46 pm](https://cloud.githubusercontent.com/assets/6293442/21824047/85124478-d732-11e6-89ad-ac9bc6963ab1.png)

I believed that the example wasn't running and wasn't sure what was wrong.

But it turns out that the example was running. When the user runs the app, please bring the simulator window forward!

## Version info

~ $ flutter doctor
[✓] Flutter (on Mac OS, channel master)
    • Flutter at /Users/shaza/GoogleDrive/TMP-flutter/flutter
    • Framework revision 98016b1c4a (7 weeks ago), 2016-11-21 09:21:49
    • Engine revision cea5ed2b9b
    • Tools Dart version 1.21.0-dev.6.0

[✓] Android toolchain - develop for Android devices (Android SDK 25.0.0)
    • Android SDK at /Users/shaza/Library/Android/sdk
    • Platform android-25, build-tools 25.0.0
    • Java(TM) SE Runtime Environment (build 1.8.0_112-b16)

[✓] iOS toolchain - develop for iOS devices (Xcode 7.3.1)
    • XCode at /Applications/Xcode.app/Contents/Developer
    • Xcode 7.3.1, Build version 7D1014

[✓] IntelliJ IDEA Community Edition (version 2016.3.1)
    • Dart plugin installed
    • Flutter plugin installed

[✓] Connected devices
    • iPhone 6s Plus • 00A1718E-C7AA-4B05-B650-6AFA1BA1F84B • ios • iOS 9.3 (simulator)

PR #623, opened 2017-01-12
Bring simulator to front on run/debug (#611).

pq:
Fixes: #611.

@devoncarew

### DIFF
--- a/src/io/flutter/actions/OpenSimulatorAction.java
+++ b/src/io/flutter/actions/OpenSimulatorAction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 The Chromium Authors. All rights reserved.
+ * Copyright 2017 The Chromium Authors. All rights reserved.
  * Use of this source code is governed by a BSD-style license that can be
  * found in the LICENSE file.
  */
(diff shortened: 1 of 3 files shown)


src/io/flutter/actions/OpenSimulatorAction.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
```
devoncarew: 2017

src/io/flutter/run/FlutterAppState.java
```diff
@@ -103,6 +105,13 @@ protected ProcessHandler startProcess() throws ExecutionException {
       }
     }
 
+    if (device != null) {
+      // Bring simulator to front.
+      if (device.emulator() && Objects.equals(device.platform(), "ios")) {
+        new OpenSimulatorAction(true).actionPerformed(null);
```
devoncarew: Just to confirm, that re-issuing the open doesn't have side effects apart from bringing the app to the front?

One way that I've solved this in the past is to generate and execute an applescript on the fly which tells the app (specified by name) to come to the front.

devoncarew:
lgtm

src/io/flutter/actions/OpenSimulatorAction.java
pq: Thanks!  Updated.

src/io/flutter/run/FlutterAppState.java
pq: Did a bit of poking around and I think this is the recommended approach.  Would like to avoid `osascript` if possible but maybe that's just superstition.  Let's see if this has any hiccups and update if needed.

Thanks!

PR #625, opened 2017-01-12
detect flutter module types in small IDEs

devoncarew:
- detect flutter module types in small IDEs

Re: the `<depends>` change, I believe we get the list of supported platforms from the JetBrains plugins site now - we can toggle on support for specific platforms.

@pq

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -15,11 +15,9 @@
   <!-- plugin compatibility -->
   <!-- see: http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html -->
 
-  <!-- available in IntelliJ IDEA (CE and UE) -->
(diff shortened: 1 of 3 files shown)


alexander-doroshko:
I guess with this change auto-detection will say that the plugin is compatible with all IDEs.
If you want you can disable autodetection and select compatible IDEs manually on the plugin page.

devoncarew:
> If you want you can disable autodetection and select compatible IDEs manually on the plugin page.

I think that's what we've done - currently (on that page) we just say we're compatible with the CE and Ultimate.

alexander-doroshko:
> I think that's what we've done

May be, I can't see your [plugin settings](https://plugins.jetbrains.com/plugin/edit?pluginId=9212) page :).
That's the 'Determine supported products by dependencies in plugin.xml' check box.

devoncarew:
Yup, looks like we're fully manual.

PR #629, opened 2017-01-12
Update plugin.xml

devoncarew:
- rev to 0.1.8.1, capture some breakpoint improvements for bazel launch configs

@pq

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -8,7 +8,7 @@
 
   <category>Custom Languages</category>
 
-  <version>0.1.8</version>
+  <version>0.1.8.1</version>
 
   <idea-version since-build="162.1" until-build="163.*"/>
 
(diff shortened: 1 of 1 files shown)


resources/META-INF/plugin.xml
```diff
@@ -21,6 +21,11 @@
 
   <change-notes>
     <![CDATA[
+0.1.8.1:
+<ul>
+  <li>improve handling of breakpoints for the bazel launch config</li>
```
pq: For completeness we might add a line item to record that the iOS simulator now gets activated/brought to front on run/debug.

devoncarew: I just cherry-picked in two breakpoint fixes to create 0.1.8.1 (i.e., it doesn't contain all the commits from the last release).

pq: Ah! 👍

Issue #636, opened 2017-01-18
No error message if Dart SDK is not set or is set incorrectly


skybrian:
Currently we find the Flutter SDK from the Dart SDK. If the Dart SDK is not set or it's set to a non-Flutter SDK, we don't show any error message, but any Run Configurations for Flutter are disabled.

We need to somehow tell the user that they should set the Dart SDK to the one in $FLUTTER_SDK/bin/cache/dart-sdk and then everything will work.

pq:
Good catch!  We actually used to have a check for this.  (I'll try and dig those old bits up.)

Moving on, we might add a check for this condition here:

    io.flutter.inspections.SdkConfigurationNotificationProvider#createNoFlutterSdkPanel

(`flutterSdk` will be `null` but `DartSdk.getDartSdk(project)` won't be.)

Self-assigning as I may tackle this quickly.  Feel free to ping me if you want to snag it.

skybrian:
I'm about to send a pull request, might want to wait for that. (Although I'm sure you'll do it better.)

skybrian:
https://github.com/flutter/flutter-intellij/pull/637

pq:
Cool!  (And don't be too sure of that!)

PR #637, opened 2017-01-18
Show error when Dart SDK is not set correctly (#636)

skybrian:


alexander-doroshko:
Flutter users do not need to use the Dart page in Preferences and to know much about Dart SDK. I'd suggest rephrasing: it is **Flutter** SDK that is not set, and it is **Flutter** page in Settings (Linux) or Preferences (Mac) where user should set **Flutter** SDK.
Yes, under the hood IDE will automatically change Dart SDK as well (to bin/cache/dart-sdk), but this is not something user should care about.
P.S. I believe this check has already been there at some point, @pq should remember better.

skybrian:
I tested error message by changing the Dart SDK, so that does work. I guess what you're saying is that modifying the Flutter SDK also works?

If the user sets the Flutter SDK and then the Dart SDK, it seems like that might lead to confusion, if they don't know they're linked? I wonder if there is anything we can do to make this more obvious. Perhaps in a Flutter module, the Dart SDK should appear to be locked down (no edit allowed).

alexander-doroshko:
You are right, there's a field for the Dart SDK in the Dart page and for the Flutter SDK in the Flutter page. And inside the model used by the IDE it is the same setting. Changing one SDK path affects another SDK path. But it may be not obvious for users.

Indeed, you may be confused if you change both SDKs in one shot. Only one of them wins and the next time you open Preferences you'll see it. 

Also you may be confused if you change one SDK path and then move to another page. For example you change Flutter SDK path and open Dart page without closing the Preferences dialog. You'll see that Dart SDK path is not updated at this moment. Only when you press OK and then open Preferences again you'll see the new path in the 2nd page.

I'm afraid I do not know a good fix for this confusion at the moment.

What I'm saying is that Flutter users do not have to care about the Dart page and Dart SDK field. So all messages related to the Flutter settings could mention only Flutter page and Flutter SDK field.

> Perhaps in a Flutter module, the Dart SDK should appear to be locked down (no edit allowed).

We may think about some solution like this. The problem is that IntelliJ IDEA a project may contain a lot of modules of different nature. Three Java modules, five Dart modules, two Flutter modules, etc. Dart/Flutter SDK in 2016.3 is set per the whole IDE - once changed it affects all projects. In 2017.1 it is improved, and SDK is set per project. Different projects may have different SDKs configured. But still not per module.

pq:
> P.S. I believe this check has already been there at some point, @pq should remember better.

Right.  It was in this commit:

https://github.com/flutter/flutter-intellij/commit/d6b1cd18b4ef8e2abf57c51f359b53ac232592ad

Given all the nuances, how do you feel about restoring `WrongDartSdkConfigurationNotificationProvider`? It seems like the case we really want to flag is where an incompatible Dart SDK is set and this in inspection does just that.

pq:
Ah, wait, it looks like that logic just got moved (see: https://github.com/flutter/flutter-intellij/commit/fcbb0908bddb72cf7ff815189650a458e82ba13a).  It's now here:

```io.flutter.inspections.SdkConfigurationNotificationProvider#createWrongSdkPanel```

this *should* generate an editor notification in this case.  If it's not we should fix that!  (Or if that's not the approach we want to take we should remove it.)

**Comment by alexander-doroshko** [comment]
With the current model I wouldn't say 'wrong Dart SDK'. I would say 'Flutter SDK not specified'. Updating `SdkConfigurationNotificationProvider` looks good to me. That's exactly the TODO that you added in its `createNoFlutterSdkPanel()` that Brian suggests to implement. I only mentioned that you can avoid word 'Dart SDK' 'Dart page in Settings', 'bin/cache/dart-sdk'.

Link to open Flutter page would be great for sure as Brian mentioned in the new TODO.

BTW, `createWrongSdkPanel()` seems to be unreachable with the current model. As well as 
`// TODO(devoncarew): Recommend to set up with Flutter's dart sdk.` - dead branch too.

**Comment by alexander-doroshko** [comment]
> this should generate an editor notification in this case. If it's not we should fix that!

it's a dead branch. And the fix is the TODO that you left in `createNoFlutterSdkPanel()` - exactly what Brian is suggesting :)

**Comment by pq** [comment]
Clearly I need more coffee...  and we have some dead code to cleanup!

**Comment by skybrian** [comment]
Trying this out, if the error tells the user to change the Flutter SDK, it will be confusing because it looks like the Flutter SDK is already set correctly in the UI.

It looks like we populate mySdkCombo in addKnownSDKPathsToCombo() and the first item is selected by default, but this doesn't actually have any effect on the rest of the system. It seems like we need to somehow represent 'Flutter SDK not set' in the preferences UI, so that the user will have something to do, and then when they save it, it will be fixed.

Also, the 'Apply' button is weird. It apparently has no effect (the UI doesn't update right away) but in the debugger I see that it does set the Dart SDK. So we need to fix that too.

**Comment by skybrian** [comment]
Another problem: the device selector does not appear when it should. If you didn't have the Flutter SDK properly configured, and you configure it, doesn't show up.

It looks like it's listening to FlutterSdkManager which doesn't get an event when the Dart SDK changes.

I *think* FlutterSdkManager.Listener should fire events for this, but that's not what the comment says. It says it should be fired when the "Flutter global library is set".

(Also, I don't understand why FlutterSdkManager uses ApplicationLibraryTable. Is that table even used for Dart?)

**Comment by skybrian** [comment]
Please take another look. This doesn't handle everything, but I made enough changes so that it mostly looks right when the user changes the Flutter SDK in response to the error.

But the Apply button still doesn't work, and it's unclear why. Current behavior is that the Flutter Run Configuration gets greyed out when you click Apply, but the device menu doesn't appear until after you close the settings dialog.

File: src/io/flutter/inspections/SdkConfigurationNotificationProvider.java [hunk 1]
```diff
@@ -47,30 +47,18 @@ public SdkConfigurationNotificationProvider(@NotNull Project project) {
     this.project = project;
   }
 
-  @Nullable
-  private static EditorNotificationPanel createWrongSdkPanel(@NotNull Project project, @Nullable Module module) {
```
pq: How is this case being handled?

alexander-doroshko: it is not possible

alexander-doroshko:
>  I'm not seeing how the case where the Dart SDK is set to a non-Flutter SDK is handled. 

It is not possible to configure both Dart SDK and Flutter SDK in such a way that they are in conflict. Because under the hood it is **ONE** setting.
If you change Dart SDK in Preferences to non-Flutter's and press OK - at this moment you loose Flutter SDK setting for this project. This is the case 'Flutter SDK is not specified'. You can fix it in 2 ways: either change Dart SDK in the Dart page or set Flutter SDK in the Flutter page. Under the hood these are absolutely equivalent actions.

skybrian:
There is still an issue that no "Flutter SDK changed" event is sent when the user sets the Dart SDK. But, this error message doesn't rely on events - it does a query each time.

pq:
>It is not possible to configure both Dart SDK and Flutter SDK in such a way that they are in conflict. Because under the hood it is ONE setting.

Something is not right then...  I just got into this state by setting my Dart SDK preferences.

screen shot 2017-01-19 at 1 55 04 pm


screen shot 2017-01-19 at 1 54 58 pm

skybrian:
Yes, when no Flutter SDK is set, the combo box is displayed wrong. (It displays the first entry.) I fixed that too.

However, I just noticed that the "Version:" text is still wrong. Working on that.

alexander-doroshko:
> Something is not right then...

Notice a 'Reset...' link in the top-right of your 2nd screenshot :). It is a marker of a bug: the page is in the incorrect state right after opening it. 'Reset...' (as well as enabled Apply button) automatically appears when UI doesn't correspond to the stored model. On page opening there must be neither Reset nor Apply,

Issue #630, opened 2017-01-13
Project view shows many files that first-time users may be confused about


mit-mit:
First-time users will likely try to make sense of each and every entry shown in the Project view. Thus it's important this only contains things they actually need to worry about.

I think we should be hiding the following:

1. `.idea` dir: IDE settings that are typically edited using the IJ UI itself
1. `.iml` file: Same as above
1. `.packages` file: Generated pub artefact
1. `pubspec.lock` file: Generated pub artefact

Additionally, the Dart plugin will automatically hide these:
1. `.pub` dir: The local pub cache, not for editing
1. `build` dir: build outputs, not a source artefact

alexander-doroshko:
My $0.02.
I'd like not to decide for users whether they want to see all files as they exist in the file system or not.

For example: for a long time we used to hide excluded folders from the Project View. But so many users were unhappy with it, so some day we introduced an option to show excluded files (gear menu in the Project View tool window header). And a bit later we even made it the default.

You may document this gear option: disabling it will hide `.pub` and `build`. Though I'm sure a lot of users will finally prefer to keep `build` folder contents visible.

As for the `*.iml` file: it doesn't have to be in the project root. Your flutter app generator may place it inside `.idea` folder with the coresponding path in `.idea/modules.xml`.

Finally. Yes, it is possible to implement some hacky magic so that some specific files and folders are not visible in the Project View. But I'm strongly against this magic. I think it would introduce much more confusion than seeing file structure as is.

mit-mit:
I want the project view to hide these exactly in the case you describe: When 'Show excluded files' is set to No. I am totally fine with them being shows when `Show excluded files` is Yes.

To get that, don't we need to mark these with `excludeFolder` in the `.iml` file? Or is what is shown/hidden depending on that toggle controlled elsewhere?

alexander-doroshko:
Unfortunately individual files exclusion is not supported.

Folder `build` and `.pub` should be excluded automatically (Dart plugin should do the trick even if auto-generated project template doesn't).

`*.iml` may be placed inside `.idea` on project creation

As for `.packages` and `pubspec.lock` - I'm afraid no way to hide them.

mit-mit:
Thanks for the tip on `build` and `.pub`, that seems to work!

pq:
Thanks for the input @alexander-doroshko !

> *.iml may be placed inside .idea on project creation

[...]

Issue #316, opened 2016-10-18
Opening an existing Flutter project that is missing .iml/.idea files causes confusing errors


devoncarew:
We should implement `ProjectOpenProcessor` for easy opening existing Flutter projects (I mean those that have never been opened in the IDE before and do not have .idea folder, so that Flutter module is created for them, not the default module type).

mit-mit:
+1, as discussed in #326 this causes a confusing Android message ("Frameworks detected: Android framework is detected in the project Configure"), and causes the debug config to be missing.

mit-mit:
From #421: This also means that the user is left with an empty IDE shell rather than with a shows Project View and the code editor open on main.dart.

pq:
As per the conversation in #530, I think the plan is to document this short-coming and potentially fix this up a bit in a future release.

devoncarew:
We should create a project with the cli, then try and `File > Open` that project (to determine how common this use case is for the user).

pq:
As per our discussion last week, short-term is to update the known issues doc.

Long-term work to be driven by a getting started user journey.

cc @mit-mit

mit-mit:
I confirmed that in 0.1.8 RC this issue only pertains to projects that are missing the `.iml` file and/or `.idea` directory.

A new project -- whether created with `flutter create` or IntelliJ has both these, and opens as expected.

pq:
Thanks!  As mentioned in #607, we've really got to double-down on this flow as it may be hard to ensure that even our sample projects have reliably sufficient metadata.

pq:
As per #621, the `OpenProcessor` should also do it's best to make sure that `pub` is run if needed.

devoncarew:
It's looking like the `ProjectOpenProcessor` could be a silver bullet for several different issues...

devoncarew:
Can we update our known issues (https://github.com/flutter/flutter-intellij#known-issues) to close this out?

goleary:
I'm trying to run this example: https://github.com/flutter/plugins/tree/master/packages/android_alarm_manager/example but when i open the folder in android studio I get all kinds of errors about various things not being configured.  I copied the flutter/dart settings from another project, but I'm still getting this message.

pq:
cc @stevemessick

goleary:
I got it working, but it reasons that it's possible to use the same config when opening flutter projects not created from the  IDE.

pq:
@goleary : do you recall what you had to do?

[...]

PR #638, opened 2017-01-18
Open processor first steps (#316)

pq:
* adds an open processor that delegates to the platform for basic project import but adds it’s own post-processing hooks
* detects missing `.packages` and runs `flutter packages get` (#630)
* updates SDK actions to allow for running w/o analytics (for programmatic access)
* analytics reporing enabled for flutter doctor calls
* `FlutterSdkUtil` cleanup and refactoring

screen shot 2017-01-18 at 1 54 19 pm

Still TODO: add detection and handling of stale `.packages` (a toast suggesting update) --- I'll pick that up in a follow-up PR with a bit more cleanup too.

@devoncarew

### DIFF
--- a/src/io/flutter/actions/FlutterSdkAction.java
+++ b/src/io/flutter/actions/FlutterSdkAction.java
@@ -11,40 +11,41 @@ import com.intellij.openapi.actionSystem.CommonDataKeys;
 import com.intellij.openapi.actionSystem.LangDataKeys;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.module.Module;
-import com.intellij.openapi.module.ModuleManager;
 import com.intellij.openapi.project.DumbAwareAction;
 import com.intellij.openapi.project.Project;
-import com.intellij.openapi.roots.ModuleRootManager;
 import com.intellij.openapi.roots.ProjectRootManager;
 import com.intellij.openapi.util.Pair;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiFile;
 import io.flutter.FlutterBundle;
-import io.flutter.FlutterConstants;
 import io.flutter.FlutterErrors;
+import io.flutter.FlutterInitializer;
 import io.flutter.sdk.FlutterSdk;
+import io.flutter.sdk.FlutterSdkUtil;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.ArrayList;
-import java.util.List;
-
 /**
  * Base class for Flutter commands.
+ * <p>
+ * In general actions are executed via {@link #actionPerformed(AnActionEvent)}, and
+ * send analytics.  In case an action is performed and analytics should not be
+ * collected, prefer {@link #perform(FlutterSdk, Project, AnActionEvent, boolean)}.
  */
 public abstract class FlutterSdkAction extends DumbAwareAction {
 
   private static final Logger LOG = Logger.getInstance(FlutterSdkAction.class);
 
   @Nullable
-  public static Pair<Module, VirtualFile> getModuleAndPubspecYamlFile(@NotNull final Project project, final AnActionEvent e)
+  public static Pair<Module, VirtualFile> getModuleAndPubspecYamlFile(@NotNull final Project project, @Nullable final AnActionEvent e)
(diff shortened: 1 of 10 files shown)


src/io/flutter/actions/FlutterSdkAction.java
```diff
@@ -120,5 +82,13 @@ public void actionPerformed(AnActionEvent event) {
     }
   }
 
-  public abstract void perform(@NotNull FlutterSdk sdk, @NotNull Project project, AnActionEvent event) throws ExecutionException;
+  public final void perform(@NotNull FlutterSdk sdk, @NotNull Project project, @Nullable AnActionEvent event, boolean logAction)
+    throws ExecutionException {
+    if (logAction) {
+      sendActionEvent();
+    }
+    perform(sdk, project, event);
```
devoncarew: I generally put the implementation in the version of the method w/ the most args, but don't think we have any specific guidelines for such -

src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
```
devoncarew: 2017 :)

src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+public class FlutterProjectOpenProcessor extends ProjectOpenProcessor {
+
+  private static void doPerform(@NotNull FlutterSdkAction action, @NotNull Project project) throws ExecutionException {
+    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
+    if (sdk == null) {
+      // Lack of SDK will be flagged elsewhere by inspection.
+      return;
+    }
+
+    action.perform(sdk, project, null, false);
```
devoncarew: You could call the perform() method w/o the `logAction` boolean here

src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -0,0 +1,105 @@
...
+import io.flutter.sdk.FlutterSdkUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.*;
+import java.util.Arrays;
+import java.util.Objects;
+
+
+public class FlutterProjectOpenProcessor extends ProjectOpenProcessor {
```
devoncarew: It's not clear to me where we add (or create) the flutter module type...

devoncarew:
Awesome! It'll be very nice to have this.

One main question re: where we create flutter modules (I imagine the default importer will not set up the right module type w/ help?).

pq:
Right.  This doesn't do anything to add modules.  We do have an inspection that handles that case though.  That said, we can improve things further by rolling that into the processor as well.  (I'll add that as a TODO.)

devoncarew:
> Right. This doesn't do anything to add modules.

I suspect that we should, in order to set things up w/ as few steps (and as llittle chance of running into issues) for the user.

> That said, we can improve things further by rolling that into the processor as well. (I'll add that as a TODO.)

sgtm!

src/io/flutter/project/FlutterProjectOpenProcessor.java
pq: d'oh!

src/io/flutter/project/FlutterProjectOpenProcessor.java
pq: Right.  We don't (yet).  That'll come in the next round.

src/io/flutter/project/FlutterProjectOpenProcessor.java
pq: Sure.  Done.

src/io/flutter/actions/FlutterSdkAction.java
pq: I'm going for a template method though. I don't want implementers to see (or care) about the extra `logAction` bit.  Or maybe I'm misunderstanding?

devoncarew: ah, gotcha

pq:
Thanks!

PR #639, opened 2017-01-19
Update .travis.yml

devoncarew:
- test against the latest IntelliJ (`2016.3.3`)

@pq


devoncarew:
Most of the work for the build is done here:

- https://github.com/flutter/flutter-intellij/blob/master/.travis.yml#L22
- https://github.com/flutter/flutter-intellij/blob/master/tool/travis.sh#L22
- https://github.com/flutter/flutter-intellij/blob/master/build.xml

PR #646, opened 2017-01-19
Show error when Dart SDK is not set correctly (#636)

skybrian:
Since the Flutter SDK is derived the from Dart SDK, we actually handle
this by reporting: 'No Flutter SDK configured'. When the user fixes
the Flutter SDK, the Dart SDK will also be fixed.

Fixed combo box to display correctly when no Flutter SDK is configured.
Fire a Flutter SDK changed event when the user changes the Flutter SDK.
Removed dead code.

Not done: when the user changes the Dart SDK, we should fire a Flutter
SDK changed event.

(Sorry about the repost; I accidentally deleted the branch for the last pull request.)

### DIFF
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.java
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.java
@@ -143,8 +143,12 @@ public class FlutterSettingsConfigurable implements SearchableConfigurable {
   public void reset() {
     final FlutterSdk sdk = FlutterSdk.getGlobalFlutterSdk();
     final String path = sdk != null ? sdk.getHomePath() : "";
-    mySdkCombo.getComboBox().getEditor().setItem(FileUtil.toSystemDependentName(path));
     FlutterSdkUtil.addKnownSDKPathsToCombo(mySdkCombo.getComboBox());
+
+    // Set this after populating the combo box to display correctly when the Flutter SDK is unset.
+    // (This can happen if the user changed the Dart SDK.)
+    mySdkCombo.getComboBox().getEditor().setItem(FileUtil.toSystemDependentName(path));
+
     updateVersionText();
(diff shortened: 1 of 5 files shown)


pq:
Cool!

I added a comment to the older PR to the effect that I don't see how we're guarding against the case where a user.

1. sets the Flutter SDK
1. updates the Dart SDK (to a different location)

https://github.com/flutter/flutter-intellij/pull/637#issuecomment-273911447

Curious what @alexander-doroshko is thinking there.

This is what the old logic to detect mismatched SDKs was intended to catch.

skybrian:
Answered already I think, but I'll answer again here to avoid confusion.

When you set the Dart SDK to a different location, the Flutter SDK is automatically unset (because the Dart SDK changed, and the Flutter SDK is derived from it). However, this wasn't visible to the user due to UI bugs - it would still show the old Flutter SDK even though the setting changed. I've mostly fixed the UI bugs in this patch - in particular see the changes in FlutterSettingsConfigurable.

pq:
Ah!  Thanks for the clarification.  It was the UI bug that was throwing me off.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -153,26 +157,38 @@ private void updateVersionText() {
...
+  }
+
+  /***
+   * Sets the version text but only if we don't have stale data.
+   *
+   * @param sdk the SDK that was current at the time.
+   */
+  private void updateVersionTextIfCurrent(@NotNull FlutterSdk sdk, @NotNull String value) {
+    FlutterSdk current = FlutterSdk.forPath(getSdkPathText());
+    if (current == null || !sdk.getHomePath().equals(current.getHomePath())) {
```
pq: Maybe consider `Objects.equals(sdk.getHomePath(), current.getHomePath()))`?  (Gracefully handles `null`s -- not that there should be any!)

Issue #621, opened 2017-01-12
Opening samples leads to lots of analysis errors


mit-mit:
When a user opens `examples\flutter_gallery` for the first time, they will see 73 errors and 14 hints. I speculate this can be quite intimidating to a new user.

Could we just auto-run `flutter packages get` if we detect these are missing in the cache?

pq:
> Could we just auto-run flutter packages get if we detect these are missing in the cache?

👍 I think we could add something like this to our custom project open support (#316).

pq:
Fixed w/ #638 and #648.

PR #648, opened 2017-01-20
Add check on project open for out of date .packages (#621).

pq:
More improvement to address #621.

On project open, if a pubspec exists and is fresher than `.packages`, pop a toast.

screen shot 2017-01-19 at 4 50 45 pm

If they choose to update, `flutter packages upgrade` is run.

screen shot 2017-01-19 at 4 51 01 pm

@devoncarew

### DIFF
--- a/src/io/flutter/project/FlutterProjectOpenProcessor.java
+++ b/src/io/flutter/project/FlutterProjectOpenProcessor.java
@@ -6,6 +6,10 @@
 package io.flutter.project;
 
 import com.intellij.execution.ExecutionException;
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationListener;
+import com.intellij.notification.NotificationType;
+import com.intellij.notification.Notifications;
 import com.intellij.openapi.extensions.Extensions;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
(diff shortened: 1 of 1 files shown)


pq:
FWIW: still looking at adding module type but hit some speedbumps.  Will follow-up with that next.

src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -102,4 +114,33 @@ private ProjectOpenProcessor getDelegateImportProvider(@Nullable VirtualFile fil
...
   }
+
+  private static class PackagesOutOfDateNotification extends Notification {
+
+    @NotNull
+    private final Project myProject;
+
+    public PackagesOutOfDateNotification(@NotNull Project project) {
+      super("Flutter Packages", FlutterIcons.Flutter, "Package updates.",
+            null, "This project's packages are ready to" +
```
devoncarew: Some bikeshedding with the message: `The pubspec.yaml file has been modified since the last time the 'flutter packages get' was run; run that now?`.

devoncarew:
lgtm!

src/io/flutter/project/FlutterProjectOpenProcessor.java
pq: Let's look at some options together.  @mit-mit: your input would be great here too.

The biggest downside to a more precise message is that if the message is longer than a line, the actionable hyperlink gets hidden beneath the fold.

For example:

screen shot 2017-01-20 at 5 58 38 am

needs to be expanded to reveal:

screen shot 2017-01-20 at 5 58 43 am

An alternative might be to have the action above the fold but provide more details below?  Anyway, yeah, let's definitely refine in another pass.  (There'll be a few more rest assured! 😄 )

mit-mit: Isn't there a pretty big difference between 'get' and 'upgrade'? And doesn't missing packages just imply 'get' is needed', not 'upgrade'?

pq: 👍 

Downgrading to a `get` call: https://github.com/flutter/flutter-intellij/pull/652

PR #649, opened 2017-01-20
Fix flutter project open to not be so greedy.

pq:
* ensures Flutter post-processing only occurs on projects with a Flutter module

@devoncarew 

(FWIW still working out how to add the Flutter module type in this flow...)


devoncarew:
lgtm

pq:
Thanks!

PR #652, opened 2017-01-23
Prefer packages `get` over `upgrade` in project open (#648).

pq:
* perform packages `get` rather than `upgrade` on project import

Follow-up on: https://github.com/flutter/flutter-intellij/pull/648#discussion_r97264115.

@devoncarew @danrubel

### DIFF
--- a/src/io/flutter/project/FlutterProjectOpenProcessor.java
+++ b/src/io/flutter/project/FlutterProjectOpenProcessor.java
@@ -127,9 +127,10 @@ public class FlutterProjectOpenProcessor extends ProjectOpenProcessor {
     private final Project myProject;
 
     public PackagesOutOfDateNotification(@NotNull Project project) {
-      super("Flutter Packages", FlutterIcons.Flutter, "Package updates.",
-            null, "This project's packages are ready to" +
-                  " <a href=\"\">update</a>.",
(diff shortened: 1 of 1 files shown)


devoncarew:
We should tweak the text as well (`This project's packages are ready to update`).

pq:
> We should tweak the text as well (This project's packages are ready to update).

Right.  Thoughts?  Different text for the two cases (missing vs. out of date)?  It'd be great to keep the message simple and on one line...

devoncarew:
Yeah, +1 to being terse. Makes sense that we might need to specialize for the 2 use cases.

What I'd written before was: `The pubspec.yaml file has been modified since the last time the 'flutter packages get' was run; run that now?`, but that doesn't handle the missing use case, and might be a bit too long.

pq:
Updated message:

screen shot 2017-01-23 at 12 45 48 pm

We can refine this down the road to see if we can get the link above the fold.  (Or not!)

(In the missing case we just run `get` without asking which I think is the right flow.)

pq:
@devoncarew : LMK if this works.  Happy to bike shed a bit after some coffee. 😉

devoncarew:
lgtm! thanks for the word-smithing

PR #653, opened 2017-01-23
Add Flutter Module type on project open (#316).

pq:
* adds flutter module type (if missing)
* reloads project
* runs `flutter packages get`
* code cleanup (new `FlutterModuleUtils` class)

Fixes: #316.

### DIFF
--- a/src/io/flutter/project/FlutterProjectOpenProcessor.java
+++ b/src/io/flutter/project/FlutterProjectOpenProcessor.java
@@ -10,7 +10,9 @@ import com.intellij.notification.Notification;
 import com.intellij.notification.NotificationListener;
 import com.intellij.notification.NotificationType;
 import com.intellij.notification.Notifications;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.extensions.Extensions;
+import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.projectImport.ProjectOpenProcessor;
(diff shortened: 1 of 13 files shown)


src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -87,18 +94,27 @@ private void doPostOpenProcessing(@Nullable Project project) {
       return;
     }
 
-    // TODO(pq): handle adding Flutter module type.
-    if (!FlutterSdkUtil.hasFlutterModule(project)) {
-      return;
+    if (!FlutterModuleUtils.hasFlutterModule(project)) {
```
pq: FWIW here's the beef.  (Mostly shuffling existing source otherwise.)

src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -87,18 +94,27 @@ private void doPostOpenProcessing(@Nullable Project project) {
...
+      if (FlutterModuleUtils.usesFlutter(project)) {
+        final List<Module> modules = FlutterModuleUtils.findModulesWithFlutterContents(project);
+        if (modules.isEmpty()) {
+          LOG.warn(MessageFormat.format("No module found for {0}", project.getName()));
+          return;
+        } else if (modules.size() > 1){
```
devoncarew: I'm curious, why do you warn if there's more than one flutter module?

devoncarew:
lgtm

src/io/flutter/project/FlutterProjectOpenProcessor.java
pq: It's an odd case.  Essentially it means we've got 2 modules that have pubspecs with flutter deps (or the module type declared) and we're just picking the first one.  I don't expect this to happen often (or ever) but in case it does it may be nice to log something for the record.

PR #655, opened 2017-01-23
Collection utilities.

pq:
Some collection utilities to make working with streams a little easier.

@devoncarew @skybrian

### DIFF
--- a/src/io/flutter/utils/CollectionUtils.java
+++ b/src/io/flutter/utils/CollectionUtils.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright  2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.utils;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class CollectionUtils {
+  private CollectionUtils() {
+  }
+
+  public static <T> boolean anyMatch(@NotNull T[] in, @NotNull final Predicate<T> predicate) {
+    return anyMatch(Arrays.stream(in), predicate);
+  }
+
+  public static <T> boolean anyMatch(@NotNull final List<T> in, @NotNull final Predicate<T> predicate) {
+    return anyMatch(in.stream(), predicate);
+  }
+
+  public static <T> boolean anyMatch(@NotNull final Stream<T> in, @NotNull final Predicate<T> predicate) {
+    return in.anyMatch(predicate);
+  }
+
+  public static <T> List<T> filter(@NotNull T[] in, @NotNull final Predicate<T> predicate) {
+    return filter(Arrays.stream(in), predicate);
+  }
+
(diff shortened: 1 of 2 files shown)


src/io/flutter/utils/FlutterModuleUtils.java
```diff
@@ -59,21 +59,24 @@ public static boolean hasFlutterModule(@NotNull Project project) {
     // If not IntelliJ, assume a small IDE (no multi-module project support).
     // Look for a module with a flutter-like file structure.
     if (!PlatformUtils.isIntelliJ()) {
-      for (Module module : ModuleManager.getInstance(project).getModules()) {
-        if (usesFlutter(module)) {
-          return true;
-        }
+      if (CollectionUtils.anyMatch(getModules(project), FlutterModuleUtils::usesFlutter)) {
```
skybrian: As a style thing, I'm not convinced this is an improvement? A simple for-each loop in a small function is self-contained and universally understood.

(But I won't block this if others disagree.)

src/io/flutter/utils/FlutterModuleUtils.java
```diff
@@ -59,21 +59,24 @@ public static boolean hasFlutterModule(@NotNull Project project) {
     // If not IntelliJ, assume a small IDE (no multi-module project support).
     // Look for a module with a flutter-like file structure.
     if (!PlatformUtils.isIntelliJ()) {
```
skybrian: Nit: could reverse the test and exit early to avoid indenting the rest.

src/io/flutter/utils/FlutterModuleUtils.java
pq: I'm not entirely in love with java8's flavor of terseness and while I wouldn't argue for being prescriptive, in some cases I think things are made clearer.  For example,

```CollectionUtils.filter(getModules(project), m -> isFlutterModule(m) || usesFlutter(m))```

is I think nicer than it's alternative(s).

But anyway, I think you're right this is a style thing and I promise not to troll your PRs with suggestions to use these idioms!

File: src/io/flutter/utils/CollectionUtils.java [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+
+  public static <T> boolean anyMatch(@NotNull T[] in, @NotNull final Predicate<T> predicate) {
+    return anyMatch(Arrays.stream(in), predicate);
+  }
+
+  public static <T> boolean anyMatch(@NotNull final List<T> in, @NotNull final Predicate<T> predicate) {
+    return anyMatch(in.stream(), predicate);
+  }
+
+  public static <T> boolean anyMatch(@NotNull final Stream<T> in, @NotNull final Predicate<T> predicate) {
```
skybrian: I'm not sure we need this one. Writing in.anyMatch(predicate) in the caller seems better?

pq: Another matter of taste but I hear you. If you care, happy to remove (or make private).

skybrian: I am indifferent enough to leave the final call to someone else :-)

pq: 😄 

@danrubel @devoncarew ?

devoncarew:
Lgtm!

pq:
Thanks! 👍

Issue #656, opened 2017-01-24
Update flutter detection logic for merged pubspec and flutter.yaml


pq:
As of https://github.com/flutter/flutter/pull/7605, we'll need to re-jigger our detection logic.

Otherwise, we could potentially consider help in transitioning existing projects, or maybe the feedback from `flutter run` is enough?

pq:
I'll take a look at this today.

pq:
TODO: confirm that the current check works.

Another consequence is that we lose a place to hang the banner advertising flutter actions.

Issue #660, opened 2017-01-24
Update Dart plugin to suppress explorer pub actions for flutter pubspecs.


pq:
We should only offer the `flutter package ...` commands.

<img width="587" alt="screen shot 2017-01-24 at 1 28 21 pm" src="https://cloud.githubusercontent.com/assets/67586/22267849/235b464a-e23a-11e6-9da5-745a6f4714f5.png">

@alexander-doroshko : how do you feel about adding logic to the Dart plugin that suppresses these contributions in the Flutter module case?

Incidentally, we should do the same for the editor notification for `pubspec.yaml`s (#509) and once we do, we can re-contribute `FlutterYamlNotificationProvider.java` removed in #661.

alexander-doroshko:
SGTM. Some simple Flutter-specific things (like labels) can be fully done by the Dart plugin. Others can be suppressed by the Dart plugin and contributed by the Flutter plugin.

pq:
Great.  Thanks @alexander-doroshko !

Do you want to tackle that or should we throw together a PR?

alexander-doroshko:
PR is welcome!

pq:
👍

pq:
PR up for review: https://github.com/JetBrains/intellij-plugins/issues/472.

* suppresses pub actions in the editor, explorer and menu for Flutter module pubspecs in case Flutter is installed

pq:
https://github.com/JetBrains/intellij-plugins/pull/472 landed. 👍 

This will be in the next stable plugin push.

PR #661, opened 2017-01-24
Remove `flutter.yaml` handling logic (#656).

pq:
* remove checks for `flutter.yaml`
* remove editor notifications for `flutter.yaml` files
* update package explorer group to enable for flutter pubspecs

Fixes: #656.

Note that we've got a pile-up of actions suggested for `pubspec`s.  

screen shot 2017-01-24 at 1 28 21 pm

See: #660 for a proposal to fix this in the Dart Plugin.  Moreover, if the Dart editor notification for `pubspec`s gets suppressed we can recontribute `FlutterYamlNotificationProvider.java` (but keyed off Flutter pubspecs).

@devoncarew


src/io/flutter/FlutterUtils.java
```diff
@@ -30,10 +32,15 @@ public static void invokeAndWait(@NotNull Runnable runnable) throws ProcessCance
 
   @SuppressWarnings("BooleanMethodIsAlwaysInverted")
   public static boolean isFlutteryFile(@NotNull VirtualFile file) {
-    final String fileName = file.getName();
-    return file.getFileType() == DartFileType.INSTANCE ||
-           fileName.equals(FlutterConstants.FLUTTER_YAML) ||
-           fileName.equals(FlutterConstants.PUBSPEC_YAML);
+    return isDartFile(file) || isPubspecFile(file);
```
devoncarew: We _may_ want to keep `flutter.yaml` in the check for a release or two (to identify the project as a flutter project), in order to give people a few weeks to migrate.

src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -186,7 +186,6 @@ public static String getErrorMessageIfWrongSdkRootPath(final @NotNull String sdk
 
   public static boolean isFlutterProjectDir(@Nullable VirtualFile dir) {
     if (dir == null || !dir.isDirectory()) return false;
-    if (dir.findChild(FlutterConstants.FLUTTER_YAML) != null) return true;
```
devoncarew: Same here re: removing the check; we may want to keep but remove after a certain time.

devoncarew:
+1 to removing the UI and moving the contribution points.

We'll probably want a brief grace period for still identifying projects with flutter.yaml files as flutter projects.

lgtm

pq:
> We'll probably want a brief grace period for still identifying projects with flutter.yaml files as flutter projects.

On thinking about it, this logic was always a little suspect --- isn't it true that all flutter projects need to have a pubspec dependency as well?  IOW: I think this check alone was never sufficient and once we started parsing pubspecs was already safe to remove.  What do you think?

src/io/flutter/FlutterUtils.java
pq: See comment below.

src/io/flutter/sdk/FlutterSdkUtil.java
pq: And here too.  Net-net: I think the check was always fishy once we started reading pubspecs for Flutter dependencies.  Anyway, see my hypothesis below.

devoncarew:
> isn't it true that all flutter projects need to have a pubspec dependency as well?

Thinking about it, that sounds likely. I think I'm still rooted in a world where we used relative paths to reference flutter.

PR #664, opened 2017-01-26
Flutter dependency inspection.

pq:
* Flutter’s version of the check performed by the Dart Plugin  (soon suppressed) on `pubspec.yaml` edits
* Offers to run `flutter packages get` or `flutter packages update`

screen shot 2017-01-26 at 10 05 15 am

(Notice the pile-up of notifications --- https://github.com/JetBrains/intellij-plugins/pull/472 will suppress the Dart one.)

Still TODO:

* word-smithing (input welcome)
* consider auto-close of inspection banner on fix execution (not done by the Dart one either)

@devoncarew

### DIFF
--- a/resources/inspectionDescriptions/FlutterDependency.html
+++ b/resources/inspectionDescriptions/FlutterDependency.html
@@ -1,0 +1,14 @@
+<!--
+  ~ Copyright  2017 The Chromium Authors. All rights reserved.
+  ~ Use of this source code is governed by a BSD-style license that can be
+  ~ found in the LICENSE file.
+  -->
+
+<html>
+<body>
+Flutter package dependency check.
+<!-- tooltip end -->
+Checks whether a Flutter <code><a href="https://www.dartlang.org/tools/pub/pubspec.html">pubspec.yaml</a></code> has been edited
+since the last <a href="https://flutter.io/upgrading/l">package update</a>.
+</body>
+</html>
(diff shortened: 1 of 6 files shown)


pq:
RE: wordsmithing, one stab

screen shot 2017-01-26 at 10 15 54 am


(But it's too long...)

resources/inspectionDescriptions/FlutterDependency.html
```diff
@@ -0,0 +1,14 @@
...
+  ~ Use of this source code is governed by a BSD-style license that can be
+  ~ found in the LICENSE file.
+  -->
+
+<html>
+<body>
+Flutter package dependency check.
+<!-- tooltip end -->
+Checks whether a Flutter <code><a href="https://www.dartlang.org/tools/pub/pubspec.html">pubspec.yaml</a></code> has been edited
+since the last <a href="https://flutter.io/upgrading/l">package update</a>.
```
devoncarew: Just `https://flutter.io/upgrading` (to trailing `l`).

devoncarew:
2 questions:
- is it possible to add a flutter icon to the editor notification? I see it's not using the same api as the EditorNotificationPanel (use in `WrongModuleTypeNotificationProvider`). A flutter icon here would help the user identify where the notification is coming from
- what do you think about dropping the 'Open pubspec' link? The notification currently seems super long

lgtm either way

resources/inspectionDescriptions/FlutterDependency.html
pq: 👍  Thanks!

pq:
> what do you think about dropping the 'Open pubspec' link? The notification currently seems super long

Done.

As for the Flutter Icon I wondered that too but  I'm not sure if there's a way to add icons to inspections like we do for editor notifications but will dig a little deeper.  (If we can't, we might consider moving this to a notification but would have to look at what that loses us.)  Anyway, I'll address that in a follow-up.

Thanks!

PR #665, opened 2017-01-26
Dart Plugin API dependency round-up.

pq:
Round up of Dart Plugin API calls.

On-going.

This (and another I've got stashed) will make integrating with `HEAD` a bit easier.

@devoncarew

### DIFF
--- a/src/io/flutter/dart/DartPlugin.java
+++ b/src/io/flutter/dart/DartPlugin.java
@@ -8,7 +8,16 @@ package io.flutter.dart;
 import com.intellij.ide.plugins.IdeaPluginDescriptor;
 import com.intellij.ide.plugins.PluginManager;
 import com.intellij.openapi.extensions.PluginId;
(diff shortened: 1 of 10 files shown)


devoncarew:
ooo, very nice! lgtm

In case it's helpful, the list of APIs used from the Dart plugin is scraped out on each build: https://travis-ci.org/flutter/flutter-intellij/jobs/195638452#L260.

pq:
> In case it's helpful, the list of APIs used from the Dart plugin is scraped out on each build: https://travis-ci.org/flutter/flutter-intellij/jobs/195638452#L260.

Super helpful.  I wonder if we can get this down to a list we feel good about and then add a test that asserts our dependencies haven't changed?  This would have the benefit of forcing us to review all newly added dependencies.  Anyway, we're not quite ready for that but maybe something to aspire to?

devoncarew:
Yeah, +1 to being very intentional about what APIs we consume.

Issue #651, opened 2017-01-23
when running flutter doctor w/o a flutter sdk installed, you don't get any feedback


devoncarew:

pq:
A few things TODO:

1. A message that says "Flutter is not configured..." and an action to open preferences.
2. A link to flutter io setup instructions in prefs.

mit-mit:
Setup link: https://flutter.io/setup/

PR #671, opened 2017-01-27
Tools group visibility fixes;  SDK configuration dialog (#651).

pq:
Hides the tools action group when no Flutter project is active.
  * this addresses a race when a project is not active yet and action stubs do nothing
  * it also prevents Flutter actions from being wrongly run against non-Flutter projects

Opens a new Dialog when no SDK is configured and SDK actions are run.

screen shot 2017-01-27 at 8 54 52 am

(We can --- _and should_ --- noodle on dialog polish.)

Fixes: #651.

@devoncarew


devoncarew:
lgtm

PR #682, opened 2017-01-30
0.1.9

pq:
@devoncarew

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -23,7 +23,6 @@
     <![CDATA[
 0.1.9:
 <ul>
-  <li>preliminary windows support</li>
   <li>fixes to the visibility of the "Tools" menu</li>
   <li>inspection to detect pubspec modifications (that may imply out of date package dependencies)</li>
   <li>key bindings fixes</li>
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
Sorry, I still didn't review `FlutterProjectOpenProcessor` thoroughly. Are you sure you want to keep `isStrongProjectInfoHolder()` overridden?

pq:
Will fix this today: https://github.com/flutter/flutter-intellij/issues/684

resources/META-INF/plugin.xml
```diff
@@ -21,6 +21,18 @@
 
   <change-notes>
     <![CDATA[
+0.1.9:
+<ul>
+  <li>preliminary windows support</li>
```
devoncarew: let's back this out - there's still a bunch of work in flutter_tools

devoncarew:
lgtm; can also add a line item about webstorm support

pq:
> lgtm; can also add a line item about webstorm support

How about:

> initial webstorm IDE support

Or should we be more assertive?

resources/META-INF/plugin.xml
pq: Done.

PR #695, opened 2017-02-01
Fix 'flutter doctor' to work the second time it's run

### LINKED ISSUE #694: Running 'flutter doctor' only works the first time
To reproduce, run 'Flutter Doctor' twice. On the first time it works. On the second time, the 'Messages Flutter' window shows up, but it's blank.

Tracing through with the debugger, it seems we're ignoring the 'runOnActive' parameter if the console already exists. 

Reproduced at HEAD on Mac 10.12.2.

skybrian:
Fixes #694


devoncarew:
lgtm! god catch

Issue #701, opened 2017-02-01
reload and restart actions don't work when app launched in 'run' mode


devoncarew:
To repro:

- start an app with the `run` button (not `debug`)
- wait for it to start
- top reload buttons don't work; keybindings don't work (but the bottom reload buttons do work)

PR #702, opened 2017-02-01
fix the reload action when the app was launched from 'run'

devoncarew:
- fix the reload action when the app was launched from 'run'
- fix https://github.com/flutter/flutter-intellij/issues/701

@pq


devoncarew:
> If this is an IntelliJ API thing

Not certain; the docs a thin here

Issue #715, opened 2017-02-07
Editing pubspec.yaml leads to persistent 'pub get' banner


mit-mit:
## Steps to Reproduce

1. Create a new project in IJ
1. Open pubspec.yaml and add an empty line between existing lines 1 and 2
1. Open `lib/main.dart`
1. Wait for the 'Pubspec has been edited' banner, and hit 'Get dependencies'
1. Close and then re-open `lib/main.dart`

Expected result: No pub banner
Actual result: You get the 'Pubspec has been edited' banner again

pq:
This looks like a dup of #667.   The rub is that IDEA is lagging behind the local filesystem.  A way I can reliably get it back up to date is by manually opening the `.packages` file.  I'm trying to do the equivalent thing in source already but that appears flaky too:

        LocalFileSystem.getInstance().refreshAndFindFileByPath(workingDir.getPath() + "/" + FlutterConstants.PACKAGES_FILE);

@alexander-doroshko : any thoughts on your end?

pq:
Incidentally, just noticed an update waiting:

<img width="642" alt="screen shot 2017-02-08 at 5 00 05 am" src="https://cloud.githubusercontent.com/assets/67586/22738275/9887a8dc-edbb-11e6-94de-48b0969aab6d.png">

Will apply and re-try.  Fingers-crossed!

pq:
After a little more investigation I think this inspection was always a little fishy.  The docs for `com.intellij.openapi.vfs.VirtualFile#getTimeStamp` are clear that the value may get cached:

    Gets the timestamp for this file. Note that this value may be cached and may differ from
    the timestamp of the physical file.

cc @alexander-doroshko : since I think this applies equally to similar checks in the Dart Plugin.

alexander-doroshko:
As I wrote in #667 I think the problem is not with the timeStamp, but more likely with the inspection that is not rerun after timeStamp changes. So it simply doesn't check the updated timeStamp and continues to show the warning that was calculated previously. I didn't investigate it further, though.

pq:
Thanks for chiming in @alexander-doroshko!

> As I wrote in #667 I think the problem is not with the timeStamp,

I'm pretty confident the timestamp is lagging.  Like I said, I can reliably get it to update by manually opening the lagging file.  I can also confirm it with printfs.  Moreover, if I skip right to using `java.io.File` I don't see the issue.

alexander-doroshko:
[...]

Issue #667, opened 2017-01-27
Out of date pubspec inspection lags behind local filesystem changes.


pq:
To repro, click `get` or `update`:

<img width="979" alt="screen shot 2017-01-27 at 5 35 25 am" src="https://cloud.githubusercontent.com/assets/67586/22372597/720a4970-e452-11e6-8e89-9f258001f225.png">

And watch the inspection *not* go away...

The issue appears to be this check:

`pubspec.getTimeStamp() > packages.getTimeStamp()`

which takes a long time to catch up with the time stamps on the local file system.

Oddly, I *thought* I was addressing this exact issue with this code:

      sdk.run(COMMAND, pair.first, workingDir, new ProcessAdapter() {
        @Override
        public void processTerminated(ProcessEvent event) {
          // Refresh to ensure Dart Plugin sees .packages and doesn't mistakenly nag to run pub.
          LocalFileSystem.getInstance().refreshAndFindFileByPath(workingDir.getPath() + "/" + FlutterConstants.PACKAGES_FILE);
        }
      });

(in `io.flutter.actions.FlutterPackagesGetAction#perform`).

@alexander-doroshko : any thoughts?  Is there a better way to refresh the `.packages` file?  I don't think doing it in the inspection is correct (since it's a read action).

alexander-doroshko:
Inspection doesn't rerun when IDE understands that .packages file has been updated. Need to fix.

pq:
So this is a platform bug?  Thanks for clarifying that.

@alexander-doroshko : if you find a tracking bug, maybe link here?

pq:
In the meantime, since this is pretty frustrating for new users especially, is there any way to dismiss or force an inspection to re-run?

alexander-doroshko:
Need to think about a fix at Dart plugin level. BTW it looks flaky. Sometimes I can't reproduce the problem.

pq:
FWIW this problem changes a bit if we transition to using an editor notification instead (#670).

@alexander-doroshko : is the flakiness that the timestamps are lagging or that the inspection doesn't get re-run?

alexander-doroshko:
Inspection. But it could be fixable at plugin end, not in platform.

pq:
While the underlying problem doesn't appear to have gone away, https://github.com/flutter/flutter-intellij/commit/75ee5755fd1d5c10764f02200d51e1f18f335aba effectively works around it.  Optimistically closing.

Lootwig:
[...]

PR #716, opened 2017-02-08
Fix pubspec timestamp checking (#715, #667).

pq:
Bypasses the Virtual FS (which lags) in favor of a trip to `java.io`.

Fixes: #715, #667

@alexander-doroshko @devoncarew


devoncarew:
👍 lgtm

PR #723, opened 2017-02-13
Add support for loading a plugin config file from a Bazel Workspace

skybrian:
Finds the workspace based on the content roots. The ancestor directories of
each content root are searched for a WORKSPACE file. There should be a unique
workspace for each project.

Also, automatically detect changes to the configuration. This can be due
to a changed WORKSPACE file, config file, or a changed module configuration.

Also, implemented some test infrastructure to test it all, based on JUnit 4.

Not hooked up to anything yet.

### DIFF
--- a/src/io/flutter/bazel/PluginConfig.java
+++ b/src/io/flutter/bazel/PluginConfig.java
@@ -47,15 +47,14 @@ public class PluginConfig {
   /**
    * Reads plugin configuration from a file, if possible.
    */
-  public static @Nullable
-  PluginConfig load(@NotNull VirtualFile file) {
+  public static @Nullable PluginConfig load(@NotNull VirtualFile file) {
     final Computable<PluginConfig> readAction = () -> {
       try {
         final InputStreamReader input = new InputStreamReader(file.getInputStream(), "UTF-8");
         final Fields fields = GSON.fromJson(input, Fields.class);
         return new PluginConfig(fields);
       } catch (IOException e) {
-        LOG.debug("failed to load flutter plugin config", e);
+        LOG.warn("failed to load flutter plugin config", e);
         return null;
       }
     };
--- a/src/io/flutter/bazel/PluginConfig.java
+++ b/src/io/flutter/bazel/PluginConfig.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright  2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.bazel;
+
+import com.google.common.base.Objects;
+import com.google.gson.Gson;
+import com.google.gson.annotations.SerializedName;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.util.Computable;
+import com.intellij.openapi.vfs.VirtualFile;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+/**
+ * An in-memory snapshot of the flutter.json file from a Bazel workspace.
+ */
+public class PluginConfig {
+  private final @NotNull Fields fields;
+
+  private PluginConfig(@NotNull Fields fields) {
+    this.fields = fields;
+  }
+
+  public @Nullable String getFlutterDaemonScript() {
+    return fields.flutterDaemonScript;
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof PluginConfig)) return false;
+    final PluginConfig other = (PluginConfig)obj;
+    return Objects.equal(fields, other.fields);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(fields);
+  }
+
+  /**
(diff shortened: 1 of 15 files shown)


skybrian:
The test pass, but there is a disconcerting warning logged to the console. It doesn't seem to have anything to do with my code; just setting up and tearing down a heavy IdeaProjectTestFixture seems to do it. Any ideas?

ERROR: Local history is broken(version:6,current timestamp:Feb 12, 2017 11:36:20 PM,storage timestamp:Jan 19, 2017 6:52:50 PM,vfs timestamp:Jan 19, 2017 6:52:50 PM)
null
java.io.IOException: Recursive records found
	at com.intellij.history.core.ChangeListStorageImpl.doReadPrevSafely(ChangeListStorageImpl.java:280)
	at com.intellij.history.core.ChangeListStorageImpl.findFirstObsoleteBlock(ChangeListStorageImpl.java:271)
	at com.intellij.history.core.ChangeListStorageImpl.purge(ChangeListStorageImpl.java:237)
	at com.intellij.history.core.ChangeList.purgeObsolete(ChangeList.java:175)
	at com.intellij.history.integration.LocalHistoryImpl.disposeComponent(LocalHistoryImpl.java:137)
	at com.intellij.history.integration.LocalHistoryImpl.lambda$initComponent$0(LocalHistoryImpl.java:67)
	at com.intellij.openapi.util.ShutDownTracker.run(ShutDownTracker.java:60)
	at java.lang.Thread.run(Thread.java:745)

devoncarew:
Hey - I haven't looked into this PR yet, but just FYI, the two bots that are failing are the two on which we run unit tests. And a current limitation for those tests is that they can't reference IntelliJ classes. I ran into issues getting the dart plugin installed on the bots into the intellij product(s), which made the flutter plugin unhappy at runtime. Going forward, we definitely want to allow unit tests to reference IntelliJ classes (and to get some heavier-weight, integration test style tests running as well).

skybrian:
Thanks!

Maybe we should just run unit tests locally for now? It seems like having better test coverage (right now there are very few tests) is more important than running them on Travis?

devoncarew:
> Maybe we should just run unit tests locally for now?

sgtm. You can make this change in this PR, or I can in a separate PR.

devoncarew:
> Also, implemented some test infrastructure to test it all, based on JUnit 4.

We should standardize on JUnit 3.8 or 4; I don't care which, but it'd be useful to be consistent. /cc @pq

src/io/flutter/bazel/PluginConfig.java
```diff
@@ -0,0 +1,90 @@
...
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(fields);
+  }
+
+  /**
+   * Reads plugin configuration from a file, if possible.
+   */
+  public static @Nullable
+  PluginConfig load(@NotNull VirtualFile file) {
```
devoncarew: This line break looks odd - `public static @Nullable PluginConfig load(` instead?

pq:
> We should standardize on JUnit 3.8 or 4; I don't care which

👍 

I have a mild preference for JUnit 4, but don't care too strongly either way.

src/io/flutter/bazel/PluginConfig.java
```diff
@@ -0,0 +1,90 @@
...
+    final Computable<PluginConfig> readAction = () -> {
+      try {
+        final InputStreamReader input = new InputStreamReader(file.getInputStream(), "UTF-8");
+        final Fields fields = GSON.fromJson(input, Fields.class);
+        return new PluginConfig(fields);
+      } catch (IOException e) {
+        LOG.debug("failed to load flutter plugin config", e);
```
devoncarew: We should verify, but I believe that debug level will show an error to the user and offer to report to an issue tracker. Perhaps use `warn` level instead?

skybrian: Hmm, that's surprising. I will try it.

alexander-doroshko: It is LOG.error that writes stack trace to the log file and suggests user to submit it.
LOG.warn writes stack trace to log file silently.
LOG.info writes only message without stack to the log file.
LOG.debug by default does nothing at all unless you manually add corresponding category to Help | Debug Log Settings. With category added it writes message to the log file wihtout any stack trace.

skybrian:
I haven't used JUnit 4 much, but I do like to avoid complicated test class hierarchies (like LightPlatformCodeInsightFixtureTestCase and so on) since it tends to evolve into a mess.

It looks like Junit 4's @Rule annotation makes test fixtures easy to initialize.

src/io/flutter/bazel/PluginConfig.java
devoncarew: Ah, I mis-remembered. I only had concerns here about using the one that would display a dialog.

src/io/flutter/bazel/PluginConfig.java
skybrian: Oops! IDEA's formatter seems to prefer it this way for some reason and I've been fixing it manually.

skybrian:
Within a unit test, it appears that only errors are logged, unfortunately.

devoncarew:
lgtm!

skybrian:
I ran across a bug with malformed JSON (PluginConfig.load() throws JsonSyntaxException instead of returning null). But since nobody is using this config file yet, I'll fix it in another patch.

Issue #609, opened 2017-01-10
Project creation should validate legal project names.


pq:
In particular, we should make sure names are pub-friendly:

<img width="727" alt="screen shot 2017-01-10 at 9 51 45 am" src="https://cloud.githubusercontent.com/assets/67586/21817860/8c89ac7c-d71a-11e6-9041-5497dd9b243b.png">

pq:
Putting on-deck as it's hit a few folks (besides me!) --- see #687.

eseidelGoogle:
We already do some of this in `flutter create`, maybe it should do more?
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/commands/create.dart#L179

pq:
Good thinking @eseidelGoogle.  Probably we should.  The rub here is that we've already created the project in IntelliJ when we call `flutter create` so it will just bomb out on us when run.  A better flow would be live validation in the wizard (e.g., before running `create`).  IOW: I think we probably need validation in both places unfortunately.  But anyway, yeah, both validations should totally agree!

pq:
Forward pointer:  Section "16.33 Identifier Reference" of the [Dart language spec](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-408.pdf) has the grammar we want to crib.

mit-mit:
@pq did you complete this?

pq:
Not yet.  My PR is hung up on build issues.  It's also won't do anything in the context of WebStorm and other "small IDEs". I'll do some more poking later this week.

PR #724, opened 2017-02-13
Validate project names during creation (#609).

pq:
Fixes: #609.

screen shot 2017-02-13 at 4 58 54 am

screen shot 2017-02-13 at 4 58 33 am

(Needless to say, word-smithing welcome!)

@devoncarew

### DIFF
--- a/src/io/flutter/module/FlutterModuleBuilder.java
+++ b/src/io/flutter/module/FlutterModuleBuilder.java
@@ -191,9 +191,7 @@ public class FlutterModuleBuilder extends ModuleBuilder {
       }
 
       if (projectName.length() > FlutterConstants.MAX_PROJECT_NAME_LENGTH) {
-        throw new ConfigurationException("Flutter project name '" +
-                                         projectName +
-                                         "' is too long; must be less than " +
+        throw new ConfigurationException("Flutter project name is too long; must be less than " +
                                          FlutterConstants.MAX_PROJECT_NAME_LENGTH +
                                          " characters.");
       }
--- a/src/io/flutter/module/FlutterModuleBuilder.java
+++ b/src/io/flutter/module/FlutterModuleBuilder.java
@@ -6,8 +6,10 @@
 package io.flutter.module;
 
 import com.intellij.execution.ExecutionException;
+import com.intellij.ide.projectWizard.ProjectSettingsStep;
 import com.intellij.ide.util.projectWizard.ModuleBuilder;
 import com.intellij.ide.util.projectWizard.ModuleWizardStep;
+import com.intellij.ide.util.projectWizard.SettingsStep;
 import com.intellij.ide.util.projectWizard.WizardContext;
 import com.intellij.openapi.Disposable;
 import com.intellij.openapi.diagnostic.Logger;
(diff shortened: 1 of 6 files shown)


src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -140,4 +153,52 @@ public boolean validate() throws ConfigurationException {
...
+      final WizardContext context = mySettingsStep.getContext();
+      final String projectName = context.getProjectName();
+
+      if (!FlutterUtils.isValidDartdentifier(projectName)) {
+        throw new ConfigurationException("Invalid Flutter project name '" + projectName + "'; must be a valid Dart identifier.");
+      }
+
+      if (projectName.length() > FlutterConstants.MAX_PROJECT_NAME_LENGTH) {
+        throw new ConfigurationException("Flutter project name '" +
+                                         projectName +
```
devoncarew: We may not want to display the project name here (given that it will be pretty long).

src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -61,6 +65,15 @@ public void setupRootModel(ModifiableRootModel model) throws ConfigurationExcept
   }
 
   @Override
+  public ModuleWizardStep modifyStep(SettingsStep step) {
+    if (getModuleType() != null && step instanceof ProjectSettingsStep) {
```
devoncarew: From the travis failure, it looks like ProjectSettingsStep doesn't exist in non-idea products. Is this whole class (FlutterModuleBuilder.java) idea specific? Perhaps we should call out such using idea-contribs.xml?

devoncarew:
We'll likely need to separate the source out so we don't try and compile idea specific source against a non-idea product...

src/io/flutter/module/FlutterModuleBuilder.java
pq: Good idea.

src/io/flutter/module/FlutterModuleBuilder.java
pq: Not sure.  Looking at this now...

Maybe @alexander-doroshko will beat me to it though? 😉

alexander-doroshko: Project Wizard paths are completely different in different IDEs.

See the location of the `ProjectSettingsStep.java` file in the repository to get a hint ;)

pq: Got it.  Thanks!

pq:
I did a bit of refactoring and (I believe) got the IDEA bits off the small IDE code path, however the WebStorm build is still failing.

@devoncarew : could you give me a quick summary of how IDEA-specific bits are supposed to be skipped in the WebStorm build?  From a quick glance, it looks like that build tries to compile everything...

devoncarew:
> could you give me a quick summary of how IDEA-specific bits are supposed to be skipped in the WebStorm build

Just seeing this now; @pq, this was before our conversation yesterday? Just to close the loop, I think we want to have Idea specific bits in an `idea` sub package, and adjust out fileset in the ant script based on idea / not-idea.

pq:
Yep.  It was before we chatted.  I've got a path forward.  👍

devoncarew:
Re: this PR, I just noticed that the `AbstractModuleBuilder` class has a `public boolean validateModuleName(@NotNull String moduleName) throws ConfigurationException` method that we may be able to implement and take advantage of.

pq:
Oh man.  Hard to imagine I missed that.  Looking now.  Thanks!

devoncarew:
Do not know what the UX of the feedback from that method looks like, but thought it worth mentioning here :)

pq:
Closing out in favor of a fresh take. 👍

PR #725, opened 2017-02-13
Update .travis.yml

devoncarew:
- disable unit testing on the bots (we'll rely on manually running the tests); will file an associated issue

@skybrian


skybrian:
LGTM. Thanks!

PR #728, opened 2017-02-13
0.1.10

pq:
* fixes to pubspec timestamp checking
* analytics events for run, debug, and process stop
* fix to `flutter doctor` to better support multiple runs
* fix to the reload action for apps launched from 'run'

@devoncarew


devoncarew:
lgtm

PR #740, opened 2017-02-15
Update known issues

### LINKED ISSUE #316: Opening an existing Flutter project that is missing .iml/.idea files causes confusing errors

mit-mit:
Remove fixed issue (Fixes #316)

Add two new issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,4 +23,6 @@ A very brief summary of the getting started guide linked from above:
 
 Please note the following known issues:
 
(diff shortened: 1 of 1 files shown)


devoncarew:
lgtm

PR #746, opened 2017-02-20
Various device menu fixes

skybrian:
- Start showing the device menu for Bazel users. In this case they don't have
to configure a Flutter SDK. Instead, the device menu will be shown based on the
project's content roots.

By default, a content root containing 'flutter' will trigger the device menu.
This can be overridden by the 'directoryPatterns' setting in flutter.json.

- Show or hide device menu when the user changes the Dart or Flutter SDK,
or the Bazel workspace changes.

- Start/stop/restart the 'flutter daemon' process when the configuration changes.

- Fixed a bug where Flutter SDK changes would not be detected when the user changes
the Dart SDK.

- Made API clearer for attaching FileRoot or Refreshable to a Disposable.

- Refreshable class improvements:
  - add a callback when a value is unpublished
  - make sure the variable doesn't change while a Swing event handler is
    running.

### DIFF
--- a/src/io/flutter/bazel/PluginConfig.java
+++ b/src/io/flutter/bazel/PluginConfig.java
@@ -41,7 +41,7 @@ public class PluginConfig {
    *
    * <p>The path should be relative to the workspace root.
    */
-  public boolean withinFlutterDirectory(String path) {
+  public boolean withinFlutterDirectory(@NotNull String path) {
     if (directoryPatterns == null) {
       // Default if unconfigured.
       return path.contains("flutter");
--- a/src/io/flutter/bazel/PluginConfig.java
+++ b/src/io/flutter/bazel/PluginConfig.java
@@ -6,7 +6,9 @@
 package io.flutter.bazel;
 
 import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
 import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
 import com.google.gson.annotations.SerializedName;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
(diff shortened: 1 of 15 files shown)


src/io/flutter/bazel/PluginConfig.java
```diff
@@ -15,21 +17,44 @@
...
+   *
+   * <p>The path should be relative to the workspace root.
+   */
+  public boolean withinFlutterDirectory(String path) {
+    if (directoryPatterns == null) {
+      // Default if unconfigured.
+      return path.contains("flutter");
+    }
+
+    for (Pattern p : directoryPatterns) {
```
devoncarew: Do the normal directoryPatterns we have configured contain 'flutter'?

devoncarew:
lgtm, but @pq should take a quick look

(as an aside, it looks like a few changes could have been done independent of this PR - perhaps next time break them out, to take the edge off a large PR?)

src/io/flutter/bazel/PluginConfig.java
```diff
@@ -15,21 +17,44 @@
...
   }
 
-  public @Nullable String getFlutterDaemonScript() {
-    return fields.flutterDaemonScript;
+  /**
+   * Returns true if the given path is within a flutter project in this workspace.
+   *
+   * <p>The path should be relative to the workspace root.
+   */
+  public boolean withinFlutterDirectory(String path) {
```
pq: Perhaps @NotNull here since it looks like it shouldn't be?

src/io/flutter/bazel/Workspace.java
```diff
@@ -39,13 +40,87 @@ private Workspace(@NotNull VirtualFile root, @Nullable PluginConfig config) {
...
+    }
+    return result.build();
+  }
+
+  /**
+   * Returns a VirtualFile's path relative to the workspace's root directory.
+   *
+   * <p>Returns null for the workspace root or anything outside the workspace.
+   */
+  public @Nullable String getRelativePath(VirtualFile file) {
```
pq: Tag as @Nullable?

src/io/flutter/bazel/Workspace.java
```diff
@@ -39,13 +40,87 @@ private Workspace(@NotNull VirtualFile root, @Nullable PluginConfig config) {
...
+    }
+    return null;
+  }
+
+  /**
+   * Returns true if the given path is within a flutter project.
+   *
+   * <p>The path should be relative to the workspace root.
+   */
+  public boolean withinFlutterDirectory(String path) {
```
pq: @NotNull?

src/io/flutter/sdk/FlutterSdkManager.java
```diff
@@ -5,10 +5,12 @@
  */
 package io.flutter.sdk;
 
+import com.google.common.base.Objects;
```
pq: I haven't checked but is this adding a new dependency?  (I guess if travis is happy, we're OK.)

pq:
> as an aside, it looks like a few changes could have been done independent of this PR 

👍 Yes, please.  This looks more like 3 independent lines of thought to me and as a general rule I'd prefer to see each in it's own PR...

Having said that, LGTM. 😄

src/io/flutter/bazel/PluginConfig.java
skybrian: Done

src/io/flutter/bazel/PluginConfig.java
skybrian: It's not checked in yet, I'm thinking of having at least "flutter" and "mobile".

src/io/flutter/bazel/Workspace.java
skybrian: Done

src/io/flutter/bazel/Workspace.java
skybrian: Done

src/io/flutter/sdk/FlutterSdkManager.java
skybrian: Guava 19 is included with IDEA CE and is used in other plugins including Dart.

PR #755, opened 2017-02-22
Work around incorrectly reported device platform

### LINKED ISSUE #754: Android emulator doesn't disappear from device menu when shut down
The problem seems to be that 'flutter daemon' reports an inconsistent 'platform' field when starting up versus shutting down.

In the debugger, when a device is added, I see:
```
id: emulator-5554
name: Android SDK built for x86 64
platform: android-x64
```

On shutdown, the "id" and "name" fields are the same, but platform is incorrectly reported as:
```
platform:  android-arm
```

This seems like a bug in 'flutter daemon'. However, we can also fix it in the Flutter plugin by just comparing device entries using the id.

skybrian:
Fixes #754

### DIFF
--- a/src/io/flutter/run/daemon/FlutterDaemonControllerHelper.java
+++ b/src/io/flutter/run/daemon/FlutterDaemonControllerHelper.java
@@ -380,11 +380,11 @@ class FlutterDaemonControllerHelper {
 
   private void eventDeviceRemoved(@NotNull DeviceRemovedEvent removed, @NotNull FlutterDaemonController controller) {
     synchronized (myLock) {
-      myController.getService().getConnectedDevices().stream().filter(
-        device -> device.deviceName().equals(removed.name) &&
-                  device.deviceId().equals(removed.id) &&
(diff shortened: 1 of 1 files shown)


eseidelGoogle:
Is flutter_tools doing anything wrong here?  Should this workaround (not really sure what it does) be in flutter_tools instead?

pq:
Maybe add a link to a new corresponding bug tracking a fix in Flutter (as per the comment in #754?)

devoncarew:
this fix lgtm (comparing by id), but we should also the issue in flutter_tools (I'll file a tracking bug)

skybrian:
I called it a workaround, but I'm not sure we'll ever need to revert it? Seems like the device ID should be a unique primary key for a device in the device list? Could we have used a LinkedHashMap here?

devoncarew:
I don't think there's a strong guarantee of uniqueness (guid), but I think they're reasonably unique. I don't think we'll need to revert this change.

PR #756, opened 2017-02-22
improve the location of the reload actions in the Run menu

devoncarew:
- improve the location of the reload actions in the Run menu

screen shot 2017-02-22 at 2 04 38 pm

@pq @skybrian

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -209,7 +209,7 @@
       <separator/>
       <add-to-group group-id="RunContextGroup" anchor="last"/>
       <add-to-group group-id="ToolbarRunGroup" anchor="last"/>
-      <add-to-group group-id="RunMenu" anchor="last"/>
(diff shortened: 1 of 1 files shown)


skybrian:
LGTM

Regarding the menu, I wonder if "Reload Flutter App" and "Restart Flutter App" would be clearer?

pq:
LGTM

devoncarew:
> Regarding the menu, I wonder if "Reload Flutter App" and "Restart Flutter App" would be clearer?

We have had a bit of user confusion in terms of the difference between hot reload and a full restart. I think `full restart` helps imply that we're doing more work than a incremental reload.

I'll land this as is, but we may want to noodle about whether these names are the ones we're happy with - that they do as much as they can to separate the two terms. /cc @johnmccutchan @mit-mit @sethladd

skybrian:
I'd like to see "Flutter" somewhere in the menu item's name, since IDEA is also used to work on non-Flutter apps.

PR #759, opened 2017-02-23
Update README.md

mit-mit:


mit-mit:
Merging

TBR @devoncarew

devoncarew:
👍 

We might also want to add a note that this will be fixed in the upcoming 2017.1 IntelliJ release.

PR #761, opened 2017-02-24
some additional console folding for ios

devoncarew:
- some additional console folding for ios

This last bit of folding means that you don't need to scroll the console to see all the initial output from an iOS simulator run.

@pq @skybrian


skybrian:
LGTM

PR #762, opened 2017-02-24
some cleanup to our use of the message bus

devoncarew:
- some cleanup to our use of the message bus

This converts to sending launch events on the project bus, not the application one. I think this is more correct, in that you don't want to have flutter views listening to launches in other windows.

@skybrian


skybrian:
LGTM

PR #766, opened 2017-02-27
Introduce DaemonApi and DaemonEvent

skybrian:
This is a rewrite of the code to interact with an external process
using the Flutter Daemon protocol [1].

The classes aren't hooked up to anything yet. (Migration will happen
in a separate patch.)

[1] https://github.com/flutter/flutter/wiki/The-flutter-daemon-mode

### DIFF
--- a/src/io/flutter/run/daemon/DaemonApi.java
+++ b/src/io/flutter/run/daemon/DaemonApi.java
@@ -1,0 +1,340 @@
+/*
+ * Copyright  2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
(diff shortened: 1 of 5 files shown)


src/io/flutter/run/daemon/DaemonApi.java
```diff
@@ -0,0 +1,340 @@
...
+    String getMessage() {
+      return message;
+    }
+  }
+
+  /**
+   * A pending command to a Flutter process.
+   */
+  private static class Command<T> {
+    final String method;
```
pq: Not a big deal but if this is nullable, maybe annotate it so?  (I notice the next field is annotated.)

pq:
As per our conversation earlier, let's check in before landing in case we want to do a quick patch to the `0.1.11` release.

PR #770, opened 2017-02-28
updates to Refreshable

skybrian:
- change API to optionally pass a Request to the callback.
This provides access to the previous version and isCancelled() method.
- expose getState() with the current status: busy/idle/closed
- Fire events for state changes, not just value changes.

Avoid starting and immediately killing the daemon process due to another event arriving.


pq:
As with the other PR, let's touch base before landing.

skybrian:
Yes, the idea with Refreshable is to have a simple foundation for reactive-style programming in the Swing environment, but it turned out to be more complicated than I hoped. Multithreaded code is hard. :-)

PR #771, opened 2017-02-28
Move device list tracking to a separate service

skybrian:
Keeps track of the device list in the DeviceDaemon class. This
avoids a possible bug due to lost events when the process
restarts.

Simpified event handling when updating the device list UI. The
same code gets executed regardless of what changed.

### DIFF
--- a/src/io/flutter/run/daemon/DeviceSelection.java
+++ b/src/io/flutter/run/daemon/DeviceSelection.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright  2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.run.daemon;
+
+import com.google.common.collect.ImmutableList;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.*;
+
+/**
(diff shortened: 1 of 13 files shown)


src/io/flutter/run/daemon/DeviceSelection.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private final ImmutableList<FlutterDevice> devices;
+  private final @Nullable FlutterDevice selection;
+
+  private DeviceSelection(ImmutableList<FlutterDevice> devices, @Nullable FlutterDevice selected) {
+    this.devices = devices;
+    this.selection = selected;
+  }
+
+  @NotNull ImmutableList<FlutterDevice> getDevices() {
```
pq: How are we ensuring that this is not null?  It seems like `devices` the field (or constructor arg) should be annotated.  Or?

skybrian: The constructor is private and used in three places, all within this file. But sure, I can add the annotation.

pq: No worries.

skybrian:
The bigger picture here is that we were using the flutter daemon protocol for two different kinds of external processes, polling for devices and running Flutter apps. For historical reasons this code was mixed together, making it harder to tell which case it's handling.  So, I thought I'd separate them.

They will still use common code for the Flutter daemon protocol itself. (After the next patch lands.)

Issue #772, opened 2017-02-28
exception from the flutter view


devoncarew:
## What happened

Closing the project window while a flutter app was running.

## Version information

Flutter plugin `0.1.11`

Flutter • channel master • https://github.com/flutter/flutter.git
Framework • revision 9a9a3ffd11 (4 hours ago) • 2017-02-27 18:01:21
Engine • revision 0edcbb7db4
Tools • Dart 1.23.0-dev.0.0

## Exception

Already disposed: Project (Disposed) untitled_12

```
java.lang.AssertionError: Already disposed: Project (Disposed) untitled_12
	at com.intellij.openapi.components.impl.ComponentManagerImpl.getComponent(ComponentManagerImpl.java:172)
	at com.intellij.openapi.wm.ToolWindowManager.getInstance(ToolWindowManager.java:37)
	at io.flutter.view.FlutterView.lambda$updateIcon$0(FlutterView.java:133)
	at com.intellij.openapi.application.TransactionGuardImpl$6.run(TransactionGuardImpl.java:318)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.a(LaterInvocator.java:417)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:401)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:726)
	at com.intellij.ide.IdeEventQueue.g(IdeEventQueue.java:843)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:679)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:391)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
[...]

PR #774, opened 2017-02-28
fix a use after dispose exception

devoncarew:
- fix https://github.com/flutter/flutter-intellij/issues/772

@pq @skybrian


pq:
I'm guessing we want to add this to a patch release?  Or a quick `0.1.12`?

devoncarew:
Yeah, we probably want to batch this up for a patch release.

PR #776, opened 2017-02-28
device list visibility fixes for Bazel

skybrian:
For Bazel, only show the device list when we are in a Flutter project (according to the Workspace).

If the Bazel repo has a script to start the device daemon, prefer that, but fall back to the Flutter SDK if configured.

Fixes a bug where the device menu would not show up because the user configured a Flutter SDK (for open source work) but forgot they did that. The device list should still appear.

Also, log when the device daemon starts and stops (at INFO level).

### DIFF
--- a/src/io/flutter/run/daemon/DeviceDaemon.java
+++ b/src/io/flutter/run/daemon/DeviceDaemon.java
@@ -60,6 +60,7 @@ class DeviceDaemon {
    * Kills the process.
    */
   public void shutdown() {
+    LOG.info("shutting down Flutter device poller: " + command.toString());
     controller.forceExit();
   }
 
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/daemon/DeviceDaemon.java
```diff
@@ -69,30 +70,42 @@ public void shutdown() {
...
+    }
 
-      try {
-        final String path = FlutterSdkUtil.pathToFlutterTool(sdk.getHomePath());
-        return new Command(sdk.getHomePath(), path, ImmutableList.of("daemon"));
-      }
-      catch (ExecutionException e) {
-        LOG.warn("Unable to calculate command to watch Flutter devices", e);
-        return null;
+    // See if the Bazel workspace provices a script.
```
pq: provices => provides

skybrian: Done

PR #778, opened 2017-03-01
Replace FlutterDaemonService with FlutterAppService

skybrian:
This migrates the implementation to use DaemonApi/DaemonEvent
for communicating with the flutter process.

Also, clean up FlutterApp and FlutterDevice.


src/io/flutter/run/daemon/FlutterAppService.java
```diff
@@ -0,0 +1,241 @@
+/*
+ * Copyright 2016 The Chromium Authors. All rights reserved.
```
pq: 2017?  (Or wait, this is a rename...)

pq: Yeah, I see it was a rename.  Feel free to ignore.

alexander-doroshko: Haven't you enabled 'Update Copyright' in the Commit dialog yet? ;-)
image

PR #781, opened 2017-03-02
Should run Flutter app by right-clicking on main.dart

### LINKED ISSUE #780: Can't run Flutter apps by right-clicking on the file with the main function
To reproduce
  - open hello world
  - delete all run configurations
  - make sure you have a Flutter device in the device menu
  - in the Project tree, right-click on main.dart and select 'Run'.

Expected to run as a Flutter app. Actually runs the main method as a Dart app.

skybrian:
Fixes #780

### DIFF
--- a/src/io/flutter/run/FlutterRunConfigurationProducer.java
+++ b/src/io/flutter/run/FlutterRunConfigurationProducer.java
@@ -6,6 +6,7 @@
 package io.flutter.run;
 
 import com.intellij.execution.actions.ConfigurationContext;
+import com.intellij.execution.actions.ConfigurationFromContext;
 import com.intellij.execution.actions.RunConfigurationProducer;
 import com.intellij.openapi.roots.ProjectRootManager;
 import com.intellij.openapi.util.Ref;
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/FlutterRunConfigurationProducer.java
```diff
@@ -51,6 +52,12 @@ public boolean isConfigurationFromContext(final @NotNull FlutterRunConfiguration
     return dartFile != null && dartFile.getPath().equals(configuration.getRunnerParameters().getFilePath());
   }
 
+  @Override
+  public boolean shouldReplace(ConfigurationFromContext self, ConfigurationFromContext other) {
+    // Prefer Flutter runner to plain Dart runner for Flutter code.
+    return other.getConfigurationType().getId().equals("DartCommandLineRunConfigurationType");
```
pq: Could we do something like this instead?

    other.getConfigurationType() == DartCommandLineRunConfigurationType.getInstance()

Or is there a good reason to use the String ID?

src/io/flutter/run/FlutterRunConfigurationProducer.java
```diff
@@ -62,7 +69,7 @@ private static VirtualFile getRunnableFlutterFileFromContext(final @NotNull Conf
         ProjectRootManager.getInstance(context.getProject()).getFileIndex().isInContent(virtualFile) &&
         !DartWritingAccessProvider.isInDartSdkOrDartPackagesFolder(psiFile.getProject(), virtualFile) &&
         DartResolveUtil.getMainFunction(psiFile) != null &&
-        hasImport((DartFile)psiFile, "package:flutter/material.dart")) { // TODO Determine flutter imports
+        hasImport((DartFile)psiFile, "package:flutter/widgets.dart")) {
```
pq: I wonder if there isn't an even more inclusive heuristic.  What about `package:flutter/**`?

src/io/flutter/run/FlutterRunConfigurationProducer.java
skybrian: I wanted to avoid any possible breakage if the Dart plugin classes change. (How do we handle dependencies like that?)

pq: Right. It's a hard call.  We don't have a good story for this.  Let's say a string is fine but maybe store it as a constant in the `DartPlugin` class (that's where we're trying to consolidate our plugin dependencies)?  Alternatively, add a method there?  Something like:

    static boolean isDartCommandLineRunConfig(ConfigurationFromContext config) { ... }

skybrian: Looking at this more closely, it seems this class already uses a lot of other Dart plugin code? For example, DartFile, DartImportStatement, DartResolveUtil.

src/io/flutter/run/FlutterRunConfigurationProducer.java
skybrian: Done

skybrian:
Okay, please take another look.

src/io/flutter/run/FlutterRunConfigurationProducer.java
```diff
@@ -6,85 +6,133 @@
...
-        !DartWritingAccessProvider.isInDartSdkOrDartPackagesFolder(psiFile.getProject(), virtualFile) &&
-        DartResolveUtil.getMainFunction(psiFile) != null &&
-        hasImport((DartFile)psiFile, "package:flutter/material.dart")) { // TODO Determine flutter imports
-      return virtualFile;
+  private static VirtualFile getFlutterEntryFile(final @NotNull ConfigurationContext context) {
+    final DartFile dart = getDartFile(context);
+    if (dart == null) return null;
+
+    if (DartResolveUtil.getMainFunction(dart) == null) return null;
+    if (findImportUrls(dart).noneMatch((url) -> url.startsWith("package:flutter/"))) return null;
```
pq: If this becomes more generally useful we may want to extract this test into a method but we can cross that bridge when we get to it.

Issue #739, opened 2017-02-15
Auto-gen a run config for imported projects with a `lib/main.dart`


mit-mit:
Opening projects with no IntelliJ metadata is a lot better with the recent fixes. Code analysis works as expected!

However, the project still has no run config, and adding one is a number of steps, incl. having to know some specific paths. Should we consider always adding a run config if there are none?

## Steps to Reproduce

1. Create a new project: `flutter create ijtest`
1. Delete IJ metadata: `cs ijtest; rm ijtest.iml; rm -rf .idea`
1. Open project: `idea .`

-> Project analyses but cannot be run

## Version info

```
Running flutter doctor...
[✓] Flutter (on Mac OS, channel master)
    • Flutter at /Users/mit/dev/github/flutter
    • Framework revision 77e53016f2 (2 hours ago), 2017-02-15 09:27:05
    • Engine revision b64c88766d
    • Tools Dart version 1.22.0-dev.10.6

[✓] Android toolchain - develop for Android devices (Android SDK 25.0.0)
    • Android SDK at /Users/mit/Library/Android/sdk
    • Platform android-25, build-tools 25.0.0
    • ANDROID_HOME = /Users/mit/Library/Android/sdk
    • Java(TM) SE Runtime Environment (build 1.8.0_112-b16)

[✓] iOS toolchain - develop for iOS devices (Xcode 8.2.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 8.2.1, Build version 8C1002
    • ios-deploy 1.9.0

[✓] IntelliJ IDEA Ultimate Edition (version 2016.3.4)
    • Dart plugin version 163.13137
    • Flutter plugin version 0.1.9.1
```

pq:
@devoncarew : is it true that the canonical entry-point will (always) be `lib/main.dart`?  If so, any reason we couldn't auto-generated a run config if that file exists?

mit-mit:
Some apps might now have that file (for example /examples/layers/ doesn't), but if that file exists it's pretty likely it's the entry point,.

devoncarew:
Auto-generation if lib/main.dart exists sounds good -

mit-mit:
Tentatively tagging this for next release

ajharry69:
Kindly check @omar1betar answer from here https://github.com/flutter/flutter-intellij/issues/943

PR #783, opened 2017-03-03
Auto create a run config on project open (#739).

pq:
Fixes: #739.

@devoncarew @skybrian

### DIFF
--- a/src/io/flutter/project/FlutterProjectOpenProcessor.java
+++ b/src/io/flutter/project/FlutterProjectOpenProcessor.java
@@ -15,6 +15,7 @@ import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.extensions.Extensions;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.Pair;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.projectImport.ProjectOpenProcessor;
 import icons.FlutterIcons;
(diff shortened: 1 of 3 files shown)


src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -103,7 +104,14 @@ else if (modules.size() > 1) {
           LOG.warn(MessageFormat.format("{0} contains {1} modules.", project.getName(), modules.size()));
```
skybrian: It's not part of your change, but I'm wondering if this should really be a warning?
If there is more than one module, perhaps just return?

src/io/flutter/utils/FlutterModuleUtils.java
```diff
@@ -74,6 +84,69 @@ public static boolean usesFlutter(@NotNull Project project) {
...
+
+      parameters.setWorkingDirectory(workingDir.getPath());
+
+      runManager.addConfiguration(settings, false);
+      runManager.setSelectedConfiguration(settings);
+    }
+  }
+
+  /**
+   * Returns a pair describing the main file (`lib/main.dart`) and it's content root for
```
skybrian: it's -> its

src/io/flutter/utils/FlutterModuleUtils.java
```diff
@@ -74,6 +84,69 @@ public static boolean usesFlutter(@NotNull Project project) {
...
+   * the given module.  Note that if there is more than one defined main in the given module,
+   * the first will be returned.  If none is found, returns `null`.
+   */
+  @Nullable
+  public static Pair<VirtualFile, VirtualFile> findFlutterMain(@NotNull Module module) {
```
skybrian: Optional: maybe a tiny class would be better, just to avoid confusing the files? (I tend to avoid Pair.)

```
public class FileWithContext {
  public final VirtualFile file;
  public final VirtualFile contextRoot;
  ...
}
```

src/io/flutter/project/FlutterProjectOpenProcessor.java
pq: I *believe* this will just get logged and so should go pretty much unnoticed unless folks are seeing issues in which case maybe it's useful?  That said, I'm up for a re-think.

src/io/flutter/utils/FlutterModuleUtils.java
pq: Thanks!

src/io/flutter/utils/FlutterModuleUtils.java
pq: 👍  I avoid things like `Pair` generally too so yeah, I'll fix it.

Thanks!

pq:
Thanks!

Issue #643, opened 2017-01-19
bazel launch config should have more info pre-populated


devoncarew:
From feedback:

> You should be able to derive working directory and list of targets etc from the project. launch script location also should probably be pre-populated.

There are several things we can infer here or pre-populate.

devoncarew:
@skybrian, this is largely done, right? The directory and the launch script are pre-populated, and we're punting on the target discovery?

skybrian:
Yes, I don't plan to do more work. For the remainder (that we're punting on), should we keep this bug open or start a new one?

devoncarew:
> list of targets etc

If we don't plan to do that work, I think we should just close the issue. If we do think we'll get to it, then opening another issue for the backlog makes sense.

PR #787, opened 2017-03-07
Populate work dir and launch script in new Bazel run configs

skybrian:
* Uses the workspace root as the work dir by default.
* Gets the launch script from flutter.json
* Sets the fields only if the corresponding template field is blank.

Partial fix for #643 
(Doesn't try to calculate the target, which seems harder.)

### DIFF
--- a/src/io/flutter/run/bazel/FlutterBazelRunConfigurationType.java
+++ b/src/io/flutter/run/bazel/FlutterBazelRunConfigurationType.java
@@ -5,6 +5,7 @@
  */
 package io.flutter.run.bazel;
 
+import com.intellij.execution.ExecutionBundle;
 import com.intellij.execution.configurations.ConfigurationFactory;
 import com.intellij.execution.configurations.ConfigurationTypeBase;
 import com.intellij.execution.configurations.RunConfiguration;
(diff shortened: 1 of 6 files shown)


src/io/flutter/run/bazel/FlutterBazelRunConfigurationType.java
```diff
@@ -32,14 +32,25 @@ public Factory(FlutterBazelRunConfigurationType type) {
...
     public RunConfiguration createConfiguration(String name, RunConfiguration template) {
-      // Override the default name which is always "Unnamed".
-      return super.createConfiguration(template.getProject().getName(), template);
+      // Called in two cases:
+      //   - When creating a non-template config from a template.
+      //   - whenever the run configuration editor is open (for creating snapshots).
+      // In the first case, we want to override the defaults from the template.
+      // In the second case, don't change anything.
+      //   (But the name doesn't matter; it will immediately be overwritten.)
+      if ((name.equals("Unnamed") || name.startsWith("Unnamed (")) && template instanceof BazelRunConfig) {
```
pq: I know you didn't add this here but I wonder if there isn't a less hacky way to test for the "Unnamed" case. (And is that name localized?)  Maybe @alexander-doroshko knows?

alexander-doroshko: As I see for other configuration types `createConfiguration()` is normally not overridden, and it is definitely not designed to tweak RC name.
I remember RC name related stuff in `LocatableConfiguration`, `LocatableConfigurationBase`, its `suggestedName()`, `setGeneratedName()` and around. I do not remember all the details though.

skybrian: Changed the hack so it should work with an internationalized name. (Not tested that way, though.)

There seems to be nothing in the run configuration framework that lets you know if you're creating a new, regular configuration from a template configuration. There are two factory methods but they don't actually work that way. Hence the comments and this hack.

It looks like LocatableConfiguration and setGeneratedName() aren't used for this codepath. As far as I can tell, they are used by automatically generated run configs (for example, right-clicking on a file or method), not when creating a new run config by clicking '+'. The name being passed in here is generated by RunConfigurable.createUniqueName() which is private.
 
(It all seems rather dubious to me, and hardly anything is commented, so it's hard to determine intent. If JetBrains wants people to do things differently they should write javadoc explaining what to do.)

Issue #784, opened 2017-03-04
Fix breakpoints in same package when running on mac bazel with READONLY


gavindoughtie:
Set a breakpoint in IntelliJ (Community edition on OSX, with Flutter and Dart plugins). Attach debugger over the port specified when running app via iOS simulator from command line.

Expected:
IntelliJ stops at breakpoint.
Actual:
IntelliJ console logs from the running app, but breakpoint does not trigger.

eseidelGoogle:
To confirm: is this using the `flutter run` based workflow or the bazel based commandline workflow?

devoncarew:
@gavindoughtie, this is using the remote debug configuration? @skybrian has done a bunch of work to get the bazel workflow working on a mac (not all of it may be in the current stable release).

skybrian:
@gavindoughtie, could you file an internal bug with the Google-specific details?

gavindoughtie:
Filed bug to devoncarew with the internal details (bazel build)

skybrian:
Okay, reassigned to me.

skybrian:
I haven't done much actual work on this lately, but did learn some things while testing that might be helpful:

It looks like we are okay for most Dart packages (when they use package: imports). In that case, we use the analysis server to resolve source positions, and it generally does the right thing for Bazel.

Things are messier for .dart files in the same package as lib/main.dart. It mostly works provided that you set the "working directory" field to the directory containing "lib/main.dart", but doesn't handle generated files. (It looks like Gavin's project recently did a refactoring so that generated code is in a different package, so they might not be seeing this anymore.)

If you set working directory incorrectly, breakpoints will work until you do a hot reload, then stop working.

I've been thinking we either need to do more error-checking on working directory, or perhaps replace the working directory field with something less ambiguous. Asking the user to choose the location of main.dart seems clearer, and we can automatically calculate the application directory and Bazel workspace directory based on that.

skybrian:
Seems low priority these days, since most people don't use a setup with a READONLY directory.

devoncarew:
assumed stale -

PR #795, opened 2017-03-09
Fix race condition when there is a breakpoint in main

skybrian:
We should not set any breakpoints until receiving the app.started
event from the flutter command, because until then, the flutter
command is still using Observatory.

To be safe, don't tell the debugger the Observatory URL until then.

Possibly related to #784 (found while attempting to reproduce it)
but the symptoms are different.


devoncarew:
lgtm

I'll file an issue about the travis bots, which look fairly [red](https://travis-ci.org/flutter/flutter-intellij/builds/209219892).

PR #796, opened 2017-03-09
Update plugin.xml

devoncarew:
Update the changelog to reflect the last version pushed from the release_0.1.11 branch.

@pq @skybrian


skybrian:
The <version> was not updated. Is that a problem?

devoncarew:
There's a bit of asymmetry, but the latest version is in the branch, and we'll rev this to a new major version on the next release. I'm OK w/ this, but if you have a better workflow happy to try that

Issue #799, opened 2017-03-09
Add support for passing flags to `flutter run` in the run config.


pq:
In particular, this will be handy for capturing bug reporting data but in general will open running up to experimentation with whatever flags `flutter run` supports.

cc @tvolkert

pq:
WIP:

<img width="246" alt="screen shot 2017-03-12 at 9 21 56 pm" src="https://cloud.githubusercontent.com/assets/67586/23841378/fe99a09a-0769-11e7-8872-7a39ce6604f8.png">

PR #811, opened 2017-03-13
Program arg support for run configs (799).

pq:
Fixes: #799.

@skybrian @devoncarew

### DIFF
--- a/src/io/flutter/run/FlutterConfigurationEditorForm.form
+++ b/src/io/flutter/run/FlutterConfigurationEditorForm.form
@@ -52,10 +52,10 @@
         </constraints>
         <properties>
           <labelFor value="e7876"/>
-          <text value="Program &amp;arguments:"/>
+          <text value="Additional &amp;arguments:"/>
         </properties>
       </component>
-      <component id="e7876" class="javax.swing.JTextField" binding="myProgramArguments">
+      <component id="e7876" class="javax.swing.JTextField" binding="myAdditionalArguments">
         <constraints>
           <grid row="2" column="1" row-span="1" col-span="1" vsize-policy="0" hsize-policy="6" anchor="0" fill="1" indent="0" use-parent-layout="false"/>
         </constraints>
--- a/src/io/flutter/run/FlutterConfigurationEditorForm.form
+++ b/src/io/flutter/run/FlutterConfigurationEditorForm.form
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <form xmlns="http://www.intellij.com/uidesigner/form/" version="1" bind-to-class="io.flutter.run.FlutterConfigurationEditorForm">
-  <grid id="27dc6" binding="myMainPanel" default-binding="true" layout-manager="GridLayoutManager" row-count="3" column-count="2" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
+  <grid id="27dc6" binding="myMainPanel" default-binding="true" layout-manager="GridLayoutManager" row-count="4" column-count="2" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
     <margin top="0" left="0" bottom="0" right="0"/>
     <constraints>
       <xy x="20" y="20" width="500" height="320"/>
@@ -10,7 +10,7 @@
     <children>
       <vspacer id="3f5de">
         <constraints>
-          <grid row="2" column="0" row-span="1" col-span="1" vsize-policy="6" hsize-policy="1" anchor="0" fill="2" indent="0" use-parent-layout="false"/>
+          <grid row="3" column="0" row-span="1" col-span="1" vsize-policy="6" hsize-policy="1" anchor="0" fill="2" indent="0" use-parent-layout="false"/>
         </constraints>
       </vspacer>
       <component id="17e7a" class="com.intellij.openapi.ui.TextFieldWithBrowseButton" binding="myFileField">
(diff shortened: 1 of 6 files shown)


src/io/flutter/run/SdkFields.java
```diff
@@ -25,13 +25,15 @@
  */
 public class SdkFields {
   private @Nullable String filePath;
+  private @Nullable String programArgs;
   private @Nullable String workDir;
 
-  public SdkFields() {}
+  public SdkFields() {
```
pq: @skybrian : sorry for the unrelated formatting changes -- I ran the formatter (which we generally do) and this happened.  Hopefully not too distracting...

src/io/flutter/run/SdkFields.java
```diff
@@ -59,7 +70,7 @@ public void setWorkingDirectory(final @Nullable String dir) {
 
   /**
    * Reports any errors that the user should correct.
-   *
+   * <p>
```
skybrian: Also here.

src/io/flutter/run/SdkFields.java
```diff
@@ -25,13 +25,15 @@
...
 
-  public SdkFields() {}
+  public SdkFields() {
+  }
 
   /**
    * Creates SDK fields from a Dart file containing a main method.
-   *
+   * <p>
    * <p>(Automatically chooses the working directory.)
```
skybrian: Remove redundant `<p>`

src/io/flutter/run/SdkFields.java
```diff
@@ -77,7 +88,8 @@ VirtualFile chooseWorkDir(VirtualFile launchFile, Project project) throws Runtim
     // Work directory is optional; if not set we must infer it.
     if (!StringUtil.isEmptyOrSpaces(workDir)) {
       return checkWorkDir(workDir);
-    } else {
+    }
+    else {
```
skybrian: Okay but I wonder if we could tweak formatter rules to use `} else {`?

This is Google style so I keep forgetting to do it the Jetbrains way.

src/io/flutter/run/daemon/FlutterAppService.java
```diff
@@ -45,17 +44,19 @@ private FlutterAppService(Project project) {
...
    */
   public FlutterApp startFlutterSdkApp(@NotNull String workDir,
+                                       @Nullable String programArgs,
                                        @Nullable FlutterDevice device,
                                        @NotNull RunMode mode,
                                        @Nullable String path)
     throws ExecutionException {
 
-    final GeneralCommandLine command = createFlutterSdkRunCommand(workDir, device, mode, path);
+    final GeneralCommandLine command = createFlutterSdkRunCommand(workDir, programArgs, device, mode, path);
```
skybrian: Suppose we move createFlutterSdkRunCommand() to SdkFields and just pass in the GeneralCommandLine as a parameter here? That we we build the arguments in one place.

(I was thinking about doing the same for Bazel, as well.)

src/io/flutter/run/FlutterConfigurationEditorForm.form
```diff
@@ -46,6 +46,21 @@
         </constraints>
         <properties/>
       </component>
+      <component id="7656f" class="javax.swing.JLabel">
+        <constraints>
+          <grid row="2" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="0" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+        </constraints>
+        <properties>
+          <labelFor value="e7876"/>
+          <text value="Program &amp;arguments:"/>
```
skybrian: Do we want to call this "Additional arguments" for consistency with Bazel?

testSrc/unit/io/flutter/run/SdkFieldsTest.java
```diff
@@ -21,18 +21,22 @@
   public void shouldReadFieldsFromXml() {
     final Element elt = new Element("test");
     addOption(elt, "filePath", "lib/main.dart");
+    addOption(elt, "programArgs", "--debug");
```
skybrian: What's a more realistic example? (I think --debug" doesn't do anything since it's the default?)

skybrian: Do we need to make sure the user doesn't set flags that are important for debugging? (For example, passing in --no-hot will interfere with hot reloading.)

src/io/flutter/run/SdkFields.java
pq: This is generated by the formatter.  As a general rule we trust it or change it's settings.

src/io/flutter/run/SdkFields.java
pq: Same here.

src/io/flutter/run/SdkFields.java
pq: No objection to updating the formatter.

> This is Google style so I keep forgetting to do it the Jetbrains way.

Yeah and you shouldn't have to.  You should always just format and move on.

src/io/flutter/run/FlutterConfigurationEditorForm.form
pq: Sure!

I'll update.

testSrc/unit/io/flutter/run/SdkFieldsTest.java
pq: Good question.  In general, I'd like to add some validation but haven't thought to hard about it.  I'll add a TODO.

src/io/flutter/run/SdkFields.java
skybrian: I'm guessing the formatter wants `<p>` to be on a separate line, but it's incorrectly not deleting the `<p>` that's already there. So, if you delete the second `<p>` and reformat, I suspect it won't add it back again?

pq: (removing but let's fix the formatter)

src/io/flutter/run/SdkFields.java
pq: (removing but let's fix the formatter)

src/io/flutter/run/daemon/FlutterAppService.java
pq: I won't have the time to do this just now; maybe you or I could follow up with another quick change?  (Need to context switch shortly 😞 )

pq:
Updated to label as "Additional args".  PTAL.

skybrian:
LGTM

pq:
Thanks!

Issue #738, opened 2017-02-15
change to use whole version numbers?


devoncarew:
We're at version `0.1.10` and rev'ing to `0.1.11`. We should probably simplify a bit, and look to switch to whole version numbers. Perhaps for version `0.1.12` / 12?

PR #815, opened 2017-03-13
Migrate version to whole numbers (#738).

pq:
As per our conversation, whole numbers FTW.

Fixes: #738.

@devoncarew @skybrian

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -8,7 +8,7 @@
 
   <category>Custom Languages</category>
 
-  <version>0.1.11.1</version>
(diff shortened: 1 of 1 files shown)


resources/META-INF/plugin.xml
```diff
@@ -8,7 +8,7 @@
 
   <category>Custom Languages</category>
 
-  <version>0.1.11.1</version>
+  <version>12</version>
```
devoncarew: `12.0`?

pq: K!

skybrian:
LGTM either way

Issue #823, opened 2017-03-15
support 2017.1


devoncarew:
We'll need to have support for 2017.1 fairly soon (and continue to support 2016.3 for a bit after that).

pq:
As discussed, we should probably bump this to `M12`.

pq:
And a good place to start might be to get us building against `2017.1` PDQ.

mit-mit:
@pq @devoncarew what is the status of this?

This launched yesterday:
https://www.jetbrains.com/idea/whatsnew/

b-strauss:
I just updated and flutter is gone in the "create new project" dialog.

pq:
Sorry for the inconvenience @b-strauss!  A bit changed in our dependencies between releases and we need to do some fancy footwork to accommodate. @devoncarew was looking at this earlier in the week; not sure of status but will follow-up.

pq:
Chatting w/ @devoncarew just now and the rub is that it won't merge cleanly into an older branch and will take more doing than we can swing in a quick patch release.  Folks should hold off on upgrading to `2017.1` until we get things sorted.

Sorry for the continued inconvenience!

mit-mit:
Website updated in https://github.com/flutter/website/pull/472

b-strauss:
Has this yet been released? I can't find flutter in 2017.1's plugin repository.

mit-mit:
No, it has not been released yet. But we are getting close; currently testing a first release candidate, and hope to release an update soon!

devoncarew:
The new version is available now; thanks for your patience!

devoncarew:
Please note the one known issue when upgrading to 2017.1 - older projects may need to re-configure their Dart SDK: https://github.com/flutter/flutter-intellij#known-issues.

PR #829, opened 2017-03-15
update our product build matrix

devoncarew:
- update our product build matrix
- support 2017.1
- initial support for separate sdks per project

fix https://github.com/flutter/flutter-intellij/issues/823

### DIFF
--- a/src/io/flutter/dart/DartSdkLibUtilDelegate.java
+++ b/src/io/flutter/dart/DartSdkLibUtilDelegate.java
@@ -6,6 +6,7 @@
 package io.flutter.dart;
 
 import com.intellij.openapi.module.Module;
+import com.intellij.openapi.project.Project;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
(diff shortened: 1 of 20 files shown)


devoncarew:
@pq @skybrian

pq:
So awesome.  LGTM! 🤘

src/io/flutter/dart/DartSdkLibUtilDelegate.java
```diff
@@ -94,7 +94,7 @@ public void ensureDartSdkConfigured(@Nullable Project project, String sdkHomePat
       }
       else {
         //noinspection unchecked
-        final Method method = delegate.getMethod("ensureDartSdkConfigured", Project.class, Collection.class);
+        final Method method = delegate.getMethod("ensureDartSdkConfigured", Project.class, String.class);
```
pq: Do we not need to support both APIs?

devoncarew: This was an error on my part in the method params. Pre 2017.1, this method in the dart plugin took one param; in 2017.1+ it takes two.

Issue #814, opened 2017-03-13
Flutter IDE Wish: wrap with a builder, a stateful builder, or a layout builder


HansMuller:
Another useful variation on the `wrap this new widget expression` with another new widget idiom: inserting a new expression for one of the common widget builder classes, Builder, StatefulBuilder, LayoutBuilder.

There are some idomatic scenarios where a Builder widget is needed. For example.

If you write something like the following and then try pressing the button the `Scaffold.of(context)` expression will fail because the context doesn't actually include the Scaffold.
```
@override
Widget build(BuildContext context) {
  return new Scaffold(
    ...
    body: new FlatButton(
      child: new Text('SHOW SNACKBAR'),
      onPressed: () {
        Scaffold.of(context).showSnackBar(new SnackBar(
          content: ...
        ));
      },
    ),
  );
}
```
Wrapping the button in a builder corrects the problem because the Scaffold is within the parent of the Builder's BuildContext.
```
@override
Widget build(BuildContext context) {
  return new Scaffold(
    ...
    body: new Builder(
      builder: (BuildContext scaffoldContext) {
        return new FlatButton(
          child: new Text('SHOW SNACKBAR'),
          onPressed: () {
            Scaffold.of(scaffoldContext).showSnackBar(new SnackBar(
              content: ...
            ));
          },
        );
      },
    ),
  );
}
```
I've called the Builder function's BuildContext parameter `scaffoldContext` for clarity. I don't think that sort of rename would be possible for the IDE to generate.

It would be really impressive if the analyzer could detect the origin `Scaffold.of(context)` and the IDE could suggest this remedy.

Just FTR, the potential difficulty of referring to inherited widgets defined in the same build function is also noted here: https://github.com/flutter/flutter/issues/4581.

ThinkDigitalRepair:
Adding support for this request. It would be good to add all the different variations of what could go there, I.E.
Widgets where the selected widget is a Child or one of the children like we already have, and widgets where the selected widget is returned from a method, as in above. I can't think of any more currently, 😄

slightfoot:
This is a feature I've need consistently for the past 2 years. I am happy to implement it. I have found the locations of where the current StreamBuilder wrapper is.

* https://github.com/dart-lang/sdk/blob/cd95fcc8109ff5d3ce40fe9ba0b2a6fef7e64a47/pkg/analysis_server/lib/src/services/correction/assist.dart#L153
[...]

PR #834, opened 2017-03-22
Add some Flutter templates to create widgets

stevemessick:
@pq @devoncarew Define three live templates to generate boilerplate code for certain types of widgets as defined in #814.

I'm open to suggestions for different template names. I chose: stless, stful, stanim. Respectively, they generate new subclasses of StatelessWidget, StatefulWidget, and StatefulWidget with an animation controller.

To generate a new StatefulWidget, for example, you would type `stful` then `tab`, then type the name of the class once. That name will be inserted everywhere it needs to be, according to the boilerplate @HansMuller gave.

#814 should remain open because we still need to provide a mechanism to add (and remove?) various attributes of Flutter classes (like animation controllers).

### DIFF
--- a/src/io/flutter/template/DartToplevelTemplateContextType.java
+++ b/src/io/flutter/template/DartToplevelTemplateContextType.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.template;
+
+import com.intellij.psi.PsiComment;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.util.PsiTreeUtil;
+import com.jetbrains.lang.dart.ide.template.DartTemplateContextType;
+import com.jetbrains.lang.dart.psi.DartClassDefinition;
+import org.jetbrains.annotations.NotNull;
+
+public class DartToplevelTemplateContextType extends DartTemplateContextType {
+  public DartToplevelTemplateContextType() {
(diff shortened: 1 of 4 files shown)


pq:
This looks great @stevemessick !

> #814 should remain open because we still need to provide a mechanism to add (and remove?) various attributes of Flutter classes (like animation controllers).

Maybe we should create a new meta-bug with a punch list of templates?

src/io/flutter/template/DartToplevelTemplateContextType.java
```diff
@@ -0,0 +1,25 @@
...
+import com.intellij.psi.PsiComment;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.util.PsiTreeUtil;
+import com.jetbrains.lang.dart.ide.template.DartTemplateContextType;
+import com.jetbrains.lang.dart.psi.DartClassDefinition;
+import org.jetbrains.annotations.NotNull;
+
+public class DartToplevelTemplateContextType extends DartTemplateContextType {
+  public DartToplevelTemplateContextType() {
+    super("DART_TOPLEVEL", "Top-level", Generic.class);
```
pq: Is there anything we can do to better scope this to Flutter contexts?  (Or maybe that's not desirable?)

stevemessick: Possibly, but I didn't come up with anything simple. About all I can think of is looking for an import statement that references Flutter, but that might not be good enough. Only people that have the Flutter plugin installed will be affected. Given how hard these things are to find I think we're safe waiting for bugs to be filed.

pq: 👍 SGTM.  Thanks!

PR #837, opened 2017-03-27
`0.12.0` change notes.

pq:
@devoncarew @skybrian @mit-mit

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -21,6 +21,16 @@
 
   <change-notes>
     <![CDATA[
+
+0.12.0:
+<ul>
+  <li>new Flutter `stless`, `stful`, and `stanim` live templates</li>
+  <li>support for specifying "Additional Args" to Flutter application launches</li>
+  <li>default run configuration creation on project open (when possible)</li>
+  <li>device menu improvements</li>
+  <li>miscellaneous bug fixes</li>
+</ul>
+
 0.1.11.2:
 <ul>
   <li>fix an NPE in the Flutter View when launching an app</li>
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -24,6 +24,7 @@
 
 0.12.0:
 <ul>
+  <li>support for IDEA `2017.1`</li>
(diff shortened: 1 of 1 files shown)


devoncarew:
lgtm! I would also add support for 2017.1

pq:
> lgtm! I would also add support for 2017.1

Added!

resources/META-INF/plugin.xml
```diff
@@ -21,6 +21,17 @@
 
   <change-notes>
     <![CDATA[
+
+0.12.0:
```
skybrian: I thought we were going to whole numbers?

devoncarew: Yeah, +1. @mit-mit had some ideas around the versioning, but without other guidance let's go with `12.0`.

mit-mit: Yes, I suggested we make this 1.0 (as in our first, complete and stable plugin). It seems little odd to me to start with 12.0.

resources/META-INF/plugin.xml
```diff
@@ -21,6 +21,16 @@
...
+<ul>
+  <li>new Flutter `stless`, `stful`, and `stanim` live templates</li>
+  <li>support for specifying "Additional Args" to Flutter application launches</li>
+  <li>default run configuration creation on project open (when possible)</li>
+  <li>device menu improvements</li>
+  <li>miscellaneous bug fixes</li>
+</ul>
```
mit-mit: Mention causal stack traces?

resources/META-INF/plugin.xml
pq: I'd rather not personally.  `1.0` suggests something more than we want to (I think).  If you guys feel strongly though I won't resist.

pq: In particular, I don't know that this assessment

> as in our first, complete and stable plugin

is really accurate.

Again, I won't fight this if you all feel strongly.

resources/META-INF/plugin.xml
pq: Good idea.  Can you think of a short sentence?

Issue #842, opened 2017-03-27
use after dispose exception


devoncarew:
To repro:

- open an project
- close the project
- create (or open) a new project

you'll see an exception, likely from accessing the old (disposed) project

```
Caused by: java.lang.AssertionError: Already disposed: Project (Disposed) flutter_gallery
	at com.intellij.openapi.components.impl.ComponentManagerImpl.getPicoContainer(ComponentManagerImpl.java:259)
	at com.intellij.openapi.components.impl.ComponentManagerImpl.getPicoContainer(ComponentManagerImpl.java:58)
	at com.intellij.openapi.components.ServiceManager.doGetService(ServiceManager.java:47)
	at com.intellij.openapi.components.ServiceManager.getService(ServiceManager.java:42)
	at com.intellij.openapi.roots.impl.libraries.ProjectLibraryTable.getInstance(ProjectLibraryTable.java:57)
	at com.jetbrains.lang.dart.sdk.DartSdk.lambda$getDartSdk$0(DartSdk.java:56)
	at com.intellij.psi.impl.PsiCachedValueImpl.doCompute(PsiCachedValueImpl.java:49)
	at com.intellij.util.CachedValueBase.getValueWithLock(CachedValueBase.java:222)
	at com.intellij.psi.impl.PsiCachedValue.getValueWithLock(PsiCachedValue.java:60)
	at com.intellij.psi.impl.PsiCachedValueImpl.getValue(PsiCachedValueImpl.java:38)
	at com.jetbrains.lang.dart.sdk.DartSdk.getDartSdk(DartSdk.java:72)
	at io.flutter.dart.DartPlugin.getDartSdk(DartPlugin.java:47)
	at io.flutter.sdk.FlutterSdk.getFlutterSdk(FlutterSdk.java:61)
	at io.flutter.sdk.FlutterSdkManager.isFlutterSdkSetAndNeeded(FlutterSdkManager.java:97)
	at io.flutter.sdk.FlutterSdkManager.checkForFlutterSdkChange(FlutterSdkManager.java:72)
	at io.flutter.sdk.FlutterSdkManager$1.projectOpened(FlutterSdkManager.java:51)
	at com.intellij.openapi.project.impl.ProjectManagerImpl.d(ProjectManagerImpl.java:686)
```

alexander-doroshko:
Got this report from a user, probably the same cause:
```
 Product:IDEA Community 163.15188 OS:Mac OS X 
[...]

PR #845, opened 2017-03-28
fix a use after dispose exception

devoncarew:
- fix https://github.com/flutter/flutter-intellij/issues/842
- fix a use after dispose exception, introduced with the work to adapt to per-project sdks

@pq @skybrian

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkManager.java
+++ b/src/io/flutter/sdk/FlutterSdkManager.java
@@ -5,71 +5,73 @@
  */
 package io.flutter.sdk;
 
-import com.google.common.base.Objects;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.project.ProjectManager;
 import com.intellij.openapi.project.ProjectManagerAdapter;
-import com.intellij.openapi.roots.RootProvider;
-import com.intellij.openapi.roots.impl.libraries.ApplicationLibraryTable;
+import com.intellij.openapi.roots.impl.libraries.ProjectLibraryTable;
 import com.intellij.openapi.roots.libraries.Library;
 import com.intellij.openapi.roots.libraries.LibraryTable;
+import com.intellij.openapi.util.Disposer;
 import com.intellij.util.EventDispatcher;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.EventListener;
+import java.util.Timer;
+import java.util.TimerTask;
 
 /**
  * Monitors the application library table to notify clients when Flutter SDK configuration changes.
  */
 public class FlutterSdkManager {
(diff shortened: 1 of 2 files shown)


src/io/flutter/sdk/FlutterSdkManager.java
```diff
@@ -5,71 +5,73 @@
...
+  private FlutterSdkManager(@NotNull Project project) {
+    final LibraryTableListener libraryTableListener = new LibraryTableListener(project);
+    ProjectLibraryTable.getInstance(project).addListener(libraryTableListener);
 
-    listenForSdkChanges();
-    // Cache initial state.
-    isFlutterConfigured = isFlutterSdkSetAndNeeded();
-  }
+    final Timer timer = new Timer();
+    timer.scheduleAtFixedRate(new TimerTask() {
```
skybrian: Polling like this is a battery drain for laptop users, so I would expect at least a TODO here to switch to using real listeners again.

src/io/flutter/sdk/FlutterSdkManager.java
```diff
@@ -5,71 +5,73 @@
...
-      watchDartSdkRoots(library);
-    }
+    // Cache initial state.
+    isFlutterConfigured = isFlutterSdkSetAndNeeded(project);
   }
 
   // Send events if Flutter SDK was configured or unconfigured.
-  public void checkForFlutterSdkChange() {
-    if (!isFlutterConfigured && isFlutterSdkSetAndNeeded()) {
+  public void checkForFlutterSdkChange(@NotNull Project project) {
```
skybrian: It seems like saving the Project to a field and removing the parameter would be better? (It eliminates the possibility of using a different project.)

src/io/flutter/sdk/FlutterSdkManager.java
```diff
@@ -117,18 +108,21 @@ default void flutterSdkRemoved() {
 
   // Listens for changes in Flutter Library configuration state in the Library table.
   private final class LibraryTableListener implements LibraryTable.Listener {
+    private @NotNull final Project myProject;
```
skybrian: Move field to top level?

devoncarew: done

src/io/flutter/sdk/FlutterSdkManager.java
devoncarew: makes sense! done

src/io/flutter/sdk/FlutterSdkManager.java
devoncarew: Added a TODO (and found a slightly better IntelliJ API for polling - `JobScheduler.getScheduler().scheduleWithFixedDelay()`).

devoncarew:
updated, ptal

skybrian:
LGTM

pq:
LGTM!

PR #846, opened 2017-03-28
show the flutter editor notification in small ides

devoncarew:
Fix an issue where the flutter pubspec.yaml editor notification would not show in small IDEs (note that this is not a recent regression).

@pq 

Phil, do you know why we only define `editorNotificationProvider`s in `idea-contribs.xml` and not the main `plugin.xml`?


devoncarew:
(And note that w/ this change, we _are_ now displaying both the Dart and the Flutter editor contributions; we'll need to tackle that as well)

pq:
> Phil, do you know why we only define editorNotificationProviders in idea-contribs.xml and not the main plugin.xml?

That was probably a mistake (thinking that was an IDEA-only extension point).  I'll dig a bit though and see if I can recover the rationale.

pq:
> (And note that w/ this change, we are now displaying both the Dart and the Flutter editor contributions; we'll need to tackle that as well)

Hmmmm.  Yeah, let's look at this today.  The logic for suppressing that notification in the Dart Plugin must need updating.

pq:
Needless to say, LGTM. 👍

PR #847, opened 2017-03-28
Switch to a newer version of Dart plugin code

skybrian:
Use the currently released version.
Pick up the source code, assuming that intellij-plugins
is checked out in the same directory.


pq:
LGTM!

Issue #861, opened 2017-03-30
convert plugin to fully 2017.1


devoncarew:
Remove support for 2016.3 (and remove code necessary to span both versions).

PR #864, opened 2017-03-30
move to only supporting 2017.1

devoncarew:
- move to only supporting 2017.1
- fix https://github.com/flutter/flutter-intellij/issues/861
- remove reflection based support for 2017.1 and 2016.3
- only test against 2017.1 versions
- update out supported intellij range
- update our min required dart plugin version (based on webstorm)

@skybrian


skybrian:
I think it's too soon, but no issues with the CL.

devoncarew:
To summarize where I think we ended up after an IM:
- if we do a 12.1 release after landing this PR, that version would be for 2017.1 only
- our internal customers likely won't need anything from 12.1
- they will be moving to 2017.1 around the same timeframe as our M13 release being available

skybrian:
I tried upgrading to 2017.1 and patching this in. Unit tests are failing.

The first failure was that the Dart plugin was missing; this is fixed by starting IDEA in the sandbox and installing it.

The second error is a TypeCastException in service.kt, where it's trying to get "com.intellij.openapi.components.impl.stores.IComponentStore" and it's returning null.

(There were also failures running on 2017.1 without this patch.)

skybrian:
I got the tests working. Here's what you need to do:

Set up the plugin SDK:
- File-> Project Structure, select SDK's
- Rename the "IntelliJ IDEA Community Edition" SDK
- Create a new IDEA SDK for 2017.1, and rename it to "IntelliJ IDEA Community Edition"
- Make sure you give it a new sandbox directory.

Set up the sandbox:
- Launch IDEA using the "Flutter Plugin" run config
- Go to plugins and install the Dart plugin

devoncarew:
Thanks! Can you update the https://github.com/flutter/flutter-intellij/blob/master/CONTRIBUTING.md docs w/ the above?

skybrian:
Done

PR #865, opened 2017-03-30
enable Dart support when setting the sdk location

devoncarew:
- enable Dart support when setting the sdk location

This will prevent the issue where users can see and set the sdk in the Flutter settings page, but the value doesn't persist (when they re-open the settings page, the sdk location isn't persisted and the project didn't have Dart enabled).

@skybrian (/cc @pq)

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkUtil.java
+++ b/src/io/flutter/sdk/FlutterSdkUtil.java
@@ -9,6 +9,8 @@ import com.intellij.execution.ExecutionException;
 import com.intellij.ide.util.PropertiesComponent;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.project.ProjectManager;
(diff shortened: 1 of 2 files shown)


src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -218,4 +220,16 @@ public static void setFlutterSdkPath(@NotNull final Project project, @NotNull fi
...
   }
+
+  // TODO(devoncarew): The Dart plugin supports specifying individual modules in the settings page.
+
+  /**
+   * Do a best-effort basis to enable Dart support for the given project.
+   */
+  public static void enableDartSdk(@NotNull final Project project) {
+    final Module[] modules = ModuleManager.getInstance(project).getModules();
+    if (modules.length == 1) {
```
skybrian: Or perhaps enable Dart for all of them?

devoncarew: Certainly an option; perhaps the best thing to do?

skybrian: On second thought I don't know enough. Depends how easy it is to undo.

devoncarew: Will keep this as is. I think as part of future work we'll want to expose controls for individual modules.

PR #868, opened 2017-04-01
Bazel: load plugin config and daemon script from READONLY directory

### LINKED ISSUE #791: bazel: should look for flutter.json in READONLY folder.
If dart/config/intellij-plugins/flutter.json doesn't exist, we should look in ../READONLY for this file. (This is to handle the case where the user is not tracking this file in git.)

For comparison, here is the lookup code in the analysis server:

https://github.com/dart-lang/sdk/blob/62b6e8995a5e4f44a670149ffcff069586fbd956/pkg/analyzer/lib/src/generated/bazel.dart#L256

skybrian:
This handles the case where dart/config/intellij-plugins is not checked out, or the directory containing the daemon script isn't checked out.

Fixes #791


devoncarew:
lgtm!

Issue #840, opened 2017-03-27
when invoking the flutter tool, we should pass in a marker env variable


devoncarew:
We want IDEs to pass in an env var to help us distinguish how the flutter tool's being invoked. Something like `FLUTTER_ENVIRONMENT=IntelliJ-IDEA`.

devoncarew:
(similar PR into the Dart plugin here: https://github.com/JetBrains/intellij-plugins/pull/487)

PR #873, opened 2017-04-03
pass in a FLUTTER_ENVIRONMENT env var when shelling out to the flutter cli

devoncarew:
- pass in a FLUTTER_ENVIRONMENT env var when shelling out to the flutter cli
- fix https://github.com/flutter/flutter-intellij/issues/840

@skybrian

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkUtil.java
+++ b/src/io/flutter/sdk/FlutterSdkUtil.java
@@ -8,6 +8,7 @@ package io.flutter.sdk;
 import com.intellij.execution.ExecutionException;
 import com.intellij.ide.util.PropertiesComponent;
 import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.application.ApplicationNamesInfo;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.module.ModuleManager;
@@ -35,11 +36,23 @@ import java.util.*;
 
 
 public class FlutterSdkUtil {
+  /** The environment variable to use to tell the flutter tool which app is driving it. */
+  public static final String FLUTTER_ENVIRONMENT = "FLUTTER_ENVIRONMENT";
+
   private static final Map<Pair<File, Long>, String> ourVersions = new HashMap<>();
   private static final String FLUTTER_SDK_KNOWN_PATHS = "FLUTTER_SDK_KNOWN_PATHS";
   private static final Logger LOG = Logger.getInstance(FlutterSdkUtil.class);
 
   private FlutterSdkUtil() {
+  }
+
+  /**
+   * Return the environment variable value to use when shelling out to the Flutter command-line tool.
+   */
+  public static String getFlutterEnvValue() {
+    final String clientId = ApplicationNamesInfo.getInstance().getFullProductName().replaceAll(" ", "-");
+    final String existingVar = System.getenv(FLUTTER_ENVIRONMENT);
+    return existingVar == null ? clientId : (existingVar + ":" + clientId);
   }
 
   public static void updateKnownSdkPaths(@NotNull final String newSdkPath) {
--- a/src/io/flutter/sdk/FlutterSdkUtil.java
+++ b/src/io/flutter/sdk/FlutterSdkUtil.java
@@ -37,7 +37,7 @@ import java.util.*;
 
 public class FlutterSdkUtil {
   /** The environment variable to use to tell the flutter tool which app is driving it. */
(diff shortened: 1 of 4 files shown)


src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -35,13 +36,25 @@
 
 
 public class FlutterSdkUtil {
+  /** The environment variable to use to tell the flutter tool which app is driving it. */
+  public static final String FLUTTER_ENVIRONMENT = "FLUTTER_ENVIRONMENT";
```
skybrian: The name seems ambiguous - "environment" could mean anything. For the analysis server we use "client_id" and "client_version". Should we do the same here?

devoncarew: Perhaps `FLUTTER_CLIENT`? `FLUTTER_HOST`? This env var is meant to given the flutter cli tool some info about the environment it's running in: IntelliJ, Travis, ...

skybrian: FLUTTER_CLIENT sounds good. Do you want to pass in the version separately? Could be FLUTTER_CLIENT_ID and FLUTTER_CLIENT_VERSION.

(Or perhaps FLUTTER_CALLER?)

skybrian: This is much like the user-agent string in browsers. Hopefully we never have to use it for anything other than analytics.

devoncarew: Perhaps `FLUTTER_HOST_ENV` then? The var could represent IntelliJ, or it could represent Travis.

devoncarew: (will update to `FLUTTER_HOST_ENV`)

src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -49,9 +49,9 @@ private FlutterSdkUtil() {
   /**
    * Return the environment variable value to use when shelling out to the Flutter command-line tool.
    */
-  public static String getFlutterEnvValue() {
+  public static String getFlutterHostEnvValue() {
     final String clientId = ApplicationNamesInfo.getInstance().getFullProductName().replaceAll(" ", "-");
```
skybrian: Maybe Flutter should handle replacing the space with a dash, so it's done consistently?

src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -49,9 +49,9 @@ private FlutterSdkUtil() {
...
-    final String existingVar = System.getenv(FLUTTER_ENVIRONMENT);
+    final String existingVar = System.getenv(FLUTTER_HOST_ENV);
     return existingVar == null ? clientId : (existingVar + ":" + clientId);
```
skybrian: What's our allowed character set? How do we escape special characters? Maybe send JSON?

src/io/flutter/sdk/FlutterSdkUtil.java
devoncarew: This is just normalizing what we're sending to the flutter cli - we'll mostly have control over what is send into the tool (we'll send it in from IntelliJ; similar modifications to the travis testing script and other callers will need to be made).

src/io/flutter/sdk/FlutterSdkUtil.java
devoncarew: Not clear, but happy to put together a regex to whitelist some ranges of chars. From the IntelliJ does, `getProductName()` will return something like:

```
shortened name of the product if it contains two words (e.g. {@code "IDEA"} for IntelliJ IDEA, {@code "WebStorm"} for WebStorm)
```

skybrian: Okay, fair enough. Maybe this is over-engineering, but I was thinking we could either send a set of hashtags as JSON:
```
["travis","intellij-idea"]
```
Or a map as JSON:
```
{"travis":true,"product":"IntelliJ IDEA"}
```
And let the JSON encoder/parser worry about special characters. (Reasonably compact, and better than getting into an ad-hoc user-agent parsing situation later.)

devoncarew: So, I think we're ok passing plain text here. We did do something very similar for invoking pub a few weeks ago. I don't think this graduate to a full user agent type thing, but just be used in analytics to get a better sense of how the flutter cli is used (exclude bots, only human cli usages and not tool driven, ...).

PR #881, opened 2017-04-04
Add a few more testing steps

mit-mit:
@devoncarew 

I changed File > Import to File > Open as that is the main flow. But should we still have File > Import in the test script?


devoncarew:
lgtm wrt the import -> open change

Don't know about testing both import and open in the script; we walk a fine line w/ completeness and time-to-get-through the whole script.

mit-mit:
OK, lets leave Import out for now

PR #885, opened 2017-04-05
show a hot reload info toast on first run

devoncarew:
The first time we run a Flutter app for a project, show a notification as a prompt to use hot reload:

screen shot 2017-04-05 at 4 33 44 pm

@pq @skybrian @mit-mit

### DIFF
--- a/src/io/flutter/run/FlutterRunNotifications.java
+++ b/src/io/flutter/run/FlutterRunNotifications.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.run;
+
+import com.intellij.ide.BrowserUtil;
+import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.notification.*;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.project.Project;
+import icons.FlutterIcons;
+import io.flutter.FlutterBundle;
+import io.flutter.view.FlutterViewMessages;
+import org.jetbrains.annotations.NotNull;
+
+import javax.swing.event.HyperlinkEvent;
+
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/FlutterRunNotifications.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  @NotNull Project myProject;
+
+  FlutterRunNotifications(@NotNull Project project) {
+    this.myProject = project;
+  }
+
+  private void checkForDisplayFirstReload() {
+    final PropertiesComponent properties = PropertiesComponent.getInstance(myProject);
```
pq: Will this get cleared for each project?  If so does that suggest that this notification will show for the first run of each project.  Is that intentional?  (The alternative I guess, would be an application-scoped property?)

skybrian: Yes, I think we need to be careful about this. Even application-level might be too often.  Consider that some users use multiple JetBrains products on multiple computers, and it would show up again if you upgrade and don't keep your settings.

Maybe touch a file somewhere in the user's home directory? That would make it once per computer; not too bad.

Another possibility would be to wait until there is some evidence that they don't know about hot reload. Here's an idea: if they're restarting their Flutter app and they have never used hot reload, we could print a message at the top of the console tab:
"Restart your Flutter app faster by using _hot reload_."

devoncarew: I did intentionally chose to display this notification for every new project. I can't say that it's the right choice, but it's what I was going for :) Note that w/ the flags I've set, this toast automatically times out after some short period of time (~10 secs).

I like the others ideas for when we could trigger this message. For some of them, we may even be able to show the toast at a slightly better place - as a balloon off the run/debug tool window. That said, there is at least nice consistency in showing the toast once per new project.

devoncarew:
I'm going to land this change, but we'll likely iterate on the best place and trigger to show it.

Issue #882, opened 2017-04-05
Remove working directory field from run config?


skybrian:
The SDK run config asks for the file containing main() and also asks for a working directory.

But usually you can find the Flutter application's directory by looking for 'lib' in the ancestors of the main file. (You don't even need the pubspec.yaml file, though we look for that too.) I suppose main.dart could be outside lib but still within a directory containing pubspec.yaml, though that seems rare too.

I'm wondering if we should just drop the working directory field altogether and require that main.dart either live within a "lib" directory, or live above a pubspec.yaml file. Is there any reason the user would need to override this?

Rationale: one less thing for the user to think about, and one less corner case to handle in the code.

skybrian:
@mit-mit

mit-mit:
Sounds good to me. I don't think the terminal tools have any notion of a 'working directory' either, just a target you are running. From a bit of experimenting:

1. `flutter create testmain`
1. `mv lib lib2`
1. `flutter run`
    =>`Target file "lib/main.dart" not found.`
1. `flutter run -t lib2/main.dart`
    => works fine

skybrian:
Well, every command line tool has the working directory provided by the shell; it's less meaningful for IDE's though.

If we look for either a pubspec.yaml or a BUILD file, I think that will cover all the cases. Maybe look for a .packages file too.

mit-mit:
What I meant was: I don't think the flutter terminal tool actually *uses* the working directory for anything.

skybrian:
Actually, it seems to be picky about this. If I run it from a different directory it doesn't find pubspec.yaml:

```
$ cd lib
$ ../../../bin/flutter run main.dart 
Error: No pubspec.yaml file found.
This command should be run from the root of your Flutter project.
Do not run this command from the root of your git clone of Flutter.
```

Similarly if I go to a a higher directory. It's not a big deal though, if we find the app dir and pass it in.

devoncarew:
done?

skybrian:
Done.

PR #887, opened 2017-04-06
remove working directory from SDK run config

skybrian:
We now calculate it from the entrypoint in the MainFile class.
Also, do more error-checking on the entrypoint; make sure it
actually contain main and a flutter import.

Wrote unit tests for the field validation (in MainFile).

Also, moved command line calculation to SdkFields and
removed FlutterAppService.

Discussion in #882.


src/io/flutter/run/MainFile.java
```diff
@@ -0,0 +1,173 @@
...
+    if (!(psi instanceof DartFile)) {
+      return error(FlutterBundle.message("entrypoint.not.dart"));
+    }
+    final DartFile dart = (DartFile)psi;
+
+    if (DartResolveUtil.getMainFunction(dart) == null) {
+      return error(FlutterBundle.message("main.not.in.entrypoint"));
+    }
+
+    if (findImportUrls(dart).noneMatch((url) -> url.startsWith("package:flutter/"))) {
```
devoncarew: I don't think this check will be reliable - users could be importing a file which itself used the flutter package.

src/io/flutter/run/MainFile.java
```diff
@@ -0,0 +1,173 @@
...
+      return error(FlutterBundle.message("entrypoint.not.dart"));
+    }
+
+    final PsiFile psi = PsiManager.getInstance(project).findFile(file);
+    if (!(psi instanceof DartFile)) {
+      return error(FlutterBundle.message("entrypoint.not.dart"));
+    }
+    final DartFile dart = (DartFile)psi;
+
+    if (DartResolveUtil.getMainFunction(dart) == null) {
```
devoncarew: This will almost always be reliable. I've only see getting-main-through-exports once.

src/io/flutter/run/MainFile.java
```diff
@@ -0,0 +1,173 @@
...
+
+  private static boolean inProject(@Nullable VirtualFile file, @NotNull Project project) {
+    return file != null && ProjectRootManager.getInstance(project).getFileIndex().isInContent(file);
+  }
+
+  /**
+   * Returns the import URL's in a Dart file.
+   */
+  @NotNull
+  private static Stream<String> findImportUrls(@NotNull DartFile file) {
```
devoncarew: My 0.02 is we should drop this check...

src/io/flutter/run/SdkFields.java
```diff
@@ -26,18 +30,15 @@
 public class SdkFields {
   private @Nullable String filePath;
   private @Nullable String additionalArgs;
-  private @Nullable String workDir;
```
devoncarew: +1

devoncarew:
One pass through - some comments about checks which may I think lead to false negatives. Removing the working dir field is nice cleanup!

devoncarew:
We do have a doc which provides an overview of the debugger arch. and classes: https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/run/README.md. We should update it as I think it's getting out of date. Possibly just delete any docs for classes that no longer exist, and add TODO:s for new ones (does not need to be for this PR).

src/io/flutter/run/MainFile.java
skybrian: Sure, I can take it out.

(I copied the logic from FlutterRunConfigurationProducer.getFlutterEntryFile. But I guess that check should be more strict?)

devoncarew: Ah, I didn't realize it was existing. I do think it will hit false negatives, so we should remove.

skybrian: The code I borrowed from is for running the app by right-clicking, so I think it needs to avoid overriding the regular Dart run config if you're running a non-Flutter main method. I guess it makes sense there but not here.

devoncarew: OK, sounds reasonable.

skybrian:
Please take another look. (Changes in a separate commit.)

src/io/flutter/run/MainFile.java
```diff
@@ -0,0 +1,198 @@
...
+   */
+  @NotNull
+  public VirtualFile getAppDir() {
+    return appDir;
+  }
+
+  /**
+   * Returns true if the file has any direct flutter imports.
+   */
+  public boolean hasFlutterImports() {
```
devoncarew: I don't believe this is used? If so, we should remove.

devoncarew:
lgtm! One comment for cleaning up unused code (if it is actually unused).

src/io/flutter/run/MainFile.java
skybrian: It is unused now, but I'd like to change FlutterRunConfigurationProducer to use MainFile in a future CL.

PR #894, opened 2017-04-07
adjust the enablement state of run and debug; re-run means full restart

devoncarew:
- when running an app, keep the run button enabled but the debug button disabled
- when debugging an app, keep the debug button enabled but the run button disabled
- when we get a request to re-run or re-debug an app, translate that to a full restart if the app is currently running

(from discussions, we decided to map a re-run intention to a full restart)

/cc @pq @skybrian @mit-mit


devoncarew:
@skybrian, @pq, this is ready for review now

src/io/flutter/run/daemon/RunMode.java
```diff
@@ -20,7 +20,7 @@
   @NonNls
   DEBUG("debug", true),
 
-  // Not used. See: https://github.com/flutter/flutter-intellij/issues/410
+  // TODO(devoncarew): Delete this - the IntelliJ profile mode is not the same as the flutter_tools' --profile mode.
```
pq: 👍

skybrian: If everyone is agreed, I think we can delete it now.

src/io/flutter/run/LauncherState.java
```diff
@@ -96,7 +96,9 @@ private RunContentDescriptor launch(@NotNull ExecutionEnvironment env) throws Ex
     final FlutterApp app = callback.createApp(device);
 
     // Remember the run configuration that started this process.
-    app.getProcessHandler().putUserData(KEY, runConfig);
+    app.getProcessHandler().putUserData(FLUTTER_RUN_CONFIG_KEY, runConfig);
+    app.getProcessHandler().putUserData(FLUTTER_APP_KEY, app);
+    app.getProcessHandler().putUserData(FLUTTER_LAUNCH_MODE_KEY, getEnvironment().getExecutor().getId());
```
skybrian: Maybe better to add the launch mode as a field to the FlutterApp class? Then it's more generally available, and you only need two keys here.

src/io/flutter/run/LauncherState.java
```diff
@@ -238,33 +254,71 @@ protected boolean canRun(C config) {
...
+      // Else, launch the app.
+      return launcherState.launch(env);
     }
 
     /**
      * Returns true if any processes are running that were launched from the given RunConfig.
      */
-    private static boolean isRunning(RunConfig config) {
+    @SuppressWarnings("SameParameterValue")
+    private static boolean isRunning(RunConfig config, String executorId, boolean ignoreSameLaunchMode) {
```
skybrian: The ignoreSameLaunchMode parameter here is confusing. How about a more general-purpose method instead?

```
/**
 * Returns the currently running app for the given RunConfig, if any.
 */
@Nullable
private static FlutterApp getRunningApp(RunConfig config);
```

Then the caller can decide what to do based on the running FlutterApp.

src/io/flutter/run/LauncherState.java
```diff
@@ -238,33 +254,71 @@ protected boolean canRun(C config) {
...
+      final String executorId = env.getExecutor().getId();
+
+      // See if we should issue a hot-reload.
+      final List<RunContentDescriptor> runningProcesses =
+        ExecutionManager.getInstance(env.getProject()).getContentManager().getAllDescriptors();
+
+      for (RunContentDescriptor descriptor : runningProcesses) {
+        final ProcessHandler process = descriptor.getProcessHandler();
+        if (process != null && !process.isProcessTerminated() && process.getUserData(FLUTTER_RUN_CONFIG_KEY) == launcherState.runConfig) {
+          final String launchMode = process.getUserData(FLUTTER_LAUNCH_MODE_KEY);
```
skybrian: Seems like this would be simpler if we had a getRunningApp() method - see below.

src/io/flutter/run/LauncherState.java
```diff
@@ -50,28 +50,28 @@
 
 /**
  * Launches a flutter app, showing it in the console.
- *
- * <p>Normally creates a debugging session, which is needed for hot reload.
+ * <p>
+ * Normally creates a debugging session, which is needed for hot reload.
  */
-public class Launcher extends CommandLineState {
+public class LauncherState extends CommandLineState {
```
skybrian: Hmm, renaming this seems unfortunate due to the inner interfaces. LauncherState.RunConfig etc get longer.

devoncarew: The `Launcher` name seemed a bit short to me; all of the IntelliJ subclasses of `RunProfileState` are suffixed with `State`. I prefer `LauncherState` to just `Launcher`, but will leave it as `Launcher` if you feel strongly.

src/io/flutter/run/LauncherState.java
devoncarew: sgtm! will so add

src/io/flutter/run/LauncherState.java
devoncarew: done!

src/io/flutter/run/daemon/RunMode.java
devoncarew: deleted

src/io/flutter/run/LauncherState.java
devoncarew: added getRunningApp()

devoncarew:
Updated! @skybrian, let me know re: `Launcher` vs `LauncherState`.

src/io/flutter/run/LauncherState.java
skybrian: Okay how about LaunchState? (Slightly shorter.)

(I think maybe I can avoid inheriting from CommandLineState, but that's a refactoring for another time.)

devoncarew: `LaunchState` sgtm; will change

src/io/flutter/run/LauncherState.java
```diff
@@ -238,33 +259,58 @@ protected boolean canRun(C config) {
...
+
+      final LauncherState launcherState = (LauncherState)state;
+      final String executorId = env.getExecutor().getId();
+
+      // See if we should issue a hot-reload.
+      final List<RunContentDescriptor> runningProcesses =
+        ExecutionManager.getInstance(env.getProject()).getContentManager().getAllDescriptors();
+
+      final ProcessHandler process = getRunningAppProcess(launcherState.runConfig);
+      if (process != null) {
```
skybrian: nit: could invert all these if statements (return early) to avoid nesting.

src/io/flutter/run/LauncherState.java
skybrian: Actually, never mind... LauncherState is okay. I'll figure it out some other time.

devoncarew:
Landing this, and we can iterate from there.

PR #895, opened 2017-04-09
change the icon for the hot reload toast

devoncarew:
Some changes to the hot reload first-run toast:
- use the hot reload icon (to help the user identify where in the ui to access reload)
- use an action instead of a hyperlink
- make the text terse to ensure it doesn't wrap and end up below the fold

screen shot 2017-04-09 at 4 36 46 pm

@pq @skybrian


devoncarew:
(updated toast text to: "Apply changes to your app in place, instantly.")

Issue #410, opened 2016-11-03
Expose --release and --profile modes


mit-mit:
Currently both the `Run > Run` and `Run > Debug` actions  produce a debug build, i.e. they correspond to `flutter run --debug` (historical background is in #362 and #404).

Thus there is no way of running `--release` or `--profile` from IntelliJ. This is concerning as developers who use just the IDE might get the impression that all flutter apps are slow, and it means that release testing and profiling is not possible from the IDE. We should find some way of exposing these from IntelliJ.

***Note***: Developers can use this workaround until this is fixed:

To run in release mode, enter this in the terminal:
```
cd <location of app>
flutter run --release
```

To run in profile mode, enter this in the terminal:
```
cd <location of app>
flutter run --profile
```

stevemessick:
I think this has been suggested elsewhere, but in case it hasn't you could extend the launch configuration to include a `Build mode`. Flutter actually has three options: debug, release, and profile. You'd need to ensure the Debug button was disabled if a Release mode launch configuration was selected.

mit-mit:
Yes, I can think of at least three different ways of doing this:

1. Add a build mode option to launch config, and create a default config for each of those
1. Add a custom build mode option box next to the Run/Debug buttons
1. Add  'Install release mode app' / 'Install profile mode app' actions somewhere in the menu / toolbar

skybrian:
Also, the FAQ is out of date:

"If you are using the Flutter plugin for IntelliJ, you can disable Slow Mode by using the Run>Run action rather than Run>Debug."

https://flutter.io/faq/#my-app-has-a-slow-mode-bannerribbon-in-the-upper-right-why-am-i-seeing-that

mit-mit:
FAQ updated in https://github.com/flutter/website/pull/444. 

Remember to update that again when this bug is resolved.

mit-mit:
@pq @devoncarew now that we have other Flutter actions in the Run menu, I think a variant of option 3 above seems natural:

* Add 'Run in release mode' and 'Run in profile mode' menu items below the current 'Full App Restart' item ?

pq:
I lean towards adding options to the Run Config to start as I worry that the menu is already kind of overwhelming.  Don't feel too strongly though!

devoncarew:
A good place to start here would be to add two new menu items to the Run menu (we already have a contributed Flutter section):

- `Run in profile mode...`
- `Run in release mode...`

pq:
WIP:

[...]

PR #898, opened 2017-04-10
Support for running release and profile modes (#410).

pq:
Fixes: #410

Not quite ready to land but some food for thought.

Still TODO:

  * [x] figure out what to do for Bazel (@skybrian?)
  * [ ] update enablement logic (to disable when running)
  * [x] finalize icons

cc @skybrian @devoncarew for early feedback


resources/META-INF/plugin.xml
```diff
@@ -256,6 +256,14 @@
               description="Restart"
               icon="AllIcons.Actions.Restart">
       </action>
+      <action id="Flutter.Menu.RunProfileAction" class="io.flutter.actions.RunProfileFlutterApp"
+              description="Profile"
+              icon="FlutterIcons.RunProfile">
+      </action>
+      <action id="Flutter.Menu.RunReleaseAction" class="io.flutter.actions.RunReleaseFlutterApp"
+              description="Release"
+              icon="FlutterIcons.RunRelease">
```
devoncarew: `AllIcons.Actions`?

resources/META-INF/plugin.xml
```diff
@@ -256,6 +256,14 @@
               description="Restart"
               icon="AllIcons.Actions.Restart">
       </action>
+      <action id="Flutter.Menu.RunProfileAction" class="io.flutter.actions.RunProfileFlutterApp"
+              description="Profile"
+              icon="FlutterIcons.RunProfile">
```
devoncarew: `AllIcons.Actions.Profile`?

src/io/flutter/FlutterBundle.properties
```diff
@@ -7,6 +7,12 @@ app.reload.action.description=Hot reload changes into the running Flutter app
 app.restart.action.text=Flutter Full Restart
 app.restart.action.description=Restart the Flutter app
 
+app.profile.action.text=Run in profile mode...
+app.profile.action.description=Run Flutter app in profile mode...
```
devoncarew: We could even include the word 'Flutter' in the action name (not just the description)? It would be more specific, if a bit long.

src/io/flutter/actions/FlutterRunModeAction.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    final RunnerAndConfigurationSettings settings = RunManagerEx.getInstanceEx(project).getSelectedConfiguration();
+    if (settings == null) {
+      return;
+    }
+
+    final RunConfiguration configuration = settings.getConfiguration();
+    if (!(configuration instanceof SdkRunConfig)) {
+      //TODO: supported in Bazel?
```
devoncarew: Could we fast-fail? Say that this mode is not supported for bazel launch configs? The user's recourse will be to run from the command-line, which seems reasonable.

src/io/flutter/actions/FlutterRunModeAction.java
```diff
@@ -0,0 +1,96 @@
...
+import io.flutter.FlutterInitializer;
+import io.flutter.run.SdkFields;
+import io.flutter.run.SdkRunConfig;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.*;
+
+
+public abstract class FlutterRunModeAction extends AnAction {
```
skybrian: Possible alternative names: RunFlutterAction? RunAppAction?

src/io/flutter/actions/RunProfileFlutterApp.java
```diff
@@ -0,0 +1,22 @@
...
+import io.flutter.FlutterBundle;
+
+
+@SuppressWarnings("ComponentNotRegistered")
+public class RunProfileFlutterApp extends FlutterRunModeAction {
+  public static final String TEXT = FlutterBundle.message("app.profile.action.text");
+  public static final String DESCRIPTION = FlutterBundle.message("app.profile.action.description");
+
+  public RunProfileFlutterApp() {
+    super(TEXT, DESCRIPTION, FlutterIcons.RunProfile, "--profile", ToolWindowId.RUN
```
skybrian: Since no methods are overridden, I wonder if a static factory method on the parent class would work? Similarly for the other subclass.

src/io/flutter/actions/FlutterRunModeAction.java
skybrian: Maybe disable the action altogether for Bazel?

That is, override update() and call e.getPresentation().setEnabled(). The docs on update() say you can change the presentation in other ways; not sure what's best.

src/io/flutter/actions/FlutterRunModeAction.java
```diff
@@ -0,0 +1,96 @@
...
+      return;
+    }
+
+    final RunConfiguration configuration = settings.getConfiguration();
+    if (!(configuration instanceof SdkRunConfig)) {
+      //TODO: supported in Bazel?
+      return;
+    }
+
+    final SdkRunConfig sdkRunConfig = (SdkRunConfig)configuration.clone();
```
skybrian: It seems like cloning the run config will defeat the check that makes sure you don't launch the same config multiple times. Instead how about using putUserData() on the ExecutionEnvironment?

resources/META-INF/plugin.xml
pq: That's what I started with actually and then convinced myself we might want to swap this out.  Anyway, I'm back to agreeing. 👍

src/io/flutter/FlutterBundle.properties
pq: Sure.  I was worried about length but I think clarity is better.

src/io/flutter/actions/FlutterRunModeAction.java
pq: Right.  I think I'll disable it entirely for now. 👍

src/io/flutter/actions/FlutterRunModeAction.java
pq: Good idea.   `RunFlutterAction` wins it.

src/io/flutter/actions/FlutterRunModeAction.java
pq: I've actually got this started in another cleanup PR.  I'll follow-up w/ it.  Thanks!

src/io/flutter/actions/RunProfileFlutterApp.java
pq: Interesting.  Can I reference factory methods in action declarations?

```
      <action id="Flutter.Menu.RunProfileAction" class="io.flutter.actions.RunProfileFlutterApp"
              description="Profile"
              icon="AllIcons.Actions.Profile">

```

pq:
Main issues addressed.  I've got another PR cooking that improves enablement and updates execution to use user data.  Let me know if this one is OK to land in the meantime.

@devoncarew  @skybrian 

Thanks!

src/io/flutter/actions/RunProfileFlutterApp.java
skybrian: Oops, I don't know. Probably not.

src/io/flutter/actions/RunFlutterAction.java
```diff
@@ -0,0 +1,109 @@
...
+    final ExecutionEnvironmentBuilder builder = ExecutionEnvironmentBuilder.create(executor, sdkRunConfig);
+    final ExecutionEnvironment env = builder.activeTarget().dataContext(e.getDataContext()).build();
+
+    ProgramRunnerUtil.executeConfiguration(env, false, true);
+  }
+
+  @Override
+  public void update(AnActionEvent e) {
+    final RunnerAndConfigurationSettings settings = getRunConfigSettings(e);
+    // TODO: add support for Bazel.
```
devoncarew: nit: `TODO(pq): ...`

devoncarew:
lgtm

eturk1:
How do we run --release in Intellij?
I don't see a menu item.

devoncarew:
This has not fully landed - support is tracked here: https://github.com/flutter/flutter-intellij/issues/930.

Issue #901, opened 2017-04-11
Hot restart unexpectedly disabled.


pq:
I'd expect restart to be enabled whenever restart is...

![screen shot 2017-04-11 at 11 45 30 am](https://cloud.githubusercontent.com/assets/67586/24925222/6d05cc26-1eac-11e7-8bde-3b0c24c5d719.png)

cc @devoncarew

PR #902, opened 2017-04-11
fix the restart action enablement

devoncarew:
- fix the restart action enablement
- fix https://github.com/flutter/flutter-intellij/issues/901

In order for the action to be enabled, we need to have an instance of `RestartFlutterApp` created w/ the right FlutterApp. Also, re-add the keybindings for full restart; we still have a bit of work left to do wrt rationalizing the keybindings.

@pq


pq:
Nice.  LGTM.

PR #903, opened 2017-04-11
Update profile/release run enablement to track running Flutter app.

pq:
@devoncarew @skybrian


src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -35,6 +36,9 @@
  * A running Flutter app.
  */
 public class FlutterApp {
+
+  public static final Key<FlutterApp> FLUTTER_APP_KEY = new Key<>("FLUTTER_APP_KEY");
```
skybrian: Seems like it could be private?

pq: 👍

Issue #782, opened 2017-03-03
Support editing a Flutter project's Android java code in IntelliJ IDEA


hotznplotz:
Currently I don't get language specific IDE support for Java code. Some features like auto complete identifiers does work. Others like detecting wrong syntax and type checks doesn't. Tried it on Linux and OSX.

## Steps to Reproduce

1. follow Intellij setup instructions: https://flutter.io/intellij-setup/
2. run 'flutter create newApp'
3. open newApp with Intellij
4. check IDE support for Java code in android/app/src/...

## Version info

[✓] Flutter (on Mac OS, channel master)
    • Flutter at /Users/*****/.flutter/flutter
    • Framework revision c9644a49f2 (31 hours ago), 2017-03-02 09:21:40
    • Engine revision 74de13c0bd
    • Tools Dart version 1.23.0-dev.2.0

[✓] Android toolchain - develop for Android devices (Android SDK 25.0.2)
    • Android SDK at /Users/*****/Library/Android/sdk
    • Platform android-25, build-tools 25.0.2
    • Java(TM) SE Runtime Environment (build 1.8.0_102-b14)

[-] iOS toolchain - develop for iOS devices (Xcode 8.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 8.1, Build version 8B62
    ✗ ideviceinstaller not available; this is used to discover connected iOS devices.
      Install via 'brew install ideviceinstaller'.
    ✗ ios-deploy not installed: install via 'brew install ios-deploy'.

[✓] Android Studio (version 2.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Gradle version 2.14.1

[✓] IntelliJ IDEA Ultimate Edition (version 2016.3)
    • Dart plugin version 163.8651
    • Flutter plugin version 0.1.11.1

[-] WebStorm (version 2016.2.4)
    • Flutter plugin not installed; this adds Flutter specific functionality.
    • For information about managing plugins, see
      https://www.jetbrains.com/help/idea/managing-plugins.html

[✓] Connected devices
    • None

I am aware that IDE support for the native languages maybe isn't high priority, but would be handy when using host message passing. How do you do this when implementing libraries?

mit-mit:
Thanks @hotznplotz, this is something we are investigating.

skybrian:
[...]

PR #904, opened 2017-04-12
Enable android module if present when creating Flutter project

skybrian:
If "flutter create" writes an android.iml file, make sure it's added to the project.

Also, make sure all files generated by the "flutter create" command are visible in IDEA. (There was a race condition where we didn't wait for "flutter create" to finish before continuing.)

Progress towards fixing #782
The android.iml file will be added in https://github.com/flutter/flutter/pull/9335

### DIFF
--- a/src/io/flutter/module/FlutterModuleBuilder.java
+++ b/src/io/flutter/module/FlutterModuleBuilder.java
@@ -5,18 +5,22 @@
  */
 package io.flutter.module;
 
+import com.google.common.collect.ImmutableList;
 import com.intellij.execution.ExecutionException;
 import com.intellij.ide.util.projectWizard.ModuleBuilder;
 import com.intellij.ide.util.projectWizard.ModuleWizardStep;
 import com.intellij.ide.util.projectWizard.WizardContext;
 import com.intellij.openapi.Disposable;
+import com.intellij.openapi.application.WriteAction;
 import com.intellij.openapi.diagnostic.Logger;
-import com.intellij.openapi.module.ModuleType;
+import com.intellij.openapi.module.*;
 import com.intellij.openapi.options.ConfigurationException;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.ContentEntry;
 import com.intellij.openapi.roots.ModifiableRootModel;
+import com.intellij.openapi.roots.ui.configuration.ModulesProvider;
 import com.intellij.openapi.util.Disposer;
+import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
 import icons.FlutterIcons;
 import io.flutter.FlutterBundle;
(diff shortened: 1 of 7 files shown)


alexander-doroshko:
Make sure to test in small IDEs (no android support + unexpected things may break in multi-modular projects there). Also makes sense to test in IntelliJ IDEA with disabled Android plugin.

src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -89,22 +156,40 @@ public ModuleType getModuleType() {
...
+  private static void runFlutterCreate(@NotNull FlutterSdk sdk, Module module, @NotNull VirtualFile baseDir)
+    throws ConfigurationException {
     // Create files.
     try {
-      sdk.run(FlutterSdk.Command.CREATE, model.getModule(), baseDir, null, baseDir.getPath());
+      final Process process =
+        sdk.startProcess(FlutterSdk.Command.CREATE, module, baseDir, null, baseDir.getPath());
+      if (process != null) {
+        // Wait for process to finish. (We overwrite some files, so make sure we lose the race.)
+        // TODO(skybrian) the user sees a short pause. Show progress somehow?
```
devoncarew: Is this running in the wizard dialog, or in message console in the new IntelliJ window?

devoncarew:
> Make sure to test in small IDEs (no android support + unexpected things may break in multi-modular projects there). Also makes sense to test in IntelliJ IDEA with disabled Android plugin.

Re: that comment, I think where we want to get to with this PR is for the Android code in a Flutter project to be editable in IDEA, and for Dart code to continue to work in small IDEs (WebStorm) w/o issues (i.e., it's not an active use case for the Android code to work in WebStorm, but we don't want exceptions during project creation there).

pq:
> > Make sure to test in small IDEs (no android support + unexpected things may break in multi-modular projects there). Also makes sense to test in IntelliJ IDEA with disabled Android plugin.

>Re: that comment, I think where we want to get to with this PR is for the Android code in a Flutter project to be editable in IDEA, and for Dart code to continue to work in small IDEs (WebStorm) w/o issues (i.e., it's not an active use case for the Android code to work in WebStorm, but we don't want exceptions during project creation there).

A quick (if not long-term) solution might be to simply skip the android module configuration in all but the IDEA case.  That should solve the IDEA problem and ensure we don't confuse the small-IDEs with unsupported multi-module projects.

Thoughts?

src/io/flutter/module/FlutterModuleBuilder.java
skybrian: For IDEA this runs after the project wizard disappears and before the new project window appears - not ideal.

For WebStorm, the new project window is open, but blank.

skybrian:
WebStorm project creation takes a different code path, so the android module isn't used. (I renamed a method to make this clearer.)

When I create a project in WebStorm, the Dart SDK is not set correctly. However, it seems that was already the case without this patch.

PR #905, opened 2017-04-12
Improve Flutter run menu item messaging.

pq:
Updates menu text with selected run config (in keeping w/ other run menu items).

screen shot 2017-04-12 at 6 13 48 am

Message is a bit long...  Ideas welcome!

@skybrian @devoncarew


devoncarew:
ooo, nice. lgtm as is. Could also consider calling it `(profile mode)` / `(release mode)`; do not have a preference for which text is better

skybrian:
Seems like "Profile" would work as a verb?

What is fp-10-10-1?

pq:
I'll land it and we can iterate.

> What is fp-10-10-1?

That's the name of the active configuration.

Thanks!

PR #910, opened 2017-04-13
add an experimental reload on save option

devoncarew:
This PR adds an experimental reload-on-save option. When resources are written to disk, we:

- check if any dart files in a flutter module were affected
- check that the files are structurally sound (that they pass IntelliJ's parser)
- and if there's an active flutter launch, we issue a reload command

There are a bunch of corner cases to work through, and decisions about how best to validate that a auto-reload should happen (should we take signals from the analysis server?).

Not clear that this PR will land; not yet for review :)


FaisalAbid:
I would <3 this!

devoncarew:
note: investigate `com.intellij.openapi.actionSystem.ex.AnActionListener` to distinguish explicit user saves

devoncarew:
Closing, but still an area of active investigation :)

PR #911, opened 2017-04-14
fix build and fix unit tests to work again

skybrian:
- Jetbrains warning about unnecessary type inference seems to be incorrect?
- Registering the same action id twice causes tests to fail

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -236,22 +236,19 @@
     <!-- main toolbar run actions -->
     <group id="Flutter.MainToolbarActions.Run">
       <separator/>
-      <!--<action id="Flutter.Toolbar.ReloadAction" class="io.flutter.actions.ReloadFlutterAppRetarget"-->
-              <!--description="Reload"-->
-              <!--icon="FlutterIcons.HotReload">-->
-        <!--<keyboard-shortcut keymap="$default" first-keystroke="ctrl alt SEMICOLON"/>-->
-        <!--<keyboard-shortcut keymap="$default" first-keystroke="ctrl BACK_SLASH"/>-->
-      <!--</action>-->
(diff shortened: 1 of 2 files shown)


skybrian:
Probably not the right fix; suggestions welcome. I'd be interested in knowing if it also failed on your workstation.

resources/META-INF/plugin.xml
```diff
@@ -236,12 +236,12 @@
...
-              icon="FlutterIcons.HotReload">
-        <keyboard-shortcut keymap="$default" first-keystroke="ctrl alt SEMICOLON"/>
-        <keyboard-shortcut keymap="$default" first-keystroke="ctrl BACK_SLASH"/>
-      </action>
+      <!--<action id="Flutter.Toolbar.ReloadAction" class="io.flutter.actions.ReloadFlutterAppRetarget"-->
+              <!--description="Reload"-->
+              <!--icon="FlutterIcons.HotReload">-->
+        <!--<keyboard-shortcut keymap="$default" first-keystroke="ctrl alt SEMICOLON"/>-->
+        <!--<keyboard-shortcut keymap="$default" first-keystroke="ctrl BACK_SLASH"/>-->
+      <!--</action>-->
```
pq: Hmmm.  We do want it in both places though right?  In that case, can we replace one with an action reference?

    <reference ref="Flutter.Toolbar.ReloadAction"/>

devoncarew: +1 to the action reference

devoncarew:
@skybrian, sorry for the breakage! +1 to Phil's suggestion about using an action reference. If that doesn't work, please merge this and I'll look at alternative ways to have the action exposed in both places.

skybrian:
Using a reference fixed it.

Issue #748, opened 2017-02-20
re-consider how we define flutterness for a module


devoncarew:
Several times I've run into a situation where I open a flutter project/module in a copy of IntelliJ that has the Dart plugin installed but not the Flutter plugin. IntelliJ displays an error message about not understanding the module type.

I think we want the Flutter modules to be a slightly different flavor of Dart modules, not an entirely different module type. So, the Dart plugin could work with Flutter modules just fine, even in the absence of the Flutter plugin. Having the Flutter plugin installed would give you additional capabilities, above and beyond just static analysis, refactorings, ... available from the Dart plugin.

skybrian:
Sounds good in principle. For Bazel, we don't use the Flutter module type at all.

Do we gain anything from it in pub-based projects? (Seems like we just need to cache what we learned from pubspec.yaml?)

pq:
Early on we had some conversations about [facets](https://www.jetbrains.com/help/idea/2016.2/facet.html), which on the surface look promising.  Curious what @alexander-doroshko thinks?

alexander-doroshko:
There are no facets in Small IDEs.

pq:
Ah!  Well that explains why we didn't push harder. Any thoughts on @devoncarew's question?

alexander-doroshko:
`[Dart support enabled for the module] === [The module has an order entry pointing to a 'Dart SDK' project-level library that in its turn contains valid Dart SDK roots]`.

So one option is to try to extend this rule:
`[Flutter support enabled for a module] === [Dart support enabled + referenced Dart SDK is the one inside a Flutter SDK]`. An obvious downside is that it will be impossible to create a regular Dart project without downloading a separate standalone Dart SDK.

Another thought is Module's setOption/getOptionValue. Module options are serialized in IMLs. I didn't play much with them.

Any other way to detect that the given Dart module is also a Flutter one could be considered as well.

pq:
Cool.  Thanks for the ideas!

devoncarew:
> Also it will be impossible to have both Flutter and Dart modules within one IntelliJ IDEA project.

I don't understand this one. If the two modules had different values for their Dart SDK, one pointed to a non-flutter dart and one to a flutter dart sdk, wouldn't re recognize the first module as a Dart one, and the 2nd as a Dart+Flutter one?

skybrian:
As I understand it, Dart SDK's will soon be per-project, not per-module. We aren't planning to switch this to per-module, are we?

[...]

PR #912, opened 2017-04-14
Limit Flutter Icon customizations to Flutter projects.

pq:
cc @alexander-doroshko @devoncarew

### DIFF
--- a/src/io/flutter/project/FlutterIconProvider.java
+++ b/src/io/flutter/project/FlutterIconProvider.java
@@ -7,12 +7,13 @@ package io.flutter.project;
 
 import com.intellij.icons.AllIcons;
 import com.intellij.ide.IconProvider;
+import com.intellij.openapi.module.ModuleUtil;
 import com.intellij.openapi.util.Iconable;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiDirectory;
 import com.intellij.psi.PsiElement;
 import io.flutter.FlutterConstants;
-import io.flutter.utils.FlutterModuleUtils;
+import io.flutter.module.FlutterModuleType;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/project/FlutterIconProvider.java
```diff
@@ -20,7 +21,7 @@
 public class FlutterIconProvider extends IconProvider {
   @Nullable
   public Icon getIcon(@NotNull final PsiElement element, @Iconable.IconFlags final int flags) {
-    if (element instanceof PsiDirectory) {
+    if (element instanceof PsiDirectory && FlutterModuleUtils.hasFlutterModule(element.getProject())) {
```
alexander-doroshko: Is it possible to detect flutter module just by looking files / folders around?
`hasFlutterModule()` looks too heavy, it even may result in reading a file (`declaresFlutterDependency()`).

BTW is `if (!PlatformUtils.isIntelliJ()) ` branch needed in `hasFlutterModule()`? I thought `FlutterModuleType` used in small IDEs as well. (If it's not the case it means that you can't work with the same project in different IDEs. I hope it;s not the case).

pq: I see your concern.  For this I think checking for a declared Flutter module should suffice.  Updating.

pq:
OK.  Updated to prefer:

    ModuleUtil.hasModulesOfType(..., FlutterModuleType.getInstance())

Let me know what you think.

alexander-doroshko:
Well, the best, of course, is to check that this concrete folder is the flutter project root. You should know better how to check it. I guess it means checking this concrete `pubspec.yaml` file for flutterness which is done in `FlutterModuleUtils#declaresFlutterDependency()`. However that method contains `file.contentsToByteArray()` which I'm not sure about performance, it might be using cached content or not, I don't know. If not - then I'd recommend not to use it here. As an option, you can add your own caching to `declaresFlutterDependency()` to be sure. For example we have similar caching in `PubspecYamlUtil#getPubspecYamlInfo`.

Side note: do you agree that the second part of `hasFlutterModule()` is just a time-consuming noop?

Sorry for too many words. I always keep in mind that there may be thousands of modules of different types, that each module may contain more than one Dart project, that Dart project root may be deeper that a content root, etc., etc. Probably you have many places in the code where you implicitly assume 'standard' flutter project structure and that people do not mix Dart and Flutter projects in one IntelliJ project. Then ok, this will be one more place. Updating and reusing `hasFlutterModule()` might make sense anyway.

Feel free to ignore everything, it's a minor comment.

devoncarew:
The updated PR lgtm.

The 2nd part of `hasFlutterModule()` is there to support the small IDE case, where we don't have control over the module type. As per https://github.com/flutter/flutter-intellij/issues/748, I think we should eventually move to not use a Flutter specific module type; this would let people work with flutter projects even w/o the Flutter plugin installed (just the Dart plugin).

PR #916, opened 2017-04-17
Validate module names during creation (#609)

pq:
Same UX as described in closed PR (https://github.com/flutter/flutter-intellij/pull/724) but without the build dependency issues. 👍 

Fixes: #609 

@devoncarew @skybrian


src/io/flutter/FlutterUtils.java
```diff
@@ -58,4 +61,15 @@ public static VirtualFile getRealVirtualFile(@Nullable PsiFile psiFile) {
...
   }
+
+  /**
+   * Checks whether a given string is a valid Dart identifier.
+   *
+   * See: https://www.dartlang.org/guides/language/spec
+   * @param id the string to check
+   * @return true if a valid identifer, false otherwise.
+   */
+  public static boolean isValidDartdentifier(@NotNull String id) {
```
skybrian: Do you want to check for Dart keywords?
Here is the list we use in the protobuf code generator:
https://github.com/dart-lang/dart-protoc-plugin/blob/master/lib/names.dart#L246

Or if we don't care whether it's a Dart keyword, perhaps this method should be renamed?

src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -107,6 +109,21 @@ private static void createProjectFiles(@NotNull final ModifiableRootModel model,
     }
   }
 
+  @Override
+  public boolean validateModuleName(@NotNull String moduleName) throws ConfigurationException {
+    if (!FlutterUtils.isValidDartdentifier(moduleName)) {
+      throw new ConfigurationException("Invalid Flutter module name '" + moduleName + "'; must be a valid Dart identifier.");
```
skybrian: It's not obvious why the IDEA module name needs to be a valid Dart identifier. Add a comment explaining this?

src/io/flutter/FlutterUtils.java
devoncarew: In the past we had issues w/ users choosing project names that were the same as common pub package names (esp. the names of flutter dependencies), and running into issues w/ the first `pub get`. For `flutter create`, we proscribe the common package names: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/commands/create.dart#L248.

devoncarew: (`isValidDartdentifier` ==> `isValidDartIdentifier`)

pq:
Lots of good comments.  Thanks!  I'll follow-up with a PR in a bit. 👍

pq:
OK, some fresh updates.  Thanks again for all the feedback.

PTAL.

@devoncarew @skybrian

src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -160,6 +162,32 @@ public boolean validate(Project current, Project dest) {
...
+    if (FlutterUtils.isDartKeword(moduleName)) {
+      throw new ConfigurationException("Invalid module name: '" + moduleName + "' - must not be a Dart keyword.");
+    }
+
+    if (!FlutterUtils.isValidDartIdentifier(moduleName)) {
+      throw new ConfigurationException("Invalid module name: '" + moduleName + "' - must be a valid Dart identifier.");
+    }
+
+    if (FlutterConstants.FLUTTER_PACKAGE_DEPENDENCIES.contains(moduleName)) {
+      throw new ConfigurationException("Invalid module name: '" + moduleName + "' - this will conflict with Flutter package dependencies.");
```
devoncarew: 👍 nice messages :)

devoncarew:
lgtm

PR #920, opened 2017-04-18
remove the FlutterSdkService class and related code

devoncarew:
- remove the FlutterSdkService class and related code

@pq, It doesn't look like this service is referenced - likely obviated from various refactorings? We can bring it back if we have a similar need in the future.


pq:
Yep.  This got simplified away.  If/when we decide to custom tailor support to the different platforms this is one way to do it and agree we can bring it back then. 

LGTM 👍

PR #935, opened 2017-04-20
Sync Flutter SDK on open if Dart SDK doesn't exist (old)

skybrian:
Work in progress. This almost works except that the analysis server
needs to be refreshed.

### DIFF
--- a/src/io/flutter/project/FlutterProjectOpenProcessor.java
+++ b/src/io/flutter/project/FlutterProjectOpenProcessor.java
@@ -16,6 +16,7 @@ import com.intellij.openapi.extensions.Extensions;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.openapi.vfs.VirtualFileManager;
 import com.intellij.projectImport.ProjectOpenProcessor;
 import icons.FlutterIcons;
 import io.flutter.FlutterBundle;
@@ -68,6 +69,14 @@ public class FlutterProjectOpenProcessor extends ProjectOpenProcessor {
       return null;
     }
 
+    final FlutterSdk sdk = FlutterSdk.getForProjectDir(file);
+    if (sdk != null && !sdk.hasDartSdk()) {
+      boolean ok = sdk.syncShowingProgress(null);
+      if (!ok) {
+        LOG.warn("failed to sync flutter SDK");
+      }
+    }
+
     final Project project = importProvider.doOpenProject(file, projectToClose, forceOpenInNewFrame);
 
     // Once open, process.
@@ -79,16 +88,6 @@ public class FlutterProjectOpenProcessor extends ProjectOpenProcessor {
   private void doPostOpenProcessing(@Nullable Project project) {
     if (project == null) {
       return;
-    }
-
-    if ( FlutterSdk.getFlutterSdk(project) == null) {
-      final FlutterSdk incomplete = FlutterSdk.getIncomplete(project);
-      if (incomplete != null) {
-        boolean ok = incomplete.syncShowingProgress(project);
-        if (!ok) {
-          LOG.warn("failed to sync flutter SDK");
-        }
-      }
     }
 
     if (!FlutterModuleUtils.hasFlutterModule(project)) {
--- a/src/io/flutter/project/FlutterProjectOpenProcessor.java
+++ b/src/io/flutter/project/FlutterProjectOpenProcessor.java
@@ -22,7 +22,6 @@ import io.flutter.FlutterBundle;
 import io.flutter.FlutterMessages;
 import io.flutter.FlutterUtils;
 import io.flutter.actions.FlutterPackagesGetAction;
-import io.flutter.actions.FlutterSdkAction;
 import io.flutter.sdk.FlutterSdk;
 import io.flutter.sdk.FlutterSdkUtil;
 import io.flutter.utils.FlutterModuleUtils;
@@ -39,16 +38,6 @@ import java.util.Objects;
 public class FlutterProjectOpenProcessor extends ProjectOpenProcessor {
 
   private static final Logger LOG = Logger.getInstance(FlutterProjectOpenProcessor.class.getName());
-
-  private static void doPerform(@NotNull FlutterSdkAction action, @NotNull Project project) throws ExecutionException {
-    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
-    if (sdk == null) {
-      // Lack of SDK will be flagged elsewhere by inspection.
-      return;
-    }
-
-    action.perform(sdk, project, null);
-  }
 
   private static void handleError(@NotNull Exception e) {
     FlutterMessages.showError("Error opening", e.getMessage());
@@ -92,6 +81,16 @@ public class FlutterProjectOpenProcessor extends ProjectOpenProcessor {
       return;
     }
 
+    if ( FlutterSdk.getFlutterSdk(project) == null) {
+      final FlutterSdk incomplete = FlutterSdk.getIncomplete(project);
+      if (incomplete != null) {
+        boolean ok = incomplete.syncShowingProgress(project);
+        if (!ok) {
+          LOG.warn("failed to sync flutter SDK");
+        }
+      }
+    }
+
     if (!FlutterModuleUtils.hasFlutterModule(project)) {
       if (FlutterModuleUtils.usesFlutter(project)) {
         final List<Module> modules = FlutterModuleUtils.findModulesWithFlutterContents(project);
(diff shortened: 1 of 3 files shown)


src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -92,6 +81,16 @@ private void doPostOpenProcessing(@Nullable Project project) {
       return;
     }
 
+    if ( FlutterSdk.getFlutterSdk(project) == null) {
+      final FlutterSdk incomplete = FlutterSdk.getIncomplete(project);
+      if (incomplete != null) {
+        boolean ok = incomplete.syncShowingProgress(project);
```
skybrian: I intended this to be modal, but it's not. While we're syncing the flutter SDK, the project window opens and the analysis server starts running. As a result, the analysis server shows a bunch of errors due to missing files. Syncing the analysis server fixes it.

skybrian:
Didn't finish tonight; feel free to take it over if you have any ideas...

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -61,6 +70,59 @@ public static FlutterSdk getFlutterSdk(@NotNull final Project project) {
...
   }
 
+
+  /**
+   * Returns the Flutter SDK for a project that has a possibly broken "Dart SDK" project library.
+   * <p>
+   * (This can happen for a newly-cloned Flutter SDK where the Dart SDK is not cached yet.)
+   */
+  @Nullable
+  public static FlutterSdk getIncomplete(@NotNull final Project project) {
```
devoncarew: So, this method will always return a FlutterSdk if one is set, whether it's fully set up or not? We may want to rethink the names for this and the current `getFlutterSdk()` to clarify that this method always returns an sdk, and the other only returns fully formed sdks.

skybrian: Yep. I didn't rename the other one yet to keep the commit small.

New plan is to write a method to get the Flutter SDK of an unopened project. Then we can sync the SDK before project open begins.

skybrian:
Okay, this seems to work. Please take another look.

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -61,6 +75,57 @@ public static FlutterSdk getFlutterSdk(@NotNull final Project project) {
...
+    final VirtualFile libPath = projectDir.findFileByRelativePath(SDK_LIBRARY_PATH);
+    if (libPath == null) {
+      return null;
+    }
+
+    try {
+      final Element elt = JDOMUtil.load(new File(libPath.getPath()));
+      final ProjectLibraryTable table = new ProjectLibraryTable();
+      table.readExternal(elt);
+      return table.getLibraryByName("Dart SDK");
```
pq: I wonder if there's a more robust way to do this lookup than by string lookup.

cc @alexander-doroshko

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -217,6 +324,18 @@ public String getDartSdkPath() throws ExecutionException {
     return FlutterSdkUtil.pathToDartSdk(getHomePath());
   }
 
+  /**
+   * Returns true if the Dart SDK has been downloaded for this Flutter SDK.
+   */
+  public boolean hasDartSdk() {
+    VirtualFile file = LocalFileSystem.getInstance().findFileByPath(myHomePath + "/bin/cache/dart-sdk");
```
pq: Aside: we use `"/bin/cache/dart-sdk"` in a few places --- maybe a good idea to stuff into a constant?

testSrc/unit/io/flutter/sdk/FlutterSdkTest.java
```diff
@@ -0,0 +1,45 @@
...
+  public final ProjectFixture fixture = Testing.makeEmptyModule();
+
+  @Rule
+  public final TestDir tmp = new TestDir();
+
+  @Test
+  public void canReadFlutterSdkFromUnopenedProjectInExampleDirectory() throws Exception {
+    final VirtualFile projectDir = tmp.ensureDir("myproject");
+    tmp.ensureDir("myproject/.idea/libraries");
+    tmp.writeFile("myproject/.idea/libraries/Dart_SDK.xml",
```
pq: @alexander-doroshko : I'm guessing this structure is not API but do you have a sense for how stable we can trust it to be?

src/io/flutter/sdk/FlutterSdk.java
pq: (It's also in `DART_CORE_SUFFIX`).

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -32,13 +38,21 @@
...
 import org.jetbrains.annotations.Nullable;
 
+import java.io.File;
+import java.io.IOException;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class FlutterSdk {
   public static final String FLUTTER_SDK_GLOBAL_LIB_NAME = "Flutter SDK";
+
+  private static final String SDK_LIBRARY_PATH = ".idea/libraries/Dart_SDK.xml";
```
pq: @alexander-doroshko : is there a better way to get this?

alexander-doroshko: You may consider using a `StartupActivity` instead of `ProjectOpenProcessor` to recover the SDK. Inside a `StartupActivity` you have an instance of `Project` and can use all the Platform API mthods, so there will be no suspicious string literals in code or hacky `loadDartSdkLibrary()` method.

src/io/flutter/sdk/FlutterSdk.java
alexander-doroshko: Getting library by name is ok (you may reuse `DartSdk.DART_SDK_LIB_NAME` constant). But the rest of the method is a bit hacky :). It's because you do not have `Project` instance in hands in `ProjectOpenProcessor`.

AFAIR normally `ProjectOpenProcessor` implementations extend `ProjectOpenProcessorBase` and do not override `doOpenProject()`. They work not more than once per project. `ProjectOpenProcessor` does its job when user opens a folder without `.idea` file inside - it helps to create initial project configuration. When you open existing project `ProjectOpenProcessor` normally do not do anything. If you need to check and fix anything in the opened project `StartupActivity` is there to help. 

I may not be remembering all the nuances, but I feel it is designed to work like this.

testSrc/unit/io/flutter/sdk/FlutterSdkTest.java
alexander-doroshko: I think it's totally fine to have such hardcoded paths and XMLs in tests. In production code, you normally don't have any need to write such.

src/io/flutter/sdk/FlutterSdk.java
skybrian: Thanks for the reference; I'll see what I can do with StartupActivity.

Neither ProjectOpenProcessor nor StartupActivity have any javadoc. Could you add some? And maybe have cross-references between them. More javadoc would save a lot of time for plugin authors.

alexander-doroshko: `StartupActivity` has one very important documentation line :). The rest, I believe is pretty much clear from its name and method signature.
In general, I agree, a lot of API methods lack inline documentation. I used to learn by example.
Some useful docs for plugin writers are [here](http://www.jetbrains.org/intellij/sdk/docs/index.html).

skybrian: It's a callback, so there are many more things could be said:

- The basics: when is this callback invoked? (What is starting up? What is the state of the Project when it's called? What runs before and after?)
 - What is blocked from running until this callback finishes? What can be running at the same time?
 - If the action is lengthy, how is progress shown?
 - If there are multiple startup activities, what order do they run?

In this case it's fairly easy to read the source code. In other cases the calling source code is complicated. Getting the details about a callback's environment often requires coming up with experiments and using the debugger.

The reference to DumbAware is itself a somewhat cryptic clue. Looking at StartupManagerImpl, it significantly changes the behavior.

(Again, your advice is very helpful - but it could help more people if recorded in javadoc.)

alexander-doroshko: > requires coming up with experiments and using the debugger

That's true. And even when javadocs are present they sometimes go out-of-date, developers may forget to update them. I usually experiment/debug/read code anyway.

Let me know which of your questions are not clear, I know all the answers.

skybrian: Thanks! I will ask if I can't figure it out.

I'll note that the results of reading source code and experiments may go out of date too (in the next release, if the source code changes). Javadoc doesn't tell you about bugs, but it gives some clue what behavior the authors intended. When plugin authors need to do detailed investigation of the implementation, rather than just relying on what was both intended and works today, it makes them more likely to break.

skybrian:
Obsolete; see #942

PR #937, opened 2017-04-20
change user agent to better handle analytics on windows

devoncarew:
- change user agent to better handle analytics on windows
- fix https://github.com/flutter/flutter-intellij/issues/879

This fix is a bit of a shot in the dark, as we have a long latency between a change here and knowing if it's correctly handled by google analytics. This change does bring our synthetic user agent in-line with the dart package:usage, which does correctly identify the platform.

@skybrian

### DIFF
--- a/src/io/flutter/analytics/Analytics.java
+++ b/src/io/flutter/analytics/Analytics.java
@@ -167,10 +167,10 @@ public class Analytics {
         return "Mozilla/5.0 (Windows; Windows; Windows; " + locale + ")";
       }
       else if (SystemInfo.isMac) {
-        return "Mozilla/5.0 (Macintosh; Intel Mac OS X; Macintosh; \" + locale + \")";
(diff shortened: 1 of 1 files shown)


devoncarew:
> Sure, okay, but I wonder why we use User Agent at all since we're not a browser. Does analytics have any other way to record this?

No unfortunately it doesn't. We can send the locale in separately from the UA, but not the platform.

> Regarding testing, maybe write a separate command to POST to a test instance of analytics and see what shows up?

Good idea; perhaps a test main() method in the Analytics class.

skybrian:
I guess the best alternative would be to use custom dimensions [1], but doesn't seem worth it.

[1] https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#customs

devoncarew:
I think is we can hone in on a synthetic UA that gets us the platform, that will work well w/ the regular analytics reports.

PR #942, opened 2017-04-22
sync Flutter SDK on open if Dart SDK doesn't exist

skybrian:
Also:
  - add cross-references between different code that runs at project open
  - FlutterSdk.getDartSdkPath() now returns null if the SDK doesn't exist
  - clean up callers

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -288,6 +288,7 @@
 
   <extensions defaultExtensionNs="com.intellij">
     <consoleInputFilterProvider implementation="io.flutter.run.daemon.DaemonJsonInputFilterProvider"/>?
+    <postStartupActivity implementation="io.flutter.ProjectOpenActivity"/>
     <postStartupActivity implementation="io.flutter.FlutterInitializer"/>
     <projectService serviceInterface="io.flutter.run.daemon.DeviceService"
                     serviceImplementation="io.flutter.run.daemon.DeviceService"/>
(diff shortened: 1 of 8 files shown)


skybrian:
Continuation from #935

alexander-doroshko:
LGTM!

Side note, not related to this PR.
I noticed that `FlutterProjectOpenProcessor` works when you open an existing folder using File -> Open (**regardless** whether the folder already contains IntelliJ project files (`.idea`) or not!), but doesn't work when you reopen existing project from the Welcome Screen recent list or using File -> Open Recent.

You may reconsider what you want to achieve using `FlutterProjectOpenProcessor`. If you want it to work not more than once (only when opening an existing folder without IntelliJ project metadata) you probably need to stop overriding `doOpenProject()` and `isStrongProjectInfoHolder()`, but instead extend `ProjectOpenProcessorBase`, override its `canOpenProject()` and implement its `doQuickImport()`.

If you want some logic to run every time project is opened (regardless of how it is being opened and whether it has been just auto-created by opening a folder without `.idea` or not) you may rewrite it to be one more `StartupActivity` or merge logic into this `ProjectOpenActivity`.

alexander-doroshko:
Disclaimer: don't trust me blindly (I may be wrong in details), read the code and debug :-).

skybrian:
Good point. Unfortunately I don't know what was intended. For example, do we always want to run "flutter packages get" on project open if the .packages file is missing? (Consider the case where the user deleted it.)

Looks like @pq wrote this code originally.

pq:
Right.   The idea was to make it easier for end-users to not have to understand the nuanced differences between the many ways to bring source code into IntelliJ (and get it mostly right for the common cases). It sounds like we have a little more work to do on that front! 😄

devoncarew:
> It sounds like we have a little more work to do on that front!

We should collect any actions items from here into a new issue so we don't lose track of the work.

skybrian:
Updated javadoc to better explain what happens now. I'll look into moving the actions in a different pull request.

Issue #947, opened 2017-04-25
Packages get runs off stale file info


mit-mit:
It appears that 'packages get' in IJ on the `pubspec.yaml` file runs off the last explicitly saved file, not the current file contents at the time the button is pressed.

This was highly surprising; @szakarias had concluded that the button was likely entirely broken until we figured out that is what was happening.

## Steps to Reproduce

1. Create a new project in IJ
1. Open `pubspec.yaml`
1. Hit Packages Get

=> succeeds as expected

1. Replace:
```
  flutter:
    sdk: flutter
```
with:
```
  flutter:
    sdk: flutter
  url_launche:
```
1. Hit Packages get

=> succeeds, but url_launcher is not listed under Dart Packages in the External Libraries node

1. Hit cmd-s
1. Hit Packages get

=> pub get failure

1. Correct `url_launche:` to `url_launcher:`
1. Hit Packages get

=> still fails

1. Hit cmd-s
1. Hit Packages get

=> finally succeeds

mit-mit:
Marking this as I/O adoption blocker, as codelabs will use plugins and this require edits to `pubspec.yaml`.

alexander-doroshko:
You need to invoke `FileDocumentManager.getInstance().saveAllDocuments();` before calling any command line tool that reads files from the disk.
For example 'pub get' action provided by the Dart plugin does it here: https://github.com/JetBrains/intellij-plugins/blob/c9e52bf8c08c1d592d3497e98fae8538594b6280/Dart/src/com/jetbrains/lang/dart/ide/actions/DartPubActionBase.java#L190

devoncarew:
Sounds like this will be an easy fix. We'll just want to determine whether it makes sense to save all files before every flutter tool operation.

PR #949, opened 2017-04-25
save all before running flutter actions

devoncarew:
- save dirty documents before running flutter actions
- fix https://github.com/flutter/flutter-intellij/issues/947

@pq @skybrian

### DIFF
--- a/src/io/flutter/actions/FlutterSdkAction.java
+++ b/src/io/flutter/actions/FlutterSdkAction.java
@@ -10,6 +10,7 @@ import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.actionSystem.CommonDataKeys;
 import com.intellij.openapi.actionSystem.LangDataKeys;
 import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.fileEditor.FileDocumentManager;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.DumbAwareAction;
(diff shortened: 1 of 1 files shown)


src/io/flutter/actions/FlutterSdkAction.java
```diff
@@ -88,11 +85,14 @@ public void actionPerformed(AnActionEvent event) {
...
-                            @SuppressWarnings("SameParameterValue") boolean logAction)
+                            @SuppressWarnings("SameParameterValue") boolean sendAnalytics)
     throws ExecutionException {
-    if (logAction) {
-      sendActionEvent();
+    if (sendAnalytics) {
+      FlutterInitializer.sendAnalyticsAction(this);
     }
+
+    FileDocumentManager.getInstance().saveAllDocuments();
```
pq: Pretty sure we want to make sure this is run on the event dispatch thread.  IOW wrap in

    ApplicationManager.getApplication().runWriteAction( ... )

?

alexander-doroshko: @pq `runWriteAction()` doesn't change thread :).

alexander-doroshko: Take a look at the implementation, it takes write action inside if needed, so we are fine here because `actionPerformed()` is called in EDT.

devoncarew: sg

PR #955, opened 2017-04-27
Simplify FlutterSdkAction and subclasses

skybrian:
Probably no user-visible changes.

### DIFF
--- a/src/io/flutter/FlutterConstants.java
+++ b/src/io/flutter/FlutterConstants.java
@@ -8,9 +8,6 @@ package io.flutter;
 
 public class FlutterConstants {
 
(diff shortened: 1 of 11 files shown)


skybrian:
@devoncarew I noticed a few places where we don't have analytics. It's not a regression, so I'm going to go ahead and commit this (with TODO's at those places).

devoncarew:
>  I noticed a few places where we don't have analytics. It's not a regression, so I'm going to go ahead and commit this (with TODO's at those places).

Ah, I see those now; good catch. We should probably add analytics here, for things that are user initiated actions.

PR #957, opened 2017-04-27
return the result of calling service protocol extensions

devoncarew:
- return the result of calling service protocol extensions
- use this in the `TogglePlatformAction` action, in order to know which platform the device is currently rendering, so we can toggle appropriately

The `app.callServiceExtension` method is now documented here: https://github.com/flutter/flutter/wiki/The-flutter-daemon-mode#callserviceextension)

@pq @skybrian

### DIFF
--- a/src/io/flutter/run/daemon/DaemonApi.java
+++ b/src/io/flutter/run/daemon/DaemonApi.java
@@ -181,8 +181,8 @@ public class DaemonApi {
     cmd.complete(result);
   }
 
-  private @Nullable
-  Command takePending(int id) {
+  @Nullable
+  private Command takePending(int id) {
     final Command cmd;
     synchronized (pending) {
       cmd = pending.remove(id);
@@ -227,8 +227,8 @@ public class DaemonApi {
     }
   }
 
-  private static @Nullable
-  PrintWriter getStdin(ProcessHandler processHandler) {
+  @Nullable
+  private static PrintWriter getStdin(ProcessHandler processHandler) {
     final OutputStream stdin = processHandler.getProcessInput();
     if (stdin == null) return null;
     return new PrintWriter(new OutputStreamWriter(stdin, Charsets.UTF_8));
--- a/src/io/flutter/run/daemon/DaemonApi.java
+++ b/src/io/flutter/run/daemon/DaemonApi.java
@@ -29,14 +29,16 @@ import java.util.function.Function;
 
 /**
  * Sends JSON commands to a flutter daemon process, assigning a new id to each one.
- *
+ * <p>
  * <p>Also handles dispatching incoming responses and events.
- *
+ * <p>
  * <p>The protocol is specified in
  * <a href="https://github.com/flutter/flutter/wiki/The-flutter-daemon-mode"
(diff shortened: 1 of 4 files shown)


src/io/flutter/run/daemon/DaemonApi.java
```diff
@@ -209,19 +217,25 @@ private static void sendCommand(String json, ProcessHandler handler) {
...
+      System.out.println("[--> " + json + "]");
+    }
+
     if (stdin.checkError()) {
       LOG.warn("can't write command to Flutter process: " + json);
     }
   }
 
-  private static @Nullable PrintWriter getStdin(ProcessHandler processHandler) {
+  private static @Nullable
```
skybrian: move "private static" to the next line

src/io/flutter/run/daemon/DaemonApi.java
```diff
@@ -174,9 +181,10 @@ private void complete(int id, @Nullable JsonElement result) {
     cmd.complete(result);
   }
 
-  private @Nullable Command takePending(int id) {
+  private @Nullable
```
skybrian: move 'private' to next line

src/io/flutter/view/FlutterView.java
```diff
@@ -254,27 +256,33 @@ protected void perform(AnActionEvent event) {
...
 
   @Override
   public void actionPerformed(AnActionEvent event) {
-    isAndroid = !isAndroid;
-
     // ext.flutter.platformOverride, value: iOS, android
-    final Map<String, Object> params = new HashMap<>();
-    params.put("value", isAndroid ? "android" : "iOS");
-    view.getFlutterApp().callServiceExtension("ext.flutter.platformOverride", params);
+    final CompletableFuture<JsonObject> result = view.getFlutterApp().callServiceExtension(
```
skybrian: Maybe move the implementation to a method on FlutterApp?

view.getFlutterApp().togglePlatform().thenAccept((boolean isAndroid) -> {
  ... print console message ...
});

src/io/flutter/view/FlutterView.java
```diff
@@ -254,27 +256,33 @@ protected void perform(AnActionEvent event) {
...
-    }
+      final boolean switchToAndroid = !"android".equals(obj.get("value").getAsString());
+
+      final Map<String, Object> params = new HashMap<>();
+      params.put("value", switchToAndroid ? "android" : "iOS");
+      view.getFlutterApp().callServiceExtension("ext.flutter.platformOverride", params);
+
+      final ConsoleView console = view.getFlutterApp().getConsole();
+      if (console != null) {
+        console.print(FlutterBundle.message("flutter.view.togglePlatform.output", switchToAndroid ? "Android" : "iOS"),
```
skybrian: The console message is: `Switched platform OS to {0}`

But, technically, we don't know that it worked since we didn't wait for the response to the second call. Maybe call thenAccept() again?

src/io/flutter/view/FlutterView.java
devoncarew: sounds good - it does make the API a little higher level

PR #965, opened 2017-04-28
Clean up FlutterSdk and FlutterConsole

skybrian:
Changes in behavior:
- Always print exit code.
- Don't set Dart SDK because "flutter create" will do it
- Make sure "pub action in progress" is set for entire flutter command


src/io/flutter/console/FlutterConsole.java
```diff
@@ -0,0 +1,107 @@
...
+    final String title = module != null ? "[" + module.getName() + "] Flutter" : "Flutter";
+    final Content content = ContentFactory.SERVICE.getInstance().createContent(panel.getComponent(), title, true);
+    Disposer.register(content, view);
+
+    return new FlutterConsole(view, content, project, module);
+  }
+
+  /**
+   * Starts displaying the output of a different process.
+   */
```
pq: Maybe annotate `@NotNull` (just to be safe)?

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -369,27 +339,12 @@ private static FlutterSdk getFlutterFromDartSdkLibrary(Library lib, VirtualFile
...
-
-    CREATE("Flutter create", "create") {
-      @Override
-      void onStart(@Nullable Module module, @Nullable VirtualFile workingDir, @NotNull String... args) {
-        // Enable Dart.
-        if (module != null) {
-          ApplicationManager.getApplication()
-            .invokeLater(() -> ApplicationManager.getApplication().runWriteAction(() -> DartPlugin.enableDartSdk(module)));
-        }
-      }
```
pq: On a quick glance, I'm not seeing where this logic moved to...  Give me a pointer?

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -369,27 +339,12 @@ private static FlutterSdk getFlutterFromDartSdkLibrary(Library lib, VirtualFile
...
     DOCTOR("Flutter doctor", "doctor"),
     PACKAGES_GET("Flutter packages get", "packages", "get"),
     PACKAGES_UPGRADE("Flutter packages upgrade", "packages", "upgrade"),
     UPGRADE("Flutter upgrade", "upgrade"),
-    VERSION("Flutter version", "--version") {
-      @Override
-      boolean attachToConsole() {
-        return false;
-      }
-    };
```
pq: Similarly, how are we conditionally attaching processes to consoles?  (Sorry I know it's right in front of me!)

src/io/flutter/sdk/FlutterSdk.java
skybrian: We don't need to do it in the plugin anymore because running "flutter create" generates the iml file with a reference to the project library.

src/io/flutter/sdk/FlutterSdk.java
skybrian: I looked at the usages of VERSION and there aren't any that pass a Project to startProcessWithoutModule(). That's enough to prevent a console from being attached, so this is basically dead code.

src/io/flutter/console/FlutterConsole.java
skybrian: Okay, I added it to a future CL since it's easier.

src/io/flutter/sdk/FlutterSdk.java
pq: 👍

src/io/flutter/console/FlutterConsole.java
pq: Thanks!

src/io/flutter/sdk/FlutterSdk.java
pq: Ah, right. My only reservation is that if there are any IML format changes running it from the IDE (possibly redundantly) ensures that things work but it sounds like we're committed to keeping this up to date so NP.

Thanks!

src/io/flutter/console/FlutterConsoles.java
```diff
@@ -0,0 +1,65 @@
...
+import com.intellij.openapi.util.Key;
+import com.intellij.ui.content.Content;
+import com.intellij.ui.content.MessageView;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Methods that use the appropriate Flutter console.
+ * <p>
+ * We share one Flutter console per Module, and have one global Flutter console (for tasks which don't
```
devoncarew: Not for this PR, but we may want to consider moving to having just one Flutter console per project (a bit simpler for the user than seeing multiple flutter console open at one time). I think this has the drawback that we won't be able to do as good a job at hyperlinking some content.

src/io/flutter/sdk/FlutterSdk.java
devoncarew: Should we verify that there is a dart sdk set up for this project? As a backup to the flutter create metadata?

skybrian: This sounds like a good thing to put in a test, once we have integration tests working.

src/io/flutter/console/FlutterConsoles.java
skybrian: Okay, maybe we could have one console per project but still pass in the Module if it's available.

PR #967, opened 2017-04-29
Handle project open more consistently

skybrian:
- If .packages file is missing, we don't need to run two flutter commands.
- On project open, check if we should create a run configuration
- On project open, check if we should open lib/main.dart

(Depends on #965; look at the second commit.)

### DIFF
--- a/src/io/flutter/ProjectOpenActivity.java
+++ b/src/io/flutter/ProjectOpenActivity.java
@@ -47,9 +47,6 @@ public class ProjectOpenActivity implements StartupActivity, DumbAware {
     if (!root.hasUpToDatePackages()) {
       Notifications.Bus.notify(new PackagesOutOfDateNotification(project));
     }
-
-    FlutterModuleUtils.autoCreateRunConfig(project, root);
-    FlutterModuleUtils.autoShowMain(project, root);
   }
 
   /**
--- a/src/io/flutter/ProjectOpenActivity.java
+++ b/src/io/flutter/ProjectOpenActivity.java
@@ -5,11 +5,20 @@
  */
 package io.flutter;
 
+import com.intellij.execution.ExecutionException;
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationType;
+import com.intellij.notification.Notifications;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.DumbAware;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.startup.StartupActivity;
+import icons.FlutterIcons;
+import io.flutter.pub.PubRoot;
 import io.flutter.sdk.FlutterSdk;
+import io.flutter.utils.FlutterModuleUtils;
 import org.jetbrains.annotations.NotNull;
 
 /**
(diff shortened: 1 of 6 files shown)


src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -55,7 +46,7 @@ public Icon getIcon() {
   @Override
   public boolean canOpenProject(@Nullable VirtualFile file) {
     if (file == null) return false;
-    final PubRoot root = PubRoot.forDirectoryWithRefresh(file);
+    final PubRoot root = PubRoot.forDirectory(file);
```
pq: Do we not need to refresh here?  Or is that happening elsewhere?

src/io/flutter/ProjectOpenActivity.java
```diff
@@ -20,12 +29,108 @@
 public class ProjectOpenActivity implements StartupActivity, DumbAware {
   @Override
   public void runActivity(@NotNull Project project) {
+    if (!FlutterModuleUtils.hasFlutterModule(project)) {
+      // This method is called in twice when importing a project, because the project is reloaded.
```
devoncarew: nice catch

src/io/flutter/utils/FlutterModuleUtils.java
```diff
@@ -80,43 +78,52 @@ public static boolean usesFlutter(@NotNull Project project) {
...
+      runManager.createRunConfiguration(project.getName(), configType.getFactory());
+
+    final SdkRunConfig config = (SdkRunConfig)settings.getConfiguration();
 
-      runManager.addConfiguration(settings, false);
-      runManager.setSelectedConfiguration(settings);
+    // Set config name.
+    final String name = config.suggestedName();
+    if (name == null) {
+      config.setName(project.getName());
```
devoncarew: I saw a thread go by (don't remember the channel) about us having two launch configs for a project - one `main.dart` and one named for the specific project. We may want to standardize on using the file name of the entrypoint lib when creating. So, something line `main.getName()`.

devoncarew:
I'll defer my portion of the review to @pq, but I had one comment about the name of the launch config we auto-create.

src/io/flutter/project/FlutterProjectOpenProcessor.java
skybrian: I forgot the exact error but we can't refresh here. Since the user is looking at the project open dialog, presumably IDEA's filesystem cache is pretty up to date.

src/io/flutter/utils/FlutterModuleUtils.java
skybrian: Okay I'll look into this.

skybrian: Reproduced.

It looks like the templates for "flutter create" use "main.dart" as the name, and it's not detected here because .idea/runConfigurations isn't aren't loaded from disk yet.

I removed the calls to create the run configuration in FlutterModuleBuilder because the template already does it. Also, when opening a project, I moved the run configuration creation logic to happen later. It looks like the run configurations are loaded by then.

Also changed the name to "main.dart" to match the template.

skybrian:
Please take a look

skybrian:
Ping...

devoncarew:
lgtm!

pq:
LGTM2!

PR #990, opened 2017-05-04
fix an issue w/ displaying 'no devices' text

devoncarew:
- fix an issue w/ displaying 'no devices' text

@skybrian


src/io/flutter/actions/DeviceSelectorAction.java
```diff
@@ -137,7 +139,9 @@ private void updateActions(@NotNull Project project, Presentation presentation)
       }
     }
 
-    presentation.setText("<no devices>");
+    if (devices.isEmpty()) {
+      presentation.setText("<no devices>");
+    }
```
skybrian: Otherwise, maybe setText(null) like we used to do?

devoncarew: sounds good

devoncarew:
updated

Issue #993, opened 2017-05-05
don't use the VERSION file from the flutter sdk


devoncarew:
It's not really maintained (and may go away in the future). If we need to do version checks, it looks like https://github.com/flutter/flutter/blob/master/packages/flutter/pubspec.yaml is now being maintained.

devoncarew:
Closing - as of https://github.com/flutter/flutter-intellij/pull/1002, we handle flutter sdks w/ and w/o VERSION files.

PR #994, opened 2017-05-05
remove the use of the VERSION file

devoncarew:
- remove the use of the VERSION file
- https://github.com/flutter/flutter-intellij/issues/993

If we need to do version checks in the future, we can restore parts of this PR, but referencing https://github.com/flutter/flutter/blob/master/packages/flutter/pubspec.yaml instead.

@pq


devoncarew:
good point - I'm going to close this in favor of a PR which gracefully handles no VERSION file, and open an issue to clarify where we should be pulling the version info from

Issue #980, opened 2017-05-02
Respond to 'daemon.showMessage' events by showing toast


tvolkert:
If we added a `--machine` flag to `flutter doctor`, then IntelliJ could invoke that and provide preemptive warnings to the user (more than just console output seen by the user when triggering doctor manually).

Like https://github.com/flutter/flutter-intellij/issues/976, this is in response to a report from a user whereby IntelliJ didn't pick up the `ANDROID_HOME` environment variable, which (combined with them having a non-standard Android SDK install location) meant that `flutter daemon` couldn't locate an Android SDK, and the devices dropdown was empty.  If IntelliJ had the doctor metadata handy, it could have provided a way better user experience in guiding them towards success.

tvolkert:
Related: https://github.com/flutter/flutter-intellij/issues/981

tvolkert:
After discussion and resulting changes in https://github.com/flutter/flutter/pull/9749, I'm re-purposing this issue to make it cover the IntelliJ side of the `'daemon.showMessage'` event.

devoncarew:
👍

PR #1003, opened 2017-05-08
handle the daemon showMessage event

devoncarew:
- handle the daemon showMessage event
- added docs for the event here: https://github.com/flutter/flutter/wiki/The-flutter-daemon-mode#showmessage
- fix https://github.com/flutter/flutter-intellij/issues/980

@skybrian (/cc @tvolkert)

### DIFF
--- a/src/io/flutter/FlutterMessages.java
+++ b/src/io/flutter/FlutterMessages.java
@@ -29,6 +29,16 @@ public class FlutterMessages {
                        NotificationType.ERROR));
   }
 
+  public static void showMessage(String title, String message) {
+    final Notification notification = new Notification(
+      FLUTTER_NOTIFICATION_GOUP_ID,
+      title,
+      message,
+      NotificationType.INFORMATION);
+    notification.setIcon(FlutterIcons.Flutter);
+    Notifications.Bus.notify(notification);
+  }
+
   public static int showDialog(@Nullable Project project,
                                @NotNull String message,
                                @NotNull @Nls(capitalization = Nls.Capitalization.Title) String title,
--- a/src/io/flutter/FlutterMessages.java
+++ b/src/io/flutter/FlutterMessages.java
@@ -29,7 +29,15 @@ public class FlutterMessages {
                        NotificationType.ERROR));
   }
 
-  public static void showMessage(String title, String message) {
+  public static void showWarning(String title, String message) {
+    Notifications.Bus.notify(
+      new Notification(FLUTTER_NOTIFICATION_GOUP_ID,
+                       title,
+                       message,
+                       NotificationType.WARNING));
+  }
+
+  public static void showInfo(String title, String message) {
     final Notification notification = new Notification(
       FLUTTER_NOTIFICATION_GOUP_ID,
       title,
--- a/src/io/flutter/run/daemon/DeviceDaemon.java
+++ b/src/io/flutter/run/daemon/DeviceDaemon.java
@@ -16,6 +16,7 @@ import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.vfs.CharsetToolkit;
+import io.flutter.FlutterMessages;
 import io.flutter.bazel.Workspace;
 import io.flutter.bazel.WorkspaceCache;
 import io.flutter.sdk.FlutterSdk;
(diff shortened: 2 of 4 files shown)


devoncarew:
screen shot 2017-05-08 at 7 41 02 am

src/io/flutter/FlutterMessages.java
```diff
@@ -29,6 +29,16 @@ public static void showError(String title, String message) {
                        NotificationType.ERROR));
   }
 
+  public static void showMessage(String title, String message) {
+    final Notification notification = new Notification(
+      FLUTTER_NOTIFICATION_GOUP_ID,
+      title,
+      message,
+      NotificationType.INFORMATION);
```
tvolkert: Should we add a `type` field to the daemon protocol so we can distinguish between informational messages and error or warning messages?

tvolkert:
LGTM

src/io/flutter/FlutterMessages.java
devoncarew: I was thinking about that - the toast API does distinguish between info, warning, and error. `NotificationType.INFORMATION` doesn't have any proscriptive UI - it's fairly bland - so good as a catch-all, but doesn't convey something like an error state.

tvolkert: I'm happy to add the info on the tools side - any preference on the serialization?

src/io/flutter/FlutterMessages.java
```diff
@@ -29,6 +29,16 @@ public static void showError(String title, String message) {
                        NotificationType.ERROR));
   }
 
+  public static void showMessage(String title, String message) {
```
skybrian: Maybe rename to showInfoToast and the previous method to showErrorToast. (Assuming that's what they do.)

src/io/flutter/run/daemon/DeviceDaemon.java
```diff
@@ -262,6 +263,11 @@ public void onDaemonLogMessage(@NotNull DaemonEvent.LogMessage message) {
       LOG.info("flutter device watcher: " + message.message);
     }
 
+    @Override
+    public void onDaemonShowMessage(@NotNull DaemonEvent.ShowMessage event) {
+      FlutterMessages.showMessage(event.title, event.message);
```
skybrian: We're driving the UI directly from flutter_tools and I worry that it's not obvious to someone working on the flutter_tools side what the UI effect will be. So if that's how we want to do it, maybe the message should be called daemon.showToast?

Alternately, if we want to abstract away from the UI in the protocol, there should probably be a unique id for each kind of message, and a switch here to decide how to display each kind of message.

src/io/flutter/FlutterMessages.java
devoncarew: Perhaps something like a `level` field, w/ values of `info`, `warning`, and `error`?

src/io/flutter/FlutterMessages.java
devoncarew: sgtm

src/io/flutter/run/daemon/DeviceDaemon.java
devoncarew: The intent of the `showMessage` event is that clients can choose how to render it - most IDE clients would likely translate it into a toast. Perhaps this could be make more clear by docs on the flutter_tools side - to indicate that `showMessage` will have a client effect (I tried to make that pretty clear in the wiki protocol docs: https://github.com/flutter/flutter/wiki/The-flutter-daemon-mode#showmessage)

skybrian: Okay, sounds good.

tvolkert: Working on a flutter_tools PR now...

tvolkert: https://github.com/flutter/flutter/pull/9889

src/io/flutter/FlutterMessages.java
devoncarew: (updated this PR)

src/io/flutter/FlutterMessages.java
devoncarew: W/ the addition of the `level` field, these method names now read as: `showError`, `showWarning`, and `showInfo`.

tvolkert:
LGTM!

PR #1004, opened 2017-05-08
fix Flutter version handling again

skybrian:
Redo fix from #1002, broken due to a bad merge.

While we're here, move this code to FlutterSdkVersion and simplify.

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkVersion.java
+++ b/src/io/flutter/sdk/FlutterSdkVersion.java
@@ -6,39 +6,85 @@
 package io.flutter.sdk;
 
 
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.util.Pair;
 import com.intellij.openapi.util.Version;
-import org.jetbrains.annotations.NonNls;
+import com.intellij.openapi.util.io.FileUtil;
+import com.intellij.openapi.vfs.VirtualFile;
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 
-public class FlutterSdkVersion implements Comparable<FlutterSdkVersion> {
+import java.io.File;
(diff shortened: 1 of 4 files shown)


src/io/flutter/sdk/FlutterSdkVersion.java
```diff
@@ -6,39 +6,85 @@
...
 
-  public static FlutterSdkVersion forVersionString(@Nullable String versionString) {
-    return versionString == null ? UNKNOWN : new FlutterSdkVersion(versionString);
+  @NotNull
+  public static FlutterSdkVersion readFromSdk(@NotNull VirtualFile sdkHome) {
+    final File versionFile = new File(sdkHome.getPath() + "/VERSION");
+
+    // Use the cache if the file's last modfication date didn't change.
+    // But we still stat the file every time this is called. Not sure the cache saves us much?
+    // TODO(skybrian) should we use the VirtualFile cache instead?
```
devoncarew: +1

skybrian: +1 for making this change?

Side effect will be that it could be out of date sometimes, depending on how often IDEA refreshes the cache.

devoncarew: Yes, I was agreeing with the TODO in favor of making the change. I don't know the caching details of `VirtualFile`, but like the idea of relying on an IntelliJ solution instead of hand-rolling. Does not need to be for this PR :)

Looks like the Dart plugin does some caching of the pubspec.yaml file here: https://github.com/JetBrains/intellij-plugins/blob/f6e42fd98e5d583f983a6efe0881a1cd8ce20249/Dart/src/com/jetbrains/lang/dart/util/PubspecYamlUtil.java#L122.

PR #1007, opened 2017-05-09
On project open, set the Java SDK based on ANDROID_HOME

### LINKED ISSUE #1001: Wrong Java version default when creating a project from IntelliJ
Not sure what is going on here, this seems a little weird...

Before reproing, make sure to `flutter upgrade` as the android templates changed recently.

## Case 1: Create from terminal

Run these steps:

```
flutter create test1
cd test1
idea .
```

Then, in project view navigate to /android/app/.../MainActivity.java and notice that there are no errors shown.

## Case 2: Create from IntelliJ

1. Close all projects, and restart IJ.
1. Create New Project > Flutter > Call it test2

 Then, in project view navigate to /android/app/.../MainActivity.java and notice that you get an error for the `@override` annotation.

You can fix this in module settings by changing to language level8. BUT, why is this needed in this case when it's not needed in case 1?

skybrian:
(This only happens if the Android SDK already exists and the project's Java SDK is unset.)

Fixes #1001

### DIFF
--- a/src/io/flutter/FlutterInitializer.java
+++ b/src/io/flutter/FlutterInitializer.java
@@ -13,11 +13,14 @@ import com.intellij.notification.*;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.application.ApplicationInfo;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.extensions.PluginId;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.roots.ProjectRootManager;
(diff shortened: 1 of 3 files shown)


src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -26,7 +26,6 @@
 import io.flutter.FlutterBundle;
 import io.flutter.FlutterConstants;
 import io.flutter.FlutterUtils;
-import io.flutter.dart.DartPlugin;
```
pq: It looks from the build failure like removing this was a mistake?

skybrian: Oops, introduced while rebasing the patch.

PR #1010, opened 2017-05-10
Use project's Java SDK to set ANDROID_HOME for flutter commands

### LINKED ISSUE #981: Add an option for the user to specify their Android SDK location
If we added a `--android-sdk` flag to flutter tools to allow the user to explicitly specify the location of their Android SDK, then we could likewise add a UI in the IntelliJ plugin whereby the user could enter the SDK location, and the plugin would wire it up to the flag when it invoked the tools.

skybrian:
This allows the user to override the setting of ANDROID_HOME from IntelliJ's
environment, and ensures that it's consistent with the one used to resolve
imports in Java code.

If the project's Java SDK isn't set to an Android SDK, we still fall back
to using ANDROID_HOME.

Fixes #981

### DIFF
--- a/src/io/flutter/android/AndroidSdk.java
+++ b/src/io/flutter/android/AndroidSdk.java
@@ -53,6 +53,15 @@ public class AndroidSdk {
   }
 
   /**
+   * Returns the Java SDK in the project's configuration, or null if not an Android SDK.
+   */
+  @Nullable
+  public static AndroidSdk fromProject(@NotNull Project project) {
+    final Sdk candidate = ProjectRootManager.getInstance(project).getProjectSdk();
+    return fromSdk(candidate);
+  }
+
+  /**
    * Returns the Android SDK that matches the ANDROID_HOME environment variable, provided it exists.
    */
   @Nullable
@@ -68,6 +77,14 @@ public class AndroidSdk {
       return null;
     }
 
+    return fromHome(file);
+  }
+
+  /**
+   * Returns the Android SDK for the given home directory, or null if no SDK matches.
+   */
+  @Nullable
+  public static AndroidSdk fromHome(VirtualFile file) {
     for (AndroidSdk candidate : findAll()) {
       if (file.equals(candidate.getHome())) {
         return candidate;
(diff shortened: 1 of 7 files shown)


src/io/flutter/android/AndroidSdk.java
```diff
@@ -78,6 +95,16 @@ public static AndroidSdk fromEnvironment() {
   }
 
   /**
+   * Returns the best value of ANDROID_HOME to use.
+   * <p>
+   * If the given project has an Android SDK set, prefer that. Otherwise get it from the environment.
+   */
+  public static String chooseAndroidHome(@Nullable Project project) {
+    final AndroidSdk sdk = project == null ? null : fromProject(project);
+    return sdk == null ? EnvironmentUtil.getValue("ANDROID_HOME") : sdk.getHome().getPath();
```
pq: Maybe consider storing this String as a constant?  Or maybe class (or enum) that encapsulates these accesses?  For example:

```
class Environment {
   public static final String ANDROID_HOME = EnvironmentUtil.getValue("ANDROID_HOME");
}
```

pq: Oh hmmm, since values can change I guess this'd have to be method call.  Hmpfh.  Feel free to ignore.

src/io/flutter/run/daemon/DeviceDaemon.java
```diff
@@ -148,10 +151,17 @@ private static boolean usesFlutter(Project p) {
...
     private final @NotNull ImmutableList<String> parameters;
 
-    private Command(@NotNull String workDir, @NotNull String command, @NotNull ImmutableList<String> parameters) {
+    /**
+     * The value of ANDROID_HOME to use when launching the command.
+     */
+    private final @Nullable String androidHome;
+
+    private Command(@NotNull String workDir, @NotNull String command, @NotNull ImmutableList<String> parameters,
+                    @Nullable String androidHome) {
```
pq: Are there other environment variables we may want to inject as well (e.g., for iOS development)?  If so, maybe we can abstract this a little in the future to be something like an env map.

skybrian: Maybe, but I think we can defer it until we find the second variable.

Issue #1009, opened 2017-05-09
Have the new project wizard show failure messages from 'flutter create'.


branflake2267:
I can't create an application with caps. Lowercase app names work. 

Same goes for the command.
` flutter create ./MyApp`

## Steps to Reproduce
1. Create app with Caps
![screen shot 2017-05-09 at 4 40 24 pm](https://cloud.githubusercontent.com/assets/1326504/25877092/38df5d28-34d6-11e7-8e10-f0fd21e5fe95.png)
2. Notice Failure
![screen shot 2017-05-09 at 4 30 23 pm](https://cloud.githubusercontent.com/assets/1326504/25877096/3f0bc15a-34d6-11e7-8162-5045fea995d2.png)

## Version info
```
[✓] Flutter (on Mac OS X 10.12.4 16E195, channel master)
    • Flutter at /Users/branflake2267/git/flutter
    • Framework revision 5b18fca053 (60 minutes ago), 2017-05-09 15:41:06 -0700
    • Engine revision cbbf956fcb
    • Tools Dart version 1.23.0-dev.11.11

[✓] Android toolchain - develop for Android devices (Android SDK 25.0.3)
    • Android SDK at /Users/branflake2267/Library/Android/sdk
    • Platform android-25, build-tools 25.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version: OpenJDK Runtime Environment (build 1.8.0_112-release-b06)

[✓] iOS toolchain - develop for iOS devices (Xcode 8.3.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 8.3.2, Build version 8E2002
    • ios-deploy 1.9.1
    • CocoaPods version 1.2.1

[✓] Android Studio (version 2.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Gradle version 3.2
    • Java version: OpenJDK Runtime Environment (build 1.8.0_112-release-b06)

[✓] IntelliJ IDEA Ultimate Edition (version 2017.1.2)
    • Dart plugin version 171.4424
    • Flutter plugin version 13.1

[✓] Connected devices
    • None
```

branflake2267:
Workaround: Use app name with all lowercase characters. 
[...]

PR #1012, opened 2017-05-10
Add package_name validation (#1009)

pq:
Fixes: #1009 

@devoncarew @skybrian

### DIFF
--- a/src/io/flutter/module/FlutterModuleBuilder.java
+++ b/src/io/flutter/module/FlutterModuleBuilder.java
@@ -168,8 +168,8 @@ public class FlutterModuleBuilder extends ModuleBuilder {
 
     // See: https://www.dartlang.org/tools/pub/pubspec#name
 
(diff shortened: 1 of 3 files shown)


src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -176,6 +176,10 @@ public boolean validateModuleName(@NotNull String moduleName) throws Configurati
       throw new ConfigurationException("Invalid module name: '" + moduleName + "' - must be a valid Dart identifier.");
     }
 
+    if (!FlutterUtils.isValidPackageName(moduleName)) {
```
skybrian: Maybe do this check first, since it's usually stricter than the previous checks and the error message is clearer. (If the user fixes the mistake based on this message, it will usually pass the others.)

pq: 👍  Will do.  Thanks.

Issue #1028, opened 2017-05-16
Unable to "Open Flutter View"


lorrainekan:
## What happened

After running my application using the Flutter tool bar, I clicked on "Open Flutter View".

## Version information

IntelliJ IDEA `2017.1.1` • Flutter plugin `13.1` • Dart plugin `171.4424`

No Flutter sdk configured.

## Exception

null

```
java.lang.NullPointerException
	at io.flutter.view.OpenFlutterViewAction.actionPerformed(OpenFlutterViewAction.java:43)
	at com.intellij.openapi.actionSystem.ex.ActionUtil$1.run(ActionUtil.java:215)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAware(ActionUtil.java:232)
	at com.intellij.openapi.actionSystem.impl.ActionButton.a(ActionButton.java:188)
	at com.intellij.openapi.actionSystem.impl.ActionButton.a(ActionButton.java:154)
	at com.intellij.openapi.actionSystem.impl.ActionButton.processMouseEvent(ActionButton.java:325)
	at java.awt.Component.processEvent(Component.java:6306)
	at java.awt.Container.processEvent(Container.java:2236)
	at java.awt.Component.dispatchEventImpl(Component.java:4897)
	at java.awt.Container.dispatchEventImpl(Container.java:2294)
	at java.awt.Component.dispatchEvent(Component.java:4719)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4888)
	at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4525)
	at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4466)
	at java.awt.Container.dispatchEventImpl(Container.java:2280)
	at java.awt.Window.dispatchEventImpl(Window.java:2746)
	at java.awt.Component.dispatchEvent(Component.java:4719)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:764)
	at java.awt.EventQueue.access$500(EventQueue.java:98)
	at java.awt.EventQueue$3.run(EventQueue.java:715)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
[...]

PR #1033, opened 2017-05-17
Null check on Flutter View open (#1028).

pq:
See https://github.com/flutter/flutter-intellij/issues/1028.

@devoncarew

### DIFF
--- a/src/io/flutter/view/OpenFlutterViewAction.java
+++ b/src/io/flutter/view/OpenFlutterViewAction.java
@@ -42,7 +42,7 @@ public class OpenFlutterViewAction extends DumbAwareAction {
 
     final ToolWindow toolWindow = ToolWindowManager.getInstance(project).getToolWindow(FlutterView.TOOL_WINDOW_ID);
     if (toolWindow == null) {
-      FlutterMessages.showError("Error Opening Flutter View", "Unable to find Flutter tool window");
+      FlutterMessages.showError("Unable to open view", "Unable to open the Flutter tool window - no Flutter modules found");
     } else {
       toolWindow.show(null);
     }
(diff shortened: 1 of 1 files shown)


src/io/flutter/view/OpenFlutterViewAction.java
```diff
@@ -39,7 +40,11 @@ public void actionPerformed(@NotNull final AnActionEvent e) {
 
     FlutterInitializer.sendAnalyticsAction(this);
 
-    final ToolWindow toolWindow = ToolWindowManager.getInstance(project).getToolWindow("Flutter");
-    toolWindow.show(null);
+    final ToolWindow toolWindow = ToolWindowManager.getInstance(project).getToolWindow(FlutterView.TOOL_WINDOW_ID);
```
devoncarew: Looks like getToolWindow() can return null if `FlutterViewCondition` returns false, which will happen if we don't recognize any modules as flutter modules.

src/io/flutter/view/OpenFlutterViewAction.java
```diff
@@ -39,7 +40,11 @@ public void actionPerformed(@NotNull final AnActionEvent e) {
...
+    if (toolWindow == null) {
+      FlutterMessages.showError("Error Opening Flutter View", "Unable to find Flutter tool window");
```
devoncarew: Perhaps modify the message in light of the above? `Unable to open View` and `Unable to open the Flutter tool window - no Flutter modules found.`

devoncarew:
Thanks for the null check! lgtm

src/io/flutter/view/OpenFlutterViewAction.java
pq: Sure thing.

Related, it seems that maybe we want the `OpenFlutterViewAction` to detect and disable itself in that case?  I may take a little poke at that.

src/io/flutter/view/OpenFlutterViewAction.java
pq: Got it.  Thanks!

PR #1048, opened 2017-05-22
Mention `alpha` explicitly.

pq:
@mit-mit @devoncarew

### DIFF
--- a/testing.md
+++ b/testing.md
@@ -5,7 +5,7 @@ Manual tests to execute before plugin releases.
 ## Setup
 
 Pre-reqs: Run through the [flutter setup](https://flutter.io/setup/) and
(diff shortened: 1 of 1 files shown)


testing.md
```diff
@@ -7,6 +7,7 @@ Manual tests to execute before plugin releases.
 Pre-reqs: Run through the [flutter setup](https://flutter.io/setup/) and
 [flutter getting started](https://flutter.io/getting-started/) guides.
 
+* Be sure you're on the  `alpha` branch.
```
devoncarew: If we align our release schedules, we'll probably want to test against the up-coming alpha (the master branch).

pq: Ah, K.  That makes sense for a few reasons.  I'll close this out.

Issue #1025, opened 2017-05-16
expose debugPaintBaselinesEnabled in the Flutter tool window


devoncarew:
We should expose the `debugPaintBaselinesEnabled` flutter framework toggle in the Flutter tool window.

Issue #1024, opened 2017-05-16
UX cleanup to the Flutter tool window toolbar


devoncarew:
We have 1/2 dozen buttons in the Flutter tool window toolbar. They're all useful, but it's difficult to know which action does what without hovering over the button. We should drop down to just 2-3 important actions, make sure they have easy to recognize icons, add an overflow menu on the toolbar, and add the other actions there (with their full text names).

PR #1067, opened 2017-05-24
re-organize the flutter view toolbar

devoncarew:
- re-organize the flutter view toolbar
- fix https://github.com/flutter/flutter-intellij/issues/1024, fix https://github.com/flutter/flutter-intellij/issues/1025

screen shot 2017-05-23 at 6 47 50 pm

screen shot 2017-05-23 at 6 47 58 pm

@pq, @skybrian, /cc @jacob314

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -229,9 +229,7 @@ class DebugDrawAction extends AbstractToggleableAction {
   }
 
   protected void perform(AnActionEvent event) {
-    final Map<String, Object> params = new HashMap<>();
-    params.put("enabled", isSelected(event));
-    view.getFlutterApp().callServiceExtension("ext.flutter.debugPaint", params);
+    view.getFlutterApp().callBooleanExtension("ext.flutter.debugPaint", isSelected(event));
   }
 }
 
@@ -241,9 +239,7 @@ class PerformanceOverlayAction extends AbstractToggleableAction {
   }
 
   protected void perform(AnActionEvent event) {
-    final Map<String, Object> params = new HashMap<>();
-    params.put("enabled", isSelected(event));
-    view.getFlutterApp().callServiceExtension("ext.flutter.showPerformanceOverlay", params);
+    view.getFlutterApp().callBooleanExtension("ext.flutter.showPerformanceOverlay", isSelected(event));
   }
 }
 
@@ -283,9 +279,7 @@ class RepaintRainbowAction extends AbstractToggleableAction {
   }
 
   protected void perform(AnActionEvent event) {
-    final Map<String, Object> params = new HashMap<>();
-    params.put("enabled", isSelected(event));
-    view.getFlutterApp().callServiceExtension("ext.flutter.repaintRainbow", params);
+    view.getFlutterApp().callBooleanExtension("ext.flutter.repaintRainbow", isSelected(event));
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -261,6 +256,9 @@ protected void perform(AnActionEvent event) {
   @Override
   public void actionPerformed(AnActionEvent event) {
     final FlutterApp app = view.getFlutterApp();
+    if (app == null) {
+      return;
+    }
```
pq: Oh!  Were we hitting NPEs here?

src/io/flutter/view/FlutterView.java
```diff
@@ -279,9 +277,65 @@ public void actionPerformed(AnActionEvent event) {
...
+
+class TimeDilationAction extends AbstractToggleableAction {
+  TimeDilationAction(@NotNull FlutterView view) {
+    super(view, "Enable Slow Animations");
+  }
+
+  protected void perform(AnActionEvent event) {
+    final Map<String, Object> params = new HashMap<>();
+    params.put("timeDilation", isSelected(event) ? 5.0 : 1.0);
+    view.getFlutterApp().callServiceExtension("ext.flutter.timeDilation", params);
```
pq: I feel slightly inclined to extract a method since there's a bit of duplicated logic in the perform methods but not a big deal.

src/io/flutter/view/FlutterView.java
```diff
@@ -279,9 +277,65 @@ public void actionPerformed(AnActionEvent event) {
...
+
+class HideSlowBannerAction extends AbstractToggleableAction {
+  HideSlowBannerAction(@NotNull FlutterView view) {
+    super(view, "Hide Slow Mode Banner");
+  }
+
+  @Override
+  protected void perform(AnActionEvent event) {
+    final Map<String, Object> params = new HashMap<>();
+    params.put("enabled", !isSelected(event));
```
pq: I assume this is right but it jumps out at me (not totally understanding the code) since the other methods aren't negating `isSelected(..)`...

src/io/flutter/view/FlutterView.java
```diff
@@ -296,9 +350,7 @@ public final void update(AnActionEvent e) {
 
 class ObservatoryTimelineAction extends AbstractFlutterAction {
   ObservatoryTimelineAction(@NotNull FlutterView view) {
-    super(view, FlutterBundle.message("flutter.view.observatoryTimeline.text"),
-          FlutterBundle.message("flutter.view.observatoryTimeline.description"),
-          AllIcons.Debugger.Value);
+    super(view, "Open Observatory Timeline");
```
pq: I assume we no longer want to pass the `AllIcons.Debugger.Value` icon up to super?

src/io/flutter/view/FlutterView.java
devoncarew: We would with the new change - the actions in the toolbar don't respect the disabled state.

src/io/flutter/view/FlutterView.java
devoncarew: This one is reversed - all the others are off by default - the slow mode banner is on by default (but good eyes!)

src/io/flutter/view/FlutterView.java
devoncarew: Yup - the actions in the overflow menu don't have icons associated with them.

src/io/flutter/view/FlutterView.java
devoncarew: Sounds good - will make that change.

Issue #66, opened 2016-08-29
Run flutter unit tests (flutter test)


stevemessick:
Need a way to run unit tests for Flutter apps. Assuming flutter continues to use package:test we probably want similar support to what Dart already has. Dart tests are presented in the IntelliJ test UI with linkage back to source and ability to filter pass/fail/skip.

eseidelGoogle:
It's not clear how IntelliJ will know to run tests in `sky_shell` (flutter's test harness) or the `dart` command line (pure package:test).  The differences are that `sky_shell` exposes `dart:ui` but `dart` exposes `dart:mirrors` (as well as dart native library extension loading).  All non-mirrors dart code works in just the same in `sky_shell` so one could always use `sky_shell`.  IntelliJ could also tell which test needed which based on it being a flutter project vs. a dart project, or if we put one typ of test in a different directory.  It's also possible from analyzing the code to know if it depends on dart:ui or not.

devoncarew:
Yup. Another resolution would be to have a different test launch configuration for flutter tests (`Flutter Test` instead of `Dart Test`). It would be good in general to avoid a parallel everything for flutter however, and just try and adapt and use existing Dart functionality where possible.

branflake2267:
It would be nice to see launching the the flutter tests, `flutter test` execution if a right click and debug unit tests on, unite_test.dart. 
![screen shot 2017-05-21 at 12 50 49 pm](https://cloud.githubusercontent.com/assets/1326504/26286925/ef53d212-3e24-11e7-9e03-99eab15ac57c.png)

And if I could right click on the unit test, it would be nice to run the single test. 
![screen shot 2017-05-21 at 12 58 42 pm](https://cloud.githubusercontent.com/assets/1326504/26286943/4d0b59e8-3e25-11e7-9cd6-7bf3ff36f6a1.png)

devoncarew:
So we don't lose track: PR out to allow for debugging `flutter test` tests: https://github.com/flutter/flutter/pull/10520

skybrian:
Debugging tests was implemented in #1106.

Still to do: implement the test UI (instead of showing progress on the console) and add the ability to run a single test (rather than all tests in a file).

devoncarew:
Initial version done! We'll track any additional work in more specific issues.

xster:
I may have missed it but is there some documentation for how to use it? 

I'm on v15.0. 
[...]

PR #1069, opened 2017-05-25
Add new run configuration type for 'flutter test'

skybrian:
For now this just executes 'flutter test' and sends the output
to the console. Still to do: show test output in IDEA's UI
and support debugging. (That will require modifying the
"flutter test" command to support --machine mode.)

Previously, running a test from within IDEA would fail because
the Dart plugin does it and it doesn't use "flutter test",
which is needed for all tests using a flutter pubspec.

### DIFF
--- a/src/io/flutter/sdk/FlutterSdk.java
+++ b/src/io/flutter/sdk/FlutterSdk.java
@@ -120,10 +120,6 @@ public class FlutterSdk {
 
   public FlutterCommand flutterRun(@NotNull PubRoot root, @NotNull VirtualFile main,
                                    @Nullable FlutterDevice device, @NotNull RunMode mode, String... additionalArgs) {
-    if (!root.contains(main)) {
-      throw new IllegalArgumentException("main isn't within the pub root: " + main.getPath());
-    }
-
     final List<String> args = new ArrayList<>();
     args.add("--machine");
     if (FlutterInitializer.isVerboseLogging()) {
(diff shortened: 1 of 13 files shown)


pq:
Exciting!  I'll take a look later this AM when I'm more awake but in the meantime it occurs to me that @stevemessick may have thoughts since he implemented the Dart `test` support.  If not now, he may be worth tapping later.  👍

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -141,14 +137,36 @@ public FlutterCommand flutterRun(@NotNull PubRoot root, @NotNull VirtualFile mai
...
 
+  public FlutterCommand flutterTest(@NotNull PubRoot root, @NotNull VirtualFile main, @Nullable FlutterDevice device) {
+
+    // We don't have machine mode yet, so just run it normally and show the output in the console.
+    final List<String> args = new ArrayList<>();
+    if (FlutterInitializer.isVerboseLogging()) {
+      args.add("--verbose");
+    }
+    if (device != null) {
+      args.add("--device-id=" + device.deviceId());
```
devoncarew: does `flutter test` take a device ID? doesn't is just run test against sky_shell?

devoncarew:
Very cool! (some work towards https://github.com/flutter/flutter-intellij/issues/66)

src/io/flutter/sdk/FlutterSdk.java
skybrian: I don't know what it does with this flag, but it didn't reject it.

PR #1075, opened 2017-05-26
Don't require a test's filename to end with _test.dart

skybrian:
Instead, require it to be in the test directory of a pubroot.
(This is closer to what the Dart plugin does.)


devoncarew:
lgtm, though have you thought about false positives? I would suspect that most (all?) test files end in `_test.dart`, and conversely, `test/` dirs often contains support libraries that are themselves not tests.

skybrian:
It won't run a file that doesn't have a main() method, so I think support libraries are okay.

The problem is that if the Flutter plugin doesn't run the test, the Dart plugin will do it, and then it fails in a confusing way that looks like a bug. I figured that running the test was least confusing.

Not ending the name in `_test.dart` has a different side-effect though; if you run all tests in a directory (currently only possible from the command line) it will be skipped. So we might want to add a banner suggesting to the user that they rename it.

PR #1076, opened 2017-05-26
Fix compilation against IntelliJ IDEA 2017.2

alexander-doroshko:
this commit doesn't break compilation against 2017.1
@devoncarew @pq

### DIFF
--- a/src/io/flutter/module/FlutterModuleBuilder.java
+++ b/src/io/flutter/module/FlutterModuleBuilder.java
@@ -54,7 +54,6 @@ public class FlutterModuleBuilder extends ModuleBuilder {
     return FlutterBundle.message("flutter.module.name");
   }
 
-  @Override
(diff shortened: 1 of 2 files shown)


src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -54,7 +54,6 @@ public String getPresentableName() {
     return FlutterBundle.message("flutter.module.name");
   }
 
-  @Override
```
devoncarew: This does not exist in 2017.2?

devoncarew:
lgtm

src/io/flutter/module/FlutterModuleBuilder.java
alexander-doroshko: No, it doesn't

Issue #1068, opened 2017-05-24
Produce more output in the IJ terminal during builds


mit-mit:
We have had a number of issues recently where flutter builds on both android and ios slowed down due to several "environment" issues (e.g., gradle setup, network throttling, Pod install). When running in the terminal it's somewhat clear where the slowdown is, and that nothing is "hung" as you get a progress spinner. In IntelliJ it's harder as 1) there is little terminal feedback, and 2) you may miss the progress indicator in the IJ window chrome.

I'd like to propose we simply print the same status to the IJ terminal that `flutter run` / `flutter build` prints to the terminal (incl. the `/` spinner for indeterminate progress).

devoncarew:
Printing the progress text sounds like a good thing to experiment with.

Whenever the cli shows a spinner, IntelliJ shows an indeterminate progress bar (in the status line or in the editor, depending on a user preference). Sounds like the progress bar is harder to notice. I'm of two minds about adding a text spinner, since that seems appropriate for a CLI, but low-fidelity for an IDE (w/ existing progress conventions).

devoncarew:
We're doing most of this now - we're printing the progress messages (in addition to showing the IntelliJ progress UI), and now have a setting to effectively enable `flutter tun -v` (which should help diagnose issues where we're not making progress when running apps).

PR #1077, opened 2017-05-26
improved feedback on reload

### LINKED ISSUE #939: not enough feedback when a hot reload has been performed
From UX studies, we've seen that there's not enough feedback when a hot reload has been performed or is in progress.

devoncarew:
improved feedback on reload:

- clear the console when a reload starts (fix #939)
- disable the reload button while a reload is occurring
- print progress messages to the console view (in addition to using IntelliJ's normal progress mechanism) (part of #1068)
- (and, some small tweaks to the FlutterView)

@skybrian @pq

### DIFF
--- a/src/io/flutter/run/daemon/FlutterAppListener.java
+++ b/src/io/flutter/run/daemon/FlutterAppListener.java
@@ -109,7 +109,9 @@ class FlutterAppListener implements DaemonEvent.Listener {
     if (event.getType().startsWith("hot.")) {
       // We clear the console view in order to help indicate that a reload is happening.
       if (app.getConsole() != null) {
-        app.getConsole().clear();
+        if (!FlutterInitializer.isVerboseLogging()) {
+          app.getConsole().clear();
+        }
       }
 
       stopwatch.set(Stopwatch.createStarted());
--- a/src/io/flutter/run/daemon/FlutterAppListener.java
+++ b/src/io/flutter/run/daemon/FlutterAppListener.java
@@ -105,8 +105,18 @@ class FlutterAppListener implements DaemonEvent.Listener {
   @Override
   public void onAppProgressStarting(@NotNull DaemonEvent.AppProgress event) {
     progress.start(event.message);
+
     if (event.getType().startsWith("hot.")) {
+      // We clear the console view in order to help indicate that a reload is happening.
+      if (app.getConsole() != null) {
+        app.getConsole().clear();
+      }
+
       stopwatch.set(Stopwatch.createStarted());
+    }
+
+    if (app.getConsole() != null) {
+      app.getConsole().print(event.message + "\n", ConsoleViewContentType.NORMAL_OUTPUT);
     }
   }
 
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -9,7 +9,9 @@ import com.intellij.execution.runners.ExecutionUtil;
 import com.intellij.execution.ui.ConsoleViewContentType;
 import com.intellij.icons.AllIcons;
 import com.intellij.openapi.Disposable;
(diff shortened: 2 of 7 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -317,7 +316,7 @@ protected void perform(AnActionEvent event) {
   }
 }
 
-abstract class AbstractFlutterAction extends DumbAwareAction {
+abstract class AbstractFlutterAction extends AnAction {
```
pq: Should we does this more universally?  I admit I don't really get what `DumbAwareAction`s are about and the docs are errr, lacking. 🙁

src/io/flutter/view/FlutterViewFactory.java
```diff
@@ -32,9 +31,6 @@ private static void initFlutterView(@NotNull Project project, FlutterViewMessage
 
   @Override
   public void createToolWindowContent(@NotNull Project project, @NotNull ToolWindow toolWindow) {
-    //noinspection CodeBlock2Expr
-    DumbService.getInstance(project).runWhenSmart(() -> {
```
pq: Wonder if there are other places we can make similar simplifications.  I don't think being "smart" gets us much.

src/io/flutter/view/FlutterView.java
pq: Ah, OK.  Did some digging.  OK, I think I agree that we generally don't want to be dumb aware.  I see a bunch of references elsewhere so maybe we should consider a bit of cleanup down the road... 👍

src/io/flutter/view/FlutterViewFactory.java
skybrian: One effect of "run when smart" is that at project open, it delays the callback until IntelliJ finished updating the project's index. So that can be useful.

src/io/flutter/view/FlutterView.java
devoncarew: Yeah, I think dumbaware is about things that need to have info from the index available before they run, which I don't think applies to the FlutterView.

devoncarew: sounds good

src/io/flutter/view/FlutterViewFactory.java
devoncarew: Yup, but not sure that the FlutterView needs to be IntelliJ index aware.

src/io/flutter/run/daemon/FlutterAppListener.java
```diff
@@ -105,9 +105,19 @@ public void onAppLog(@NotNull DaemonEvent.AppLog message) {
   @Override
   public void onAppProgressStarting(@NotNull DaemonEvent.AppProgress event) {
     progress.start(event.message);
+
     if (event.getType().startsWith("hot.")) {
+      // We clear the console view in order to help indicate that a reload is happening.
```
devoncarew: Yeah, I'm really of two minds about this. It does to its job of making reload clearer - when it's happening, and when it's finished. But it also has the drawback of removing (potentially valuable) console history.

Preserving history w/ verbose sounds useful. A larger 'reloading' banner is a good idea but may not move the needle greatly.

I _think_ I'm leaning towards clearing the log on reload when verbose is not on. We can try that for a while and see if it's a net benefit.

skybrian: Okay, yeah, this seems fine; if we're debugging we probably want the verbose log anyway.

alexander-doroshko:
Normally we try to make actions dumb aware when possible. If an action doesn't use an index, why make it disabled during indexing.

alexander-doroshko:
(my comment is not related to this PR, I didn't review it, just noticed dumb awareness discussion :))

Issue #880, opened 2017-04-04
we should have an easy way to open XCode from a flutter project


devoncarew:
- in the Tools > Flutter menu
- as a context menu item off the <project>/ios directory

pq:
First cut in #1079.  We can swizzle UI bits around as needed.

Navigator:

![image](https://cloud.githubusercontent.com/assets/67586/26505237/d0ded38c-41fc-11e7-9f8b-f31071a25ca9.png)

Top-level menu:

![image](https://cloud.githubusercontent.com/assets/67586/26505266/ee2a692e-41fc-11e7-8dae-64bf3cb7d3b6.png)

pq:
From @mit-mit , it looks like we're opening the wrong project... Look at:  

 https://flutter.io/platform-plugins/#ios-platform-code-hmswift

for details.

pq:
@mit-mit : I may have misunderstood when we chatted earlier.  Could you verify?  Currently, given a project called `hello_world`, this action will open `hello_world/ios/Runner.xcodeproj`.  Is that correct?

mit-mit:
Sorry, I answered out of context. The issue is that it depends on whether you are editing a Flutter app, or a Flutter plugin. For an app, you have it right. For a plugin, you need to open the native projects for the `/example/` app located *inside* the plugin project.

pq:
Ah OK, thanks.  Let's follow-up on plugin support in another set of issues.

mit-mit:
@pq did you open that second bug?

pq:
No.  Feel free to open plugin support enhancements.  I'm guessing there will be a host of them...

PR #1079, opened 2017-05-26
Open in Xcode… actions

pq:
Fixes: #880.

@devoncarew


resources/META-INF/plugin.xml
```diff
@@ -252,6 +252,10 @@
               text="Flutter Packages Upgrade"
               description="Run 'flutter packages upgrade'"/>
       <separator/>
+      <action id="flutter.xcode.open" class="io.flutter.actions.OpenInXcodeAction"
+              text="Open in Xcode..."
```
devoncarew: Assuming the menu in IntelliJ generally use the ellipses char, we should use that here (available a few lines down).

devoncarew: And - to bikeshed a bit - `Open Project in Xcode…`?

devoncarew:
Looks good!

resources/META-INF/plugin.xml
pq: 👍

PR #1080, opened 2017-05-26
re-add DumbAwareAction

devoncarew:
Revert parts of https://github.com/flutter/flutter-intellij/pull/1077 - re-add `DumbAwareAction`.

From the docs: `A marker interface for the things that are allowed to run in dumb mode` - so it sounds since these actions do not need to access bits of the index, we want to keep them dumb aware.

### DIFF
--- a/src/io/flutter/view/AbstractToggleableAction.java
+++ b/src/io/flutter/view/AbstractToggleableAction.java
@@ -5,18 +5,18 @@
  */
 package io.flutter.view;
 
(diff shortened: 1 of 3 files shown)


devoncarew:
And, make `FlutterViewFactory` DumbAware. This prevents the flashing of the Flutter tool window on startup, which is what I was trying to do in the previous PR.

@skybrian @alexander-doroshko

Issue #1064, opened 2017-05-23
display error messages from flutter create when project creation fails


mit-mit:
Repro steps:

This was found on Windows; I suspect this will rarely happen elsewhere

1. make sure git is not in path
1. file > new project

=>

![image](https://cloud.githubusercontent.com/assets/13644170/26369952/8d8a1478-3ff6-11e7-8450-ab1f0ec913b6.png)

From the command line:
```
C:\Users\mit\tmp>flutter create test2
Error: Unable to find git in your PATH.
```

Shouldn't we be printing that error message also in IJ?

devoncarew:
> Shouldn't we be printing that error message also in IJ?

Yup! We should do.

pq:
We can clean things up but #1081, ensures we don't miss anything useful by reproducing _all_ of stderr.  For example, here's an example where flutter_tools falls over entirely.

![image](https://cloud.githubusercontent.com/assets/67586/26512990/f13782de-421d-11e7-88df-1fdedbaf6d64.png)

A little crude but hopefully only encountered infrequently and at least we have lots of context around the failure.

PR #1081, opened 2017-05-26
Capture and report `flutter create` error details (#1064).

pq:
Fixes: #1064.

@devoncarew @skybrian 

This is a little inelegant but expedient (and I didn't see a better way that wouldn't involve a bunch of swizzling of bits around).

Happy to revisit.

### DIFF
--- a/src/io/flutter/sdk/FlutterCommand.java
+++ b/src/io/flutter/sdk/FlutterCommand.java
@@ -73,11 +73,15 @@ public class FlutterCommand {
    * <p>
    * If unable to start (for example, if a command is already running), returns null.
    */
-  public Process start(@Nullable Consumer<ProcessOutput> onDone) {
+  public Process start(@Nullable Consumer<ProcessOutput> onDone, @Nullable ProcessListener processListener) {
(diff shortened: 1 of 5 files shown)


src/io/flutter/sdk/FlutterCommand.java
```diff
@@ -73,13 +73,17 @@ public String getDisplayCommand() {
    * <p>
    * If unable to start (for example, if a command is already running), returns null.
    */
-  public Process start(@Nullable Consumer<ProcessOutput> onDone) {
+  public Process start(@Nullable Consumer<ProcessOutput> onDone, @Nullable ProcessListener processListener) {
```
skybrian: Seems fine for now. If this gets any more complicated, it might be a bit more elegant to use a builder pattern to give optional parameters to a FlutterCommand. Something like:

sdk.flutterCreate(baseDir).withProcessListener(listener).whenDone(callback).start()

pq: Neat idea! 👍

PR #1082, opened 2017-05-27
Add support for running all tests in a directory

skybrian:
The directory can be the pub root, the "test" directory, or any
subdirectory of "test".

If it's the pub root, we just run "flutter test".

Also, simplify PubRoot.fromPsiFile() slightly (remove fallback).

### DIFF
--- a/src/io/flutter/pub/PubRoot.java
+++ b/src/io/flutter/pub/PubRoot.java
@@ -119,28 +119,27 @@ public class PubRoot {
   }
 
   /**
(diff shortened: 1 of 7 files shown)


devoncarew:
lgtm!

Can you add a screenshot of the launch config UI, and/or confirm that the spacing and padding makes sense? The `Dart Test` launch config UI would be a good point of comparison.

skybrian:
I started making a few UI tweaks, but I think it makes sense to land them as a separate CL.

PR #1086, opened 2017-05-30
Improve “Open in Xcode” filtering (#880)

pq:
Allows for opening from:
  * project dir
  * `ios` dir

Follow-up on #880
 
@devoncarew


devoncarew:
lgtm

PR #1087, opened 2017-05-31
Test configs: make switching between scopes smoother

skybrian:
- set a default value so the field doesn't become blank
- increase margin so the fields don't move

### DIFF
--- a/src/io/flutter/run/test/TestForm.java
+++ b/src/io/flutter/run/test/TestForm.java
@@ -17,6 +17,7 @@ import org.jetbrains.annotations.NotNull;
 import javax.swing.*;
 
 import java.awt.event.ActionEvent;
+import java.io.File;
 
 import static com.jetbrains.lang.dart.ide.runner.server.ui.DartCommandLineConfigurationEditorForm.initDartFileTextWithBrowse;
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/test/TestForm.java
```diff
@@ -97,11 +105,41 @@ private Scope getScope() {
...
+      }
+    }
+  }
+
+  /**
+   * Show and hide fields as appropriate for the next scope.
+   */
+  private void render(Scope next) {
+
+    testDirLabel.setVisible(next == Scope.DIRECTORY);
```
devoncarew: Probably 6 of one and 1/2 dozen of another, but you could also one path field component, and toggle it's behavior when you hit the browse button (instead of showing and hiding sets of fields)..

skybrian: There are some slight user-visible differences: once a file is set, the form will remember it even if you switch to directory scope and back again. Also, once a directory is set, if you go to "file", clear it, and go back again, it will remember the previous directory.

I'm not sure whether it's better behavior, though.

devoncarew: OK, I don't have a real opinion here, but wanted to mention it.

Issue #1092, opened 2017-06-06
Packages get doesn't handle nested packages


matanlurey:
## Steps to Reproduce

1. I cloned [flutter/plugins](https://github.com/flutter/plugins).
2. I opened `packages/firebase_database` in IntelliJ.
3. I got analysis errors after running `packages get`:

<img width="770" alt="screen shot 2017-06-05 at 8 22 20 pm" src="https://cloud.githubusercontent.com/assets/168174/26812465/b8034310-4a2c-11e7-88b6-7a9b031f3a48.png">

I attempted to fix this by double clicking on `example/pubspec.yaml` and hitting `packages get`, but based on the terminal prompt (`Running 'flutter packages get' in firebase_database...`), it looks like it just ran on the root, not the nested package. Right clicking on `example/pubspec.yaml` and clicking `Flutter > Packages Get` did the same thing.

I ended up having to navigate with the command line to resolve this analysis warning.

## Version info

```
[✓] Flutter (on Mac OS X 10.12.4 16E195, channel alpha)
    • Flutter at /Users/matanl/git/flutter
    • Framework revision 00214fa7bd (3 weeks ago), 2017-05-16 09:31:35 -0700
    • Engine revision f99d3ef108
    • Tools Dart version 1.23.0-dev.11.11

[✓] Android toolchain - develop for Android devices (Android SDK 25.0.3)
    • Android SDK at /Users/matanl/Library/Android/sdk
    • Platform android-25, build-tools 25.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version: OpenJDK Runtime Environment (build 1.8.0_112-release-b06)

[✓] iOS toolchain - develop for iOS devices (Xcode 8.2.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 8.2.1, Build version 8C1002
    • ios-deploy 1.9.1
    • CocoaPods version 1.2.1

[✓] Android Studio (version 2.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Gradle version 3.2
    • Java version: OpenJDK Runtime Environment (build 1.8.0_112-release-b06)

[✓] IntelliJ IDEA Ultimate Edition (version 2017.1.3)
    • Dart plugin version 171.4424.63
    • Flutter plugin version 14.0

[✓] Connected devices
    • None
```

pq:
cc @alexander-doroshko

[...]

PR #1094, opened 2017-06-06
support nested packages

devoncarew:
- support nested packages for `flutter packages get / upgrade`
- fix https://github.com/flutter/flutter-intellij/issues/1092

This specifically addresses the use case of the pubspec.yaml file editor banner, and a context click on a pubspec.yaml file in the project navigator.

@pq

### DIFF
--- a/src/io/flutter/editor/FlutterPubspecNotificationProvider.java
+++ b/src/io/flutter/editor/FlutterPubspecNotificationProvider.java
@@ -77,33 +77,11 @@ public class FlutterPubspecNotificationProvider extends EditorNotifications.Prov
       text("Flutter commands");
 
       // "flutter.packages.get"
-      HyperlinkLabel label = createActionLabel("Packages get", () -> {
-        // We can assert below since we know we're on a pubspec.yaml file in a project, with a valid Flutter SDK.
-        final Project project = ProjectLocator.getInstance().guessProjectForFile(myFile);
-        assert (project != null);
-        final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
-        assert (sdk != null);
-
-        final PubRoot root = PubRoot.forDirectory(myFile.getParent());
-        if (root != null) {
-          sdk.startPackagesGet(root, project);
-        }
-      });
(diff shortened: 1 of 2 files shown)


src/io/flutter/editor/FlutterPubspecNotificationProvider.java
```diff
@@ -76,10 +76,34 @@ public EditorNotificationPanel createNotificationPanel(@NotNull VirtualFile file
...
+        final Project project = ProjectLocator.getInstance().guessProjectForFile(myFile);
+        assert (project != null);
+        final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
+        assert (sdk != null);
+
+        final PubRoot root = PubRoot.forDirectory(myFile.getParent());
+        if (root != null) {
+          sdk.startPackagesUpgrade(root, project);
+        }
+      });
```
pq: Maybe extract this into a method (`doUpgrade(...)` or something) so you can share it with above?

src/io/flutter/pub/PubRoot.java
```diff
@@ -126,8 +126,14 @@ public static PubRoot forEventWithRefresh(@NotNull final AnActionEvent event) {
    * Based on the filesystem cache; doesn't refresh anything.
    */
   @Nullable
-  public static PubRoot forPsiFile(@NotNull PsiFile file) {
-    return forDescendant(file.getVirtualFile(), file.getProject());
+  public static PubRoot forPsiFile(@NotNull PsiFile psiFile) {
+    final VirtualFile file = psiFile.getVirtualFile();
+    if (isPubspec(file)) {
```
pq: Looking at `PsiFileImpl`, I see `getVirtualFile()` may technically return `null`. Though I'm not sure how likely it would be in practice, maybe consider guarding all the same?

devoncarew: Makes sense! added

src/io/flutter/editor/FlutterPubspecNotificationProvider.java
devoncarew: done

PR #1115, opened 2017-06-20
Display a proper UI for test progress

skybrian:
Depends on: https://github.com/flutter/flutter/pull/10848

### DIFF
--- a/src/io/flutter/run/daemon/DaemonConsoleView.java
+++ b/src/io/flutter/run/daemon/DaemonConsoleView.java
@@ -81,12 +81,6 @@ public class DaemonConsoleView extends ConsoleViewImpl {
       return;
     }
 
-    // "flutter test" prints this when using --start-paused.
-    // (We should probably change it to stop printing it in machine mode.)
-    if (trimmed.startsWith("You should first set appropriate breakpoints, then resume the test in the debugger.")) {
-      return;
-    }
-
     super.print(text, contentType);
   }
 }
--- a/src/io/flutter/run/test/DebugTestRunner.java
+++ b/src/io/flutter/run/test/DebugTestRunner.java
@@ -25,6 +25,7 @@ import com.jetbrains.lang.dart.ide.runner.ObservatoryConnector;
 import com.jetbrains.lang.dart.util.DartUrlResolver;
 import io.flutter.FlutterInitializer;
 import io.flutter.run.PositionMapper;
+import io.flutter.sdk.FlutterSdk;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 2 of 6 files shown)


src/io/flutter/run/daemon/DaemonConsoleView.java
```diff
@@ -81,12 +81,6 @@ public void print(@NotNull String text, @NotNull ConsoleViewContentType contentT
       return;
     }
 
-    // "flutter test" prints this when using --start-paused.
-    // (We should probably change it to stop printing it in machine mode.)
-    if (trimmed.startsWith("You should first set appropriate breakpoints, then resume the test in the debugger.")) {
-      return;
-    }
-
```
pq: Has this been fixed in `flutter test`?

pq: Ah. Ok, I see the change in https://github.com/flutter/flutter/pull/10848.

What happens if that change hasn't landed?  IOW: what's the behavior with the current Flutter alpha?

skybrian: The test runs and output appears in the console, but it the Test UI claims that there was a tool error since it didn't see any progress events.

But I'm not sure if the `--machine` flag made it to the alpha branch either, in which case testing wouldn't work at all. Good point; we should make sure these two changes land in alpha before releasing the plugin.

pq: Since we're shooting for a release next week (and can't ensure they're on alpha besides), is there an easy way to make this fail gracefully on unsupported SDKs?

skybrian: I suppose if the `--machine` flag fails we could try again without it, using a plain-text console. In that case, the debug button should be disabled altogether. Or maybe we could use the Flutter SDK version number, since we already have code for that.

More generally, it seems like we should have a generic way for the IDE to ask the Flutter SDK for a set of properties. This would let us turn on new features based on the presence of flags, and there are other things we could use this for, such as getting the Java classpath. (See https://github.com/flutter/flutter/pull/10802.)

pq: cc @devoncarew whose thought a bit about querying properties/capabilities IIRC.

In the short term maybe something coarse like SDK number will suffice.

skybrian: Okay, it now checks the version number and only provides minimal support for testing on the alpha branch. Verified by switching back and forth. (The debug icon is enabled and disabled appropriately.)

The code assumes 0.0.11 supports `--machine` mode and 0.0.10 does not. That's currently true at HEAD on both branches.

pq: Awesome.  Thanks! 👍

PR #1120, opened 2017-06-22
Add missing guaranteed assertions about nullability

jwren:
Also, reorder assertions to be before declarations and signatures to be consistent with other JetBrains plugin source.

@pq @devoncarew

### DIFF
--- a/src/io/flutter/bazel/Workspace.java
+++ b/src/io/flutter/bazel/Workspace.java
@@ -45,7 +45,7 @@ public class Workspace {
   /**
    * Returns true for a project that uses Flutter code within this workspace.
    */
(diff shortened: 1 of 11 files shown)


src/io/flutter/bazel/Workspace.java
```diff
@@ -45,7 +45,7 @@ private Workspace(@NotNull VirtualFile root, @Nullable PluginConfig config, @Nul
   /**
    * Returns true for a project that uses Flutter code within this workspace.
    */
-  public boolean usesFlutter(Project project) {
+  public boolean usesFlutter(@NotNull final Project project) {
```
pq: 👍

Issue #764, opened 2017-02-27
the new flutter project wizard should offer to install the flutter sdk


devoncarew:
The new flutter project wizard should offer to install the flutter sdk in-line if there's not one configured.

pq:
#1105: 

![image](https://user-images.githubusercontent.com/67586/27148434-238a95ce-50f5-11e7-9dff-5020011c9fae.png)

pq:
First cut done.  We can track refinements in separate issues.

PR #1124, opened 2017-06-23
SDK install flow improvements (#764).

pq:
Still some noodling to do but this gets us a bit further...

* adds cancel UI
* hooks up errors to wizard status line
* filters unfriendly progress details
* automatically advances to project configuration page on successful installation
* adds install action link icons

image

image

@devoncarew 

See: #764


pq:
Thanks!  As for follow-on work, I think testing is the main thing.  It may also be nice to hook up a way to cancel installation if the wizard is cancelled but that's not obviously supported at the framework level.  (We may be able to hack it in at a lower-level but I haven't dug too deeply.)

Issue #1108, opened 2017-06-15
set the 'use dartfmt on whole file formatting' setting


devoncarew:
We should (one-time) automatically set the 'use dartfmt for whole file formatting'. This setting existing in the Dart plugin, starting in 2017.2. If the user unsets the setting, we want to respect that, but from the flutter plugins perspective we'll act as if the default for the setting is true.

<img width="376" alt="screen shot 2017-06-15 at 7 10 44 am" src="https://user-images.githubusercontent.com/1269969/27185353-e92ec558-5199-11e7-88d0-8b47d89ba6c6.png">

Issue #816, opened 2017-03-14
make it easier to run dartfmt


devoncarew:
There's no keybinding for dartfmt - the default binding invokes the IntelliJ formatter. We should make it easier to run dartfmt; one idea here is to add an option to format on save.

devoncarew:
Note that one implementation of save actions is here: https://github.com/dubreuia/intellij-plugin-save-actions/blob/master/src/com/dubreuia/SaveActionManager.java.

jwren:
One idea/ possible progress: https://github.com/JetBrains/intellij-plugins/pull/510

dvdwasibi:
Is there an update on this?

We are doing some live coding on stage for Google I/O using intellij and this would be nice.

devoncarew:
It's a surprisingly difficult problem. We're working on an API so that another plugin (the Flutter plugin) can expose a user preference for this. We may also change the Dart plugin to invoke dartfmt when the user requests a full file format (there are several different formatting modes in IntelliJ).

> We are doing some live coding on stage for Google I/O using intellij and this would be nice.

For now, you can edit your settings to create a local keybinding to the dartfmt action. In the 'Keymap' setting, you can search for the dartfmt action, and add a new keybinding to it.

mit-mit:
Can we make that dartfmt option on by default?

devoncarew:
In the Dart plugin, the resolution was not to. We may be able to change that decision w/ sufficient user input.

We can also choose to set that option from the flutter plugin as a one-time thing, so flutter users do see whole-file formats using dartfmt by default, but can adjust the setting if they choose.

InMatrix:
Will our analytics/logs include how many users turned on that option? In addition, do we know the percentage of users who have used dartfmt? I'm trying to understand how we might get "user input".

devoncarew:
> I'm trying to understand how we might get "user input".

I suspect it's going to be more anecdotal rather than data driven. @alexander-doroshko can speak to the data available via IntelliJ's analytics. In the Flutter plugin, we're not currently tracking user preferences (we also have very few to potentially track).

alexander-doroshko:
As for now, no dartfmt metrics gathered. We'll probably have to make the decision about the default value based on the user feedback and expert feelings.

mit-mit:
@devoncarew can we close this bug now that we have https://github.com/flutter/flutter-intellij/issues/1108 ?

devoncarew:
Yup, closing.

PR #1125, opened 2017-06-25
toggle the default for whole file format

devoncarew:
Toggle the default for whole file format (fix https://github.com/flutter/flutter-intellij/issues/1108):

- if the `DartCodeStyleSettings` class exists
- we use reflection to set the default value of the 'use dartfmt for whole file formats'
- and set a setting so we remember we changed the default, and only do it once (the user can then override)

@pq @skybrian

### DIFF
--- a/src/io/flutter/FlutterInitializer.java
+++ b/src/io/flutter/FlutterInitializer.java
@@ -21,6 +21,7 @@ import com.intellij.openapi.startup.StartupActivity;
 import io.flutter.analytics.Analytics;
 import io.flutter.analytics.ToolWindowTracker;
 import io.flutter.android.AndroidSdk;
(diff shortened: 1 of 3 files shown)


devoncarew:
I have been using the mode in WebStorm EAP - it's really very nice. You can now use the std IntelliJ keybinding for formatting - great when working across both Java and Dart code.

For the Dart plugin default, see discussion in https://github.com/flutter/flutter-intellij/issues/816.

pq:
Oh, awesome.  REALLY looking forward to this. 👍

alexander-doroshko:
I'm afraid it's too late to rethink the defaut value for 2017.2, but if we get positive feedback and no major issues I'm fine with toggling the default value in 2017.3.

PR #1127, opened 2017-06-26
SDK install flow fixes and bit flip (#764).

pq:
* implement `hasGit` and flip the bit
* fix local filesystem refresh issues w/ newly cloned SDKs

See: #764.

@devoncarew

### DIFF
--- a/src/io/flutter/actions/InstallSdkAction.java
+++ b/src/io/flutter/actions/InstallSdkAction.java
@@ -20,6 +20,7 @@ import com.intellij.openapi.fileChooser.FileChooserDescriptorFactory;
 import com.intellij.openapi.project.DumbAwareAction;
 import com.intellij.openapi.util.Key;
 import com.intellij.openapi.vfs.VirtualFile;
+import io.flutter.FlutterUtils;
 import io.flutter.module.FlutterGeneratorPeer;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -29,6 +30,9 @@ import java.io.File;
 
 @SuppressWarnings("ComponentNotRegistered")
 public class InstallSdkAction extends DumbAwareAction {
+
+  //TODO(pq): add support for "git.exe"
+  private static final String GIT_EXECUTABLE = "git";
 
   public interface CancelActionListener {
     void actionCanceled();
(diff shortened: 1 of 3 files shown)


src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -269,7 +269,19 @@ public void dispose() {
...
     @Nullable
     public FlutterSdk getFlutterSdk() {
-      return FlutterSdk.forPath(myPeer.getSdkComboPath());
+      final String sdkPath = myPeer.getSdkComboPath();
+
+      //Ensure the local filesystem has caught up to external processes (e.g., git clone).
+      if (!sdkPath.isEmpty()) {
+        try {
+          LocalFileSystem
+            .getInstance().refreshAndFindFileByPath(sdkPath);
```
devoncarew: Should this refresh instead be done at the end of the call to `flutter precache`?

src/io/flutter/actions/InstallSdkAction.java
```diff
@@ -30,6 +31,9 @@
 @SuppressWarnings("ComponentNotRegistered")
 public class InstallSdkAction extends DumbAwareAction {
 
+  //TODO(pq): add support for "git.exe"
+  private static final String GIT_EXECUTABLE = "git";
```
devoncarew: This may do it? `SystemInfo.isWindows ? "git.exe" : "git"`

devoncarew:
I suspect that we'll want to do that file system refresh at the end of the sdk step (after the sdk's written out) instead of on every access to `getFlutterSdk()`.

lgtm

src/io/flutter/module/FlutterModuleBuilder.java
pq: I tried this a few different ways but couldn't get the threading right.  The good news is that this method is only accessed in this wizard step so it's effectively the same as calling it after `flutter precache`.

src/io/flutter/actions/InstallSdkAction.java
pq: 👍 

I'm slightly hesitant to flip this on for Windows w/o doing some testing first.  Happy to defer to you if you say bombs away though.

src/io/flutter/module/FlutterModuleBuilder.java
devoncarew: Ah! I didn't check which class this was in. lgtm (though, perhaps reduce the visibility of the method?)

pq: Good call.  Done!

src/io/flutter/actions/InstallSdkAction.java
pq: FYI: I didn't flip this (in the absence of an easy way to test today) but can in a follow-up if you'd like to see it.

Issue #536, opened 2016-12-13
Enable starting an Android simulator from IntelliJ


mit-mit:
If no Android virtual device is running, you have to start it from the Android AVD Manager. Would be good to have en entry in tools to open that, similar to what Android Studio has (we should probably call it `Android AVD Manager`):

<img width="568" alt="screen shot 2016-12-13 at 12 14 37" src="https://cloud.githubusercontent.com/assets/13644170/21138322/c9a0d8c2-c12d-11e6-97c2-9027f1345726.png">

mit-mit:
Upon closer look, looks like the AVD Manager is an Android Studio dialog, so would we have to recreate that also?

devoncarew:
There's probably a way to interact with the command-line tool. We may not (easily) get device creation, but we should be able to launch them,

mit-mit:
OK. Launching is definitely the most important part.

mit-mit:
Looks like the following might work?

```
> $ANDROID_HOME\tools\emulator -list_avds
Nexus_5X

> $ANDROID_HOME\tools\emulator -avd Nexis_5X
```

devoncarew:
That worked for me locally (`$ANDROID_HOME/tools/emulator -list-avds`). Seems like a good addition, and a nice parallel to the iOS support.

devoncarew:
(just fyi, I ran into a case yesterday where I could have used this)

pq:
👍 Looking at it now!

pq:
I'm thinking we'll depend on `$ANDROID_HOME` being set. In case it's not... should we simply tell them that?  I'm assuming we don't want to support path setting.  For instance like:

<img width="713" alt="screen shot 2017-01-27 at 10 33 03 am" src="https://cloud.githubusercontent.com/assets/67586/22382774/045287ce-e47c-11e6-8228-3085831250f0.png">

I guess another approach would be configuring an external tool?

<img width="265" alt="screen shot 2017-01-27 at 10 32 50 am" src="https://cloud.githubusercontent.com/assets/67586/22382802/24ce113a-e47c-11e6-88ac-9e1ecbfd2a39.png">

(On second thought I don't think this is quite right.)

@devoncarew

devoncarew:
[...]

PR #1129, opened 2017-06-26
add the ability to start android emulators from the device pulldown

devoncarew:
- use the `AndroidSdk` class to find the best android sdk for a paritcular project
- look for the `tools/emulator` binary inside the sdk
- list and display any existing emulator avds in the device pulldown menu
- on selection, launch the indicated avd; display errors as appropriate on launching
- fix https://github.com/flutter/flutter-intellij/issues/536

@skybrian @pq 

screen shot 2017-06-26 at 4 28 16 pm

### DIFF
--- a/src/io/flutter/android/AndroidEmulator.java
+++ b/src/io/flutter/android/AndroidEmulator.java
@@ -5,19 +5,75 @@
  */
 package io.flutter.android;
 
+import com.intellij.execution.ExecutionException;
+import com.intellij.execution.configurations.GeneralCommandLine;
+import com.intellij.execution.process.OSProcessHandler;
+import com.intellij.execution.process.ProcessAdapter;
+import com.intellij.execution.process.ProcessEvent;
+import com.intellij.execution.process.ProcessOutputTypes;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.util.Key;
+import com.intellij.openapi.vfs.VirtualFile;
+import io.flutter.FlutterMessages;
+
 public class AndroidEmulator {
+  private static final Logger LOG = Logger.getInstance(AndroidEmulator.class);
+
   final AndroidSdk androidSdk;
   final String id;
-  final String name;
 
   AndroidEmulator(AndroidSdk androidSdk, String id) {
     this.androidSdk = androidSdk;
     this.id = id;
-    // TODO: How to best clean up the name?
-    this.name = id.replaceAll("_", "-");
   }
 
   public String getName() {
-    return name;
+    return id.replaceAll("_", " ");
+  }
+
+  public void startEmulator() {
+    final VirtualFile emulator = androidSdk.getEmulatorToolExecutable();
+    if (emulator == null) {
+      FlutterMessages.showError(
+        "Error Opening Emulator",
+        "Unable to locate the emulator tool in the Android SDK.");
+      return;
+    }
+
+    final String emulatorPath = emulator.getCanonicalPath();
+    assert (emulatorPath != null);
+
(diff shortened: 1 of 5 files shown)


src/io/flutter/android/AndroidEmulator.java
```diff
@@ -0,0 +1,79 @@
...
+import com.intellij.execution.process.ProcessOutputTypes;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.util.Key;
+import com.intellij.openapi.vfs.VirtualFile;
+import io.flutter.FlutterMessages;
+
+public class AndroidEmulator {
+  private static final Logger LOG = Logger.getInstance(AndroidEmulator.class);
+
+  final AndroidSdk androidSdk;
```
pq: Aside: we've got a mix of field naming conventions (some w/ the `my` prefix); maybe worth a style wiki entry?

skybrian:
Great start!

To avoid a bit of repetition, I wonder if the available emulators should be a submenu:

Open Android Emulator ->
    Nexus 5X API N
 
Ideally we'd grey out or remove the emulators that are already running. This would require figuring out the unique ID for the launched emulator and matching it against the ids returned by the device daemon.

Also, I suspect the device names won't match the ones returned by `flutter daemon`, which could be confusing. These names look better, so it might make more sense to change the names returned by `flutter daemon`.

src/io/flutter/android/AndroidEmulator.java
skybrian: That's probably my fault. I didn't follow JetBrain's my prefix convention, since I try to avoid common prefixes and nobody else does that.  (So many standards to choose from...)

Rationale: if you have both common prefixes and common suffixes, names become longer and harder to distinguish because you have to look at a few characters in the middle of the identifier. Of the two, I'd rather have common suffixes and distinct prefixes.

(But at least we're not mixing conventions in a single file.)

devoncarew:
@skybrian, good points! I'll capture this in a bug so we don't forget.

mit-mit:
Shouldn't we push some of this logic down to flutter tools, so that it can also work at some point for `flutter run`?

devoncarew:
Yes, but as discussed this morning, this at least gets us going. We can incrementally improve, and move much of the logic into flutter_tools.

mit-mit:
Yup, that makes a lot of sense, thanks!

DanTup:
> > Shouldn't we push some of this logic down to flutter tools

> Yes, but as discussed this morning, this at least gets us going

Is this likely to happen anytime soon or should I consider using the Android SDK to add something similar to Dart Code? (requested at https://github.com/Dart-Code/Dart-Code/issues/490)

src/io/flutter/android/AndroidSdk.java
```diff
@@ -138,14 +159,68 @@ private static AndroidSdk fromSdk(@Nullable Sdk candidate) {
     return new AndroidSdk(candidate, home);
   }
 
+  @Nullable
+  public VirtualFile getEmulatorToolExecutable() {
+    // Look for $ANDROID_HOME/tools/emulator.
+    return home.findFileByRelativePath("tools/" + (SystemInfo.isWindows ? "emulator.exe" : "emulator"));
```
TheNightmanCodeth: In newer android SDK (After v25.3.0), the emulator tool was moved to ${ANDROID_HOME}/emulator/emulator(.exe). This should probably be accounted for, since the old emulator tool doesn't start emulators. Throws a `PANIC: Missing emulator engine program for 'x86' CPU.`.

PR #1130, opened 2017-06-26
refactor - move InstallSdkAction.java into flutter/module

devoncarew:
@pq, for your consideration: move InstallSdkAction.java into flutter/module


devoncarew:
> Maybe a good precedent for keeping actions that are not contributed declaratively out of the actions package?

I didn't take a close look at the others, but this one was only used by the flutter.module package...

Issue #1128, opened 2017-06-26
Flutter test support conflicts w/ running VM tests.


pq:
Trying to run Dart tests for a simple console app

![image](https://user-images.githubusercontent.com/67586/27563930-4fde1eaa-5a89-11e7-9b7b-0308170a4e16.png)

yields an error that a Flutter SDK is not set,

![image](https://user-images.githubusercontent.com/67586/27563895-1cddc046-5a89-11e7-877c-d5321941f1c2.png)

(but one shouldn't be required for a Dart console app).

cc @skybrian

devoncarew:
This is something we'll want to fix before shipping M15.

PR #1131, opened 2017-06-26
Scope Flutter test run config contrib (#1128).

pq:
Quick stab at scoping test config over-contribution.

I'm not sure `FlutterModuleUtils.hasFlutterModule` is the long-term fix but maybe OK short-term?

Fixes: #1128.

@skybrian @devoncarew

### DIFF
--- a/src/io/flutter/run/test/TestConfigProducer.java
+++ b/src/io/flutter/run/test/TestConfigProducer.java
@@ -48,7 +48,7 @@ public class TestConfigProducer extends RunConfigurationProducer<TestConfig> {
     final PubRoot root = PubRoot.forPsiFile(file);
     if (root == null) return false;
 
-    if (!FlutterModuleUtils.hasFlutterModule(file.getProject())) return false;
+    if (!FlutterModuleUtils.isFlutterModule(context.getModule())) return false;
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/test/TestConfigProducer.java
```diff
@@ -47,6 +48,8 @@ private boolean setupForDartFile(TestConfig config, ConfigurationContext context
     final PubRoot root = PubRoot.forPsiFile(file);
     if (root == null) return false;
 
+    if (!FlutterModuleUtils.hasFlutterModule(file.getProject())) return false;
```
devoncarew: Do we need to do a similar check in `setupForDirectory`?

Thanks for the fix! I would get Brian's eyes on this. If this style of check looks good, I would use the module available on the `ConfigurationContext`, and call the `FlutterModuleUtils .usesFlutter` w/ that module (a bit more scoped than the project).

pq: Nice catch.  `setupForDirectory` only gives us the directory (no config context and no module) so I think we're stuck with project scope there but here you're totally right we should be more picky. 👍

devoncarew:
lgtm!

PR #1134, opened 2017-06-27
SDK install workflow analytics.

pq:
@devoncarew

### DIFF
--- a/src/io/flutter/module/InstallSdkAction.java
+++ b/src/io/flutter/module/InstallSdkAction.java
@@ -100,13 +100,15 @@ class InstallSdkAction extends DumbAwareAction {
   }
 
   private static class ViewDocsAction extends InstallAction {
+    private static final String ANALYTICS_KEY = "InstallSdkViewDocsAction";
+
     ViewDocsAction(FlutterGeneratorPeer peer) {
       super(peer);
     }
 
     @Override
     void perform() {
-      FlutterInitializer.sendAnalyticsAction(getClass().getSimpleName());
+      FlutterInitializer.sendAnalyticsAction(ANALYTICS_KEY);
       BrowserUtil.browse("https://flutter.io/setup/");
     }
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/module/InstallSdkAction.java
```diff
@@ -143,6 +145,7 @@ public void validateSelectedFiles(VirtualFile[] files) throws Exception {
 
       final VirtualFile installTarget = FileChooser.chooseFile(descriptor, null, null);
       if (installTarget != null) {
+        FlutterInitializer.sendAnalyticsAction(getClass().getSimpleName());
```
devoncarew: This is sending `GitCloneAction `; do we need to do that? Is this always called when `InstallSdkAction` runs?

devoncarew:
The first call to analytics lgtm. Not clear that we need the second call (unless calling git clone is conditional).

src/io/flutter/module/InstallSdkAction.java
pq: There are 2 actions: git clone only gets called if `git` is found, otherwise we just open a browser on the docs.  This way we get a separate call for each action.

devoncarew: On mobile now, but perhaps we want separate analytics calls for each case then - 'InstallSdkAction' and 'ShowInstallDocsAction'?

pq: Exactly.  I *think* this is what we've got here. Separate calls for:

1. ViewDocsAction, and
1. GitCloneAction

But I may be misunderstanding.  Anyway, no need to try and review this on your phone.  Let's look at it tomorrow. 👍

devoncarew: OK, slightly more grounded now :)

The places we're collecting analytics lgtm. Let's rename the actions so we get more semantic events into analytics. I'd suggest `GitCloneAction` => `InstallSdkRunGitAction` and `ViewDocsAction` => `InstallSdkViewDocsAction` or similar. The distinction I'm trying to make is that they're both an attempt to install the sdk - one redirects to git, and one to opening a browser.

pq: Got it.  String updated.  PTAL. 👍

PR #1138, opened 2017-06-27
Harden git existence check.

pq:
Prefer testing for clean execution over mere PATH presence check.

@devoncarew

### DIFF
--- a/src/io/flutter/FlutterUtils.java
+++ b/src/io/flutter/FlutterUtils.java
@@ -5,6 +5,9 @@
  */
 package io.flutter;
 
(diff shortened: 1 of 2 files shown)


devoncarew:
lgtm!

With this, the windows PR should be fine? (Or should we just not windows-ify it in anticipation of future windows improvements?)

pq:
> (Or should we just not windows-ify it in anticipation of future windows improvements?)

This was my thinking based on @mit-mit's feedback but I'm open either way.

mit-mit:
If we think it works for Windows, I'm OK keeping it

PR #1140, opened 2017-06-28
fix launching android emulators

devoncarew:
- fix an issue launching android emulators

We would only locate an android sdk if we could both locate the sdk, _and_ an `Sdk` for it had been configured in IntelliJ. We now separate out the concepts of a wrapper around an `AndroidSdk` on disk and the `IntelliJAndroidSdk` wrapper, that references IntelliJ's Sdk instance.

@pq

### DIFF
--- a/src/io/flutter/android/AndroidSdk.java
+++ b/src/io/flutter/android/AndroidSdk.java
@@ -24,6 +24,9 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+/**
+ * A wrapper around an Android sdk on disk.
+ */
 public class AndroidSdk {
   private static final Logger LOG = Logger.getInstance(AndroidSdk.class);
 
--- a/src/io/flutter/android/AndroidSdk.java
+++ b/src/io/flutter/android/AndroidSdk.java
@@ -25,7 +25,7 @@ import java.util.Collections;
 import java.util.List;
 
 /**
- * A wrapper around an Android sdk on disk.
+ * A wrapper around an Android SDK on disk.
  */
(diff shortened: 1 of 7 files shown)


src/io/flutter/android/AndroidSdk.java
```diff
@@ -30,133 +25,37 @@
 import java.util.List;
 
 /**
- * An Android SDK and its home directory.
+ * A wrapper around an Android sdk on disk.
```
pq: => SDK? (or lower case the rest in this file?)

src/io/flutter/android/IntelliJAndroidSdk.java
```diff
@@ -0,0 +1,142 @@
...
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.util.EnvironmentUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An Android SDK and its home directory; this references an IntelliJ @{@link Sdk} instance.
```
pq: => {@link Sdk} ?

src/io/flutter/android/AndroidSdk.java
devoncarew: done!

src/io/flutter/android/IntelliJAndroidSdk.java
devoncarew: Here I think the first SDK reference is more to the idea of an android sdk (rather than a specific class).

PR #1162, opened 2017-06-30
Improvements to running a single Flutter test

skybrian:
- Allow running it in the debugger.
- Allow user to run a single test by right-clicking, either in
  source code or on a test in the test tree widget.

### DIFF
--- a/src/io/flutter/run/test/TestConfigProducer.java
+++ b/src/io/flutter/run/test/TestConfigProducer.java
@@ -12,12 +12,15 @@ import com.intellij.openapi.util.Ref;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiDirectory;
 import com.intellij.psi.PsiElement;
-import com.jetbrains.lang.dart.psi.DartFile;
+import com.intellij.psi.impl.source.tree.LeafPsiElement;
+import com.jetbrains.lang.dart.DartTokenTypes;
+import com.jetbrains.lang.dart.psi.*;
 import io.flutter.dart.DartPlugin;
 import io.flutter.pub.PubRoot;
 import io.flutter.run.FlutterRunConfigurationProducer;
 import io.flutter.utils.FlutterModuleUtils;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/test/TestConfigProducer.java
```diff
@@ -35,31 +38,113 @@ protected TestConfigProducer() {
...
+      }
+      elt = elt.getParent();
+    }
+    return null; // not found
+  }
+
+  @Nullable
+  private String getCalledFunctionName(@NotNull DartCallExpression call) {
+    if (!(call.getFirstChild() instanceof DartReference)) return null;
+    return call.getFirstChild().getText();
```
pq: Do we need to consider malformed ASTs?  That is,  might there be a case where `getFirstChild()` returns null?

pq: (Or maybe we'd never get to this path in that case?)

src/io/flutter/run/test/TestConfigProducer.java
```diff
@@ -35,31 +38,113 @@ protected TestConfigProducer() {
...
+
+  @Nullable
+  private String getStringIfNotTemplate(@NotNull DartStringLiteralExpression expr) {
+    if (!expr.getShortTemplateEntryList().isEmpty() || !expr.getLongTemplateEntryList().isEmpty()) {
+      return null; // is a template
+    }
+    // We expect a quote, string part, quote.
+    if (expr.getFirstChild() == null) return null;
+    final PsiElement second = expr.getFirstChild().getNextSibling();
+    if (second.getNextSibling() != expr.getLastChild()) return null; // not three items
```
pq: Is this something we could easily get under test?

src/io/flutter/run/test/TestConfigProducer.java
skybrian: In this case, the instanceof check handles  the null check as well.

src/io/flutter/run/test/TestConfigProducer.java
skybrian: I'll take a look (and send another patch).

PR #1167, opened 2017-07-01
update README about running Flutter apps

skybrian:


skybrian:
I'm not sure we should keep this file since it somewhat duplicates what should really be in class-level JavaDoc. But at least it won't be blatantly out of date.

Issue #1136, opened 2017-06-27
improve the open android emulator actions


devoncarew:
From https://github.com/flutter/flutter-intellij/pull/1129, we should:

- consider using a submenu for the 'open android emulators' actions
- ideally we'd grey out the emulators that are already running

devoncarew:
Looking into this, I don't think we're going to be able to use ids to identify android emulators. We're getting different ids back from the `emulator` tools and `adb`.

I have a PR out for the first item, but it may not land - it's not clear that the UI is an improvement.

PR #1182, opened 2017-07-06
android emulator actions behind a submenu

devoncarew:
- put android emulator actions behind a submenu
- fix https://github.com/flutter/flutter-intellij/issues/1136 (we're not going to be able to address the portion of the bug related to disabling menu items for open emulators)

I'm a bit torn about the UI - I'm not 100% sure it's better than what we have now. This PR:

screen shot 2017-07-06 at 12 15 09 pm

and the before:

screen shot 2017-06-26 at 4 28 16 pm

Note that the user has to explicitly click on the `Open Android Emulators` submenu item for it to open - it doesn't auto-expand like a regular menu item  in the system menus.

@pq @skybrian @stevemessick thoughts?

### DIFF
--- a/src/io/flutter/actions/DeviceSelectorAction.java
+++ b/src/io/flutter/actions/DeviceSelectorAction.java
@@ -107,6 +107,9 @@ public class DeviceSelectorAction extends ComboBoxAction implements DumbAware {
       actions.add(new NoDevicesAction(message));
     }
 
+    // Show simulator and emulator actions.
+    final List<AnAction> additionalActions = new ArrayList<>();
+
     // Show the 'Open iOS Simulator' action.
     if (SystemInfo.isMac) {
       boolean simulatorOpen = false;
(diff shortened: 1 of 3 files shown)


stevemessick:
What is the motivation for adding a menu hierarchy?

devoncarew:
It was to help group the android emulator devices. It made sense in the context of the initial PR feedback, but seeing the UI in person, I think I'm leaning towards keeping the pulldown menu as it.

stevemessick:
First off, I hate submenus, so keep that in mind. :)
In this context I don't think it adds value. You can group them in the main menu. You can add an icon to further identify emulators vs simulators vs hardware. You could add a separator to clearly indicate the grouping. I could probably go on, but you get the idea. I'd look for a way to eliminate or shorten common text ("Open Android Emulator"). So, there are some alternatives to consider that help address feedback without adding submenus. :)

devoncarew:
I think I'll close this :)

PR #1183, opened 2017-07-06
sort the open emulator actions

devoncarew:
Follow up from https://github.com/flutter/flutter-intellij/pull/1182, sort the emulator actions.

@pq, @stevemessick


stevemessick:
LGTM

PR #1180, opened 2017-07-06
Flutter test run line markers (#1174).

pq:
Basic support for running tests from editor line markers.

image

In general, we should add some better scoping to our test discovery (at the moment we're not picky about the provenance of functions called `test`); TODO added and I'll follow-up with a fix, as well as some noodling on support, in general, for `testWidgets`.

Fixes: #1174.

@devoncarew

Issue #1174, opened 2017-07-05
Add editor annotations for running tests by name.


pq:
For example, see the Java editor annotations for JUnit tests:

![image](https://user-images.githubusercontent.com/67586/27887427-8f10896a-6195-11e7-893b-309693a7f8a1.png)

PR #1189, opened 2017-07-07
Test state icons.

pq:
Follow-up from #1180 and continuing #1174, adds state awareness to run marker icons to distinguish:

1. previously green tests
1. unrun tests
1. previously red tests
1. previously skipped tests

For example:

image

A future arc of work would get this under test but I think we have some integration test fleshing out for that.

@devoncarew @stevemessick

### DIFF
--- a/src/io/flutter/run/test/FlutterTestLineMarkerContributor.java
+++ b/src/io/flutter/run/test/FlutterTestLineMarkerContributor.java
@@ -36,7 +36,7 @@ import java.util.Map;
 
 public class FlutterTestLineMarkerContributor extends RunLineMarkerContributor {
 
-  private static final int TEST_LIMIT = 1024;
+  private static final int SCANNED_TEST_RESULT_LIMIT = 1024;
 
   @Nullable
   @Override
@@ -52,6 +52,7 @@ public class FlutterTestLineMarkerContributor extends RunLineMarkerContributor {
     return null;
   }
 
+  @NotNull
   private static Icon getTestStateIcon(@NotNull PsiElement element) {
 
     // SMTTestProxy maps test run data to a URI derived from a location hint produced by `package:test`.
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/test/FlutterTestLineMarkerContributor.java
```diff
@@ -20,25 +29,94 @@
...
+      // e.g., dart_location:///Users/pq/IdeaProjects/untitled1298891289891/test/unit_test.dart,3,2,["my first unit test"]
+      final String path = containingFile.getVirtualFile().getPath();
+      final String testLocationPrefix = "dart_location://" + path + "," + lineNumber;
+
+      final TestStateStorage storage = TestStateStorage.getInstance(project);
+      if (storage != null) {
+        final Map<String, TestStateStorage.Record> tests =
+          storage.getRecentTests(TEST_LIMIT, getSinceDate());
+        if (tests != null) {
+          for (Map.Entry<String, TestStateStorage.Record> entry : tests.entrySet()) {
```
skybrian: It looks like this is iterating over all recent tests (up to TEST_LIMIT and within the last day) once per PsiElement that matches isTestCall(). It seems like it might be expensive in a project has has many test files open?

src/io/flutter/run/test/FlutterTestLineMarkerContributor.java
```diff
@@ -20,25 +29,94 @@
...
 import org.jetbrains.annotations.Nullable;
 
+import javax.swing.*;
+import java.util.Date;
+import java.util.Map;
+
 
 public class FlutterTestLineMarkerContributor extends RunLineMarkerContributor {
+
+  private static final int TEST_LIMIT = 1024;
```
pq: Good idea.  Done!

src/io/flutter/run/test/FlutterTestLineMarkerContributor.java
pq: I don't think this is too expensive actually and should only occur for `PsiElement`s that correspond w/ a test call and only for active editors but you bring up a good point that we should be mindful.  I'll do a little exploring and add a TODO to keep me honest. 👍

pq:
Thanks @skybrian 🎉 !

PR #1192, opened 2017-07-07
locate the flutter sdk from the .packages file

### LINKED ISSUE #1039: locate sdks using the .packages file
Similar to https://github.com/Dart-Code/Dart-Code/issues/294, if we're opening a project that has no Flutter SDK set, instead of using the last one set by a previous project for the sdk, we could check for the presence of a .packages file and use it to infer the flutter sdk location.

devoncarew:
- when a flutter sdk isn't set, look for and parse a .packages file to infer the sdk location
- similarly, fallback on PATH detection of the flutter sdk
- fix #1039

@pq @stevemessick


src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -213,4 +219,73 @@ public static void enableDartSdk(@NotNull final Project project) {
...
+
+    return null;
+  }
+
+  /**
+   * Locate the Flutter SDK using the user's PATH.
+   */
+  @Nullable
+  public static String locateSdkFromPath() {
+    final String flutterBinPath = Which.which("flutter");
```
stevemessick: Which.which?
I'd have picked a different class name, maybe Sys, so other functions could be added.

devoncarew: Makes sense - I'll rename the class to something other than `Which`.

stevemessick:
Nice!

PR #1193, opened 2017-07-10
add an experimental reload on save option

### LINKED ISSUE #563: have an option to enable reload-on-save
The current design requires the developer to explicitly press 'hot reload' to see the effect of changes. It has been suggested that this should happen implicitly & automatically. This should probably be tied to the state of the analyzer so that we only push changes that are valid.

Not fully convinced this is going to work (it may cause more issues then expected), but enough users have suggested it that we aught to make an experiment.

devoncarew:
Add a reload-on-save implementation (fix #563). This is behind a user setting (defaulting to off) to allow us to get feedback and iterate on the UI.

This hooks into the explicit `SaveAllAction` action in IntelliJ - it doesn't fire on implicit saves. We'll reload if:

- the setting is on
- an app is running
- the current file (if any) can be parsed without syntactic errors
- there had been any unsaved files when save all was invoked, or
- if there are any files in the module are are newer then the last successful reload

The above means that we don't always reload on save. This might be hard to predict for the user, and is I think the main thing we want to test from a UX perspective. Some options are to:

- always reload on the save all action, irrespective of whether any files were saved, or if there had been any changes since the last reload
- provide visual feedback whenever hitting save all would issue a reload. This might be something like badging the reload button after the first file modification after a reload. This could mean the UI is always flashing after the user types after a reload, and might in practice be too twitchy.

We may also want an easier way to toggle the reload-on-save option than the user preferences - it may be something that you want to toggle on and off easily during any one app run.

We provide feedback on reload status in-line in the editor when reloading on save, and as a tooltip on the run or debug tool windows when reloading directly via the button or keybinding.

screen shot 2017-07-09 at 7 27 27 pm

screen shot 2017-07-09 at 7 29 48 pm

screen shot 2017-07-09 at 7 30 54 pm

screen shot 2017-07-09 at 7 31 10 pm

(I'll be out for a bit, so won't respond to feedback right away)

@pq and @stevemessick for review; @mit-mit general fyi, and @lukechurch and @InMatrix - this is something we'll probably want to test at some point.


src/io/flutter/run/FlutterReloadManager.java
```diff
@@ -0,0 +1,192 @@
...
+    return action instanceof FlutterAppAction ? ((FlutterAppAction)action).getApp() : null;
+  }
+
+  private boolean supportsPauseAfterReload() {
+    return DartPluginCapabilities.isSupported("supports.pausePostRequest");
+  }
+
+  // TODO(devoncarew): This technique still allows some syntax errors through; can we
+  // improve the IntelliJ grammer? Are there other issues in the AST we should look for?
+  private boolean documentHasErrors(@NotNull Document document) {
```
stevemessick: I'm curious why you don't just analyze the content of the analyzer errors view. Is it due to the complexity and timing of update files > analyze files > update error view?

src/io/flutter/FlutterInitializer.java
```diff
@@ -132,6 +133,9 @@ public void runActivity(@NotNull Project project) {
 
     FlutterRunNotifications.init(project);
 
+    // Watch save actions.
+    FlutterReloadManager.getInstance(project);
```
pq: It's not obvious to me what's happening here, I'm guessing there's a side effect?  Maybe update the comment to make it more obvious what's going on?

src/io/flutter/run/SdkRunConfig.java
```diff
@@ -83,10 +83,11 @@ public LaunchState getState(@NotNull Executor executor, @NotNull ExecutionEnviro
     final MainFile main = MainFile.verify(launchFields.getFilePath(), env.getProject()).get();
     final Project project = env.getProject();
     final RunMode mode = RunMode.fromEnv(env);
+    final Module module = ModuleUtil.findModuleForFile(main.getFile(), env.getProject());
```
pq: Aside: it looks like `MainFile.verify` is sure to return non-null; maybe we should annotate it that way?

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -338,6 +384,33 @@ public void removeStateListener(@NotNull StateListener listener) {
...
+        return true;
+      }
+      if (FlutterUtils.isFlutteryFile(file) || ".packages".equals(file.getName())) {
+        latest[0] = Math.max(latest[0], file.getTimeStamp());
+      }
+      return true;
+    };
+
+    if (myModule != null) {
+      ModuleRootManager.getInstance(myModule).getFileIndex().iterateContent(iterator);
```
pq: Neat! I had no idea how to do this. 👍

src/io/flutter/run/FlutterReloadManager.java
devoncarew: > Is it due to the complexity and timing of update files > analyze files > update error view?

Exactly. Feedback from the analysis server would be a much stronger signal, but currently much harder to get at due to the timing issues. Likely something we'd want to push on in future iterations.

src/io/flutter/FlutterInitializer.java
devoncarew: Yup, a side-effect; I'll add a comment.

src/io/flutter/run/SdkRunConfig.java
devoncarew: done

PR #1194, opened 2017-07-10
Bazel workspace configuration support for running a flutter doctor script

jwren:
@pq @devoncarew

### DIFF
--- a/src/io/flutter/actions/FlutterDoctorAction.java
+++ b/src/io/flutter/actions/FlutterDoctorAction.java
@@ -5,15 +5,56 @@
  */
 package io.flutter.actions;
 
+import com.intellij.execution.ExecutionException;
+import com.intellij.execution.configurations.GeneralCommandLine;
+import com.intellij.execution.process.OSProcessHandler;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.io.FileUtil;
+import com.intellij.openapi.vfs.CharsetToolkit;
+import io.flutter.bazel.Workspace;
+import io.flutter.console.FlutterConsoles;
 import io.flutter.pub.PubRoot;
 import io.flutter.sdk.FlutterSdk;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 public class FlutterDoctorAction extends FlutterSdkAction {
+
+  private static final Logger LOG = Logger.getInstance(FlutterDoctorAction.class.getName());
+
   @Override
   public void startCommand(@NotNull Project project, @NotNull FlutterSdk sdk, @Nullable PubRoot root) {
     sdk.flutterDoctor().startInConsole(project);
+  }
+
+  @Override
+  public void startCommandInBazelContext(@NotNull Project project, @NotNull Workspace workspace) {
+    final String doctorScript = workspace.getDoctorScript();
+    if (doctorScript != null) {
+      runWorkspaceFlutterDoctorScript(project, workspace.getRoot().getPath(), doctorScript);
+    } else {
+      LOG.error("No \"doctorScript\" script in the flutter.json file.");
+    }
+  }
+
+  private void runWorkspaceFlutterDoctorScript(@NotNull Project project, @NotNull String workDir, @NotNull String doctorScript) {
+    final GeneralCommandLine cmdLine = new GeneralCommandLine().withWorkDirectory(workDir);
+    cmdLine.setCharset(CharsetToolkit.UTF8_CHARSET);
+    cmdLine.setExePath(FileUtil.toSystemDependentName(doctorScript));
+
+    final OSProcessHandler handler;
+    try {
+      handler = new OSProcessHandler(cmdLine);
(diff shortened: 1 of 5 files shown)


src/io/flutter/actions/FlutterSdkAction.java
```diff
@@ -30,19 +32,36 @@
   public void actionPerformed(AnActionEvent event) {
     final Project project = DumbAwareAction.getEventProject(event);
 
-    final FlutterSdk sdk = project != null ? FlutterSdk.getFlutterSdk(project) : null;
-    if (sdk == null) {
-      showMissingSdkDialog(project);
-      return;
-    }
+    if(enableActionInBazelContext() && project != null) {
```
pq: Looks like we're missing a space after `if`?  Perhaps re-format?

src/io/flutter/actions/FlutterDoctorAction.java
```diff
@@ -5,15 +5,56 @@
...
+      FlutterConsoles.displayProcessLater(handler, project, null, handler::startNotify);
+    }
+    catch (ExecutionException e) {
+      LOG.error(e);
+    }
+  }
+
+  @Override
+  public boolean enableActionInBazelContext() {
+    return true;
```
pq: Doesn't this mean that `FlutterDoctorAction` will always run in the Bazel context? Doesn't this mean suggest that `startCommand` will never get run?

src/io/flutter/actions/FlutterSdkAction.java
```diff
@@ -30,19 +32,36 @@
...
+      FlutterInitializer.sendAnalyticsAction(this);
+      FileDocumentManager.getInstance().saveAllDocuments();
+      startCommand(project, sdk, PubRoot.forEventWithRefresh(event));
+    }
   }
 
   public abstract void startCommand(@NotNull Project project, @NotNull FlutterSdk sdk, @Nullable PubRoot root);
 
+  public void startCommandInBazelContext(@NotNull Project project, @NotNull Workspace workspace) {
+  }
```
pq: Maybe make this abstract?  Or if it's only appropriate for `FlutterDoctorAction`s, consider putting the conditional logic in there?

src/io/flutter/actions/FlutterSdkAction.java
jwren: Re-formatted

src/io/flutter/actions/FlutterDoctorAction.java
jwren: Currently yes, but that is a bug... Fixing up for the PR.

src/io/flutter/actions/FlutterSdkAction.java
jwren: Not all actions can/should be run in the bazel context.  The idea here is to have methods override this method and enableActionInBazelContext if they should have the opportunity to run in this mode.

Added this commment to startCommandInBazelContext:


  /**
   * Implemented by actions which are used in the Bazel context ({@link #enableActionInBazelContext()} returns true), by default this method
   * is a no-op.
   */

jwren:
PTAL

src/io/flutter/actions/FlutterSdkAction.java
```diff
@@ -30,19 +32,43 @@
...
+    }
   }
 
   public abstract void startCommand(@NotNull Project project, @NotNull FlutterSdk sdk, @Nullable PubRoot root);
 
+  /**
+   * Implemented by actions which are used in the Bazel context ({@link #enableActionInBazelContext()} returns true), by default this method
+   * is a no-op.
+   */
+  public void startCommandInBazelContext(@NotNull Project project, @NotNull Workspace workspace) {
```
pq: How about throwing an exception?  It seems like calling the no-op version would be a logic error?

src/io/flutter/actions/FlutterSdkAction.java
```diff
@@ -30,19 +32,43 @@
   public void actionPerformed(AnActionEvent event) {
     final Project project = DumbAwareAction.getEventProject(event);
 
-    final FlutterSdk sdk = project != null ? FlutterSdk.getFlutterSdk(project) : null;
-    if (sdk == null) {
-      showMissingSdkDialog(project);
-      return;
+    // See if the Bazel workspace exists for this project.
+    final Workspace workspace = project != null ? WorkspaceCache.getInstance(project).getNow() : null;
```
pq: Nit: don't we only need to access the workspace if `enableActionInBazelContext` returns true?

src/io/flutter/actions/FlutterDoctorAction.java
pq: 👍

src/io/flutter/actions/FlutterSdkAction.java
jwren: I like that idea, done.

src/io/flutter/actions/FlutterSdkAction.java
jwren: Correct, a small optimization, but an optimization non-the-less.  Done.

PR #1196, opened 2017-07-10
Centralized test function mapping.

pq:
Centralizes mapping of function names to test types (to ease adding `testWidgets` support once location issues are resolved).

@stevemessick @jwren


pq:
Thanks!

PR #1197, opened 2017-07-11
Icon annotations for test files.

pq:
Identifies test files (w/ a simple heuristic) in the project view and editor tabs.

image

image

@stevemessick

### DIFF
--- a/src/io/flutter/project/FlutterIconProvider.java
+++ b/src/io/flutter/project/FlutterIconProvider.java
@@ -12,6 +12,9 @@ import com.intellij.openapi.util.Iconable;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiDirectory;
 import com.intellij.psi.PsiElement;
+import com.jetbrains.lang.dart.DartFileType;
+import com.jetbrains.lang.dart.psi.DartFile;
+import io.flutter.FlutterUtils;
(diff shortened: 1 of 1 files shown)


pq:
Thanks!  Regarding transitive dependency checking, I looked into that and I'm concerned that I don't know whether we have resolution available when Icons are drawn.  We might also look for test calls to winnow out even more false positives.  That said, I chatted with @scheglov and @jwren for a bit this AM and the consensus is that if we are to do that we should push that into server proper and that I take to be an exercise for another day! 😄

src/io/flutter/project/FlutterIconProvider.java
```diff
@@ -33,6 +42,19 @@ public Icon getIcon(@NotNull final PsiElement element, @Iconable.IconFlags final
...
     }
 
+    // Files.
+    if (element instanceof DartFile && ModuleUtil.hasModulesOfType(element.getProject(), FlutterModuleType.getInstance())) {
+      DartFile dartFile = (DartFile)element;
+      final VirtualFile file = dartFile.getVirtualFile();
+      if (!file.isInLocalFileSystem()) return null;
+
+      // Use a simple naming convention heuristic to identify test files.
+      // TODO(pq): consider pushing up to the Dart Plugin.
```
devoncarew: @alexander-doroshko, fyi

PR #1200, opened 2017-07-11
Test discovery optimization.

pq:
* test call check short-circuiting to avoid needless Psi Element traversals
* minor cleanup

@stevemessick @jwren

### DIFF
--- a/src/io/flutter/FlutterUtils.java
+++ b/src/io/flutter/FlutterUtils.java
@@ -14,6 +14,7 @@ import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.progress.ProcessCanceledException;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiFile;
 import com.jetbrains.lang.dart.DartFileType;
 import com.jetbrains.lang.dart.psi.DartFile;
@@ -73,6 +74,19 @@ public class FlutterUtils {
   @Nullable
   public static VirtualFile getRealVirtualFile(@Nullable PsiFile psiFile) {
     return psiFile != null ? psiFile.getOriginalFile().getVirtualFile() : null;
+  }
+
+
+  /**
+   * Returns the Dart file for the given PsiElement, or null if not a match.
+   */
+  public static @Nullable DartFile getDartFile(final @Nullable PsiElement elt) {
+    if (elt == null) return null;
+
+    final PsiFile psiFile = elt.getContainingFile();
+    if (!(psiFile instanceof DartFile)) return null;
+
+    return (DartFile)psiFile;
   }
 
   public static void openFlutterSettings(@Nullable Project project) {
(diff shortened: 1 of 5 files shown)


src/io/flutter/dart/DartSyntax.java
```diff
@@ -70,21 +70,20 @@ public static DartCallExpression findEnclosingFunctionCall(@NotNull PsiElement e
    *
    * @return true if the given element is a call to function, false otherwise
    */
-  public static boolean isCallToFunctionNamed(@Nullable PsiElement element, @NotNull String function) {
-    if (!(element instanceof DartCallExpression)) return false;
-    final String name = getCalledFunctionName((DartCallExpression)element);
+  public static boolean isCallToFunctionNamed(@NotNull DartCallExpression element, @NotNull String function) {
```
stevemessick: Doesn't this require a change to DartSuyntaxTest?

src/io/flutter/dart/DartSyntax.java
```diff
@@ -70,21 +70,20 @@ public static DartCallExpression findEnclosingFunctionCall(@NotNull PsiElement e
...
    *
-   * @return true if the given element is a declaration, false otherwise
+   * @return true if the given element is a main declaration, false otherwise
    */
-  public static boolean isFunctionDeclarationNamed(@Nullable PsiElement element, @NotNull String name) {
+  public static boolean isMainFunctionDeclaration(@Nullable PsiElement element) {
     if (!(element instanceof DartFunctionDeclarationWithBodyOrNative)) return false;
     final String functionName = ((DartFunctionDeclarationWithBodyOrNative)element).getComponentName().getId().getText();
-    return Objects.equals(functionName, name);
+    return Objects.equals(functionName, "main");
```
jwren: nit: "main" could be a public static field (referenced anywhere else?)

src/io/flutter/FlutterUtils.java
```diff
@@ -75,6 +76,19 @@ public static VirtualFile getRealVirtualFile(@Nullable PsiFile psiFile) {
     return psiFile != null ? psiFile.getOriginalFile().getVirtualFile() : null;
   }
 
+
+  /**
+   * Returns the Dart file for the given PsiElement, or null if not a match.
+   */
+  public static @Nullable DartFile getDartFile(final @Nullable PsiElement elt) {
```
jwren: Nit: Nullable annotation should be first token in signature, and parameter statement, like lines 57 and 73 above

src/io/flutter/dart/DartSyntax.java
pq: No, but there should be more tests for sure.

Also, interestingly, I've got classpath issues and can't run these tests in EAP; will look into fixing that.

src/io/flutter/FlutterUtils.java
pq: Updated.

src/io/flutter/dart/DartSyntax.java
pq: In general we haven't been creating static fields for Strings that will just get interned (which is just to say there shouldn't be any significant performance difference).  As to whether it'd be clearer, I used to agree but now tend to think 

    Objects.equals(functionName, "main")

is clearer than

    Objects.equals(functionName, MAIN_FUNC_NAME)

where we have to define

    static final String MAIN_FUNC_NAME = "main";

elsewhere.

src/io/flutter/dart/DartSyntax.java
pq: Test added.  Thanks! 👍

PR #1201, opened 2017-07-11
Add a new module for AS customization

stevemessick:
@pq @jwren @devoncarew 

This will allow us to use a single repo yet not break the basic plugin with AS dependencies.


flutter-studio/flutter-studio.iml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/src" isTestSource="false" />
+      <sourceFolder url="file://$MODULE_DIR$/testSrc/unit" isTestSource="true" />
```
pq: I'm guessing our integration tests are not included because they don't make sense to run in AS?

stevemessick: We can add integration tests if they make sense. This PR is almost content-free, though :)

PR #1206, opened 2017-07-12
Restore New Project wizard

stevemessick:
@pq @devoncarew 
Replace the simplified 'New Project' wizard of Android Studio with the traditional IntelliJ one. This allows creation of Flutter projects (and many others) in addition to Android projects.

Consider this proof-of-concept and subject to revision as we understand the requirements better.

### DIFF
--- a/flutter-studio/src/io/flutter/startup/FlutterStudioInitializer.java
+++ b/flutter-studio/src/io/flutter/startup/FlutterStudioInitializer.java
@@ -19,6 +19,7 @@ public class FlutterStudioInitializer implements StartupActivity {
     NewProjectAction newProject = new NewProjectAction();
     newProject.getTemplatePresentation().setText("New &Project...", true);
     newProject.getTemplatePresentation().setDescription("Create a new project from scratch");
+    // TODO(messick): Design a New Project wizard for Android Studio + Flutter.
     replaceAction("NewProject", newProject);
   }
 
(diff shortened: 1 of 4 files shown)


flutter-studio/src/io/flutter/startup/FlutterStudioInitializer.java
```diff
@@ -0,0 +1,34 @@
...
+import javax.swing.*;
+
+public class FlutterStudioInitializer implements StartupActivity {
+
+  @Override
+  public void runActivity(@NotNull Project project) {
+    NewProjectAction newProject = new NewProjectAction();
+    newProject.getTemplatePresentation().setText("New &Project...", true);
+    newProject.getTemplatePresentation().setDescription("Create a new project from scratch");
+    replaceAction("NewProject", newProject);
```
devoncarew: Can you add a TODO: here? (my understanding is that replacing the project wizard is useful for us in the short-term, but not something we'd want to ship)

devoncarew:
lgtm (w/ the todo: change, if my understanding of the intend is correct)

flutter-studio/src/io/flutter/startup/FlutterStudioInitializer.java
pq: I was going to say just same thing!

PR #1207, opened 2017-07-13
Build for AS if -Dstudio=true on cmd line

stevemessick:
`ant -Dstudio-true` will build the plugin for Android Studio. This is not finished. The AS-specific plugin.xml entries still need to be merged into the main Flutter plugin.xml, and the name of the plugin should be changed (or not?). In any case, we will have different plugins for IntelliJ and Android Studio, but there will be only one that a user has to add to each platform.

@pq @devoncarew

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -4,7 +4,11 @@
      plugin in the IntelliJ sandbox in order to run Flutter IntelliJ unit tests. -->
 
 <!--suppress AntResolveInspection -->
-<project name="flutter-intellij" default="build">
+<project
+    name="flutter-intellij"
+    xmlns:if="ant:if"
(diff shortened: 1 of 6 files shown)


devoncarew:
If we do some detection at runtime (iff the product == android studio), we may be able to ship one plugin that can work in both IntelliJ IDEA and Android Studio.

Currently the pieces that would have to be android studio specific would be adding a new `File > New > Flutter Project...` menu item. @stevemessick, do you have a sense for other changes we'd need to make? The above one we could do as a check at runtime.

stevemessick:
Looks like travis has a really old ant distro. We can't use this technique even if we want to.

PR #1210, opened 2017-07-13
Combine all code into a single module

stevemessick:
Rearrange things so there is a single module that works for IntelliJ and Android Studio.

@pq @devoncarew


stevemessick:
PTAL. All checks pass now, but I always like someone to double-check code that uses reflection.

src/io/flutter/FlutterStudioInitializer.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+    actionManager.registerAction(actionId, newAction);
+  }
+
+  public static void runActivity(@NotNull Project project) {
+    try {
+      @SuppressWarnings("unchecked")
+      Class<AnAction> clazz = (Class<AnAction>)Class.forName("com.intellij.ide.actions.NewProjectAction");
+      AnAction newProject = clazz.newInstance();
+      Method method = clazz.getMethod("getTemplatePresentation");
```
devoncarew: Could you simplify this as `Presentation present = newProject. getTemplatePresentation()`?

devoncarew:
lgtm for the reflection :) thanks for the refactor

src/io/flutter/FlutterStudioInitializer.java
stevemessick: Yes, I'll do that then commit.

PR #1211, opened 2017-07-13
Add check for Flutter usage in the FlutterSdkAction bazel check

jwren:


src/io/flutter/actions/FlutterSdkAction.java
```diff
@@ -35,7 +35,7 @@ public void actionPerformed(AnActionEvent event) {
     if (enableActionInBazelContext()) {
       // See if the Bazel workspace exists for this project.
       final Workspace workspace = project != null ? WorkspaceCache.getInstance(project).getNow() : null;
-      if (workspace != null) {
+      if (workspace != null && workspace.usesFlutter(project)) {
```
devoncarew: OK, so we'll only perform these actions iff `/flutter/` is in the dir sturcture of any of the modules.

PR #1212, opened 2017-07-14
FlutterPubspecNotificationProvider optimization

jwren:
In the FlutterPubspecNotificationProvider, have the constructor receive the Project to avoid the call to ProjectLocator.getInstance().guessProjectForFile(file).

The other two NotificationProviders already do this.

### DIFF
--- a/src/io/flutter/editor/FlutterPubspecNotificationProvider.java
+++ b/src/io/flutter/editor/FlutterPubspecNotificationProvider.java
@@ -29,6 +29,12 @@ import javax.swing.*;
 public class FlutterPubspecNotificationProvider extends EditorNotifications.Provider<EditorNotificationPanel> implements DumbAware {
   private static final Key<EditorNotificationPanel> KEY = Key.create("flutter.pubspec");
 
+  @NotNull private final Project project;
+
+  public FlutterPubspecNotificationProvider(@NotNull Project project) {
+    this.project = project;
+  }
+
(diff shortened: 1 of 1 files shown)


src/io/flutter/editor/FlutterPubspecNotificationProvider.java
```diff
@@ -29,6 +29,12 @@
 public class FlutterPubspecNotificationProvider extends EditorNotifications.Provider<EditorNotificationPanel> implements DumbAware {
   private static final Key<EditorNotificationPanel> KEY = Key.create("flutter.pubspec");
 
+  @NotNull private final Project project;
+
+  public FlutterPubspecNotificationProvider(@NotNull Project project) {
+    this.project = project;
```
pq: Since we're extending `AbstractProjectComponent` I think we can just do `super(project)` (and update access below to `myProject`)?

jwren: We are extending `EditorNotifications.Provider`, (not `EditorNotifications` which extends `AbstractProjectComponent`)

pq: Ah, I see.  In that case I'm confused how the constructor gets called.  But I see that `SdkConfigurationNotificationProvider` works that way so, assuming you've verified that this works, LGTM!

PR #1213, opened 2017-07-14
Removal of the FlutterModuleType, issue #748.

jwren:
Flutter modules are no longer defined as modules with the FlutterModuleType, instead they are defined as modules with the Dart SDK enabled on the module, where the SDK path is defined inside of a Flutter SDK.  See FlutterModuleUtils.isFlutterModule().

Flutter Bazel projects are defined by FlutterModuleUtils.getFlutterBazelWorkspace(), as a project that has a Dart SDK enabled on some module and the module has flutter or module in the directory path.  The requirement to have the Dart SDK is a stronger requirement, but is already required for analysis from the Dart Analysis Server, so this will not be a problem.

Small IDEs support (aka WebStorm) for project creation and support has not changed.

This PR has updated and refactored FlutterModuleUtils to be more of a common interface and definition for determining the Flutterness of some Module or Project.

The FlutterInitializer detects the now deprecated module type , "FLUTTER_MODULE_TYPE", and converts these modules into the type of "WEB_MODULE".  On the first load of an existing project, users will see a notification from the framework "Unknown Module Type".


devoncarew:
Thanks! Will look at this today.

src/io/flutter/utils/FlutterModuleUtils.java
```diff
@@ -172,8 +214,23 @@ public static boolean usesFlutter(@NotNull Module module) {
     return CollectionUtils.filter(getModules(project), m -> isFlutterModule(m) || usesFlutter(m));
   }
 
+  public static void convertFromDeprecatedModuleType(@NotNull Project project) {
+    for (Module module : getModules(project)) {
+      if (isDeprecatedFlutterModuleType(module)) {
+        setFlutterModuleType(module);
```
devoncarew: When we do this, we should show a toast (`FlutterMessages.showInfo`) letting the user know - it'll help take the edge off the unknown module type toast that will also likely appear.

devoncarew:
Patching in locally and tested against:
- existing projects (it works, and creates the diffs on disk that you'd expect)
- new projects from intellij
- new projects from the cli
- new projects from the cli, w/ the intellij metadata deleted

lgtm!

The only feedback I have is that we should experiment with a toast - using the flutter icon and some very brief verbiage - to let the user know that their project matadata has been updated, and that the other (error) toast they're seeing from intellij is related and  can be safely ignored.

src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -211,8 +217,17 @@ public String getParentGroup() {
   }
 
   @Override
+  @NotNull
+  public String getBuilderId() {
+    // The builder id is used to distingish between different builders with the same module type, see
```
jwren: Done

src/io/flutter/utils/FlutterModuleUtils.java
jwren: Done

jwren:
Comments addressed: `FlutterMessages.showInfo` and spelling error.

PTAL at the changes.

PR #1218, opened 2017-07-17
tweak the reload action enablement

devoncarew:
Some small tweaks the reload action enablement:
- disable the reload button if there's no action app session
- have the 'open flutter view' and 'open observatory' buttons be enabled if there's an open session (previously they were enabled and disabled during a reload)

@pq @stevemessick @jwren


jwren:
LGTM

Issue #1181, opened 2017-07-06
Test support for Widget Tests.


pq:
Follow-up from #1180.  Test discovery and execution should be broadened to support calls to Widget tests (e.g., `testWidgets(..)` from `'package:flutter_test/flutter_test.dart'`).

PR #1219, opened 2017-07-18
Widget test discovery and execution (#1181).

pq:
Adds:
  * line marker support for `testWidgets`
  * test tree rewriting to elide synthetic framework-generated test groups and fix location mapping

Fixes: #1181.

image

image

image

@stevemessick


third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/util/DartTestLocationProviderZ.java
```diff
@@ -0,0 +1,161 @@
...
+            }
+          }
+
+          return true;
+        }
+
+        @Nullable
+        private DartCallExpression getGroup(final DartCallExpression expression) {
+          return (DartCallExpression)PsiTreeUtil.findFirstParent(expression, true,
+                                                                 element -> element instanceof DartCallExpression &&
```
stevemessick: Things like this make me wish we had more flexibility with formatting rules.

PR #1221, opened 2017-07-18
fix a few analysis issues

devoncarew:
@stevemessick

### DIFF
--- a/src/io/flutter/FlutterStudioInitializer.java
+++ b/src/io/flutter/FlutterStudioInitializer.java
@@ -7,9 +7,6 @@ import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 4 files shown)


src/io/flutter/FlutterStudioInitializer.java
```diff
@@ -7,9 +7,6 @@
 import com.intellij.openapi.project.Project;
 import org.jetbrains.annotations.NotNull;
 
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-
 public class FlutterStudioInitializer {
```
stevemessick: Could you drop the changes to this file? I found a way to add the menu using XML so most of this is going away. But until I get further with New Module extensions I don't know if this class will be needed.

devoncarew: 👍 will do

devoncarew:
I'll drop this PR and recreate the unused imports changes later.

Issue #1209, opened 2017-07-13
set up an android studio bot


devoncarew:
As part of android studio support, we should have a bot building the plugin against the android studio binaries, in order to ensure that the set of APIs we're using is happy wrt Android Studio.

PR #1222, opened 2017-07-18
rev versions we build against; re-add android studio

devoncarew:
- rev versions we build against
- re-add android studio (fix https://github.com/flutter/flutter-intellij/issues/1209)


devoncarew:
@stevemessick

PR #1224, opened 2017-07-20
move the android studio docs to the contributing page

devoncarew:
- move the android studio docs to the contributing page

@stevemessick


devoncarew:
> the removal of excludeFolders is intentional?

Yup! New with the Dart plugin in 2017.2.

Issue #1155, opened 2017-06-29
IntelliJ being open interferes with switching Flutter branches


collinjackson:
I was switching Flutter branches using `git bisect` while IntelliJ was also open.  The IDE shows an error:

```
Flutter device daemon #1: process exited during startup. Exit code: 1, stderr:
/Users/jackson/git/flutter/bin/flutter: line 89: /Users/jackson/git/flutter/bin/cache/dart-sdk/bin/dart: No such file or directory

com.intellij.execution.ExecutionException: Flutter device daemon #1: process exited during startup. Exit code: 1, stderr:
/Users/jackson/git/flutter/bin/flutter: line 89: /Users/jackson/git/flutter/bin/cache/dart-sdk/bin/dart: No such file or directory

	at io.flutter.run.daemon.DeviceDaemon$Command.start(DeviceDaemon.java:222)
	at io.flutter.run.daemon.DeviceService.chooseNextDaemon(DeviceService.java:220)
	at io.flutter.utils.Refreshable.runInBackground(Refreshable.java:214)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

I then got a different error trying to `flutter run` on the command line.

```
$ flutter run -d i
Downloading Dart SDK 1.24.0...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 57.9M  100 57.9M    0     0  13.9M      0  0:00:04  0:00:04 --:--:-- 13.9M
warning:  cannot set modif./access times for /Users/jackson/git/flutter/bin/cache/dart-sdk/lib/dev_compiler/common/dart_sdk.js.map
          No such file or directory
  error:  invalid compressed data to inflate /Users/jackson/git/flutter/bin/cache/dart-sdk/lib/front_end/lib/src/fasta/source/outline_builder.dart
[...]

PR #1227, opened 2017-07-20
work towards fixing an sdk upgrade issue

devoncarew:
One half of fixing an issue when upgrading the dart sdk in use by the flutter sdk (work towards https://github.com/flutter/flutter-intellij/issues/1155). This PR changes errors from the device daemon to not be hard errors - things that we show to users in dialogs.

@stevemessick

### DIFF
--- a/src/io/flutter/run/daemon/DeviceService.java
+++ b/src/io/flutter/run/daemon/DeviceService.java
@@ -205,10 +205,11 @@ public class DeviceService {
       return previous; // Don't do anything; current daemon is what we want.
     }
 
-    // Wait a bit to see if we get cancelled.
-    // This is to try to avoid starting a process only to immediately kill it.
+    // Wait a bit to see if we get cancelled. This is to try to avoid starting a process only to
+    // immediately kill it. Also, delay a bit in case the flutter tool just upgraded the sdk;
+    // we'll need a bit more time to start up.
     try {
-      Thread.sleep(50);
+      Thread.sleep(2000);
     }
     catch (InterruptedException e) {
       return previous;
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/daemon/DeviceService.java
```diff
@@ -205,10 +205,11 @@ private DeviceDaemon chooseNextDaemon(Refreshable.Request<DeviceDaemon> request)
...
     }
 
-    // Wait a bit to see if we get cancelled.
-    // This is to try to avoid starting a process only to immediately kill it.
+    // Wait a bit to see if we get cancelled. This is to try to avoid starting a process only to
+    // immediately kill it. Also, delay a bit in case the flutter tool just upgraded the sdk;
+    // we'll need a bit more time to start up.
     try {
-      Thread.sleep(50);
+      Thread.sleep(2000);
```
stevemessick: I don't understand why this timeout is increased. Is 2s enough to d/l a new SDK? Or is that some slack time to let the system stabilize after the SDK is installed?

devoncarew: It's a good question; I'm going to think more about the timing of the sdk upgrades (this delay is in here to work around some exceptions I saw if we run the `flutter` tool too soon after the new sdk is downloaded - possibly a race condition with the flutter cli tool?).

devoncarew: I think this is related to the time to snapshot flutter_tools. So:
- lock the flutter lock
- test to update the sdk
- build a snapshot of flutter_tools
- unlock the flutter lock
- run the flutter_tools dart code

However, it looks like we don't lock on the mac (https://github.com/flutter/flutter/blob/master/bin/flutter#L74), which is likely the cause of this issue.

devoncarew: (needing to introduce a delay)

PR #1229, opened 2017-07-21
adjust device list refresh

devoncarew:
- adjust the device list refresh (some work towards https://github.com/flutter/flutter/issues/8439)
- don't use a fixed delay before we show devices in the pulldown
- do introduce a bit more delay when restarting the deamon process

@stevemessick @pq

### DIFF
--- a/src/io/flutter/run/daemon/DeviceDaemon.java
+++ b/src/io/flutter/run/daemon/DeviceDaemon.java
@@ -219,11 +219,6 @@ class DeviceDaemon {
           try {
             ready.get(100, TimeUnit.MILLISECONDS);
 
-            // Try not to show a partial device list.
-            // It currently takes 4+ seconds for the devices to show up.
-            // Remove when fixed: https://github.com/flutter/flutter/issues/8439
-            Thread.sleep(4200);
-
             succeeded = true;
             return new DeviceDaemon(daemonId, this, process, listener, devices);
           }
--- a/src/io/flutter/run/daemon/DeviceService.java
+++ b/src/io/flutter/run/daemon/DeviceService.java
@@ -28,6 +28,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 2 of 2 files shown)


src/io/flutter/run/daemon/DeviceDaemon.java
```diff
@@ -218,11 +218,6 @@ else if (process.isProcessTerminated()) {
           try {
             ready.get(100, TimeUnit.MILLISECONDS);
 
-            // Try not to show a partial device list.
-            // It currently takes 4+ seconds for the devices to show up.
-            // Remove when fixed: https://github.com/flutter/flutter/issues/8439
```
pq: Is https://github.com/flutter/flutter/issues/8439 fixed?

src/io/flutter/run/daemon/DeviceService.java
```diff
@@ -230,9 +232,12 @@ private DeviceDaemon chooseNextDaemon(Refreshable.Request<DeviceDaemon> request)
   public void restart() {
     if (project.isDisposed()) return;
 
-    deviceDaemon.refresh(this::shutDownDaemon);
+    JobScheduler.getScheduler().schedule(this::shutDown, 0, TimeUnit.SECONDS);
+    JobScheduler.getScheduler().schedule(this::refreshDeviceDaemon, 4, TimeUnit.SECONDS);
```
pq: Ah, or is this the delay for https://github.com/flutter/flutter/issues/8439?  (If so, maybe a TODO to remove?)

devoncarew: This mystery delay allows the daemon to shutdown and restart successfully. Without this it would shutdown, but wouldn't start back up. I'm not sure if this is due to a race condition in the processes, or some bug in (or misuse of) the `Refreshable` class.

:(

src/io/flutter/run/daemon/DeviceDaemon.java
devoncarew: Not yet, but I have a local change in progress to fix it, and this delay is not actually necessary IMHO. We only show one device at a time, so trying not to display a partial list doesn't make a lot of sense.

Issue #1240, opened 2017-07-26
M16: switching devices in the device selector is now slow


devoncarew:
It seems to block the UI thread for 1-2 seconds. This is likely due to re-building the device selector combo, and the delay while we query flutter config for the android sdk location.

devoncarew:
@pq, do you mind doing a little investigation here? Some ideas are that we're spending time in the UI threading shelling out to list the android emulator devices, and that doing some poor man's profiling of the method that populates the device pulldown menu could ID some culprits.

devoncarew:
@pq, nevermind, I think this may be an issue w/ my last PR.

PR #1241, opened 2017-07-26
faster device switching

devoncarew:
- cache the value of the android sdk returned from flutter_tools; make device switching faster (fix https://github.com/flutter/flutter-intellij/issues/1240)

There is still a delay in redrawing the device selector combo box (which I wasn't able to suss out), but we're no longer blocking the UI thread.

@pq


devoncarew:
We may still be doing some work on the UI thread (while rebuilding the device menu), so we should continue to investigate after this lands.

Issue #1239, opened 2017-07-26
M16: exception thrown opening flutter_gallery with the new module types


devoncarew:
```
com.intellij.webcore.moduleType.PlatformWebModuleType cannot be cast to com.intellij.openapi.module.WebModuleType
java.lang.ClassCastException: com.intellij.webcore.moduleType.PlatformWebModuleType cannot be cast to com.intellij.openapi.module.WebModuleType
	at com.intellij.openapi.module.WebModuleType.getInstance(WebModuleType.java:13)
	at io.flutter.utils.FlutterModuleUtils.getModuleTypeForFlutter(FlutterModuleUtils.java:54)
	at io.flutter.utils.FlutterModuleUtils.setFlutterModuleType(FlutterModuleUtils.java:236)
	at io.flutter.utils.FlutterModuleUtils.convertFromDeprecatedModuleType(FlutterModuleUtils.java:221)
	at io.flutter.FlutterInitializer.runActivity(FlutterInitializer.java:119)
	at com.intellij.ide.startup.impl.StartupManagerImpl.a(StartupManagerImpl.java:176)
	at com.intellij.ide.startup.impl.StartupManagerImpl.b(StartupManagerImpl.java:162)
	at com.intellij.ide.startup.impl.StartupManagerImpl.a(StartupManagerImpl.java:150)
	at com.intellij.ide.startup.impl.StartupManagerImpl.a(StartupManagerImpl.java:367)
	at com.intellij.ide.startup.impl.StartupManagerImpl.c(StartupManagerImpl.java:182)
```

PR #1242, opened 2017-07-26
fix an exception upgrading the module type in webstorm

devoncarew:
- fix an exception upgrading the module type in webstorm; the `WebModuleType.getInstance()` will always throw an exception in webstorm (2017.2?) - we work around the issue by using IDs
- fix https://github.com/flutter/flutter-intellij/issues/1239
- wordsmithing of the module upgrade text

@pq

### DIFF
--- a/src/io/flutter/module/FlutterModuleBuilder.java
+++ b/src/io/flutter/module/FlutterModuleBuilder.java
@@ -227,7 +227,7 @@ public class FlutterModuleBuilder extends ModuleBuilder {
   @Override
   @NotNull
   public ModuleType getModuleType() {
-    return FlutterModuleUtils.getModuleTypeForFlutter();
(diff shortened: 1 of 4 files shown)


src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -227,7 +227,7 @@ public String getBuilderId() {
   @Override
   @NotNull
   public ModuleType getModuleType() {
-    return FlutterModuleUtils.getModuleTypeForFlutter();
+    return ModuleTypeManager.getInstance().findByID(FlutterModuleUtils.getModuleTypeIDForFlutter());
```
pq: Any reason not to push this logic into `io.flutter.utils.FlutterModuleUtils#getModuleTypeForFlutter`?

devoncarew: I was avoiding creating instances of ModuleTypes were I could, but don't have a strong preference here -

PR #1243, opened 2017-07-26
improve the speed of the device pulldown

devoncarew:
- improve the speed of the device pulldown (fix https://github.com/flutter/flutter-intellij/issues/1240); we now remember FlutterSdk instances for project:sdk_path combos
- when the device menu is open, show the currently selected device's menu as selected

@pq


pq:
LGTM!

🎉 🎉 🎉

Issue #1244, opened 2017-07-26
Device Selector should show emulator device name (currently generic)


eturk1:
## Steps to Reproduce
1. create more than one emulator for testing
2. open two or more
3. all emulators are listed with same name, can't be easily distinguished to control Run
![image](https://user-images.githubusercontent.com/525280/28643824-f90351f0-720b-11e7-8dd5-c0b04b0fe546.png)

Showing device emulated (i.e. Nexus 5X, Nexus 7) would be preferable to support testing on multiple device formats.

## Version info

C:\Users\<name>\IdeaProjects\mary_bp>flutter doctor

[√] Flutter (on Microsoft Windows [Version 10.0.15063], locale en-US, channel alpha)
    • Flutter at C:\Users\erolf\git\flutter
    • Framework revision d36e2f6191 (5 weeks ago), 2017-06-21 15:09:10 -0700
    • Engine revision b0dee695ec
    • Tools Dart version 1.24.0-dev.6.7

[√] Android toolchain - develop for Android devices (Android SDK 26.0.0)
    • Android SDK at C:\Users\erolf\AppData\Local\Android\sdk
    • Platform android-26, build-tools 26.0.0
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 1.8.0_112-release-b06)

[√] Android Studio (version 2.3)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Gradle version 4.0.1
    • Java version OpenJDK Runtime Environment (build 1.8.0_112-release-b06)

[√] IntelliJ IDEA Community Edition (version 2017.2)
    • Flutter plugin version 15.2
    • Dart plugin version 172.3317.48

[√] Connected devices
    • Android SDK built for x86 • emulator-5556    • android-x86 • Android 7.1.1 (API 25) (emulator)
    • Android SDK built for x86 • emulator-5554    • android-x86 • Android 8.0.0 (API 26) (emulator)
    • Nexus 6P                  • 84B5T15A29020348 • android-arm • Android 8.0.0 (API 26)

devoncarew:
We're getting the names of the emulators we can open with `emulator -list-avds`, and the open android devices with `adb devices -l`. I suspect that the avd images could also be updated to change the names they report to adb?

eturk1:
The text after "Open..." in screen shot bottom is what's desired in at top next to phone icon. Possible?

[...]

PR #1251, opened 2017-07-29
de-dupe device names

devoncarew:
- ensure that we show a unique name in the device dropdown - if there are different devices with the same name, display the name with the device id as well (some work towards https://github.com/flutter/flutter-intellij/issues/1244)

So, Instead of `Foo Device`, show `Foo Device (device-id)`.

### DIFF
--- a/src/io/flutter/actions/DeviceSelectorAction.java
+++ b/src/io/flutter/actions/DeviceSelectorAction.java
@@ -99,8 +99,8 @@ public class DeviceSelectorAction extends ComboBoxAction implements DumbAware {
 
     final Collection<FlutterDevice> devices = service.getConnectedDevices();
 
-    for (FlutterDevice item : devices) {
-      actions.add(new SelectDeviceAction(item));
+    for (FlutterDevice device : devices) {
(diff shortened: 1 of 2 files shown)


devoncarew:
This PR just shows the short device name in the combo box, and the full name in the pulldown list. It is a bit of a disconnect, but the longer name seemed to take up too much name in the toolbar....

devoncarew:
@stevemessick @pq

eturk1:
16.0 plugin got pushed out to my machine last night. Now there's a bunch of "emulators" listed... but I only have one AVD built.
image


adb devices (while Nexus 5 emulator is running) >
List of devices attached
emulator-5554   device
emulator-5568   host
emulator-5566   host
emulator-5564   host

devoncarew:
I don't think this is related to the update. It's listing one AVD (at the end of the menu), and 4 running android devices.

`adb devices -l` _may_ give you a bit more information. Restarting the emulator might clear out the others? This sounds like an adb specific issue, and perhaps something that a google search may turn up info for.

Issue #729, opened 2017-02-14
show analysis issues on app run


devoncarew:
When running apps, if there are analysis issues, should we increase the visibility of errors. THis could be:

- showing the Dart Analysis view
- opening a toast telling the users about the errors
- stopping the launch?

eseidelGoogle:
Other mobile development contexts (e.g. Xcode or Android Studio) don't even let you run with compile errors (due to constraints of their development languages).  It's kinda neat that Dart does... but might cause confusion for the user.

pq:
Java Run Configurations perform a "build" pre-flight (which includes an error-check):

<img width="744" alt="screen shot 2017-02-14 at 6 12 27 am" src="https://cloud.githubusercontent.com/assets/67586/22932410/ab7f5e9c-f27c-11e6-8cc1-a7ad92387487.png">

<img width="235" alt="screen shot 2017-02-14 at 6 14 52 am" src="https://cloud.githubusercontent.com/assets/67586/22932491/f555d2f8-f27c-11e6-8769-f07144fac79a.png">

we could consider something similar?

alexander-doroshko:
For example, you can proceed with the app launch but show a yellow balloon stuck to the Dart Analysis tool window saying that it contains errors.
We already show such a balloon to attract user attentions when some error is printed to the Pub Serve tool window.

devoncarew:
> you can proceed with the app launch but show a yellow balloon stuck to the Dart Analysis tool window

Sounds like a good solution! Do you have a link to the impl. for the pub server window? And what would the behavior be like if the user was not showing tool buttons?

alexander-doroshko:
> Do you have a link to the impl. for the pub server window?

see `PubServerService#NOTIFICATION_GROUP`

> And what would the behavior be like if the user was not showing tool buttons?

As I remember they are shown anyway while balloon is there

devoncarew:
(hereabouts: https://github.com/JetBrains/intellij-plugins/blob/master/Dart/src/com/jetbrains/lang/dart/pubServer/PubServerService.java#L411)

devoncarew:
The resolution here is:

- show a notification if you run and there are errors
- force open of the Dart Analysis view

We should also show the notification on a reload.

devoncarew:
We need to figure out the right state machine here so that we have a speed-bump if there are analysis issues, but it's not too intrusive to their flow.

pq:
Work ongoing in the Dart Plugin.

[...]

PR #1264, opened 2017-08-04
display any analysis issues on run

devoncarew:
- when we launch an app, check for analysis errors, and display a tool window tooltip on the analysis view if there are any
- fix https://github.com/flutter/flutter-intellij/issues/729

@pq 

We're using reflection for now - we can move to using at compile time when we rev to 2017.2,


src/io/flutter/run/LaunchState.java
```diff
@@ -113,6 +142,15 @@ private RunContentDescriptor launch(@NotNull ExecutionEnvironment env) throws Ex
     }
   }
 
+  private static Class classForName(String className) {
```
pq: Surprising this isn't in `ReflectionUtil`.

devoncarew: There is one, but it throws a runtime exception (not a checked one).

PR #1267, opened 2017-08-07
Eliminate dep on commons-lang3

stevemessick:
Android Studio 3.0 canary 9 does not include the Apache commons-lang3 library. We could distribute it, but it seems easier to in-line the single function that is used.

@pq @devoncarew

### DIFF
--- a/src/io/flutter/sdk/FlutterCreateAdditionalSettings.java
+++ b/src/io/flutter/sdk/FlutterCreateAdditionalSettings.java
@@ -6,12 +6,72 @@
 package io.flutter.sdk;
 
 import com.intellij.openapi.util.text.StringUtil;
-import org.apache.commons.lang3.BooleanUtils;
 import org.jetbrains.annotations.Nullable;
+
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


src/io/flutter/sdk/FlutterCreateAdditionalSettings.java
```diff
@@ -6,12 +6,72 @@
...
+    this.description = description;
+    this.org = org;
+    this.swift = swift;
+    this.kotlin = kotlin;
+  }
+
+  public List<String> getArgs() {
+    final List<String> args = new ArrayList<>();
+
+    if (Boolean.TRUE.equals(includeDriverTest)) {
```
branflake2267: nifty

devoncarew:
(currently at canary 7: https://github.com/flutter/flutter-intellij/blob/master/.travis.yml#L21)

stevemessick:
I just checked: C7 doesn't have the commons-lang3 library either. It does have 2.6 but that wouldn't work with the given import. We should probably rev the bot version anyway.

Issue #1254, opened 2017-07-31
run action not directed to the correct device


wmleler:
If you are running an app in IntelliJ on one device, say the iOS simulator, then add a second device, say a Nexus 6P and hit run, the iOS simulator restarts. Nothing happens on the 6P.

devoncarew:
I think what's happening here is that the run command is being passed to the active running app, when the device pulldown has been changed to different device (w/ no current running app).

LarkAscending:
Have also seen this recently...

devoncarew:
Cool; this is something that we'll like fix in a patch release to the current plugin.

PR #1272, opened 2017-08-08
allow running on a second device

devoncarew:
- fix https://github.com/flutter/flutter-intellij/issues/1254

This fixes an issue with the run action when there is a currently running app, and the user has two devices connected. Previously we ignored the device selection when deciding to issue a restart command (the user hits running with a currently running app). We now pay attention to the device selection, and will instead launch an app on that second device.

As part of fixed that issue, I also fixed one where the retargetable actions (reload and restart) would send their command to the last launched app. They now target the foreground process (which we determine via the last process that queried for `update()` in their actions, there being no real API in IntelliJ to determine the active launched process).

@pq @stevemessick

### DIFF
--- a/src/io/flutter/actions/FlutterAppAction.java
+++ b/src/io/flutter/actions/FlutterAppAction.java
@@ -21,6 +21,8 @@ abstract public class FlutterAppAction extends DumbAwareAction {
 
   @NotNull private final FlutterApp myApp;
   @NotNull private final Computable<Boolean> myIsApplicable;
+  @NotNull private final String myActionId;
+
   private final FlutterApp.StateListener myListener = new FlutterApp.StateListener() {
     @Override
     public void stateChanged(FlutterApp.State newState) {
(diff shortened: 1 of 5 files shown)


src/io/flutter/actions/FlutterAppAction.java
```diff
@@ -36,26 +38,39 @@ public FlutterAppAction(@NotNull FlutterApp app,
...
 
-  private void registerAction(@NotNull String actionId) {
+  private void updateActionRegistration(boolean isConnected) {
     final ActionManager actionManager = ActionManager.getInstance();
-    final AnAction action = actionManager.getAction(actionId);
-    // New debug sessions create new actions, requiring us to overwrite existing ones in the registry.
-    // TODO(pq): consider moving actions to our own registry for lookup.
-    if (action != null) {
-      actionManager.unregisterAction(actionId);
+    final AnAction action = actionManager.getAction(myActionId);
```
pq: I think I'm missing something but I don't see `action` being used... Or?

src/io/flutter/run/LaunchState.java
```diff
@@ -320,16 +334,25 @@ protected final RunContentDescriptor doExecute(@NotNull RunProfileState state, @
...
-          if (launchMode.supportsReload() && app.isStarted()) {
-            // Map a re-run action to a flutter full restart.
-            FileDocumentManager.getInstance().saveAllDocuments();
-            app.performRestartApp();
-          }
+
+        String selectedDeviceId = null;
+        final FlutterDevice selectedDevice = DeviceService.getInstance(env.getProject()).getSelectedDevice();
+        if (selectedDevice != null) {
+          selectedDeviceId = selectedDevice.deviceId();
```
pq: Nit: maybe extract this into a method (`getSelectedDeviceId()`?) and share w/ `canRun(..)`?

src/io/flutter/actions/FlutterAppAction.java
devoncarew: Ah, good catch (and sorry for not waiting or the review!)

src/io/flutter/run/LaunchState.java
devoncarew: sgtm

PR #1275, opened 2017-08-08
Null checks and API updates.

pq:
Fixes a few possible NPEs and updates API contracts.


devoncarew:
These looks good in the large. I have not worked through the logic of how these new null checks will handle at runtime (our plugin, compiled with these new null checks, running in the context of 2017.1).

Issue #1104, opened 2017-06-12
Create an initial "Open inspector" action


pq:
A minimal action to integrate with the initial on-device visual inspector.

cc @jacob314

mit-mit:
Let's put this somewhere not too prominent initially.

pq:
_**UPDATE:** this is hinging on work in flutter proper._

The first enabling PR to watch is https://github.com/flutter/flutter/pull/11147; after that we'll need a change to take advantage of the offstage information in the diagnostics to perform better hit testing.

pq:
PR to watch: https://github.com/flutter/flutter/pull/10332 (_Bare bones widget inspector support_).

PR #1276, opened 2017-08-08
Initial “toggle inspect mode action” (#1104).

pq:
For iteration purposes.

* provisional icon : image

* functionality a no-op until https://github.com/flutter/flutter/pull/10332 lands

image

Fixes: #1104.

@devoncarew 

FYI @jacob314

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -93,6 +93,7 @@ public class FlutterView implements PersistentStateComponent<FlutterView.State>,
     toolWindowEx.setTitleActions(
       new DebugDrawAction(this),
       new PerformanceOverlayAction(this),
+      new ToggleInspectModeAction(this),
       new TogglePlatformAction(this)
     );
 
@@ -234,6 +235,16 @@ class TimeDilationAction extends AbstractToggleableAction {
     final Map<String, Object> params = new HashMap<>();
     params.put("timeDilation", isSelected(event) ? 5.0 : 1.0);
     view.getFlutterApp().callServiceExtension("ext.flutter.timeDilation", params);
+  }
+}
+
+class ToggleInspectModeAction extends AbstractToggleableAction {
+  ToggleInspectModeAction(@NotNull FlutterView view) {
+    super(view, "Toggle Inspect Mode", "Toggle Inspect Mode", AllIcons.Actions.Menu_find);
+  }
+
+  protected void perform(AnActionEvent event) {
+    view.getFlutterApp().callBooleanExtension("ext.flutter.widgetInspector", isSelected(event));
   }
 }
 
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -92,7 +92,6 @@ public class FlutterView implements PersistentStateComponent<FlutterView.State>,
     final ToolWindowEx toolWindowEx = (ToolWindowEx)toolWindow;
     toolWindowEx.setTitleActions(
       new DebugDrawAction(this),
-      new PerformanceOverlayAction(this),
       new ToggleInspectModeAction(this),
       new TogglePlatformAction(this)
     );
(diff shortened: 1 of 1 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -237,6 +238,16 @@ protected void perform(AnActionEvent event) {
   }
 }
 
+class ToggleInspectModeAction extends AbstractToggleableAction {
```
devoncarew: Does the `ext.flutter.widgetInspector` toggle anything on device, or is it a one-shot thing to request an on-device inspector? (/cc @jacob314)

devoncarew:
Icon looks good! Let's move the fps icon (the target) into the overflow menu in order to avoid confusion between the two.

src/io/flutter/view/FlutterView.java
```diff
@@ -237,6 +238,16 @@ protected void perform(AnActionEvent event) {
...
+  ToggleInspectModeAction(@NotNull FlutterView view) {
+    super(view, "Toggle Inspect Mode", "Toggle Inspect Mode", AllIcons.Actions.Menu_find);
+  }
+
+  protected void perform(AnActionEvent event) {
+    view.getFlutterApp().callBooleanExtension("ext.flutter.widgetInspector", isSelected(event));
```
devoncarew: Boolean extensions return a bool, and can take an optional bool arg (for example, turning on or off repaint rainbow, or querying whether repaint rainbow is on). I'm wondering if the `ext.flutter.widgetInspector` service protocol extension is a boolean extension, or just a method call to switch into inspector mode.

pq: From looking at https://github.com/flutter/flutter/pull/10332/files#diff-e158ae61bb7a8dd81f710e8f59aa23ca, I think it'll be a boolean extension.

pq:
`PerformanceOverlayAction` moved to overflow.  Screenshot updated.

src/io/flutter/view/FlutterView.java
devoncarew: kk. Unless (or until) we can get an event when the inspector closes, we may want to treat it as a one-shot, show inspector action.

src/io/flutter/view/FlutterView.java
devoncarew: If we decide the behavior is more of a one-shot, open action, this should inherit from `FlutterViewAction`, and the action name change to something like `Show Widget Inspector`.

jacob314: It is toggle-able. The device won't be completely out of inspect mode until the inspector is turned off either by flutter tool or intellij

src/io/flutter/view/FlutterView.java
jacob314: from the flutter tool, press 'i' to enter inspect mode, press 'i' again to exit inspect mode.

src/io/flutter/view/FlutterView.java
devoncarew: Ah, ok. I had imagined that there was an on-device way to exit inspector mode, and that a 'toggle' button in IntelliJ, w/o an event or such from the device, would get out of sync.

devoncarew:
> PerformanceOverlayAction moved to overflow

Can you push the changes to the PR?

lgtm! Sounds like I had a not-quite-accurate view of how the on-device inspector would work.

pq:
> Can you push the changes to the PR?

Whoops, yeah. Thought I had! Done now.

devoncarew:
@pq, `PerformanceOverlayAction` needs to move into the setAdditionalGearActions() list -

pq:
image

👍

src/io/flutter/view/FlutterView.java
jacob314: Correct. It is possible in the future we'll want to add a way to exit inspect mode on the device.

PR #1281, opened 2017-08-10
Add module for AS plugin

stevemessick:
@pq @devoncarew Code for the Android Studio version of the Flutter plugin.

Adds `New Flutter Project` to the `File` menu. I don't think there's any easy way to change the new project wizard to handle Flutter, since Flutter is not specific to Android. The current UI isn't too pretty, but it is specific to Flutter. We might want to entirely re-write this. The next page is the standard Flutter configuration page, with all the latest improvements.

screen shot 2017-08-10 at 2 48 45 pm

The new module wizard is fully customized to create Flutter modules using its standard UI.

screen shot 2017-08-10 at 2 49 11 pm

screen shot 2017-08-10 at 2 49 22 pm

FlutterModuleModel has a TODO, which is mostly completed.

The build mechanism needs to be extended to handle this.


flutter-studio/src/io/flutter/module/FlutterDescriptionProvider.java
```diff
@@ -0,0 +1,62 @@
...
+    @Nullable
+    @Override
+    public Icon getIcon() {
+      return FlutterIcons.AndroidStudioNewModule;
+    }
+
+    @NotNull
+    @Override
+    public String getName() {
+      return FlutterBundle.message("module.wizard.package_title");
```
pq: FWIW I think we've been leaning towards inlining strings but I don't feel strongly.

flutter-studio/src/io/flutter/module/FlutterModuleModel.java
```diff
@@ -0,0 +1,150 @@
...
+    if (myBuilder.commitModule(getProject(), null) == null) {
+      // A message was displayed to the user, so just clean up.
+      cleanupAfterValidationFailure();
+    }
+  }
+
+  private void cleanupAfterValidationFailure() {
+    String filePath = myModuleContentRoot.get().trim();
+    //noinspection ResultOfMethodCallIgnored
+    new File(filePath).delete();
```
pq: I was going to say we should talk to the virtual filesystem to mediate this but on second thought that's probably not a good idea since there may be thread access issues.  In any event, the IDE has no business knowing about this file anyway. Which is just to say, feel free to ignore. 👍

flutter-studio/resources/META-INF/studio-contribs.xml
```diff
@@ -0,0 +1,18 @@
...
+<idea-plugin>
+
+  <extensions defaultExtensionNs="com.android">
+    <moduleDescriptionProvider implementation="io.flutter.module.FlutterDescriptionProvider"/>
+  </extensions>
+
+  <actions>
+
+    <action id="flutter.NewProject" class="io.flutter.actions.FlutterNewProjectAction"
+            text="New Flutter Project..."
```
stevemessick: That would be inconsistent with AS and IDEA. I just did a quick (and possibly incomplete) search and the only menu items I found that use the ellipsis character are in the Flutter plugin :)

flutter-studio/src/io/flutter/module/FlutterDescriptionProvider.java
stevemessick: I'll leave that TODO item active for now. I'll probably do it soon-ish.

stevemessick:
Thanks! I'll merge this Monday.

flutter-studio/resources/META-INF/studio-contribs.xml
pq: ACK! 😊

flutter-studio/flutter-studio.iml
```diff
@@ -0,0 +1,27 @@
...
+  <component name="NewModuleRootManager" inherit-compiler-output="true">
+    <exclude-output />
+    <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/src" isTestSource="false" />
+      <sourceFolder url="file://$MODULE_DIR$/testSrc/unit" isTestSource="true" />
+      <sourceFolder url="file://$MODULE_DIR$/resources" type="java-resource" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="module" module-name="adt-branding" />
```
devoncarew: Is this where the Android Studio APIs we're using live?

flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
```diff
@@ -0,0 +1,60 @@
...
+import com.intellij.ide.projectWizard.ProjectTypeStep;
+import com.intellij.ide.util.projectWizard.ModuleWizardStep;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.roots.ui.configuration.ModulesProvider;
+import com.intellij.ui.CollectionListModel;
+import com.intellij.ui.components.JBList;
+import com.intellij.util.ReflectionUtil;
+
+public class FlutterNewProjectAction extends AnAction {
```
devoncarew: I think we should add a TODO in this class that we'll eventually want to move to a mechanism that just shows our wizard, w/o a list on the left.

flutter-studio/flutter-studio.iml
stevemessick: Some, yes. Others in adt-ui, android, wizard-model and possibly other modules. AS is distributed across many modules. I included dependencies on those that were needed to build, but it is possible that some are no longer needed. I should check that.

flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
stevemessick: I will add a TODO in the next update. Thanks for taking a look.

PR #1287, opened 2017-08-16
additional changes to the reload on save workflow

devoncarew:
additional changes to the reload on save workflow:
- send analytics information about what experimental features users have enabled (reload on save, improved memory view)
- ask the user for feedback the first time they disable reload on save
- provide info about reload-on-save reload starts and stops as balloons on the run/debug view, instead of as inline tooltips in the editor. The inline tooltips were a bit too distracting as you were typing, and made the UI seem a bit jerky.
- some refactoring of the code for how we persist user settings

screen shot 2017-08-16 at 12 58 56 pm

@stevemessick

### DIFF
--- a/src/io/flutter/FlutterInitializer.java
+++ b/src/io/flutter/FlutterInitializer.java
@@ -26,6 +26,7 @@ import io.flutter.pub.PubRoot;
 import io.flutter.run.FlutterReloadManager;
 import io.flutter.run.FlutterRunNotifications;
 import io.flutter.run.daemon.DeviceService;
+import io.flutter.settings.FlutterSettings;
 import io.flutter.utils.FlutterModuleUtils;
 import io.flutter.view.FlutterViewFactory;
 import org.jetbrains.annotations.NotNull;
@@ -44,8 +45,6 @@ public class FlutterInitializer implements StartupActivity {
   private static final String analyticsClientIdKey = "io.flutter.analytics.clientId";
   private static final String analyticsOptOutKey = "io.flutter.analytics.optOut";
   private static final String analyticsToastShown = "io.flutter.analytics.toastShown";
-  private static final String verboseLoggingKey = "io.flutter.verboseLogging";
-  private static final String memoryDashboardKey = "io.flutter.memoryDashboard";
 
   private static Analytics analytics;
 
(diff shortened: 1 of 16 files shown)


src/io/flutter/run/FlutterReloadManager.java
```diff
@@ -136,29 +171,35 @@ private void handleSaveAllNotification(AnActionEvent event) {
       return;
     }
 
-    if (hasErrors(project, module, editor.getDocument())) {
-      showNotification(editor, "Reload not performed", true);
-      showAnalysisNotification("Analysis issues found");
-    }
-    else {
-      final LightweightHint hint = showNotification(editor, "Reloading…", false);
+    // Add an arbitrary 125ms delay to allow analysis to catch up.
```
stevemessick: 125 or 100? Consider adding a TODO for tuning the delay.

src/io/flutter/FlutterInitializer.java
```diff
@@ -174,7 +155,7 @@ public void runActivity(@NotNull Project project) {
 
       final Notification notification = new Notification(
         Analytics.GROUP_DISPLAY_ID,
-        FlutterBundle.message("flutter.analytics.notification.title"),
+        "Welcome to Flutter!",
```
stevemessick: It is unusual to see strings being inlined from a bundle.

devoncarew: Phil, Brian S. and I had a discussion a few months ago, and generally agreed that the overhead of maintaining the i18n file doesn't really pay off until we start actually localizing the tool. So, it's currently dealer's choice whether each PR choose to i18n strings, or put them in the code raw.

For this case, I thought the shorter string was a bit clearer -

src/io/flutter/run/FlutterReloadManager.java
devoncarew: will do

InMatrix:
Can we add a "Feedback" link next to the preference of Save on Reload? 

image

The toaster shown in this PR is great, but we'd like to capture feedback from users who might have mixed feelings about the feature in addition to users who have decided to turn it off.

devoncarew:
> Can we add a "Feedback" link next to the preference of Save on Reload?

Sounds good! Will do.

devoncarew:
screen shot 2017-08-16 at 2 57 02 pm

PR #1292, opened 2017-08-23
Build AS

stevemessick:
@devoncarew @pq

A few changes:
- Rewrite module definition page for the AS new module wizard.
- Update build script.
- Added TODOs

It still requires a special build command to create a plugin that works in Android Studio (see the comment in build.xml) but I think we can eventually get to a single artifact. The one for AS also works in IntelliJ, so we already have something that can be distributed, if needed. If you never import the flutter-studio module you shouldn't need to have any form of AS around, either.


flutter-intellij-community.iml
```diff
@@ -12,10 +12,6 @@
       <sourceFolder url="file://$MODULE_DIR$/third_party/intellij-plugins-dart/testSrc" isTestSource="true" />
       <excludeFolder url="file://$MODULE_DIR$/.pub" />
       <excludeFolder url="file://$MODULE_DIR$/build" />
-      <excludeFolder url="file://$MODULE_DIR$/packages" />
```
stevemessick: I made a special commit (Fix IML) to ensure these lines did not get removed. I don't know how they got removed. Are they really necessary?

devoncarew: I think this churn goes away with the 2017.2 version of the Dart plugin. Shouldn't matter either way.

flutter-studio/src/io/flutter/module/FlutterPackageStep.java
```diff
@@ -9,51 +9,86 @@
...
-  private ModuleNameLocationComponent myModuleNameLocationComponent;
+  private JPanel myModulePanel;
+  private JTextField myModuleName;
+  private TextFieldWithBrowseButton myModuleContentRoot;
+  private TextFieldWithBrowseButton myModuleFileLocation;
+  private boolean myModuleNameChangedByUser = false;
+  private boolean myModuleNameDocListenerEnabled = true;
+  private boolean myContentRootChangedByUser = false;
+  private boolean myContentRootDocListenerEnabled = true;
+  private boolean myImlLocationChangedByUser = false;
```
devoncarew: Are these for implementing the 'advanced' section of the wizard page? We need to inline them because the form definition is in a different module?

We could also choose to not have an advanced section - most (all?) users won't use it, and that's a lot of code we then don't have to maintain.

resources/META-INF/plugin.xml
```diff
@@ -261,6 +261,9 @@
 
   <!-- Contributes IDEA-specific features and implementations. -->
   <depends optional="true" config-file="idea-contribs.xml">com.intellij.modules.java</depends>
+  <!-- Contributes Android Studio-specific features and implementations. -->
+  <!-- depends optional="true" config-file="studio-contribs.xml">com.intellij.modules.androidstudio</depends -->
```
devoncarew: Do we need this commented out line, or is it covered by the below line?

flutter-studio/src/io/flutter/module/FlutterPackageStep.java
stevemessick: This is main page content, not the advanced section.

resources/META-INF/plugin.xml
stevemessick: The commented-out line is the documented way to depend on android studio. It doesn't work for 3.0 beta. It may work in the future; I left the comment so we don't have to look up the name again.

flutter-studio/src/io/flutter/module/FlutterPackageStep.java
stevemessick: To clarify, I did not change the UI from what we had before. I just reimplemented it to not use an embedded form. It needs work, eg the new options like language choices should be added.

devoncarew: OK, sounds like this is not from the 'More settings' expandable area (which I had originally thought).

PR #1294, opened 2017-08-24
Build zip file for AS, jar for others

stevemessick:
@devoncarew @pq Changes to the build script to conditionally compile the flutter-studio module for AS and produce a ZIP file in that case only.


stevemessick:
Good news: the simplification means it is already getting run! Check this out:
https://travis-ci.org/flutter/flutter-intellij/jobs/268070121

devoncarew:
🎉 awesome!

PR #1296, opened 2017-08-25
Switch to Project view after creating a Flutter project

stevemessick:
@devoncarew As requested, switch to the Project view after creating a new Flutter project in AS.

It would be nice, I think, to autoscroll the project tree and select the file that is open in the editor. Haven't figured out how to do that yet.

### DIFF
--- a/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
+++ b/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
@@ -35,6 +35,24 @@ public class FlutterNewProjectAction extends AnAction {
     super("New Flutter Project...");
   }
 
+  private static void createNewProject(Project projectToClose, AbstractProjectWizard wizard) {
+    final boolean proceed = ProgressManager.getInstance().runProcessWithProgressSynchronously(() -> {
+      ProjectManager.getInstance().getDefaultProject(); //warm up components
+    }, ProjectBundle.message("project.new.wizard.progress.title"), true, null);
+    if (!proceed) return;
+    if (!wizard.showAndGet()) {
+      return;
+    }
+
+    final Project newProject = NewProjectUtil.createFromWizard(wizard, projectToClose);
+    StartupManager.getInstance(newProject).registerPostStartupActivity(() -> {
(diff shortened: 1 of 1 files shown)


flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
```diff
@@ -51,11 +61,28 @@ public void actionPerformed(AnActionEvent e) {
         }
       }
     }
-    NewProjectUtil.createNewProject(getEventProject(e), wizard);
+    createNewProject(getEventProject(e), wizard);
```
devoncarew: Perhaps a comment explaining where the the inlined `createNewProject` came from?

flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
```diff
@@ -51,11 +61,28 @@ public void actionPerformed(AnActionEvent e) {
...
+  private static void createNewProject(Project projectToClose, AbstractProjectWizard wizard) {
+    final boolean proceed = ProgressManager.getInstance().runProcessWithProgressSynchronously(() -> {
+      ProjectManager.getInstance().getDefaultProject(); //warm up components
+    }, ProjectBundle.message("project.new.wizard.progress.title"), true, null);
+    if (!proceed) return;
+    if (!wizard.showAndGet()) {
+      return;
+    }
+
+    final Project newProject = NewProjectUtil.createFromWizard(wizard, projectToClose);
```
devoncarew: A comment here saying that we're the `project` view (as opposed to the default `android` view) in order for Dart resources to be visible at project creation?

flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
stevemessick: Done

flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
stevemessick: Good idea, done. Also, reformat, reorganize, optimize imports. But no code changes so I'll just merge it.

PR #1297, opened 2017-08-28
measure users using reload vs restart

devoncarew:
- measure the number of sessions using reload vs restart in their workflow (at the end of an app run, the session was a 'reload' session if they reloaded at least once; a 'restart' session if they restarted at least once; and otherwise, we record 'none' for that session)
- fix a bug where some restart events were recorded as reload (this only happened if the user had hit the reload button with the shift key pressed)

@stevemessick

### DIFF
--- a/src/io/flutter/run/daemon/FlutterApp.java
+++ b/src/io/flutter/run/daemon/FlutterApp.java
@@ -51,6 +51,9 @@ public class FlutterApp {
   private @Nullable String myBaseUri;
   private @Nullable ConsoleView myConsole;
 
+  private int reloadCount;
+  private int restartCount;
+
   /**
    * Non-null when the debugger is paused.
    */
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -143,17 +146,24 @@ public static FlutterApp start(@NotNull ExecutionEnvironment env,
...
+    final FlutterApp app = new FlutterApp(project, module, mode, device.deviceId(), process, env, api);
+
     process.addProcessListener(new ProcessAdapter() {
       @Override
       public void processTerminated(ProcessEvent event) {
         LOG.info(analyticsStop + " " + project.getName() + " (" + mode.mode() + ")");
         FlutterInitializer.sendAnalyticsAction(analyticsStop);
+
+        // Send analytics about whether this session used the reload workflow, the restart workflow, or neither.
+        final String workflowType = app.reloadCount > 0 ? "reload" : (app.restartCount > 0 ? "restart" : "none");
```
stevemessick: What if the user does both reload & restart? Either we don't care, and it will be recorded as a reload. Or we do care, and we need a fourth identifier.

devoncarew: The way we're measuring _is_ a bit coarse. What we really want to know is the % of people using hot reload in their workflow (and even more specifically, the % that aren't, and then try and improve that). This counts a reload+restart in one session as a 'reload' workflow, but I think that's fine from the POV of what we really want to measure.

Issue #1291, opened 2017-08-23
Support creating a new 'package' project


mit-mit:
The new option for selecting the project type should be extended with the new 'package' type:

https://github.com/flutter/flutter/pull/11105

```
-t, --template=<type>          Specify the type of project to create.

          [app]                
             (default) Generate a Flutter application.
          [package]            
             Generate a shareable Flutter project containing modular Dart code.
          [plugin]             
             Generate a shareable Flutter project containing an API in Dart code
             with a platform-specific implementation for Android, for iOS code, or for both.
```

mit-mit:
cc @devoncarew

devoncarew:
For FYI, what the flutter tool currently supports:

```
[app]                (default) Generate a Flutter application.
[package]            Generate a shareable Flutter project containing modular Dart code.
[plugin]             Generate a shareable Flutter project containing an API in Dart code
                               with a platform-specific implementation for Android, for iOS code, or for both.
```

devoncarew:
Note that the options above are a bit confusing.

PR #1299, opened 2017-08-28
Add Package option

stevemessick:
@devoncarew @pq 

Add an option to create packages, in addition to apps and plugins.

The help text could be improved (in another PR).

fix https://github.com/flutter/flutter-intellij/issues/1291

### DIFF
--- a/src/io/flutter/FlutterBundle.properties
+++ b/src/io/flutter/FlutterBundle.properties
@@ -93,6 +93,7 @@ flutter.pop.frame.action.description=Pop the current frame off the stack
 flutter.module.create.settings.type.label=Project type:
 flutter.module.create.settings.type.application=Application
 flutter.module.create.settings.type.plugin=Plugin
+flutter.module.create.settings.type.package=Package
 flutter.module.create.settings.type.tip=Select a project type.
 flutter.module.create.settings.radios.org.label=Organization:
 flutter.module.create.settings.org.default_text=com.yourcompany
@@ -117,7 +118,8 @@ flutter.module.create.settings.help.org.label=Organization:
 flutter.module.create.settings.help.org.description=Enter the organization domain name in reverse domain notation.
 flutter.module.create.settings.help.project_type.label=Project type:
 flutter.module.create.settings.help.project_type.description.app=Select an "Application" when building for end users.
-flutter.module.create.settings.help.project_type.description.plugin=Select a "Plugin" when building system hooks for developers.
+flutter.module.create.settings.help.project_type.description.plugin=Select a "Plugin" when exposing an Android or iOS API for developers.
+flutter.module.create.settings.help.project_type.description.package=Select a "Package" when creating a cross-platform component, like a new Widget.
 
 module.wizard.package_title=Flutter Package
 module.wizard.package_description=Creates a Flutter package
--- a/src/io/flutter/module/FlutterProjectType.java
+++ b/src/io/flutter/module/FlutterProjectType.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.module;
+
+import io.flutter.FlutterBundle;
+
+public enum FlutterProjectType {
+  APP(FlutterBundle.message("flutter.module.create.settings.type.application"), "app"),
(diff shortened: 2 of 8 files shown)


devoncarew:
@stevemessick, lgtm, thanks!

Can you take a screenshot of the wizard page and include in the PR? @mit-mit can suggest wording changes if necessary post-commit, and we can follow up.

stevemessick:
Here's a screen shot of the relevant section:
screen shot 2017-08-28 at 5 14 13 pm
@mit-mit If you have suggestions for improvement to the help text I'd like to know it. I think including some kind of example would be good, since the difference between plugin and package is subtle.

src/io/flutter/FlutterBundle.properties
```diff
@@ -117,7 +118,8 @@ flutter.module.create.settings.help.org.label=Organization:
 flutter.module.create.settings.help.org.description=Enter the organization domain name in reverse domain notation.
 flutter.module.create.settings.help.project_type.label=Project type:
 flutter.module.create.settings.help.project_type.description.app=Select an "Application" when building for end users.
-flutter.module.create.settings.help.project_type.description.plugin=Select a "Plugin" when building system hooks for developers.
+flutter.module.create.settings.help.project_type.description.plugin=Select a "Plugin" when exposing an Android or iOS API for developers.
```
mit-mit: I would really like to keep these "help" strings consistent with [`flutter --help create`](https://github.com/flutter/flutter/blob/41378e20322b9fc4c317bfd1b4c33beeafcd6981/packages/flutter_tools/lib/src/commands/create.dart#L45) and the [website docs](https://flutter.io/developing-packages/#types).

Also, `when creating a cross-platform component` is confusing; a plugin package can also be cross-platform (i.e. support both iOS and Android).

But would love suggestions for how to improve the flutter/website ones also!

devoncarew:
Here's the cli ones:

```
app:
  (default) Generate a Flutter application.
package:
  Generate a shareable Flutter project containing modular Dart code.
plugin:
  Generate a shareable Flutter project containing an API in Dart code
  with a platform-specific implementation for Android, for iOS code, or for both.
```

Perhaps `package` changes to `Generate a shareable Flutter project containing modular Dart code (for example, a reusable Widget).`

If we want the same descriptions in the wizard and the cli, we'll likely need a shorter description for `plugin`. Even the `package` desc. above may be too long.

stevemessick:
What if we change the UI entirely? It seems to me we have two options: an application or a component of an application. The component may or may not include platform-specific code.

Instead of using a drop-down for `Project type` we change that two two radio buttons. The top one is `Application`, the bottom one is `Plugin`. Underneath `Plugin` indented a bit is a check box. The check box is only enabled when the `Plugin` is selected. It either enables or disables generation of iOS/Android code (not sure which should be default nor what to call the check box). If vertical space is an issue the check box could be to the right of `Plugin`.

The help text would of course be updated to match the UI.

Furthermore, and I should have thought of this before, the radio buttons for language selection should be disabled if native code is not being generated. Ideally, they would be moved closer to the project type selection area. Might be better to move `Project type` down below the text fields.

As a reminder, this all has to be reimplemented anyway to work in the new framework that Android Studio uses, so I'm not really proposing a lot of new work. Just some new up-front design :) And I'm intending to make it as reusable as possible. I'm pretty sure an embedded form can be used but the validation logic is framework-specific so that can't be shared.

When `Application` is select the language selection area is enabled. When `Plugin` is selected the language selection area is only enabled if native code is being generated. So, the language selection won't be indented any more than it is, but it should be visually adjacent to the check box that disables it.

For reference, here's a screen shot of the current page:
screen shot 2017-08-28 at 5 13 59 pm

devoncarew:
So, something like:

- [ ] Application
- [ ] Package
  - [ ] Pluuin

(but with radio buttons for the first level)

stevemessick:
I was thinking of putting `Plugin` where you put `Package`. Actually, does anyone else think "package" is confusing to Dart programmers? (If Flutter devs already had this discussion I'll drop it.)

Maybe more like:
- Application
- Plugin
  [ ] Generate native code to hook up a new platform API

(Pretend the first two have radio buttons.)
Perhaps because I've been writing plugins for so long, I tend to think "plugin" is more like "component" than "package" is.

stevemessick:
Hmm, I forgot that `Organization` is only needed for native code...

I think this will work a lot better in Android Studio if we have three separate top-level module types.

Maybe it isn't worth the trouble to redesign the UI at this point.

PR #1316, opened 2017-09-05
Update sources to beta 4; add additional fields and simplify

### LINKED ISSUE #1304: Organization and programming language options not relevant for package project type
1. File -> New project
1. Select Flutter + Next
1. Select 'Package' as the project type

=> UI for programming language and organization is not relevant. Should be hidden or disabled.

stevemessick:
@devoncarew @pq 

Lotsa stuff
- Add option to create plugin module type in AS
- Add additional settings to package and plugin for AS
- Add disablement of settings to IJ plugin for packages (fixes #1304)
- Reorganize settings to make disabling more sensible
- Refactor additional settings support code to be less framework-dependent
- Update to AS beta 4 sources

Still need to differentiate icons for AS module types. Package and plugin use the same one now.
The help link points to the package and plugin page, rather than getting started. But only for AS modules; the project page link still uses the getting starting page.

Screen shot of package page in the New Module wizard:
screen shot 2017-09-05 at 1 40 09 pm

Screen shot of plugin page in the New Module wizard:
screen shot 2017-09-05 at 1 40 29 pm


devoncarew:
Looking at this now; one quick question though - is there room under the title (`Configure the new Flutter plugin`) for a description? That might be a natural place to put the description of what a plugin / package is.

stevemessick:
We could add a label above the "Module name:" label. None of the other
Android modules have anything like that. They don't have any sort of
descriptive text at all.

I put it at the bottom to keep some conformity with AS and because once
people have created one or two they won't need to see it. Personally, I'd
prefer it at the top and I don't know why AS doesn't have any descriptions.

devoncarew:
If there's a good way to move the description either into the page title itself (under `'New Flutter Package'`) or under `'Configure the new Flutter Package'`, that might be good. The field descriptions - how to use the individual fields like organization and module name - could move under the respective fields.

lgtm w/o this change however, just food for thought

stevemessick:
I think that's a good idea.

PR #1320, opened 2017-09-07
New icons for project and plugin module types

stevemessick:
@devoncarew @pq 

Here's what they look like in context.

screen shot 2017-09-07 at 10 26 35 am

Not great, but it will do for now (hopefully will be done better before release). The package icon has a small Dart icon added to indicate pure Dart code. The plugin icon has a Material Design "devices" icon added to indicate device-specific code.


devoncarew:
lgtm

Do we control the order of the entries in the module list, or are they lexical? If we do control it, it would be nice for the plugin entry to show first.

pq:
LGTM!

stevemessick:
The sorting appears to be lexical. I couldn't find a way to change it.

PR #1326, opened 2017-09-11
use distinct icons for the android and ios source folders

devoncarew:
- use distinct icons for the android and ios source folders

@pq, @stevemessick


devoncarew:
Generally, just to visually distinguish it from the other directories - to say, 'native bits here'. Happy to try other icons - I'm not sure any of them will have much meaning to users.

PR #1338, opened 2017-09-15
Project

stevemessick:
@devoncarew @pq 

This isn't finished, but it does generate a new project. I haven't tested it much yet.

There's a to-do list in FlutterProjectModel, and a few TODO's sprinkled here and there. I plan to keep working on it.

I updated the TODO format, as requested, and added a progress indicator. That required a bit more refactoring in "standard" plugin code.

The first page of the wizard changes slightly depending on which icon is selected. I'll only show it once.

screen shot 2017-09-15 at 2 24 49 pm

Application 1
screen shot 2017-09-15 at 2 25 04 pm

Application 2
screen shot 2017-09-15 at 2 25 17 pm

Plugin 1
screen shot 2017-09-15 at 2 25 34 pm

Plugin 2
screen shot 2017-09-15 at 2 30 27 pm

Package 1-and-only
screen shot 2017-09-15 at 2 30 45 pm

Progress indicator (same for all)
screen shot 2017-09-15 at 2 31 20 pm

### DIFF
--- a/flutter-studio/src/io/flutter/project/ChoseProjectTypeStep.java
+++ b/flutter-studio/src/io/flutter/project/ChoseProjectTypeStep.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright 2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.project;
+
+import com.android.tools.adtui.ASGallery;
+import com.android.tools.adtui.util.FormScalingUtil;
+import com.android.tools.idea.npw.module.ModuleGalleryEntry;
+import com.android.tools.idea.wizard.model.ModelWizard;
+import com.android.tools.idea.wizard.model.ModelWizardStep;
+import com.google.common.collect.Lists;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.ui.VerticalFlowLayout;
+import com.intellij.ui.JBColor;
+import com.intellij.ui.components.JBList;
+import com.intellij.ui.components.JBScrollPane;
+import com.intellij.util.IconUtil;
+import com.intellij.util.containers.HashMap;
+import io.flutter.module.FlutterDescriptionProvider;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.accessibility.AccessibleContext;
+import javax.swing.*;
+import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+import static com.android.tools.idea.wizard.WizardConstants.DEFAULT_GALLERY_THUMBNAIL_SIZE;
+import static org.jetbrains.android.util.AndroidBundle.message;
+
+public class ChoseProjectTypeStep extends ModelWizardStep<FlutterProjectModel> {
+  private static final Logger LOG = Logger.getInstance(ChoseProjectTypeStep.class.getName());
+
+  private final List<ModuleGalleryEntry> myModuleGalleryEntryList;
+  private final JComponent myRootPanel;
+  private JLabel[] helpLabels;
+
+  private ASGallery<ModuleGalleryEntry> myProjectTypeGallery;
+  private Map<ModuleGalleryEntry, FlutterProjectStep> myModuleDescriptionToStepMap;
+
+  public ChoseProjectTypeStep(@NotNull FlutterProjectModel model) {
+    this(model, FlutterDescriptionProvider.getGalleryList(true));
+  }
+
+  private ChoseProjectTypeStep(@NotNull FlutterProjectModel model, @NotNull List<ModuleGalleryEntry> moduleGalleryEntries) {
+    super(model, "New Flutter Project");
+
+    myModuleGalleryEntryList = moduleGalleryEntries;
+    myRootPanel = new JPanel();
+    myRootPanel.setLayout(new VerticalFlowLayout());
+    myRootPanel.add(createGallery());
+
+    final JPanel section = new JPanel();
+    section.setLayout(new VerticalFlowLayout());
+    helpLabels = new JLabel[moduleGalleryEntries.size()];
+    int idx = 0;
+    for (ModuleGalleryEntry entry : moduleGalleryEntries) {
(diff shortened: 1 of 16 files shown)


src/io/flutter/FlutterBundle.properties
```diff
@@ -123,15 +123,30 @@ flutter.module.create.settings.help.project_type.label=Project type:
 flutter.module.create.settings.help.project_type.description.app=Select an "Application" when building for end users.
 flutter.module.create.settings.help.project_type.description.plugin=Select a "Plugin" when exposing an Android or iOS API for developers.
 flutter.module.create.settings.help.project_type.description.package=Select a "Package" when creating a pure Dart component, like a new Widget.
+flutter.module.create.settings.help.type.application=An "application" is a stand-alone app that can be installed and run on a mobile device.
 flutter.module.create.settings.help.type.package=A "package" is a pure-Dart component, like a new Widget, with no native code.
 flutter.module.create.settings.help.type.plugin=A "plugin" is used to expose an Android or iOS API for developers.
 
+module.wizard.language.name_kotlin=Use Kotlin for Android code (defaults to Java)
```
branflake2267: I might suggest "Use the Kotlin language for Android. (defaults to the Java language)"

src/io/flutter/FlutterBundle.properties
```diff
@@ -123,15 +123,30 @@ flutter.module.create.settings.help.project_type.label=Project type:
...
+module.wizard.language.name_swift=Use Swift for iOS code (defaults to Objective-C)
```
branflake2267: I would suggest, "Use the Swift language for iOS. (defaults to Objective-C)"

src/io/flutter/FlutterBundle.properties
```diff
@@ -123,15 +123,30 @@ flutter.module.create.settings.help.project_type.label=Project type:
 flutter.module.create.settings.help.project_type.description.app=Select an "Application" when building for end users.
 flutter.module.create.settings.help.project_type.description.plugin=Select a "Plugin" when exposing an Android or iOS API for developers.
 flutter.module.create.settings.help.project_type.description.package=Select a "Package" when creating a pure Dart component, like a new Widget.
+flutter.module.create.settings.help.type.application=An "application" is a stand-alone app that can be installed and run on a mobile device.
```
branflake2267: I would suggest removing "mobile" to make it more general. Maybe a device means it's a chrome device that runs android.

devoncarew:
> This isn't finished

Gotcha. The screenshots look fantastic!

src/io/flutter/FlutterBundle.properties
stevemessick: Using "Android/iOS code" links the button to "platform-specific code" in the description at the top of the page.

src/io/flutter/FlutterBundle.properties
stevemessick: I'd like a way to indicate that if Kotlin is selected, Java can still be used. That's what the paren phrase is there for but I don't think it's quite right.

src/io/flutter/FlutterBundle.properties
stevemessick: Good point. I noticed that the flutter.io site has "mobile device" a few times, but "device" is used far more often.

stevemessick:
@mit-mit might be interested, too.

src/io/flutter/FlutterBundle.properties
mit-mit: Actually, I'd suggest changing the type of control, I think this would be easier to understand with regular radio buttons:

```
Android platform: (*) Java          ( ) Kotlin
iOS platform:     (*) ObjectiveC    ( ) Swift
```

src/io/flutter/FlutterBundle.properties
mit-mit: I don't think we need to explain that; we can assume knowledge of the local platform of someone who is using this capability.

mit-mit:
This looks great! Some smaller comments above.

One general comment: Can we move the "Creating Flutter project progress prompt" into the wizard itself so that it would be a last step in that, and so that right after the wizard closes the project would be available for editing. Is that possible?

stevemessick:
@mit-mit About the progress bar: I'm not sure we can do what you suggest. I tried to do something like that and was unsuccessful. Rather than putting a lot of time into it I adopted the style used in the current plugin release. I agree, it would be better to do it as part of the page.

src/io/flutter/FlutterBundle.properties
stevemessick: Android Studio uses a check box to determine language. Wouldn't it be better to follow the same style as the host IDE?

stevemessick:
@devoncarew @pq ping! Could I get a review soon, please?

flutter-studio/src/io/flutter/project/ChoseProjectTypeStep.java
```diff
@@ -0,0 +1,171 @@
...
+  @NotNull
+  @Override
+  protected JComponent getComponent() {
+    return myRootPanel;
+  }
+
+  @NotNull
+  @Override
+  public Collection<? extends ModelWizardStep> createDependentSteps() {
+    List<ModelWizardStep> allSteps = Lists.newArrayList();
```
pq: Old chestnut: this is not declared `final`l but above, for example, the JPanel is.  I never really drank the declare locals final doctrine but I do like consistency.

Thoughts?

(And sorry if I'm missing context --- maybe best to chat in person?)

flutter-studio/src/io/flutter/project/ChoseProjectTypeStep.java
```diff
@@ -0,0 +1,171 @@
...
+    myProjectTypeGallery = new ASGallery<ModuleGalleryEntry>(
+      JBList.createDefaultListModel(),
+      image -> image == null ? null : image.getIcon() == null ? null : IconUtil.toImage(image.getIcon()),
+      label -> label == null ? message("android.wizard.gallery.item.none") : label.getName(), DEFAULT_GALLERY_THUMBNAIL_SIZE,
+      null
+    ) {
+
+      @Override
+      public Dimension getPreferredScrollableViewportSize() {
+        // The default implementations assigns a height as tall as the screen.
```
pq: implementations => implementation

pq:
Sorry for the slow review!

flutter-studio/src/io/flutter/project/ChoseProjectTypeStep.java
stevemessick: I'm happy to remove 'final' from the local var.

PR #1339, opened 2017-09-18
Open correct file for plugin and packages; generate run config for pl…

stevemessick:
@devoncarew @pq 

While working on project generation for Android Studio I discovered some problems common to all versions of the plugin: no run config was being generated for plugin, and neither plugins nor projects were opening a file in the editor. This fixes both problems. It is independent of the work I'm doing for Android Studio (i.e. modifies the main plugin sources).

### DIFF
--- a/src/io/flutter/pub/PubRoot.java
+++ b/src/io/flutter/pub/PubRoot.java
@@ -292,11 +292,46 @@ public class PubRoot {
   }
 
   /**
+   * Returns a file in lib if it exists.
+   */
+  @Nullable
+  public VirtualFile getFileToOpen() {
+    VirtualFile main = getLibMain();
+    if (main != null) {
+      return main;
+    }
+    if (lib != null) {
+      VirtualFile[] files = lib.getChildren();
+      if (files.length != 0) {
+        return files[0];
+      }
+    }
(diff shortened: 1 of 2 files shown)


src/io/flutter/pub/PubRoot.java
```diff
@@ -292,13 +292,48 @@ public VirtualFile getLib() {
   }
 
   /**
+   * Returns a file in lib if it exists.
+   */
+  @Nullable
+  public VirtualFile getFileToOpen() {
+    VirtualFile main = getLibMain();
```
pq: Make `main` final?

src/io/flutter/pub/PubRoot.java
```diff
@@ -292,13 +292,48 @@ public VirtualFile getLib() {
...
+    if (main != null) {
+      return main;
+    }
+    if (lib != null) {
+      VirtualFile[] files = lib.getChildren();
```
pq: And here (and below)?  (Or are we not following that convention anymore?)

stevemessick: I'll send you a new PR with 'final' added.

PR #1353, opened 2017-09-26
Add an Install SDK button to the new module wizard.

stevemessick:
@devoncarew @pq 

Adds the same style of Install SDK button to the Android Studio wizards as is found in the IntelliJ wizards.

InstallSdkAction needed somne refactoring to become reusable. Getting the event-based validation in the new wizard framework to work required adding a few calls that could change how it behaves w.r.t. the old wizards. I verified that it still works, but I didn't check that cancelling has exactly the same before-and-after behavior. But it seems reasonable. It may look worse than it is because it got reorganized during reformating. Everything is still there.

Also, this removes the error message that @mit-mit found inhibiting his progress.

Static screenshot:
screen shot 2017-09-26 at 1 35 13 pm

Action-packed screenshot:
screen shot 2017-09-26 at 1 32 35 pm

The validation error goes away once validation runs after the installation is finished, and the Next button is enabled.

### DIFF
--- a/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
+++ b/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
@@ -23,6 +23,8 @@ import io.flutter.sdk.FlutterSdkUtil;
 import org.jetbrains.android.sdk.AndroidSdkUtils;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.NoSuchElementException;
+
 public class FlutterNewProjectAction extends AnAction implements DumbAware {
   public FlutterNewProjectAction() {
     this("New Flutter Project...");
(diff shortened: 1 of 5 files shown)


flutter-studio/src/io/flutter/project/FlutterProjectStep.java
```diff
@@ -94,10 +115,38 @@ public FlutterProjectStep(FlutterProjectModel model, String title, Icon icon, Fl
...
+
+    myInstallActionLink.setIcon(myInstallSdkAction.getLinkIcon());
+    myInstallActionLink.setDisabledIcon(IconLoader.getDisabledIcon(myInstallSdkAction.getLinkIcon()));
+
+    myInstallActionLink.setText(myInstallSdkAction.getLinkText());
+
+    //noinspection unchecked
+    myInstallActionLink.setListener((label, linkUrl) -> myInstallSdkAction.actionPerformed(null), null);
+
+    myProgressText.setFont(UIUtil.getLabelFont(UIUtil.FontSize.NORMAL).deriveFont(Font.ITALIC));
```
pq: I might be mis-seeing but the screenshot makes it look like the progress has a white background...  Can we / should we set it here to be consistent with the wizard background?

stevemessick: Perhaps, probably could. But isn't it the same in the IntelliJ wizards?

stevemessick: Another thing I wonder about is the visibility of the Install SDK button. I didn't even notice until I was making screenshots that it disappears during the download. That makes more sense in the original layout, I think. But, like progress bar location, it is something we can iterate on.

stevemessick: I checked the text box. It has the same appearance in both AS and IJ versions of the wizard. Using the Default theme it has the (questionable) appearance you see in the screenshot. Using the Darkula theme it has a dark charcoal-grey background, as does the dialog box.

pq: 👍  Thanks for following up!

flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
```diff
@@ -39,23 +41,16 @@ public void update(@NotNull AnActionEvent e) {
...
     ModelWizard wizard = new ModelWizard.Builder()
       .addStep(new ChoseProjectTypeStep(model))
       .build();
     ModelWizardDialog dialog =
       new StudioWizardDialogBuilder(wizard, "Create New Flutter Project").setUseNewUx(true).build();
-    dialog.show();
+    try {
+      dialog.show();
+    } catch (NoSuchElementException ex) {
+      // This happens if no Flutter SDK is installed and the user cancels the FlutterProjectStep.
```
devoncarew: Perhaps log this, in the off chance that this is not the expected exception?

devoncarew:
Oops, just missed the merge

PR #1354, opened 2017-09-27
Disallow flutter_test module name

### LINKED ISSUE #1352: Add validation on 'flutter_test' module name
## Steps to Reproduce
1. Create a new Flutter project and set module name to 'flutter_test'.
Actual result: Process fails with error: 
A package may not list itself as a dependency.
flutter_test:
^^^^^^^^^^
pub get failed (65)

I understand that package 'flutter_test' already exists and automatically added as dev_dependencies.
So, it isn't an actually issue but validation on module name can improve user experience.

Expected result: Validation on 'flutter_test' module name with the same validation message as for 'flutter'.

## Version info
[√] IntelliJ IDEA Community Edition (version 2017.2)
    • Flutter plugin version 17.0
    • Dart plugin version 172.3757.55

stevemessick:
Fixes #1352 

Disallows 'flutter_test' as a module or project name.

### DIFF
--- a/src/io/flutter/FlutterConstants.java
+++ b/src/io/flutter/FlutterConstants.java
@@ -78,23 +78,25 @@ public class FlutterConstants {
   );
 
   // From: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/commands/create.dart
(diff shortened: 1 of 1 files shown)


stevemessick:
Thanks. Sorry I jumped the gun by merging the Install SDK PR before you reviewed it, but I'm happy to continue working on layout etc.

PR #1357, opened 2017-09-27
Initialize the SDK path cache

stevemessick:
@pq @devoncarew This fixes a problem that caused the SDK path text to be cleared if the `Install SDK` button was clicked, then the file chooser was cancelled.

### DIFF
--- a/flutter-studio/src/io/flutter/project/FlutterProjectStep.java
+++ b/flutter-studio/src/io/flutter/project/FlutterProjectStep.java
@@ -59,10 +59,10 @@ public class FlutterProjectStep extends SkippableWizardStep<FlutterProjectModel>
   private final ValidatorPanel myValidatorPanel;
   private final BindingsManager myBindings = new BindingsManager();
   private final ListenerManager myListeners = new ListenerManager();
+  private final InstallSdkAction myInstallSdkAction;
   private boolean hasEntered = false;
   private OptionalValueProperty<FlutterProjectType> myProjectType;
   private StringValueProperty myDownloadErrorMessage = new StringValueProperty();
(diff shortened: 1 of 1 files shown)


flutter-studio/src/io/flutter/project/FlutterProjectStep.java
```diff
@@ -279,6 +297,7 @@ protected void onEntering() {
     }
     myHeading.setText(heading);
     myDownloadErrorMessage.set("");
+    mySdkPathContent = (String)myFlutterSdkPath.getComboBox().getEditor().getItem();
```
devoncarew: just checking that `myFlutterSdkPath.getComboBox()` can't be generified in some way (to avoid the cast)?

stevemessick: Not that I know of. ComboboxWithBrowseButton doesn't define any type variables.

flutter-studio/src/io/flutter/project/FlutterProjectStep.java
```diff
@@ -188,6 +187,25 @@ private static String findProjectLocation(@NotNull String appName) {
     return new Validator.Result(Validator.Severity.ERROR, message);
   }
 
+  private static void ensureComboModelContainsCurrentItem(@NotNull final JComboBox comboBox) {
+    final Object currentItem = comboBox.getEditor().getItem();
+
+    boolean contains = false;
+    for (int i = 0; i < comboBox.getModel().getSize(); i++) {
```
pq: Maybe store model in a local and simplify this access and the two below?

PR #1359, opened 2017-09-27
Add a TODO

stevemessick:
@devoncarew Add a TODO, as promised to @pq


stevemessick:
I signed it!

stevemessick:
Anyone know what's up with the `cla:no` label?

stevemessick:
@mit-mit Something got lost and the googlebot is claiming I need to sign a CLA. First, do I need to do that? If so, individual or company?

devoncarew:
You don't need to sign it. This generally means that the email address on your commit doesn't match what google has on file. You may want to run some variant of `git log` locally to see if the email address looks right.

devoncarew:
If no, this may get you back on track: `git config --global user.email email@example.com`.

mit-mit:
There was a CLAbot problem yesterday; it should have been fixed now.

PR #1360, opened 2017-09-28
Refresh the VFS if the SDK is not found

stevemessick:
@devoncarew @pq

A newly-installed SDK might not be found if the virtual file system doesn't know about it. Since it can be installed without IDE assistance it seems best to make sure the VFS is updated before it claims there is no SDK in the given location.

### DIFF
--- a/src/io/flutter/sdk/FlutterSdk.java
+++ b/src/io/flutter/sdk/FlutterSdk.java
@@ -29,6 +29,7 @@ import io.flutter.settings.FlutterSettings;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.io.File;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -102,7 +103,10 @@ public static FlutterSdk getIncomplete(@NotNull final Project project) {
 
   @Nullable
   public static FlutterSdk forPath(@NotNull final String path) {
-    final VirtualFile home = LocalFileSystem.getInstance().findFileByPath(path);
+    VirtualFile home = LocalFileSystem.getInstance().findFileByPath(path);
+    if (home == null) {
+      home = LocalFileSystem.getInstance().refreshAndFindFileByIoFile(new File(path));
```
pq: This call worries me as I recall issues where it blocked the UI thread.  Having said that I'm racking my brain for more context.  Does this ring any bells?

pq: How's that for an unhelpful comment! 😉

stevemessick: Good point. This could be called from anywhere (I guess). I'll move the refresh code into the wizard.

PR #1361, opened 2017-09-28
Shorten labels to improve readability

stevemessick:
@devoncarew @pq

Reduce the complexity of the labels on the check boxes to make them easier to read.


pq:
Works for me but will defer to @mit-mit or @devoncarew who may have stronger thoughts.

mit-mit:
I would phrase this as 'Include Kotlin support for Android code'

stevemessick:
Done. PTAL

stevemessick:
Oops! Thanks for catching that. Updated.

PR #1364, opened 2017-09-29
Change default names

stevemessick:
@pq @devoncarew @mit-mit Uses more conservative default names for projects. They still illustrate the snake_case naming convention.

### DIFF
--- a/src/io/flutter/FlutterBundle.properties
+++ b/src/io/flutter/FlutterBundle.properties
@@ -134,19 +134,19 @@ module.wizard.app_title=Flutter Application
 module.wizard.app_description=Creates a Flutter application
 module.wizard.app_step_title=New Flutter Application
 module.wizard.app_step_body=Configure the new Flutter application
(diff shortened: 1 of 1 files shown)


src/io/flutter/FlutterBundle.properties
```diff
@@ -134,19 +134,19 @@ module.wizard.app_title=Flutter Application
 module.wizard.app_description=Creates a Flutter application
 module.wizard.app_step_title=New Flutter Application
 module.wizard.app_step_body=Configure the new Flutter application
-module.wizard.app_name=awesome_app
+module.wizard.app_name=flutter_app
```
devoncarew: `sample_app`? `my_app`?

stevemessick: @mit-mit suggested flutter_app. If you two agree it should be changed again then I'll be happy to change it :) For now, I'll merge this.

Issue #1347, opened 2017-09-21
fix the 'frameworks detected' notification


devoncarew:
We should address the 'frameworks detected' notification; this patch is a good starting point.

```
diff --git a/packages/flutter_tools/templates/create/projectName.iml.tmpl b/packages/flutter_tools/templates/create/projectName.iml.tmpl
index dafb00113..4a51e50ba 100644
--- a/packages/flutter_tools/templates/create/projectName.iml.tmpl
+++ b/packages/flutter_tools/templates/create/projectName.iml.tmpl
@@ -1,13 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module type="WEB_MODULE" version="4">
+  <component name="FacetManager">
+    <facet type="android" name="Android">
+      <configuration>
+        <option name="GEN_FOLDER_RELATIVE_PATH_APT" value="/../../{{projectName}}/gen" />
+        <option name="GEN_FOLDER_RELATIVE_PATH_AIDL" value="/../../{{projectName}}/gen" />
+        <option name="MANIFEST_FILE_RELATIVE_PATH" value="/../../{{projectName}}/AndroidManifest.xml" />
+        <option name="RES_FOLDER_RELATIVE_PATH" value="/../../{{projectName}}/res" />
+        <option name="ASSETS_FOLDER_RELATIVE_PATH" value="/../../{{projectName}}/assets" />
+        <option name="LIBS_FOLDER_RELATIVE_PATH" value="/../../{{projectName}}/libs" />
+        <option name="PROGUARD_LOGS_FOLDER_RELATIVE_PATH" value="/../../{{projectName}}/proguard_logs" />
+      </configuration>
+    </facet>
+  </component>
   <component name="NewModuleRootManager" inherit-compiler-output="true">
     <exclude-output />
     <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/gen" isTestSource="false" generated="true" />
       <excludeFolder url="file://$MODULE_DIR$/.idea" />
       <excludeFolder url="file://$MODULE_DIR$/.pub" />
       <excludeFolder url="file://$MODULE_DIR$/build" />
       <excludeFolder url="file://$MODULE_DIR$/packages" />
     </content>
+    <orderEntry type="jdk" jdkName="Android API {{androidSdkVersion}} Platform" jdkType="Android SDK" />
     <orderEntry type="sourceFolder" forTests="false" />
[...]

Issue #1370, opened 2017-10-04
Should open two files for plugin projects


stevemessick:
The .idea/workspace.xml file is primed to open editors on two files when a plugin project is created. If the project is created from the command line, then opened in IntelliJ, the two files are opened in editors.

If the project is created from the `New` menu item in the IDE then only one file is opened.

@alexander-doroshko does this sound at all familiar to you?

devoncarew:
Steve, I ran into some issues months ago that some similar. I believe theres a race condition on project creation and `flutter create`. If run from IntelliJ, I think it ends up using the metadata files that it creates in preference to the ones that `flutter create` does. So, when driven from IntelliJ, we don't get all (any?) of the settings that flutter create is trying to write; when run from the command line, and later opened in IntelliJ, we get all the settings. Does this sound like what you're seeing?

alexander-doroshko:
Yes, **existing project opening** (which effectively happens if you create a project from command line) and **new project creation** from the IDE are 2 different branches of code that is executed. On new project creation IDE used to think that it **writes** the project metadata files, not **reads** them. AFAIK scenario, when new project wizard calls 3rd party tool to generate project metadata files, has not been implied and therefore tested before.

stevemessick:
What Alex said :) When Android Studio (and probably any IDEA variant) creates a project it writes several files before starting the `flutter create` process. Flutter won't overwrite existing files, so the lines we add to workspace.xml that cause two files to be opened do not get written, since workspace.xml is one of the files created early.

There may be a way to initialize the project with two files open. I'll look into that. Someday. I don't think it is high priority.

stevemessick:
Fixed with https://github.com/flutter/flutter-intellij/pull/1375

PR #1375, opened 2017-10-05
Create projects to open as they would when created from the cmd line

stevemessick:
@pq @devoncarew @mit-mit 

This depends on https://github.com/flutter/flutter/pull/12391

Rather than allowing Android Studio to open a newly created project, this forces a reload so the files written by Flutter are used. AS writes a few files that just confuse its world before running `flutter create`. This gets rid of them before running flutter since flutter won't overwrite them.

This also adds an Android facet to apps and plugins.

Fixes: #1347, #1370

### DIFF
--- a/flutter-studio/src/io/flutter/project/FlutterProjectCreator.java
+++ b/flutter-studio/src/io/flutter/project/FlutterProjectCreator.java
@@ -5,7 +5,12 @@
  */
 package io.flutter.project;
 
+import com.android.builder.model.SourceProvider;
 import com.android.repository.io.FileOpUtils;
+import com.android.tools.idea.gradle.npw.project.GradleAndroidProjectPaths;
+import com.android.tools.idea.npw.project.AndroidSourceSet;
+import com.intellij.facet.FacetManager;
+import com.intellij.facet.ModifiableFacetModel;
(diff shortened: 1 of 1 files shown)


flutter-studio/src/io/flutter/project/FlutterProjectCreator.java
```diff
@@ -135,7 +193,7 @@ public ModuleWizardStep modifySettingsStep(@NotNull SettingsStep settingsStep) {
       }
     };
     builder.setName(myModel.projectName().get());
-    builder.setModuleFilePath(FileUtil.toSystemIndependentName(contentRoot) + "/" + moduleName + ModuleFileType.DOT_DEFAULT_EXTENSION);
+    builder.setModuleFilePath(FileUtilRt.toSystemIndependentName(contentRoot) + "/" + moduleName + ModuleFileType.DOT_DEFAULT_EXTENSION);
```
pq: Maybe we can just replace this with `toSystemIndependentName` since it's statically imported above?

stevemessick: Good catch. Next PR

Issue #1308, opened 2017-08-30
Opening Observatory is broken


mit-mit:
1. Create new Flutter app in IJ
1. Hit Run
1. Open the Flutter view
1. Click the 'Open Observatory timeline' button

=> a 'blank' browser window is opened

devoncarew:
@mit-mit, I'm not able to repro this, either run or debug, w/ iOS simulator, android devices, or the android emulator.

mit-mit:
Can someone else try also? I got a consistent repro.

devoncarew:
Does the address look well formed? Does the observatory web ui respond in general? And, can you get the the observatory timeline page when you flutter run from the command-line?

mit-mit:
It's not timeline specific, the general observatory button in the Console doesn't work either. The address is empty, so no. Also, I don't see the observatory URL printed in the Console.

mit-mit:
If I `flutter run` from the terminal, the URL is printed and it opens just fine.

devoncarew:
OK, that should be for the general observatory URL. From that page, can you navigate to the timeline page, and does the url for that match the one that IntelliJ it trying to open?

Is this a 127.0.0.1 / localhost thing?

devoncarew:
> The address is empty

Ok, that sounds like a IntelliJ platform bug. 

https://github.com/flutter/flutter-intellij/blob/4118d2df4b768c3ddf7ebd9b1b82db7d08d4a998/src/io/flutter/run/OpenObservatoryAction.java#L53

We're trying to open the url with any chrome browser (that IntelliJ knows about), and then fallback to any browser.

devoncarew:
@mit-mit, I don't think this is a regression in the plugin per se, so moving out of M17. We can patch the release if we do find an issue in the flutter plugin.

mit-mit:
This is still a 100% repro for me. Can you try in macOS against an iOS Simulator? I do not get an Observatory URL in the console output when running from IJ, and the Observatory button still launches an empty browser tab.

devoncarew:
I'll re-try, but I originally did try macos and ios simulator. What version of IntelliJ are you using? Do you see any obvious issues in your 'Web Browsers' pref page in IntelliJ?

> I do not get an Observatory URL in the console output when running from IJ

You won't - we filter this for less noise (user's can open Observatory other ways).

mit-mit:
I'm using IDEA Community 2017.2.1.

I don't see any issues in the browser preference.

[...]

PR #1385, opened 2017-10-09
use the default BrowserLauncher to open a url

devoncarew:
- use the default BrowserLauncher to open a url (fix https://github.com/flutter/flutter-intellij/issues/1308)

@pq @stevemessick


devoncarew:
Observatory works best in chrome...

PR #1394, opened 2017-10-12
Testing

stevemessick:
@devoncarew @pq 

Ready for review!

The labelFor tags are added to JLabel's in the form editor so you can easily find the associated JTextField.

I may remove FlutterCreateState if it isn't needed. I was going to hook it into the FlutterProjectCreator so tests could know when it was done, but the trick of waiting for the event queue to drain seems to be working.

The *big* classes were copied from the framework, and they have a class comment saying so. I used the same package FQN as the framework does for fixtures since most of them are package-private, but general enough to reuse.


devoncarew:
> No need to review yet

Gotcha, but one quick question. For the classes from the android test framework (from the `com.android.tools.idea.tests.gui.framework` package and related) - have you looked into whether we can use this code directly from the android studio as a platform sdk, or if we can commit one or a few jars from their test framework into our repo?

stevemessick:
I haven't specifically looked into extracting the test framework. It's pretty big, and I doubt it can be extracted easily since it has direct references to all the Android Studio UI classes. If we wanted to do it, we'd have to copy all the classes we needed into a new jar file.

devoncarew:
We could also use the Android Studio instance as the platform sdk to build and run tests against. That would be less duplicated code from the AS test framework, with the downside that we could only run our UI integration tests in android studio (not IntelliJ, ...).

stevemessick:
We might even be able to run IntelliJ tests, with some extra work. We'd need to launch IntelliJ rather than Android Studio, and create the proper fixture to control that IDE frame. I think it should be possible, though. It would be nice to have tests for the project and module wizards that IntelliJ uses.

flutter-studio/README.md
```diff
@@ -17,3 +17,40 @@ To set up a development environment:
 6. Select the `community-main` module and add module
    dependencies to `Dart-community`, `flutter-intellij-community`,
    and `flutter-studio`.
+
+The GUI tests must be run in the same version of IntelliJ used by
```
devoncarew: We have some run configurations checked in already - one for unit tests, and a provisional one for integration tests. At some point, we should add a checked in run config for ui integration tests (and likely delete the old integration tests run config).

flutter-studio/README.md
```diff
@@ -17,3 +17,40 @@ To set up a development environment:
...
+-Dmrj.version=mac 
+-Dcom.apple.macos.useScreenMenuBar=true 
+-Dapple.laf.useScreenMenuBar=true 
+-Dflutter.home=/path/to/flutter
+```
+If you're not using a Mac then delete these:
+ - mrj.version
+ - com.apple.macos.useScreenMenuBar
+ - apple.laf.useScreenMenuBar
+
```
pq: Great instructions.  Thank you!

flutter-studio/testSrc/io/flutter/tests/gui/NewProjectTest.java
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * If flakey tests are found try adjusting these settings:
+ * Settings festSettings = myGuiTest.robot().settings();
+ * festSettings.delayBetweenEvents(50); // 30
```
devoncarew: Is this comment still relevant, and, should we just make these delays the default?

flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/fixture/newProjectWizard/NewFlutterProjectWizardFixture.java
```diff
@@ -0,0 +1,79 @@
...
+    return new FlutterSettingsStepFixture(robot(), rootPane);
+  }
+
+  @NotNull
+  @Override
+  public NewFlutterProjectWizardFixture clickFinish() {
+    super.clickFinish();
+
+    // Wait for 'flutter create' to finish
+    Wait.seconds(30).expecting("Modal Progress Indicator to finish")
```
pq: This feels like a place for a potential race condition.  If that turns out to be true, another thought would be to wait for one of the flutter created artifacts to be found on disk (e.g., `main.dart`).

Needless to say we can cross that bridge when we get to it!

stevemessick:
Needs work.

devoncarew:
> Needs work.

urk - redoing due to git issues, or other? Definitely excited to see this land!

PR #1400, opened 2017-10-17
Inspect functionality

jacob314:
I'm still cleaning up InspectorPanel and FlutterView. All other files are generally ready for review.

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -299,6 +299,18 @@
   <!-- See: http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html -->
 
   <actions>
+    <group id="Flutter.InspectorActions">
+      <action id="Flutter.JumpToTypeSource" class="io.flutter.inspector.JumpToTypeSourceAction"
+              description="Jump to Type Source"
+              text="Jump to Type Source">
+        <keyboard-shortcut keymap="$default" first-keystroke="shift F4"/>
+      </action>
+      <action id="Flutter.JumpToSource" class="io.flutter.inspector.JumpToSourceAction"
+              text="Jump to Source">
+        <keyboard-shortcut keymap="$default" first-keystroke="control DOWN"/>
+      </action>
+    </group>
+
     <group id="Flutter.MainToolbarActions">
       <separator/>
       <action id="Flutter.DeviceSelector" class="io.flutter.actions.DeviceSelectorAction"
--- a/src/io/flutter/inspector/DiagnosticLevel.java
+++ b/src/io/flutter/inspector/DiagnosticLevel.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.inspector;
+
+/**
+ * The various priority levels used to filter which diagnostics are shown and
+ * omitted.
+ * <p>
+ * Trees of Flutter diagnostics can be very large so filtering the diagnostics
+ * shown matters. Typically filtering to only show diagnostics with at least
+ * level debug is appropriate.
+ * <p>
+ * See https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/foundation/diagnostics.dart
+ * for the corresponding Dart enum.
+ */
+public enum DiagnosticLevel {
+  /**
+   * Diagnostics that should not be shown.
+   * <p>
+   * If a user chooses to display [hidden] diagnostics, they should not expect
+   * the diagnostics to be formatted consistently with other diagnostics and
+   * they should expect them to sometimes be be misleading. For example,
+   * [FlagProperty] and [ObjectFlagProperty] have uglier formatting when the
+   * property `value` does does not match a value with a custom flag
+   * description. An example of a misleading diagnostic is a diagnostic for
+   * a property that has no effect because some other property of the object is
+   * set in a way that causes the hidden property to have no effect.
+   */
+  hidden,
+
+  /**
+   * A diagnostic that is likely to be low value but where the diagnostic
+   * display is just as high quality as a diagnostic with a higher level.
+   * <p>
+   * Use this level for diagnostic properties that match their default value
+   * and other cases where showing a diagnostic would not add much value such
+   * as an [IterableProperty] where the value is empty.
+   */
+  fine,
+
+  /**
+   * Diagnostics that should only be shown when performing fine grained
+   * debugging of an object.
+   * <p>
+   * Unlike a [fine] diagnostic, these diagnostics provide important
+   * information about the object that is likely to be needed to debug. Used by
+   * properties that are important but where the property value is too verbose
+   * (e.g. 300+ characters long) to show with a higher diagnostic level.
+   */
+  debug,
+
+  /**
+   * Interesting diagnostics that should be typically shown.
+   */
+  info,
+
+  /**
+   * Very important diagnostics that indicate problematic property values.
+   * <p>
+   * For example, use if you would write the property description
+   * message in ALL CAPS.
+   */
+  warning,
+
+  /**
+   * Diagnostics that indicate errors or unexpected conditions.
+   * <p>
+   * For example, use for property values where computing the value throws an
+   * exception.
+   */
+  error,
+
+  /**
+   * Special level indicating that no diagnostics should be shown.
+   * <p>
+   * Do not specify this level for diagnostics. This level is only used to
+   * filter which diagnostics are shown.
+   */
+  off,
+}
--- a/src/io/flutter/inspector/DiagnosticsNode.java
+++ b/src/io/flutter/inspector/DiagnosticsNode.java
@@ -146,8 +146,6 @@ public class DiagnosticsNode {
     return getStringMember("type");
   }
 
-  // The follow fields are only specified for properties.
-
   /**
    * Whether the description is enclosed in double quotes.
    * <p>
@@ -167,6 +165,7 @@ public class DiagnosticsNode {
    * Unit must be acceptable to display immediately after a number with no
    * spaces. For example: 'physical pixels per logical pixel' should be a
    * [tooltip] not a [unit].
+   *
    * Only specified for Number properties.
    */
   public String getUnit() {
@@ -179,6 +178,8 @@ public class DiagnosticsNode {
 
   /**
    * String describing just the numeric [value] without a unit suffix.
+   *
+   * Only specified for Number properties.
    */
   public String getNumberToString() {
     return getStringMember("numberToString");
--- a/src/io/flutter/inspector/DiagnosticsNode.java
+++ b/src/io/flutter/inspector/DiagnosticsNode.java
@@ -1,0 +1,503 @@
+/*
+ * Copyright 2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.inspector;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonNull;
+import com.google.gson.JsonObject;
+
+import java.util.ArrayList;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Defines diagnostics data for a [value].
+ * <p>
+ * [DiagnosticsNode] provides a high quality multi-line string dump via
+ * [toStringDeep]. The core members are the [name], [toDescription],
+ * [getProperties], [value], and [getChildren]. All other members exist
+ * typically to provide hints for how [toStringDeep] and debugging tools should
+ * format output.
+ * <p>
+ * See also:
+ * <p>
+ * * DiagnosticsNode class defined at https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/foundation/diagnostics.dart
+ * The difference is the class hierarchy is collapsed on the Java side as
+ * the class hierarchy on the Dart side exists more to simplify creation
+ * of Diagnostics  than because the class hierarchy of Diagnostics is
+ * important. If you need to determine the exact Diagnostic class on the
+ * Dart side you can use the value of type. The raw Dart object value is
+ * also available via the getValue() method.
+ */
+public class DiagnosticsNode {
+  public DiagnosticsNode(JsonObject json, InspectorService inspectorService) {
+    this.inspectorService = inspectorService;
+    this.json = json;
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (other instanceof DiagnosticsNode) {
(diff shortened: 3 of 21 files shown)


src/io/flutter/utils/DiagnosticsNode.java
```diff
@@ -0,0 +1,502 @@
...
+
+  /**
+   * Reference to the actual Dart DiagnosticsNode object this Java DiagnosticsNode
+   * describes.
+   *
+   * We cannot use observatory protocol InstanceRef ids as they are not adequately long lived.
+   */
+  private String dartDiagnosticId;
+
+  // XXX document
```
pq: Possibly migrate to the `TODO(user_name):` convention?

src/io/flutter/view/FlutterView.java
```diff
@@ -150,6 +192,13 @@ private void updateIcon() {
       else {
         toolWindow.setIcon(ExecutionUtil.getLiveIndicator(FlutterIcons.Flutter_13));
       }
+
+      System.out.println("App changed... " + (app == null));
```
pq: Remove `System.out`?

pq: Alternatively, consider using a `Logger` or putting this debugging output behind a flag?

devoncarew:
Awesome! Some comments on the UI; nothing that would block this PR, but things we'll want to address as follow-on work. Some are likely uncontroversial, others likely would need to be discussed:

screen shot 2017-10-17 at 12 41 53 pm

- (as noted via IM) we'll want the flutter view's toolbar always visible - so the widget and render tree tab area should go under the toolbar
- we'll want users to be able to work on their app and use the inspector side-by-side; I think we'll want to optimize the UI for a taller, narrowing amount of screen real estate. this will be somewhat challenging given the depth of the trees :)
- from use cases I've seen that the inspector can help with (from a doc from you :) ), most issues are solved via the widget tree. The render tree is used less often. Do we want to give them equal weight in the UI? This will indicate to the user that they should be using the render tree for something and could be a source of confusion. Options here are to not have a render tab - just show a widget tree under the toolbar with no tabs - or find some way to toggle to render tree mode.

src/io/flutter/utils/DiagnosticLevel.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * The various priority levels used to filter which diagnostics are shown and
+ * omitted.
+ *
+ * Trees of Flutter diagnostics can be very large so filtering the diagnostics
+ * shown matters. Typically filtering to only show diagnostics with at least
+ * level debug is appropriate.
+ */
+public enum DiagnosticLevel {
```
devoncarew: This is related to APIs exposed in the flutter framework? Perhaps reference the code / classes here? We may also want the javadoc to be terser to reduce our maintenance burden, esp. if these classes are well documented in the flutter framework.

src/io/flutter/utils/DiagnosticsNode.java
```diff
@@ -0,0 +1,502 @@
...
+
+  /**
+   * Value as a List of strings.
+   *
+   * The raw value can always be extracted with the regular observatory protocol.
+   *
+   * Only applies to IterableProperty.
+   */
+  public ArrayList<String> getValues() {
+    ArrayList<String> values = null;
```
devoncarew: probably don't need to init this

src/io/flutter/utils/DiagnosticsNode.java
```diff
@@ -0,0 +1,502 @@
...
+    return new InspectorInstanceRef(json.get("objectId").getAsString());
+  }
+
+  public CompletableFuture<ArrayList<DiagnosticsNode>> getProperties() {
+    if (properties == null) {
+      properties = inspectorService.getProperties(getDartDiagnosticRef());
+    }
+    return properties;
+  }
+}
```
devoncarew: just to reduce future diffs, I would do a formatting pass on the files

src/io/flutter/utils/EvalOnDartLibrary.java
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+   * Workaround until the version of the VmService 'eval' method supports
+   * the scope argument.
+   */
+  private void callVmServiceRequest(VmService vmService, String methodName, JsonObject params, EvaluateConsumer consumer) {
+    try {
+      final Class<?> c = Class.forName("org.dartlang.vm.service.VmServiceBase");
+
+      final Method[] allMethods = c.getDeclaredMethods();
```
devoncarew: There is a `com.intellij.util.ReflectionUtil` class which can reduce a bit of the work here (used for example by `FlutterReloadManager`)

src/io/flutter/utils/EvalOnDartLibrary.java
```diff
@@ -0,0 +1,195 @@
...
+  public CompletableFuture<InstanceRef> eval(String expression, Map<String, String> scope) {
+    final CompletableFuture<InstanceRef> future = new CompletableFuture<>();
+    myRequestsScheduler.addRequest(() -> {
+      libraryRef.thenAcceptAsync((LibraryRef ref) -> {
+        evaluateHelper(isolateInfo.getIsolateId(), ref.getId(), expression, scope,
+         new EvaluateConsumer() {
+           @Override
+           public void onError(RPCError error) {
+             // TODO(jacobr): what is the right way to log this case?
+             System.err.println("Got RPCError: " + error);
```
devoncarew: Presumably the client will listen to the error result and display (or ignore) appropriately to the user? If you do need to log, you can use the intellij logging system: https://github.com/flutter/flutter-intellij/blob/335e25cc3c97d515305ec72ec4dd48dd3ebba8e8/src/io/flutter/run/daemon/DaemonEvent.java#L290

src/io/flutter/utils/DiagnosticsNode.java
jacob314: My convention is all XXX comments are one I expect to remove before landing.

src/io/flutter/utils/DiagnosticLevel.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.utils;
```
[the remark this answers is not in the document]
jacob314: moved to a new ` .inspector.` package

src/io/flutter/utils/DiagnosticsNode.java
jacob314: I get a "values" might not have been initialized java lint warning otherwise.
Cleaned up the method to use an early return style to reduce nesting and avoid the = null.

src/io/flutter/utils/DiagnosticLevel.java
jacob314: I copied and pasted the docs from the Flutter as you guessed. Added a link to the Flutter file with the API.

src/io/flutter/utils/DiagnosticsNode.java
devoncarew: odd

src/io/flutter/utils/DiagnosticLevel.java
devoncarew: kk. going forward, we may want to tersify the docs (or eliminate them?) to reduce drift and maintenence

src/io/flutter/utils/DiagnosticsNode.java
jacob314: done

src/io/flutter/utils/EvalOnDartLibrary.java
jacob314: That is handy. Done.

src/io/flutter/utils/EvalOnDartLibrary.java
jacob314: Done

src/io/flutter/utils/DiagnosticLevel.java
jacob314: I don't expect these particular classes will change much in the future. If there are changes likely it will just be adding extra logging levels in methods but I could be wrong. Hopefully these pretty carefully written docs will be helpful for other people contributing to this code on the IntelliJ. If not, they should be slimmed down or removed.

src/io/flutter/view/FlutterView.java
jacob314: Removed. That was left over debugging code.

src/io/flutter/utils/DiagnosticLevel.java
devoncarew: gotcha

devoncarew:
Two other thoughts:

- post-landing, we should update the release testing script (https://github.com/flutter/flutter-intellij/blob/master/docs/testing.md) to include a brief smoke-test of the inspector (brevity is important to make sure the overall script takes < 10 min)
- per a discussion, most of the automated testing of this will happen once we have general UI integration test support and a headless flutter device available. Looking over this PR, I think the json parsing code is amenable to unit tests; perhaps file an issue to add unit tests for these?

jacob314:
PTAL. Added a setting menu item to configure whether the Inspector is displayed. Moved the inspector so it is the main panel. Made the toolbar show on both panels.

src/io/flutter/inspector/DiagnosticLevel.java
```diff
@@ -0,0 +1,83 @@
...
+ * for the corresponding Dart enum.
+ */
+public enum DiagnosticLevel {
+  /**
+   * Diagnostics that should not be shown.
+   * <p>
+   * If a user chooses to display [hidden] diagnostics, they should not expect
+   * the diagnostics to be formatted consistently with other diagnostics and
+   * they should expect them to sometimes be be misleading. For example,
+   * [FlagProperty] and [ObjectFlagProperty] have uglier formatting when the
```
stevemessick: Did we adopt a new javadoc convention? I'm not familiar with [...] in javadoc. I thought we used {@code ...}. If those references should be rewritten then I think it can wait until a later PR.

src/io/flutter/inspector/EvalOnDartLibrary.java
```diff
@@ -0,0 +1,185 @@
...
+
+  public CompletableFuture<InstanceRef> eval(String expression, Map<String, String> scope) {
+    final CompletableFuture<InstanceRef> future = new CompletableFuture<>();
+    myRequestsScheduler.addRequest(() -> {
+      libraryRef.thenAcceptAsync((LibraryRef ref) -> {
+        evaluateHelper(isolateInfo.getIsolateId(), ref.getId(), expression, scope,
+                       new EvaluateConsumer() {
+                         @Override
+                         public void onError(RPCError error) {
+                           // TODO(jacobr): what is the right way to log this case?
```
stevemessick: Are you wondering if an error here is serious enough to bring to the user's attention? If so, then maybe an error balloon in the style of IntelliJ would work.

resources/META-INF/plugin.xml
```diff
@@ -299,6 +299,18 @@
...
   <actions>
+    <group id="Flutter.InspectorActions">
+      <action id="Flutter.JumpToTypeSource" class="io.flutter.inspector.JumpToTypeSourceAction"
+              description="Jump to Type Source"
+              text="Jump to Type Source">
+        <keyboard-shortcut keymap="$default" first-keystroke="shift F4"/>
+      </action>
+      <action id="Flutter.JumpToSource" class="io.flutter.inspector.JumpToSourceAction"
+              text="Jump to Source">
+        <keyboard-shortcut keymap="$default" first-keystroke="control DOWN"/>
```
devoncarew: If this is not frequently used, we may not want a keybinding.

src/io/flutter/inspector/InspectorService.java
```diff
@@ -0,0 +1,352 @@
...
+      }
+      return nodes;
+    });
+  }
+
+  /**
+   * Converts an inspector ref to value suitable for use by generic intellij
+   * debugging tools.
+   * <p>
+   * Warning: DartVmServiceValue references do not make any lifetime gaurantees
```
pq: gaurantees => guarantees

resources/META-INF/plugin.xml
```diff
@@ -299,6 +299,18 @@
   <!-- See: http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html -->
 
   <actions>
+    <group id="Flutter.InspectorActions">
+      <action id="Flutter.JumpToTypeSource" class="io.flutter.inspector.JumpToTypeSourceAction"
+              description="Jump to Type Source"
+              text="Jump to Type Source">
+        <keyboard-shortcut keymap="$default" first-keystroke="shift F4"/>
```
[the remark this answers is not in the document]
jacob314: I've verified it doesn't conflict with the existing binding. I believe the reason is that the action is only activated within the context of the inspector panel not across the whole application.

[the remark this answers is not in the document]
devoncarew: 👍

resources/META-INF/plugin.xml
devoncarew: If this doesn't conflict (as per the item above), then no real need to remove.

src/io/flutter/inspector/InspectorService.java
jacob314: done

resources/META-INF/plugin.xml
jacob314: I think we might as well be consistent with the existing debugger panel keyboard shortcuts that do the same thing. Personally I find having keybindings for these handy as jumping to the source is really handy.

devoncarew: makes sense

src/io/flutter/inspector/EvalOnDartLibrary.java
jacob314: Removed the TODO as the LOG.error call is sufficient for this unlikely exceptional case.

src/io/flutter/inspector/DiagnosticsNode.java
```diff
@@ -0,0 +1,511 @@
...
+   * Dart class defining the diagnostic node.
+   * For example, DiagnosticProperty<Color>, IntProperty, StringProperty, etc.
+   * This should rarely be required except for cases where custom rendering is desired
+   * of a specific Dart diagnostic class.
+   */
+  String getType() {
+    return getStringMember("type");
+  }
+
+  // The follow fields are only specified for properties.
```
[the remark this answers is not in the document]
jacob314: Removed this comment and made sure all fields relevant to properties have comments specifying what kind of properties they apply to.

src/io/flutter/inspector/DiagnosticLevel.java
jacob314: These are copy and pasted from the Dart comments for these properties so the style is a bit wrong. I'll clean it up with a followup PR.

brianegan:
Whaaaaaat! This is getting next level :D As a user, this looks amazing. Nice work!

pq:
🎉 🎂 🎉

PR #1403, opened 2017-10-18
add a class to create custom icons

devoncarew:
- add a class to create custom icons

This class creates small circular icons with a letter in the center. It uses the same style as the IntelliJ Structure view icons (also used a few other places in the UI). The indent here is to use this for something like the flutter inspector's widget tree, where you have a large number of elements, want the UI to fit in with IntelliJ, want some variety in the icons but don't want the user to remember hundreds of specific icons.

@jacob314 @pq @stevemessick


src/io/flutter/utils/CustomIconMaker.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Map;
+
+public class CustomIconMaker {
+  private static final Color normalColor = ColorUtil.fromHex("231F20");
+
+  private final Map<String, Icon> iconCache = new HashMap<>();
+
+  public CustomIconMaker() {
+
+  }
```
pq: Nit: why the explicit default constructor?

src/io/flutter/utils/CustomIconMaker.java
```diff
@@ -0,0 +1,110 @@
...
+    return getCustomIcon(fromText, IconKind.kClass, false);
+  }
+
+  public Icon getCustomIcon(String fromText, IconKind kind) {
+    return getCustomIcon(fromText, kind, false);
+  }
+
+  public Icon getCustomIcon(String fromText, IconKind kind, boolean isAbstract) {
+    if (fromText == null || fromText.length() == 0) {
+      return null;
```
pq: Maybe `StringUtils.isEmpty(fromText)`?  Or:

    if (fromText == null || fromText.isEmpty()) { ...

src/io/flutter/utils/CustomIconMaker.java
```diff
@@ -0,0 +1,110 @@
...
+            g2.setRenderingHint(RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON);
+            g2.setColor(normalColor);
+
+            final Font font = UIUtil.getFont(UIUtil.FontSize.MINI, UIUtil.getTreeFont());
+            g2.setFont(font);
+
+            final Rectangle2D bounds = g2.getFontMetrics().getStringBounds(text, g2);
+            final float offsetX = (getIconWidth() - (float)bounds.getWidth()) / 2.0f;
+            // Some black magic here for vertical centering.
+            final float offsetY = getIconHeight() - ((getIconHeight() - (float)bounds.getHeight()) / 2.0f) - 2.0f;
```
pq: Curious if you've tried this at different display resolutions?

src/io/flutter/utils/CustomIconMaker.java
devoncarew: Hmm, habit? Should we avoid?

src/io/flutter/utils/CustomIconMaker.java
devoncarew: done

src/io/flutter/utils/CustomIconMaker.java
devoncarew: No (just HD), but this is similar to code that we do know works across resolutions.

src/io/flutter/utils/CustomIconMaker.java
pq: Nope.  Just curious!

src/io/flutter/utils/CustomIconMaker.java
pq: Cool.  Again, I was just curious!

Issue #1371, opened 2017-10-05
the flutter test editor decoration arrow shows up for regular dart projects


devoncarew:
The flutter test execution editor contribution shows for regular dart projects, and we try and run those tests with `flutter test`.

<img width="377" alt="screen shot 2017-10-04 at 5 11 40 pm" src="https://user-images.githubusercontent.com/1269969/31205600-46801fec-a927-11e7-8dae-188c5bba5004.png">

@pq

pq:
Hmmm.  So this is curious.  I can't repro locally.

![image](https://user-images.githubusercontent.com/67586/31369475-a6d8efb6-ad39-11e7-82fd-f51fbf8db1f4.png)

Is this a vanilla linter checkout?  IDEA?

devoncarew:
This doesn't repro in android studio or my runtime workbench in IntelliJ. It does repro in my regular IntelliJ. I speculatively tried 'invalidate caches' and restarted, but didn't see a change.

pq:
OK.  So I've narrowed this down.  The issue is in how we check for flutteriness. In
`io.flutter.utils.FlutterModuleUtils#isFlutterModule` we use this heuristic:

      // [Flutter support enabled for a module] ===
      // [Dart support enabled && referenced Dart SDK is the one inside a Flutter SDK]

If we want to support vanilla Dart modules using a Flutter Dart SDK we'll need to change this.

devoncarew:
OK, sounds like we'll have to narrow the check somehow. /cc @jwren 

I know this will affect several use cases, so we'll need to do it carefully. Perhaps a narrower check which specific call sites in the code can opt into?

PR #1404, opened 2017-10-18
Ensure vanilla dart tests are not mis-indentified as fluttery (#1371).

pq:
Ensures tests run in a vanilla dart project (with nonetheless an SDK vended into a Flutter install) are not run w/ `flutter test` and that new test run configs are only fluttery if they should be.

Fixes: #1371 

@devoncarew

### DIFF
--- a/src/io/flutter/FlutterUtils.java
+++ b/src/io/flutter/FlutterUtils.java
@@ -67,6 +67,9 @@ public class FlutterUtils {
     return file != null && file.exists();
   }
 
+  /**
+   * Test if the given element is contained in a module with a pub root that declares a flutter dependency.
+   */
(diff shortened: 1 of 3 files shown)


src/io/flutter/FlutterUtils.java
```diff
@@ -64,6 +67,16 @@ public static boolean exists(@Nullable VirtualFile file) {
     return file != null && file.exists();
   }
 
+  public static boolean isInFlutterProject(@NotNull PsiElement element) {
```
devoncarew: Can we add some docs here (and in the other 'is flutter module/project' method) to say that this is the definitely-flutter-project method, and the other is the might-be-flutter one? That'll help the next person through :)

pq: Sure! 👍

Issue #1395, opened 2017-10-13
Flutter threw an uncaught NoClassDefFoundError. 


seven5566:
Android Studio 3.0 Canary 8
flutter plugin is installed.
File -> New Flutter Project,then it will show it
`Flutter threw an uncaught NoClassDefFoundError.`
Please help me.

```
java.lang.NoClassDefFoundError: com/android/tools/idea/observable/core/OptionalValueProperty
	at io.flutter.project.FlutterProjectModel.<init>(Unknown Source)
	at io.flutter.actions.FlutterNewProjectAction.actionPerformed(Unknown Source)
	at com.intellij.openapi.actionSystem.ex.ActionUtil$1.run(ActionUtil.java:215)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAware(ActionUtil.java:232)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.lambda$actionPerformed$0(ActionMenuItem.java:309)
	at com.intellij.openapi.wm.impl.FocusManagerImpl.runOnOwnContext(FocusManagerImpl.java:929)
	at com.intellij.openapi.wm.impl.IdeFocusManagerImpl.runOnOwnContext(IdeFocusManagerImpl.java:136)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.actionPerformed(ActionMenuItem.java:299)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem.lambda$fireActionPerformed$0(ActionMenuItem.java:116)
	at com.intellij.openapi.application.TransactionGuardImpl.runSyncTransaction(TransactionGuardImpl.java:86)
	at com.intellij.openapi.application.TransactionGuardImpl.lambda$submitTransaction$1(TransactionGuardImpl.java:109)
	at com.intellij.openapi.application.TransactionGuardImpl.submitTransaction(TransactionGuardImpl.java:118)
	at com.intellij.openapi.application.TransactionGuard.submitTransaction(TransactionGuard.java:122)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem.fireActionPerformed(ActionMenuItem.java:116)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
[...]

PR #1410, opened 2017-10-20
Check that Android Studio is up-to-date

### LINKED ISSUE #1396: add a programatic min. version check for Android Studio
We may want to consider a programatic min. version check for AS - if we don't meet it, we can do something like show a toast to that effect.

@stevemessick 

(see https://github.com/flutter/flutter-intellij/issues/1395 for context)

stevemessick:
Check to see if the version of Android Studio is too old to use. This check runs before the splash screen is removed. The error will be displayed even if the welcome screen is the only UI that is shown. Currently, beta 7 and later are acceptable.

Ideally, we'd disable the plugin. I experimented with that but it is hard to get working. It would require shutdown and restart anyway, so doesn't actually prevent the user from doing things that will generate stack traces as in #1395.

Fixes #1396
@devoncarew @pq

### DIFF
--- a/flutter-studio/src/io/flutter/FlutterStudioInitializer.java
+++ b/flutter-studio/src/io/flutter/FlutterStudioInitializer.java
@@ -5,11 +5,25 @@
  */
 package io.flutter;
 
+import com.intellij.openapi.application.ApplicationInfo;
+import com.intellij.openapi.ui.Messages;
+
 public class FlutterStudioInitializer implements Runnable {
+
+  private static void reportVersionIncompatibility(ApplicationInfo info) {
+    Messages.showErrorDialog("The Flutter plugin requires a more recent version of Android Studio.", "Version Mesmatch");
+  }
+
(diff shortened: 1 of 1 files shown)


flutter-studio/src/io/flutter/FlutterStudioInitializer.java
```diff
@@ -5,11 +5,25 @@
  */
 package io.flutter;
 
+import com.intellij.openapi.application.ApplicationInfo;
+import com.intellij.openapi.ui.Messages;
+
 public class FlutterStudioInitializer implements Runnable {
+
+  private static void reportVersionIncompatibility(ApplicationInfo info) {
+    Messages.showErrorDialog("The Flutter plugin requires a more recent version of Android Studio.", "Version Mesmatch");
```
pq: Mesmatch => Mismatch

stevemessick:
Fixed, thanks!

Issue #1411, opened 2017-10-23
doing a full restart should re-focus the simulator to the foreground


devoncarew:
When we run an app against the ios simulator, we focus the simulator to the foreground. We should do the same on a full restart, since the entry point to that now is to click the run/debug buttons.

pq:
I was thinking about this actually so I'll take a look. 👍

pq:
@devoncarew: hope you weren't already do this! 😃

PR #1415, opened 2017-10-24
Give restarted apps focus on iOS (#1411).

pq:
Fixes: #1411.

@devoncarew


src/io/flutter/run/FlutterReloadManager.java
```diff
@@ -217,14 +220,19 @@ public void saveAllAndReload(@NotNull FlutterApp app) {
     }
   }
 
-  public void saveAllAndRestart(@NotNull FlutterApp app) {
+  public void saveAllAndRestart(@NotNull FlutterApp app, @NotNull Project project) {
```
devoncarew: as a nit, I believe that `FlutterApp` already has a reference to the project

pq: Fantastic.  Thanks!

PR #1422, opened 2017-10-24
No need to check for identifier clash

stevemessick:
Remove check for module name matching identifier syntax.

@pq @devoncarew


pq:
If I'm remembering right, this check is to ensure that we don't create projects whose names make `flutter create` or `pub` calls fall over.  Was it being too careful?

stevemessick:
Do you have an example of a name that passes the package name test but fails the identifier validity test? I couldn't come up with one, but if there is such I'll delete this PR.

pq:
Whoops.  Scrolled up.  I didn't realize we were also validating package name format.

Ignore my comment.  I think we're good! 👍

PR #1424, opened 2017-10-24
add some descriptive text to the flutter run launch config

devoncarew:
- add descriptive text to the flutter run launch config; this is to help clarify that 'build flavors' are optionsl
- sort the build flavor last in the dialog

@pq @stevemessick @mit-mit 

screen shot 2017-10-24 at 1 18 37 pm


devoncarew:
It does look busy now, which I don't love. I could remove the label text for the entrypoint field, which is a bit more straightforward as to what it's for.

pq:
Awesome.  Besides a little padding, what about making the text gray?  Or would that be too hard to read?

stevemessick:
I used label enable/disable depending on what was selected in the AS wizards. You could trigger from text field focus, but remember the button to the right of the entry-point field.

devoncarew:
I tried de-emphasizing the labels a bit (I toggled the disabled bit):

screen shot 2017-10-24 at 2 27 30 pm

PR #1443, opened 2017-10-26
Fix compile errors

stevemessick:
@pq @devoncarew

Add xdebugger-impl to the module deps.

Am I the first to compile with the new inspector?


stevemessick:
Ah, that explains it. I do have all the sources available.

Issue #1445, opened 2017-10-26
Flutter actions wrongly contributed to non-Flutter pubspecs.


pq:
Vanilla Dart project:

![image](https://user-images.githubusercontent.com/67586/32077319-d9f49f6e-ba57-11e7-8894-a58b044b21e7.png)

PR #1448, opened 2017-10-26
Scope Flutter pubspec action contributions (#1445).

pq:
Fixes: #1445.

@devoncarew @jwren 

FYI @jwren, @alexander-doroshko : we'll want to update the check in the Dart Plugin (if we haven't already) to make sure that it verifies that projects are declaring Flutter when we test for flutteriness.


jwren:
LGTM

PR #1450, opened 2017-10-26
Changes to support multiple plugin ids

stevemessick:
@pq @devoncarew

The idea here is that PLUGINID, SINCE, and UNTIL will be defined in build.xml. We'll use the <filter> tag in ant to replace those strings with the associated values when the plugin.xml file is copied during the build.

The two possible plugin ids are defined in FlutterUtils. The other code changes are required to eliminate hard-coded plugin ids.

I started from the build.xml one version prior to current. The build-all.sh script runs it twice with the various parameters set to produce two artifacts.

Removing the module dependency for the IntelliJ version allows it to load into all IntelliJ products but no others. Define the androidstudio dependency for the Android Studio version pins that one to AS.


devoncarew:
lgtm! (@stevemessick, sorry for the delay in reviewing)

PR #1454, opened 2017-10-30
Change build script to leave usable plugin.xml

stevemessick:
Template plugin.xml added to ant script to produce usable plugin.xml. The args used in build-all.sh can be used to make a plugin.xml for Android Studio (no one but me needs to).

@devoncarew @pq @jwren

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -194,11 +194,9 @@ ant -Ddart.plugin.version=171.4424.10 -Didea.version=171.4333198 -Didea.product=
     </copy>
 
     <delete file="build/src/META-INF/plugin.xml"/>
-    <copy todir="build/src/META-INF">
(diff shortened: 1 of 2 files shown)


stevemessick:
Yeah, it is getting harder 😣 Full details on the wiki:
https://github.com/flutter/flutter-intellij/wiki/Releasing

I don't think this is likely to be our long-term solution; rather, this gets us past the current hurdle.

devoncarew:
lgtm

re: the changelog - we could maintain that in a separate file, and that could be one of the `@` variables that gets substituted

stevemessick:
That would make editing easier, for both release notes and plugin content. We should do that as part of a larger rewrite of the build script. I still want to do a little investigation on how other plugins manage plugin.xml composition. There may be other tricks we haven't learned.

Issue #1418, opened 2017-10-24
add a feedback link for the inspector preview


devoncarew:
Similar to what we did for reload on save.

PR #1462, opened 2017-11-01
add a feedback link for the widget inspector

devoncarew:
- add a placeholder for a feedback link for the widget inspector (fix https://github.com/flutter/flutter-intellij/issues/1418)

Note: I won't land this until we have a goo.gl short link for the feedback.

@jacob314 @InMatrix


devoncarew:
@jacob314, updated with a real link; ptal.

Issue #1449, opened 2017-10-26
back-port flutter check to the dart plugin


devoncarew:
I think we need to update the Dart plugin's version of the 'is flutter' check. The current Flutter plugin one is here: https://github.com/flutter/flutter-intellij/blob/42c1bf74ca59b41d692b6ab0fa65931aa45fd904/src/io/flutter/pub/PubRoot.java#L277

@jwren is this something you can take care of? My understanding is that this will help out a few things, like double-notification of pubspec.yaml editor header contributions.

(details in https://github.com/flutter/flutter-intellij/issues/1261#issuecomment-339808741)

jwren:
@devoncarew Yes, I'll take it.

PR #1468, opened 2017-11-03
Fix for https://github.com/flutter/flutter-intellij/issues/1449

jwren:
Fix for https://github.com/flutter/flutter-intellij/issues/1449 and a change to the Dart Plugin and Flutter Plugin to check for the flutteriness of a pubspec.  We now look to see if the individual pubspec declares flutter when determining if a notification should appear for a pubspec, _instead_ of asking the module if some pubspec.yaml declares flutter.

Nit: when comparing a file name to "pubspec.yaml" the two plugins are not consistent in _not_ ignoring the case of the filename.

Corresponding PR: https://github.com/JetBrains/intellij-plugins/pull/557

### DIFF
--- a/src/io/flutter/editor/FlutterPubspecNotificationProvider.java
+++ b/src/io/flutter/editor/FlutterPubspecNotificationProvider.java
@@ -55,13 +55,12 @@ public class FlutterPubspecNotificationProvider extends EditorNotifications.Prov
       return null;
     }
 
-    final Module module = ModuleUtilCore.findModuleForFile(file, project);
-    if (module == null || !FlutterModuleUtils.usesFlutter(module)) {
+    // Check that this pubspec file declares flutter
+    if (!PubRoot.declaresFlutter(file)) {
       return null;
     }
 
-    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
-    if (sdk == null) {
+    if(FlutterSdk.getFlutterSdk(project) == null) {
       return null;
     }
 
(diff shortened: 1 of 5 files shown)


src/io/flutter/editor/FlutterPubspecNotificationProvider.java
```diff
@@ -55,13 +55,12 @@ public EditorNotificationPanel createNotificationPanel(@NotNull VirtualFile file
...
-    final Module module = ModuleUtilCore.findModuleForFile(file, project);
-    if (module == null || !FlutterModuleUtils.usesFlutter(module)) {
+    // Check that this pubspec file declares flutter
+    if (!PubRoot.declaresFlutter(file)) {
       return null;
     }
 
-    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
-    if (sdk == null) {
+    if(FlutterSdk.getFlutterSdk(project) == null) {
```
pq: Perhaps format?  I'd expect a space after the `if`...

src/io/flutter/editor/FlutterPubspecNotificationProvider.java
```diff
@@ -55,13 +55,12 @@ public EditorNotificationPanel createNotificationPanel(@NotNull VirtualFile file
       return null;
     }
 
-    final Module module = ModuleUtilCore.findModuleForFile(file, project);
-    if (module == null || !FlutterModuleUtils.usesFlutter(module)) {
+    // Check that this pubspec file declares flutter
+    if (!PubRoot.declaresFlutter(file)) {
```
pq: Perhaps put this method in `FlutterUtils`?  (We've got a bunch of similar methods there and I think it's a slightly more obvious place to look.)

src/io/flutter/editor/FlutterPubspecNotificationProvider.java
jwren: Done

src/io/flutter/editor/FlutterPubspecNotificationProvider.java
jwren: Done.  I had to also moved the isFlutterPlugin method, since I didn't want 2 versions of loadPubspecInfo.

jwren:
I think I have both of your comments addressed.

Issue #1436, opened 2017-10-25
provide feedback in the new project wizard when the user has entered a correct flutter sdk location


devoncarew:
Like in the Flutter settings page, when a correct sdk is entered in the new project wizard, we should provide some feedback in the wizard page. The settings page displays the results of running `flutter --version`. Not sure if that is necessary for the project wizard, but some positive feedback would be great.

stevemessick:
Fixed with https://github.com/flutter/flutter-intellij/pull/1469

PR #1469, opened 2017-11-03
Change SDK background

stevemessick:
#1436 asks for some way to inform the user that a valid Flutter SDK path has been entered.
With this change the SDK field has a normal background color when the path is correct. Otherwise, it changes as shown in the screen shots. This is only for the Android Studio version. Once we determine this is viable it (or whatever it morphs into) can be applied to the other.

@devoncarew @pq 

screen shot 2017-11-03 at 3 29 14 pm

screen shot 2017-11-03 at 3 28 23 pm

### DIFF
--- a/flutter-studio/src/io/flutter/project/FlutterProjectStep.java
+++ b/flutter-studio/src/io/flutter/project/FlutterProjectStep.java
@@ -30,8 +30,7 @@ import com.intellij.openapi.roots.ui.configuration.ProjectStructureConfigurable;
 import com.intellij.openapi.ui.TextComponentAccessor;
 import com.intellij.openapi.ui.TextFieldWithBrowseButton;
 import com.intellij.openapi.util.IconLoader;
-import com.intellij.ui.ComboboxWithBrowseButton;
-import com.intellij.ui.JBProgressBar;
+import com.intellij.ui.*;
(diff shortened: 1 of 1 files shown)


flutter-studio/src/io/flutter/project/FlutterProjectStep.java
```diff
@@ -209,6 +220,20 @@ public static void ensureComboModelContainsCurrentItem(@NotNull final JComboBox
     comboBox.getEditor().setItem(currentItem); // to set current item in combo itself
   }
 
+  private void updateSdkField(JTextComponent sdkEditor) {
+    FlutterSdk current = FlutterSdk.forPath(sdkEditor.getText());
+    Color color = sdkBackgroundColor;
+    if (current == null) {
+      if (ColorUtil.isDark(sdkBackgroundColor)) {
+        color = ColorUtil.darker(JBColor.YELLOW, 5);
```
devoncarew: It seems to me that there may be some way of getting an 'error' or 'invalid' background color from the current theme, but some very brief looking didn't turn up anything.

stevemessick: Yeah, I found a bunch of named colors, but nothing useful. We can keep looking.

PR #1473, opened 2017-11-06
Suppress missing (optional) androidstudio module error.

pq:
This error appears to prevent sources from loading in a runtime workbench and occurs when the androidstudio is not installed.

@stevemessick

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -300,6 +300,7 @@
   <depends optional="true" config-file="idea-contribs.xml">com.intellij.modules.java</depends>
 
   <!-- Contributes Android Studio-specific features and implementations. -->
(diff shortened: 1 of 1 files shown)


resources/META-INF/plugin.xml
```diff
@@ -300,6 +300,7 @@
   <depends optional="true" config-file="idea-contribs.xml">com.intellij.modules.java</depends>
 
   <!-- Contributes Android Studio-specific features and implementations. -->
+  <!--suppress PluginXmlValidity -->
```
stevemessick: Do we need this in plugin.xml.template too? Asking because I don't know, but we can add if needed.

pq: I'm not sure.  If it's just for development maybe not?  As you say, we can add if needed.

PR #1474, opened 2017-11-06
add a keybinding for shift-cmd-s

devoncarew:
- add add a keybinding for shift-cmd-s (for a full restart - a companion to cmd-s and hot reload on save). While writing Flutter code I found myself wanting this :)
- remove the semi-colon keybinding for reload and restart; we now have three keybinding and this one seems the most esoteric. perhaps a resolution is to update the docs to mention the two keybindings, and to point people to existing intellij docs on customizing keybindings if the existing ones don't work for them

@pq @mit-mit


mit-mit:
Love it!

Now we just need to make it easier to call format!

devoncarew:
We'll need to update https://flutter.io/intellij-ide/#intellij-keyboard-shortcuts w/ the latest keybindings once M19 is out -

PR #1476, opened 2017-11-06
fix how we retarget flutter actions

devoncarew:
This addresses an issue I saw this weekend. I had two project windows open against two separate flutter projects. I hit run in one project; when the app came up, the reload icon became enabled both in the current project, but also in the 2nd open project window. Hitting reload in either window would reload the app running for the first window.

- switch from using `ActionManager` to track globally registered actions, to using a new tracker that's scoped to projects
- additionally, fix an issue in reload on save where we didn't reload when the focused dart editor wasn't in the module for the app currently running. From dogfooding, it's surprising when the app doesn't reload (users get a sense that cmd-s means reload).

@pq

### DIFF
--- a/src/io/flutter/run/FlutterReloadManager.java
+++ b/src/io/flutter/run/FlutterReloadManager.java
@@ -5,7 +5,6 @@
  */
 package io.flutter.run;
 
-import com.android.annotations.NonNull;
 import com.intellij.codeInsight.hint.HintManager;
 import com.intellij.codeInsight.hint.HintManagerImpl;
 import com.intellij.codeInsight.hint.HintUtil;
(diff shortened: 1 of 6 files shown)


src/io/flutter/run/FlutterReloadManager.java
```diff
@@ -327,14 +319,14 @@ private boolean hasErrors(@NotNull Project project, @NotNull Module module, @Not
     }
   }
 
-  private static boolean shouldBlockReload(@NotNull DartServerData.DartError error, @NotNull Module module) {
+  private static boolean shouldBlockReload(@NotNull DartServerData.DartError error, @NonNull Project project, @Nullable Module module) {
```
pq: I'm assuming `NonNull` was a typo and you mean `NotNull`?

src/io/flutter/run/FlutterReloadManager.java
```diff
@@ -5,6 +5,7 @@
  */
 package io.flutter.run;
 
+import com.android.annotations.NonNull;
```
pq: See below: I think we don't want this import?

devoncarew: oops - removing

src/io/flutter/run/FlutterReloadManager.java
devoncarew: Good catch! done

PR #1481, opened 2017-11-07
Update DartVmServiceDebugProcessZ.java

devoncarew:
@jwren, increase the timeout when waiting for an app to start up.


jwren:
LGTM

PR #1487, opened 2017-11-08
First cut of plugin tool

stevemessick:
@pq @devoncarew 

Add a new module for the plugin tool. Most of the interesting code is in plugin.dart. The spurious XML files will need to be removed, but I don't know how we're handling them yet.

WIP. A lot of code exists that is unused because this first version drives the ant builder. It is a sort of verbose build-all.sh, with lots of room for improvement. All the commands have been stubbed out, if not fully implemented.


.idea/libraries/Dart_Packages.xml
```diff
@@ -0,0 +1,316 @@
+<component name="libraryTable">
```
devoncarew: Can we remove this file from the PR? It looks like it's not covered by the .gitignore, but it likelt should be.

devoncarew:
(looking at this now)

.idea/libraries/Dart_Packages.xml
```diff
@@ -0,0 +1,316 @@
...
+          <value>
+            <list>
+              <option value="$USER_HOME$/.pub-cache/hosted/pub.dartlang.org/yaml-2.1.13/lib" />
+            </list>
+          </value>
+        </entry>
+      </option>
+    </properties>
+    <CLASSES>
+      <root url="file://$USER_HOME$/.pub-cache/hosted/pub.dartlang.org/analyzer-0.25.0+1/lib" />
```
pq: Is there any way to wild card this?

.idea/libraries/Dart_SDK.xml
```diff
@@ -0,0 +1,26 @@
+<component name="libraryTable">
```
devoncarew: ditto re: removing it

.idea/libraries/Flutter_Plugins.xml
```diff
@@ -0,0 +1,7 @@
+<component name="libraryTable">
```
devoncarew: same

tool/plugin/pubspec.yaml
```diff
@@ -0,0 +1,14 @@
+name: plugin
```
devoncarew: This PR is creating a new `tool/plugin/` sub-dir and module. There's already a pubspec.yaml at the top level, and some dart code in `tool/`. It seems like we should use that existing dart project, put the `plugin` cli code in tool, and - more medium term - plan to migrate the current ad-hoc script in tool/ to be driven through `plugin` (currently, they re-generate our support for displaying flutter icons and colors in the IDE).

Thoughts?

tool/plugin/lib/plugin.dart
```diff
@@ -0,0 +1,535 @@
...
+  String get description => 'Run the tests for the Flutter plugin.';
+
+  Future<int> run() async {
+    super.run();
+    for (var spec in specs) {
+      await spec.artifacts.provision();
+
+      // TODO(devoncarew): implement
+    }
+    return 0;
```
devoncarew: perhaps currently fail with `'unimplemented'`?

tool/plugin/lib/plugin.dart
```diff
@@ -0,0 +1,535 @@
...
+
+  String get name => 'deploy';
+
+  String get description => 'Upload the Flutter plugin to the JetBrains site.';
+
+  Future<int> run() async {
+    super.run();
+
+    // TODO(messick): implement
+    return 0;
```
devoncarew: fail with 'unimplemented'?

tool/plugin/lib/plugin.dart
```diff
@@ -0,0 +1,535 @@
...
+  GenCommand(this.runner);
+
+  String get name => 'gen';
+
+  String get description =>
+      'Generate a valid plugin.xml for the Flutter plugin from the template.';
+
+  Future<int> run() async {
+    // TODO(messick): implement
+    return 0;
```
devoncarew: fail with 'unimplemented'?

.idea/libraries/Dart_Packages.xml
stevemessick: This was generated; the file should be removed from the PR, I think.

tool/plugin/test/plugin_test.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:plugin/plugin.dart';
+import 'package:test/test.dart';
+import 'package:mockito/mockito.dart';
+import 'package:mockito/mirrors.dart';
```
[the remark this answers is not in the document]
stevemessick: Good point. I haven't formatted the tests yet.

.idea/libraries/Dart_Packages.xml
stevemessick: I don't think it should be here but I'm not sure where it came from or how it (or the others) should be managed.

tool/plugin/lib/plugin.dart
stevemessick: I can change that for all these.

tool/plugin/pubspec.yaml
stevemessick: When I started working on this I was getting frustrated by the lack of Dart editing support. Rather than trying to solve the project configuration problem (which I did spend some time on) I decided to side-step it by generating a properly-configured module.

Merging existing tool functionality into the plugin tool makes sense. That looks straight-forward.

I think its cleaner to keep Java code in a Java project and use a Dart module to handle Dart code. We could remove pubspec.yaml and friends from the top level once we've merged everything. But, I'm happy to rearrange things the way you suggest if the project is configured to support Dart editing -- sorting out the configuration issues was a distraction I didn't want.

devoncarew: It is perhaps messier (to combine both Dart and Java). Happy either way I suppose.

We may want to consider a re-org of the repo a bit. I think it might be a bit cleaner to have 3 sibling modules (rather than one at the root and two - AS and Dart - as children). That would be a big re-org, so may not be justified.

stevemessick:
I'm having merge problems. I'm going to delete this PR and start fresh. I think I've addressed all review comments so the next PR should be mergeable.

PR #1490, opened 2017-11-09
Build

stevemessick:
@pq @devoncarew 

Dup of previous sources for `plugin` with comments addressed. PTAL

One thing I forgot to mention was where to put tests for this code. That was actually the deciding factor for creating a new module. I realized that a pure Dart module made questions about organization go away.

Following up on a comment by @devoncarew: It is an interesting question to ask how hard it would be to reorganize the project with three sibling modules instead of two nested ones. The Android Studio project is organized that way, although they do have a couple nested modules for tests. We should consider doing that, I think.


stevemessick:
So, just for grins I tried to reorganize the modules. I had a 'flutter' group with all three modules within it. Looked fine. Then I decided to undo it. Moving a module back to top-level totally destroyed all the  other organization. Now I have this huge list that is impossible to find anything. I'll have to checkout the sources and rebuild to fix it, I guess. The useful point of that was that reorganizing them did not cause any visible changes for git. We'll have to remove .idea from .gitignore to make those changes stick.

devoncarew:
Interesting, so much of the work could be done just by manipulating IntelliJ metadata?

If there are change to files that are committed, but covered by a gitignore, you should still see diffs for them. Files that are not committed and covered by a gitignore can be added to git via `git add foo -f`.

But yeah, removing the exclusion of .idea would let us better see the changes, even if we added it back after.

PR #1491, opened 2017-11-09
update the build to allow the android studio zip file binary to work for IntelliJ as well

devoncarew:
- update the build to allow the android studio zip file binary to work for IntelliJ as well
- extend the version range of the AS binary to span 2017.1 through 2017.3
- ensure that the intellij source are compiled against 2017.1
- make the AS module dep listed in plugin.xml be the `com.android.tools.apk` one (important for the wizard dialog to function)

@stevemessick 

Steve, this is a bit of a pragmatic change, but does get us though the next release.

### DIFF
--- a/build-all.sh
+++ b/build-all.sh
@@ -1,15 +1,30 @@
 #!/bin/bash
 
-# Build two distribution files
+# Build two distribution files:
 #   artifacts/flutter-intellij.jar
 #   artifacts/flutter-studio.zip
 
+# fast fail the script on failures.
+set -e
+
+# clean the build
 rm -rf build/*
-ant -Ddart.plugin.version=171.4424.10 -Didea.version=171.4408382 -Didea.product=android-studio-ide \
-  -DDEPENDS=\<depends\>com.android.tools.apk\</depends\> \
-  -DPLUGINID=io.flutter.as -DSINCE=171.1 -DUNTIL=171.\*
+
+# build for Android Studio
+# TODO(devoncarew): The version range below (supported for Android Studio) is
+# a temporary measure.
+ant \
+  -Ddart.plugin.version=171.4424.10 \
+  -Didea.version=171.4408382 \
+  -Didea.product=android-studio-ide \
+  -DSINCE=171.1 -DUNTIL="173.*"
 mv build/flutter-studio.zip artifacts
 
+# build for IntelliJ
+# Note: built against IntelliJ 2017.1 for runtime compatibility with Android Studio.
 rm -rf build/*
-ant -Ddart.plugin.version=172.3317.48 -Didea.version=2017.2 -Didea.product=ideaIC
+ant \
+  -Ddart.plugin.version=172.3317.48 \
+  -Didea.version=2017.1 \
+  -Didea.product=ideaIC
 mv build/flutter-intellij.jar artifacts
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -8,20 +8,35 @@
 
   <category>Custom Languages</category>
 
-  <version>18.3</version>
+  <version>18.4</version>
 
   <idea-version since-build="171.1" until-build="173.*"/>
 
+  <depends>Dart</depends>
+
   <!-- plugin compatibility -->
   <!-- see: http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html -->
-  
+
+  <!-- Contributes IDEA-specific features and implementations. -->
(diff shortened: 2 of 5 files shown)


resources/META-INF/plugin.xml.template
```diff
@@ -8,20 +8,35 @@
...
-  <version>18.3</version>
+  <version>18.4</version>
 
   <idea-version since-build="@SINCE@" until-build="@UNTIL@"/>
 
+  <depends>Dart</depends>
+
   <!-- plugin compatibility -->
   <!-- see: http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html -->
-  @DEPENDS@
```
stevemessick: Won't this cause 'Flutter' to show up twice in the plugin manager?

resources/META-INF/plugin.xml.template
```diff
@@ -8,20 +8,35 @@
 
   <category>Custom Languages</category>
 
-  <version>18.3</version>
+  <version>18.4</version>
```
stevemessick: 18.5 or 19?

resources/META-INF/plugin.xml
```diff
@@ -8,20 +8,35 @@
...
   <!-- see: http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html -->
-  
+
+  <!-- Contributes IDEA-specific features and implementations. -->
+  <depends optional="true" config-file="idea-contribs.xml">com.intellij.modules.java</depends>
+
+  <!-- Contributes Android Studio-specific features and implementations. -->
+  <!--suppress PluginXmlValidity -->
+  <depends optional="true" config-file="studio-contribs.xml">com.android.tools.apk</depends>
+  <!-- note that com.android.tools.apk as implies com.intellij.modules.androidstudio -->
```
stevemessick: also instead of as?

build-all.sh
```diff
@@ -1,15 +1,30 @@
...
-  -DPLUGINID=io.flutter.as -DSINCE=171.1 -DUNTIL=171.\*
+
+# build for Android Studio
+# TODO(devoncarew): The version range below (supported for Android Studio) is
+# a temporary measure.
+ant \
+  -Ddart.plugin.version=171.4424.10 \
+  -Didea.version=171.4408382 \
+  -Didea.product=android-studio-ide \
+  -DSINCE=171.1 -DUNTIL="173.*"
```
stevemessick: Are you sure loading a plugin compiled for 2017.1 into 2017.2 is able to create projects?

resources/META-INF/plugin.xml
```diff
@@ -8,20 +8,35 @@
 
   <category>Custom Languages</category>
 
-  <version>18.3</version>
+  <version>18.4</version>
```
stevemessick: 18.4 is the current version. I thought I heard 19 mentioned.

build-all.sh
devoncarew: Yup! I did check in Android Studio, IntelliJ, and IntelliJ EAP. I was running through lots of builds locally, so definitely would be good to double-check after this lands (as part of the normal release process).

resources/META-INF/plugin.xml
devoncarew: Yup; I'm updating here to the current released version (we only rev'd to 18.4 on the release_18 branch, so this is moving that info to master).

@pq will likely follow up with a PR to rev to 19.0

resources/META-INF/plugin.xml
devoncarew: oops, yup

build-all.sh
stevemessick: Good to know, thanks!

resources/META-INF/plugin.xml
pq: 👍

resources/META-INF/plugin.xml.template
devoncarew: 18.4, to match the version currently in `release_18`

pq: I'll push out a 19.0 once this lands.

resources/META-INF/plugin.xml.template
devoncarew: This is shipping support for both under the `io.flutter` id; I think we'll just see this once total. Totally happy to re-evaluate the plugin distro plan once we have this release under our belts :)

devoncarew:
> LGTM, assuming this does not cause a duplicate listing in the plugin manager.

Thanks for the review!

PR #1492, opened 2017-11-09
Use the other Nullable

stevemessick:
@jacob314 

I just changed the import statement to use JetBrains Nullable. I got compile errors from the javax one.


devoncarew:
I have also imported in the wrong NotNull a few times. I wonder if we should have a task on travis that will fail if we import `import javax.annotation.Nullable;` or the wrong NotNull one?

stevemessick:
@devoncarew Good idea. Makes me wonder if maybe IntelliJ doesn't already have something built-in.

PR #1494, opened 2017-11-09
M19

pq:
<li>fixed an issue with reload when mutiple project windows are open</li>
  <li>fixed running Flutter tests in nested groups</li>
  <li>fixed miscellaneous project wizard issues</li>
  <li>fix to ensure we don't create Flutter library entries for non-Flutter projects</li>
  <li>fixed project name validation in the new project creation wizard to be more performant</li>
  <li>fixed project opening to only open main.dart if no other editors are open</li>
  <li>fix to limit Flutter icon contributions to Flutter projects</li>
  <li>reload on save updated to ignore errors in test files</li>
  <li>IDEA EAP support</li>
  <li>fix to give restarted apps focus on iOS</li>
  <li>miscellaneous Android Studio support fixes</li>
  <li>fixed check for Flutter tests to not mis-identify vanilla Dart tests</li>
  <li>improved error reporting on project creation failures</li>

@stevemessick @devoncarew

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -8,9 +8,9 @@
 
   <category>Custom Languages</category>
 
-  <version>18.4</version>
+  <version>19.0</version>
 
-  <idea-version since-build="171.1" until-build="173.*"/>
+  <idea-version since-build="172.1" until-build="173.*"/>
 
   <depends>Dart</depends>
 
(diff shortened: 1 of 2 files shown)


resources/META-INF/plugin.xml
```diff
@@ -8,9 +8,9 @@
 
   <category>Custom Languages</category>
 
-  <version>18.4</version>
+  <version>19.0</version>
 
-  <idea-version since-build="171.1" until-build="173.*"/>
+  <idea-version since-build="172.1" until-build="173.*"/>
```
stevemessick: Please keep this at 171.1. It causes me much grief if it is 172.1.

pq: Oddly, I didn't do this intentionally.  Is it possible that `ant plugin.xml` did?

stevemessick: Probably. I think the build script is still producing two files but we're only publishing the zip file.

devoncarew:
👍 lgtm

PR #1501, opened 2017-11-10
plugin deploy

stevemessick:
@pq @devoncarew 

Implement `plugin --release=NN deploy`. The argument is required. The current git branch should be named "release_NN" -- but that is not being enforced yet.

The `upload()` method has not been tested yet.

There are some formatting changes from running it through dartfmt.

### DIFF
--- a/tool/plugin/lib/plugin.dart
+++ b/tool/plugin/lib/plugin.dart
@@ -11,6 +11,16 @@ import 'package:path/path.dart' as p;
 
 import 'src/lint.dart';
 
+const plugins = const {
+  'io.flutter': '9212',
+  'io.flutter.as': '10139',
+};
+
+const files = const {
+  'io.flutter': 'flutter-intellij.jar',
+  'io.flutter.as': 'flutter-studio.zip',
+};
+
 main(List<String> args) async {
   BuildCommandRunner runner = new BuildCommandRunner();
 
@@ -30,11 +40,9 @@ main(List<String> args) async {
 }
 
 class BuildCommandRunner extends CommandRunner {
-
   BuildCommandRunner()
-      : super(
-      'plugin',
-      'A script to build, test, and deploy the Flutter IntelliJ plugin.') {
+      : super('plugin',
+            'A script to build, test, and deploy the Flutter IntelliJ plugin.') {
     argParser.addOption(
       'release',
       abbr: 'r',
@@ -43,10 +51,11 @@ class BuildCommandRunner extends CommandRunner {
     );
   }
 
-  Future<int> javac({List sourcepath,
-    String destdir,
-    List classpath,
-    List<String> sources}) async {
+  Future<int> javac(
+      {List sourcepath,
+      String destdir,
+      List classpath,
+      List<String> sources}) async {
     //final Directory javacDir = new Directory('artifacts/${artifacts.javac.output}');
 
     final List<String> args = [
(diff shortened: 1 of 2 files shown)


tool/plugin/lib/plugin.dart
```diff
@@ -81,14 +90,28 @@ abstract class ProductCommand extends Command {
 
   bool get isForIntelliJ => argResults['ij'];
 
-  String get release => globalResults['release'];
+  String get release {
+    var rel = globalResults['release'];
+    if (rel != null && rel.startsWith('=')) {
```
devoncarew: This should come in from the args package w/o the leading `=`.

tool/plugin/lib/plugin.dart
```diff
@@ -97,12 +120,17 @@ class AntBuildCommand extends ProductCommand {
 
   AntBuildCommand(this.runner);
 
-  String get name => 'build';
+  String get name => 'abuild';
```
devoncarew: is `abuild` intended?

tool/plugin/lib/plugin.dart
```diff
@@ -210,9 +250,62 @@ class DeployCommand extends ProductCommand {
...
+    Directory directory = Directory.systemTemp.createTempSync('plugin');
+    tempDir = directory.path;
+    var file = new File('$tempDir/.content');
+    file.writeAsStringSync(password, flush: true);
+
+    var value = 0;
+    for (var spec in specs) {
+      String filePath = archiveFilePath(spec);
+      value = await upload(filePath, plugins[spec.pluginId]);
+      if (value != 0) return value;
```
devoncarew: Will `upload()` have already printed a failure message?

tool/plugin/lib/plugin.dart
```diff
@@ -210,9 +250,62 @@ class DeployCommand extends ProductCommand {
...
+    log("uploading $filePath");
+    String args = '''
+-i 
+-F userName="${username}" 
+-F password="<${tempDir}/.content" 
+-F pluginId="$pluginNumber" 
+-F file="@$filePath" 
+"https://plugins.jetbrains.com/plugin/uploadPlugin"
+''';
+    return await exec('curl', args.split(r'\w'));
```
devoncarew: Since this is calling a command with a password, we should make sure that `exec` doesn't log the command. I would either parameterize `exec`, or in-line the body of exec here. That 2nd option might be safer in case someone comes along in the future and refactors exec.

tool/plugin/lib/plugin.dart
stevemessick: True, but I can't keep track of when I need '=' and when I don't. If I type "-r=19" then there's an additional '=' to be removed. I dislike programs that are too unnecessarily strict in parsing options. I can remove it you want the strict behavior.

tool/plugin/lib/plugin.dart
stevemessick: For now, to distinguish from 'build'. It will be removed eventually.

tool/plugin/lib/plugin.dart
stevemessick: Possibly. The most likely failure is that the file hasn't been built and it checks for that. There is currently no error printed if `curl` fails but there should be.

tool/plugin/lib/plugin.dart
stevemessick: Good catch. Thanks.

stevemessick:
Thanks for the suggestions. PTAL.

tool/plugin/lib/plugin.dart
devoncarew: No, I'm just surprised that this isn't somehow already handled by `package:args`; sounds like it's not -

tool/plugin/lib/plugin.dart
devoncarew: 👍

tool/plugin/lib/plugin.dart
```diff
@@ -11,6 +11,16 @@ import 'package:path/path.dart' as p;
 
 import 'src/lint.dart';
 
+const plugins = const {
+  'io.flutter': '9212',
+  'io.flutter.as': '10139',
```
devoncarew: Sounds like this is encoding the current ant build.xml functionality (and we'll likely move to one plugin ID, and building for specific platform version in the future: 2017.1, 2017.2, 2017.3).

stevemessick: Exactly. I'll change this around when implementing `plugin build`. Including putting constants like this into a file.

PR #1503, opened 2017-11-14
Add release validity checks

stevemessick:
@devoncarew @pq 

Implement the release-mode validity checks. Add --cwd to allow testing. Add tests.

Most travis builds passed. One failed with:
```bash
HTTP error 500: Internal Server Error at https://pub.dartlang.org/api/packages/collection.
This is likely a transient error. Please try again later.
```

### DIFF
--- a/tool/plugin/lib/plugin.dart
+++ b/tool/plugin/lib/plugin.dart
@@ -8,6 +8,7 @@ import 'dart:io';
 import 'package:args/args.dart';
 import 'package:args/command_runner.dart';
 import 'package:path/path.dart' as p;
+import 'package:git/git.dart';
 
 import 'src/lint.dart';
 
@@ -20,6 +21,8 @@ const files = const {
   'io.flutter': 'flutter-intellij.jar',
   'io.flutter.as': 'flutter-studio.zip',
 };
+
+String rootPath;
 
 main(List<String> args) async {
   BuildCommandRunner runner = new BuildCommandRunner();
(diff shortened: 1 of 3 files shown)


tool/plugin/lib/plugin.dart
```diff
@@ -8,6 +8,7 @@ import 'dart:io';
 import 'package:args/args.dart';
 import 'package:args/command_runner.dart';
 import 'package:path/path.dart' as p;
+import 'package:git/git.dart';
```
pq: Sort?

tool/plugin/pubspec.yaml
```diff
@@ -8,6 +8,7 @@ environment:
 dependencies:
   args: "^0.13.7"
   path: "^1.4.2"
+  git: "^0.4.4+2"
```
pq: Sort?

tool/plugin/lib/plugin.dart
```diff
@@ -160,7 +177,7 @@ class BuildCommand extends ProductCommand {
 
   Future<int> doit() async {
```
pq: General question: do we need to be async?  What bits will take advantage of running asynchronously?  (Sorry if this is obvious!)

pq:
FWIW: I restarted the build and all is green now 👍

tool/plugin/lib/plugin.dart
stevemessick: These methods run a lot of shell commands. Like curl, javac, jar, cp, git, etc. We need to use async to keep it all in order.

stevemessick:
Oops, forgot to run the formatter. I'll do that for the next PR. Thanks.

PR #1505, opened 2017-11-14
Add a product matrix file

stevemessick:
@devoncarew @pq 

Add `product-matrix.json`, which includes all the values that change when building for different platforms. Also, formatting.


stevemessick:
Starting over, after sorting.

PR #1510, opened 2017-11-16
Inspector View widget filtering basics.

pq:
A little hacky but gets us into a place to start experimenting in earnest.

@devoncarew @stevemessick 

FYI: @jacob314

### DIFF
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.java
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.java
@@ -39,6 +39,7 @@ import java.net.URISyntaxException;
 
 public class FlutterSettingsConfigurable implements SearchableConfigurable {
   private static final boolean DISABLE_INSPECTOR = true;
+  public static final boolean WIDGET_FILTERING_ENABLED = false;
 
   private static final Logger LOG = Logger.getInstance(FlutterSettingsConfigurable.class);
 
--- a/src/io/flutter/inspector/DiagnosticsNode.java
+++ b/src/io/flutter/inspector/DiagnosticsNode.java
@@ -5,6 +5,7 @@
  */
 package io.flutter.inspector;
 
+import com.google.common.collect.Lists;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonNull;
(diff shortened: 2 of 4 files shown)


src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -39,6 +39,7 @@
 
 public class FlutterSettingsConfigurable implements SearchableConfigurable {
   private static final boolean DISABLE_INSPECTOR = true;
+  public static final boolean WIDGET_FILTERING_ENABLED = false;
```
devoncarew: Looks good for getting a sense of how the filtering works. Longer term, do we think this is something the user would want to toggle? If so, it should potentially exist as an option in the overflow menu in the Flutter view.

pq: Absolutely.  We're not ready to support toggling of filters just yet (I need to plumb in invalidation and redraw) but once that's there, for sure we want this in the UI.

jacob314:
Looks good. You also need to fix handling the case where a path of
DiagnosticsNode objects I'd sent down from the server. I suspect that leads
to the conclusion that filtering should be done server side as with this
interpretation of filtering as following children of filtered nodes, the
server payload for a path is no longer sufficient so you'd need potentially
a large number of server round trips in that case to get the nodes that
should be displayed.

On Nov 17, 2017 6:15 AM, "Phil Quitslund" <notifications@github.com> wrote:

Merged #1510 <https://github.com/flutter/flutter-intellij/pull/1510>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/flutter/flutter-intellij/pull/1510#event-1346106104>,
or mute the thread

jacob314:
The other option that would scale without server side changes is to show
filtered children as (filtered) in grey before they are expanded. When
expanded the logic in your code to show a single child list could reapply.
Other meta discussion question is whether we should make sure filtering
does not make two nodes that we're not siblings appear to be siblings. That
could be misleading.

pq:
> I suspect that leads
to the conclusion that filtering should be done server side as with this
interpretation of filtering as following children of filtered nodes

@jacob314 : a few other things suggest this to me as well.  I expected this approach to be just temporary but I wonder if getting it working on the server side sooner than later might not make sense after all....

Issue #1513, opened 2017-11-17
NPE in FlutterIconProvider.getIcon


bwilkerson:
## What happened

I brought IntelliJ to the front after deleting several directories from disk. I don't think that any of the deleted files/directories were open in IntelliJ, but it's possible I'm wrong.

## Version information

IntelliJ IDEA `2017.2.6` • Flutter plugin `io.flutter 19.1` • Dart plugin `172.4343.25`

Flutter • channel master • https://github.com/flutter/flutter.git
Framework • revision d5d2cdfeef (13 hours ago) • 2017-11-16 22:53:22 -0800
Engine • revision 25912b8482
Tools • Dart 1.25.0-dev.11.0

## Exception

null

```
java.lang.NullPointerException
	at io.flutter.project.FlutterIconProvider.getIcon(FlutterIconProvider.java:44)
	at com.intellij.ide.projectView.impl.nodes.PsiDirectoryNode.setupIcon(PsiDirectoryNode.java:171)
	at com.intellij.ide.projectView.impl.nodes.PsiDirectoryNode.updateImpl(PsiDirectoryNode.java:119)
	at com.intellij.ide.projectView.impl.nodes.AbstractPsiBasedNode.lambda$update$0(AbstractPsiBasedNode.java:164)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:921)
	at com.intellij.ide.projectView.impl.nodes.AbstractPsiBasedNode.update(AbstractPsiBasedNode.java:139)
	at com.intellij.ide.util.treeView.PresentableNodeDescriptor.getUpdatedPresentation(PresentableNodeDescriptor.java:88)
	at com.intellij.ide.util.treeView.PresentableNodeDescriptor.update(PresentableNodeDescriptor.java:41)
	at com.intellij.ide.util.treeView.AbstractTreeBuilder.updateNodeDescriptor(AbstractTreeBuilder.java:577)
	at com.intellij.ide.util.treeView.AbstractTreeUi$22.perform(AbstractTreeUi.java:935)
	at com.intellij.ide.util.treeView.TreeRunnable.run(TreeRunnable.java:36)
	at com.intellij.ide.util.treeView.AbstractTreeUi.execute(AbstractTreeUi.java:1856)
	at com.intellij.ide.util.treeView.AbstractTreeUi.update(AbstractTreeUi.java:931)
[...]

PR #1515, opened 2017-11-19
fix an NPE in the icon provider when deleting projects

devoncarew:
- guard against a null getBaseDir() from a Project when the user is deleting projects (fix https://github.com/flutter/flutter-intellij/issues/1513)

@pq @stevemessick

### DIFF
--- a/src/io/flutter/actions/FlutterExternalIdeActionGroup.java
+++ b/src/io/flutter/actions/FlutterExternalIdeActionGroup.java
@@ -47,14 +47,11 @@ public class FlutterExternalIdeActionGroup extends DefaultActionGroup {
   }
 
   private static boolean isProjectDirectory(@NotNull VirtualFile file, @Nullable Project project) {
-    if (!file.isDirectory()) {
-      return false;
-    }
-    //noinspection SimplifiableIfStatement
-    if (project == null || project.getBaseDir() == null) {
+    if (!file.isDirectory() || project == null) {
       return false;
     }
 
-    return project.getBaseDir().getPath().equals(file.getPath());
+    final VirtualFile baseDir = project.getBaseDir();
+    return baseDir != null && baseDir.getPath().equals(file.getPath());
   }
 }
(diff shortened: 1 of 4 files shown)


src/io/flutter/project/FlutterIconProvider.java
```diff
@@ -41,8 +41,10 @@ public Icon getIcon(@NotNull final PsiElement element, @Iconable.IconFlags final
       if (!file.isInLocalFileSystem()) return null;
 
       // Project root.
-      if (Objects.equals(file.getPath(), project.getBaseDir().getPath())) {
-        return FlutterIcons.Flutter;
+      if (project.getBaseDir() != null) {
```
pq: In case there's a race condition, maybe store `project.getBaseDir()` in a local?

    basedir = project.getBaseDir()

?

src/io/flutter/run/test/TestLaunchState.java
```diff
@@ -117,7 +117,13 @@ protected ConsoleView createConsole(@NotNull Executor executor) throws Execution
 
   private String getBaseDir() {
     final PubRoot root = config.getFields().getPubRoot(config.getProject());
-    return root != null ? root.getPath() : config.getProject().getBaseDir().getPath();
+    if (root != null) {
+      return root.getPath();
+    }
+    if (config.getProject().getBaseDir() == null) {
```
pq: Same here?

src/io/flutter/project/FlutterIconProvider.java
devoncarew: makes sense, will do

src/io/flutter/run/test/TestLaunchState.java
devoncarew: done!

src/io/flutter/actions/FlutterExternalIdeActionGroup.java
```diff
@@ -47,6 +47,14 @@ protected static boolean isIOsDirectory(@NotNull VirtualFile file) {
   }
 
   private static boolean isProjectDirectory(@NotNull VirtualFile file, @Nullable Project project) {
-    return file.isDirectory() && project != null && project.getBaseDir().getPath().equals(file.getPath());
+    if (!file.isDirectory()) {
+      return false;
+    }
+    //noinspection SimplifiableIfStatement
+    if (project == null || project.getBaseDir() == null) {
```
pq: Probably not a big deal, but perhaps consider doing the same here?

devoncarew: good catch; updated

Issue #1520, opened 2017-11-21
Options are not getting passed in project creation.


pq:
See: https://github.com/flutter/flutter/issues/12565.

Reproduced in IDEA CE 2017.2

PR #1521, opened 2017-11-21
Fix project creation to respect custom options.

pq:
Fixes: #1520.

@stevemessick


pq:
Excellent idea.  I'll follow-up with fixes to the test script.

Thanks!

Issue #1526, opened 2017-11-22
"Open in Xcode..." opens project (not workspace) for plugin projects.


pq:
Currently when opening an Xcode project for a plugin project from IntelliJ, we scan `example/ios/` and look for the first project file.  I *believe* this would work better if we preferred the workspace file (e.g., `Runner.xcworkspace`).

Paging @xster, @collinjackson, or @cbracken to confirm.

cbracken:
@pq the workspace file is definitely what you want to open, if present.

If the logic in question is relatively old, it's possible it dates back to the pre-CocoaPods days when the workspace file may not have been part of our `flutter create` template.

pq:
Awesome.  Thanks a million! 👍

PR #1527, opened 2017-11-22
Prefer opening xcode workspace files (#1526).

pq:
Fixes: #1526.

@stevemessick


pq:
Aha!  Actually, yes it likely should. 👍

PR #1530, opened 2017-11-27
improve parsing of the daemon json stream

devoncarew:
Improve parsing of the daemon json stream:
- we now handle cases where larger json blobs would be split across two or more messages; previously we'd drop the messages on the floor; now we combine and parse correctly
- handle a case where flutter_tools occasionally sends a newline before any other output when starting an app

@stevemessick @jwren @pq

### DIFF
--- a/src/io/flutter/run/test/DebugTestRunner.java
+++ b/src/io/flutter/run/test/DebugTestRunner.java
@@ -26,6 +26,7 @@ import com.jetbrains.lang.dart.util.DartUrlResolver;
 import io.flutter.run.PositionMapper;
 import io.flutter.sdk.FlutterSdk;
 import io.flutter.settings.FlutterSettings;
+import io.flutter.utils.StdoutJsonParser;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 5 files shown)


src/io/flutter/run/test/DebugTestRunner.java
```diff
@@ -102,17 +103,21 @@ public void onTextAvailable(ProcessEvent event, Key outputType) {
...
-          if (!text.startsWith("[{") || !text.endsWith("}]")) {
-            return; // Ignore anything not in our expected format.
-          }
+          stdoutParser.appendOutput(text);
 
-          final String json = text.substring(1, text.length() - 1);
-          dispatchJson(json);
+          for (String line : stdoutParser.getAvailableLines()) {
+            if (line.startsWith("[{")) {
+              line = line.trim();
```
stevemessick: Is there any chance the json producer could break the stream mid-string? If so, this trim() could discard meaningful whitespace.

devoncarew: I didn't document, but `getAvailableLines()` will return whole, concatenated lines (so, they will include the `[{...}]`.

stevemessick: Thanks. I haven't read the parser code yet.

PR #1532, opened 2017-11-27
Add some lint rules and clean up the code

stevemessick:
As promised, code clean-up driven by additional lint rules.

I'm a bit conflicted about "type_annotate_public_apis". It caused addition of types where I had actually not wanted them. Oh well...perhaps the style guide suggests them -- I don't have it memorized.

@pq @devoncarew

### DIFF
--- a/tool/plugin/analysis_options.yaml
+++ b/tool/plugin/analysis_options.yaml
@@ -11,3 +11,9 @@ linter:
     - test_types_in_equals
     - unrelated_type_equality_checks
     - valid_regexps
+    - await_only_futures
+    - directives_ordering
+    - omit_local_variable_types
+    - type_annotate_public_apis
+    - type_init_formals
+    - unawaited_futures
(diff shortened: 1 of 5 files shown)


tool/plugin/analysis_options.yaml
```diff
@@ -11,3 +11,9 @@ linter:
     - test_types_in_equals
     - unrelated_type_equality_checks
     - valid_regexps
+    - await_only_futures
+    - directives_ordering
+    - omit_local_variable_types
+    - type_annotate_public_apis
+    - type_init_formals
+    - unawaited_futures
```
pq: FWIW folks generally alphabetize these since it makes the list a bit easier to maintain.

tool/plugin/test/plugin_test.dart
```diff
@@ -136,18 +136,18 @@ BuildCommandRunner makeTestRunner() {
 class TestAntBuildCommand extends AntBuildCommand {
   TestAntBuildCommand(runner) : super(runner);
 
-  doit() {}
+  Future<int> doit() async => new Future(() => 0);
```
pq: Curious if `new Future.value(0);` is just as good?

stevemessick: Ha! I missed that. Looks like another clean-up pass is needed, someday. Too bad there's no lint rule for that.

I sorted the rules and now I can't find that comment to reply to, so this will have to do.

PR #1537, opened 2017-11-28
GUI Testing

stevemessick:
@devoncarew @pq 

It took a while to get the GUI tests working again. Even thought this still rough in places I'd like to guard against future bit-rot by getting this into the repo.

FlutterCreateState is not yet being used and may be deleted. I thought it would be needed to sync project creation with test states but so far things work without it.

The changes to the forms, adding <labelFor>, make widget matching work. This is a trick we need to add to each label associated with a text field in a form that has tests.

ProjectWizardTest (in the main Flutter test directory) is the beginning of an experiment to make tests for non-Android Studio UI elements. It isn't implemented yet.

IIRC these tests have about 90% coverage for the code specific to Android Studio. (There's a comment somewhere that details classes with less, most have more.)


devoncarew:
Sorry for the delay! Will take a look today.

flutter-studio/src/io/flutter/project/FlutterCreateState.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.android.annotations.concurrency.GuardedBy;
+import com.intellij.openapi.components.ServiceManager;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.project.Project;
+import org.jetbrains.annotations.NotNull;
+
+import static com.intellij.openapi.util.text.StringUtil.isNotEmpty;
+
+public class FlutterCreateState {
```
devoncarew: This is a test fixture for use when testing flutter create?

flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/FlutterGuiTestRule.java
```diff
@@ -0,0 +1,351 @@
...
+  private final RobotTestRule myRobotTestRule = new RobotTestRule();
+  private final LeakCheck myLeakCheck = new LeakCheck();
+  private final PropertyChangeListener myGlobalFocusListener = e -> {
+    Object oldValue = e.getOldValue();
+    if ("permanentFocusOwner".equals(e.getPropertyName()) && oldValue instanceof Component && e.getNewValue() == null) {
+      Window parentWindow = oldValue instanceof Window ? (Window)oldValue : SwingUtilities.getWindowAncestor((Component)oldValue);
+      if (parentWindow instanceof Dialog && ((Dialog)parentWindow).isModal()) {
+        Container parent = parentWindow.getParent();
+        if (parent != null && parent.isVisible()) {
+          System.out.println("Focus Listener: Request focus!");
```
devoncarew: I see a few `System.out.println()`s in this PR; d owe still need them (or where they for development)? Should we switch to using Log.info() or similar instead?

flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/FlutterGuiTestRule.java
```diff
@@ -0,0 +1,351 @@
...
+import java.util.ArrayList;
+import java.util.Hashtable;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static com.google.common.truth.TruthJUnit.assume;
+import static org.fest.reflect.core.Reflection.*;
+
+@SuppressWarnings("Duplicates") // Adapted from com.android.tools.idea.tests.gui.framework.GuiTestRule
+public class FlutterGuiTestRule implements TestRule {
```
devoncarew: Can you add brief dartdoc so readers know what this class is for?

flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/FlutterGuiTestRule.java
```diff
@@ -0,0 +1,351 @@
...
+    return new File(GuiTests.getTestProjectsRootDirPath(), projectDirName);
+  }
+
+  @NotNull
+  protected File getTestProjectDirPath(@NotNull String projectDirName) {
+    return new File(GuiTests.getProjectCreationDirPath(), projectDirName);
+  }
+
+  public void cleanUpProjectForImport(@NotNull File projectPath) {
+    File dotIdeaFolderPath = new File(projectPath, Project.DIRECTORY_STORE_FOLDER);
```
devoncarew: I imagine some of this code is tickling our lint settings (like, preferring `final`)?

devoncarew: If we're making fewer changes to this file - to minimize diffs - that makes sense.

flutter-studio/testSrc/io/flutter/tests/util/WizardUtils.java
```diff
@@ -0,0 +1,81 @@
...
+import com.android.tools.idea.tests.gui.framework.fixture.newProjectWizard.NewFlutterProjectWizardFixture;
+import io.flutter.module.FlutterProjectType;
+import org.jetbrains.annotations.NotNull;
+
+public class WizardUtils {
+  private WizardUtils() {
+  }
+
+  public static void createNewApplication(@NotNull FlutterGuiTestRule guiTest) {
+    createNewProject(guiTest, FlutterProjectType.APP, null, null, null, null, null);
```
devoncarew: Perhaps overload `createNewProject()` so we don't need to pass as many `null`s in here?

stevemessick:
Thanks for the review :) I know this was a lot to cover.

flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/FlutterGuiTestRule.java
stevemessick: I'll do that in the next PR. BTW this was copied from a similar class in AS.

flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/FlutterGuiTestRule.java
stevemessick: These are some of the rough spots that need to be cleaned up.

flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/FlutterGuiTestRule.java
stevemessick: Yeah, we may have some conflicts between minimum diffs and lints. That's one aspect I have not thought about yet. If you have a preference I'll be glad to follow it.

flutter-studio/testSrc/io/flutter/tests/util/WizardUtils.java
stevemessick: Do you mean creating a bunch of methods, each calling the next, and adding another null parameter? Or maybe just one more createNewProject() that takes two parameters? That would be used in these three methods. I think that makes sense.

stevemessick:
Wow, I didn't know github had its own merge conflict editor. That was so simple to resolve!

flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/FlutterGuiTestRule.java
devoncarew: Minimizing diff s makes sense to me. We can perhaps turn off specific lints for whole files.

flutter-studio/testSrc/io/flutter/tests/util/WizardUtils.java
devoncarew: Yeah, likely just one overload of createNewProject(), that the three methods could call.

PR #1538, opened 2017-11-28
Add a template variable for version

stevemessick:
The value of the --release argument will be used as the version number. If there is no --release argument then the version tag will be omitted. Also, fix a range error.

Apparently, if all the TODOs in a file have a space before TODO then the TODO viewer of IntelliJ won't find any of them. In other words, do this:
`//TODO`
and not this:
`// TODO`
Fixed that, too.

@devoncarew @pq


stevemessick:
I can't say. We might not need to update any. I forgot I'm using an old version of the Dart plugin, and I haven't checked to see if there is a problem with the current release. (Switching is time-consuming.)

PR #1545, opened 2017-11-29
Fix build

stevemessick:
Make `plugin build` work again.

@devoncarew @jacob314

### DIFF
--- a/tool/plugin/test/plugin_test.dart
+++ b/tool/plugin/test/plugin_test.dart
@@ -120,6 +120,16 @@ void main() {
       var loc = content.indexOf('@');
       expect(loc, -1);
     });
+    test('provision', () async {
+      var runner = makeTestRunner();
+      TestBuildCommand cmd;
+      await runner.run(["-d../..", "build"]).whenComplete(() {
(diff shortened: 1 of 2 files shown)


tool/plugin/test/plugin_test.dart
```diff
@@ -120,6 +120,16 @@ void main() {
       var loc = content.indexOf('@');
       expect(loc, -1);
     });
+    test('provision', () async {
+      var runner = makeTestRunner();
+      TestBuildCommand cmd;
+      await runner.run(["-d../..", "build"]).whenComplete(() {
```
devoncarew: you intend `-d../..` and not something like `-d=../..`?

devoncarew: Nevermind, looks like `-dfoo` is correct, and `--cwd=foo`.

stevemessick: I added a check that makes the '=' allowable but the args parser expects none. If you use the '--' form then it wants the '=' too. I think I use both forms in test code, intentionally. See the tests named 'test' and 'deploy'.

PR #1550, opened 2017-11-30
Fix provision; convert doc

stevemessick:
@devoncarew @jacob314 @pq 

Fix various bugs in provisioning. Make it more resilient to errors.

It can still get into a bad state. The simplest fix is:
```shell
rm -rf artifacts
plugin build
```
Also, I converted the documentation to markdown and added it. I couldn't find an easy way to wrap the lines so the raw markdown is unreadable. However, it looks good in a proper reader. As I update it I will wrap the lines.

### DIFF
--- a/tool/plugin/lib/plugin.dart
+++ b/tool/plugin/lib/plugin.dart
@@ -432,6 +432,13 @@ class ArtifactManager {
           log('download failed');
           break;
         }
+        var archiveFile = new File(path);
+        if (!archiveFile.existsSync() || archiveFile.lengthSync() < 200) {
+          log('archive file not found: $base/${artifact.file}');
+          archiveFile.deleteSync();
+          result = 1;
+          break;
+        }
       }
 
       // clear unpacked cache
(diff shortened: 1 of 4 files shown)


tool/plugin/lib/plugin.dart
```diff
@@ -432,6 +432,13 @@ class ArtifactManager {
           log('download failed');
           break;
         }
+        var archiveFile = new File(path);
+        if (!archiveFile.existsSync() || archiveFile.lengthSync() < 200) {
+          log('archive file not found: $base/${artifact.file}');
```
jacob314: nit: why do files less than length 200 need to be treated as missing?

stevemessick: If the file is missing then the downloader returns an XML file, length 127, that contains an error message. I considered analyzing the XML to make sure it was reporting a missing-file error, but didn't bother. The files are all quite large.

jacob314:
Makes sense. Might want to add a comment explaining why 200 is used for the
next person who touches the code.

stevemessick:
Good point! Done locally.

PR #1553, opened 2017-12-01
Make loading Widget Catalog work in Android Studio.

jacob314:
Thanks Steve for identifying the issue!
Problem was multiple jars in Android Studio mean that only some ways of loading resources will work.

### DIFF
--- a/src/io/flutter/inspector/FlutterWidget.java
+++ b/src/io/flutter/inspector/FlutterWidget.java
@@ -5,6 +5,8 @@
  */
 package io.flutter.inspector;
 
+import com.google.common.base.Charsets;
+import com.google.common.io.ByteStreams;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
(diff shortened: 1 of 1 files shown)


src/io/flutter/inspector/FlutterWidget.java
```diff
@@ -160,7 +160,8 @@ private void init() {
       try {
         // Local copy of: https://github.com/flutter/website/tree/master/_data/catalog/widget.json
         final URL resource = getClass().getResource("widgets.json");
-        final String content = new String(Files.readAllBytes(Paths.get(resource.toURI())));
+        final byte[] contentBytes = ByteStreams.toByteArray((InputStream)resource.getContent());
```
devoncarew: DBC: Perhaps add a comment about why we're using resource.getContent() instead of toUri()? This would reduce the chance of re-introducing this issue, here or elsewhere.

PR #1560, opened 2017-12-04
misc inspector UI tweaks

devoncarew:
Misc inspector UI tweaks:
- add generated icons for widgets without specific icons
- bug fix to the CustomIcon class
- move the ability to bring a simulator to the foreground into a new `bringToFront()` method
- when the user starts inspect mode, bring the ios simulator to the foreground
- don't show the top level `[root]` node
- reduce the width of the 'Properties' table (give more space to the values)
- give the table rows slightly more vertical space
- in the elements tab, have `child n` prefixes have the same look as the `child` prefix
- tweak the look of nodes in the widget tree; node types are no longer bold, and the suffix id has a space between he name, and shows as gray (supplemental information)
- null values show as empty instead of w/ text `null`

screen shot 2017-12-03 at 12 42 33 pm

screen shot 2017-12-03 at 12 42 38 pm

screen shot 2017-12-03 at 12 49 12 pm

Comments and thoughts welcome, esp. wrt whether the icon changes are improvements, and the table l+f changes :)

@jacob314 @pq

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -29,10 +29,9 @@ import io.flutter.FlutterBundle;
 import io.flutter.FlutterInitializer;
 import io.flutter.inspector.InspectorService;
 import io.flutter.run.daemon.FlutterApp;
+import io.flutter.run.daemon.FlutterDevice;
 import io.flutter.settings.FlutterSettings;
 import io.flutter.utils.VmServiceListenerAdapter;
-import org.dartlang.vm.service.VmServiceListener;
-import org.dartlang.vm.service.element.Event;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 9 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -280,6 +285,14 @@ protected void perform(AnActionEvent event) {
 
   protected void perform(AnActionEvent event) {
     view.getFlutterApp().callBooleanExtension("ext.flutter.debugWidgetInspector", isSelected(event));
+
+    // If toggling inspect mode on, bring any device to the forground.
```
pq: => foreground

src/io/flutter/view/InspectorPanel.java
```diff
@@ -45,6 +47,9 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+// TODO(devoncarew): Should we filter out the CheckedModeBanner node type?
+// TODO(devoncarew): Should we filter out the WidgetInspector node type?
```
pq: 👍

jacob314: sgtm. 
Filtering out the WidgetInspector shouldn't be needed at some point once an inversion of control occurs on the package:flutter side but it is reasonable to filter it for now. Question is though, what happens if you trigger an inspect action on a widget under the WidgetInspector which can currently happen.

src/io/flutter/view/InspectorPanel.java
```diff
@@ -518,9 +524,11 @@ public void showProperties(DiagnosticsNode diagnostic) {
           // be loaded.
           return;
         }
+
+        properties.sort(Comparator.comparing(DiagnosticsNode::getName));
```
jacob314: If we sort by name, we should still show high priority properties first over debug level properties.

devoncarew: How common is the different levels of priorities? Are the priorities for a property always the same or can they change (like changing to an error)?

I like lexical sorting as it's always a consistent order, and predictable for the user (they'll know where a property is in the list w/o having to know its priority). I'm wondering if we should instead call out importance via color/icon/styles instead of order.

devoncarew:
landing; we can keep iterating (esp. re: property display)

PR #1562, opened 2017-12-04
Move GUI test to studio module

stevemessick:
@devoncarew @pq 

Moves experimental code to where it belongs. Still unimplemented.

We need to run UI tests using the framework in AS, which means we need the classpath of the studio module.


devoncarew:
I think you can delete this line now?

https://github.com/flutter/flutter-intellij/blob/master/flutter-intellij.iml#L18

stevemessick:
I added Devon's suggestion.

PR #1570, opened 2017-12-05
Show color properties with a nice color swatch icons

jacob314:
Gray out unimportant properties and highlight warning and error properties.
The color icon code is a bit uglier than I'd like because I don't see an obvious way to center the icons so I had to make them have a margin so centering vertically wasn't needed.

### DIFF
--- a/src/io/flutter/view/InspectorPanel.java
+++ b/src/io/flutter/view/InspectorPanel.java
@@ -12,10 +12,7 @@ import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.ui.Splitter;
 import com.intellij.openapi.util.Computable;
 import com.intellij.openapi.util.Disposer;
-import com.intellij.ui.ColoredTreeCellRenderer;
-import com.intellij.ui.PopupHandler;
-import com.intellij.ui.ScrollPaneFactory;
-import com.intellij.ui.SimpleTextAttributes;
+import com.intellij.ui.*;
 import com.intellij.ui.dualView.TreeTableView;
 import com.intellij.ui.treeStructure.Tree;
 import com.intellij.ui.treeStructure.treetable.ListTreeTableModelOnColumns;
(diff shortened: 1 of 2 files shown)


jacob314:
Screenshots:
image

image

src/io/flutter/view/InspectorPanel.java
```diff
@@ -522,6 +529,7 @@ public void showProperties(DiagnosticsNode diagnostic) {
         if (throwable != null) {
           // TODO(jacobr): show error message explaining properties could not
           // be loaded.
```
pq: `FlutterMessages` might be your friend there.

src/io/flutter/view/InspectorPanel.java
```diff
@@ -539,6 +547,91 @@ public void showProperties(DiagnosticsNode diagnostic) {
     }
   }
 
+  private static SimpleTextAttributes textAttributesForLevel(DiagnosticLevel level) {
+    switch (level) {
+      case hidden:
+      case fine:
+        return SimpleTextAttributes.GRAYED_ATTRIBUTES;
+      case warning:
+        // TODO(jacobr): would be nice to use a yellow color for level warning.
```
pq: 👍 

Maybe just cache one in this class?


    new SimpleTextAttributes(STYLE_PLAIN, Color.ORANGE);

src/io/flutter/view/InspectorPanel.java
```diff
@@ -539,6 +547,91 @@ public void showProperties(DiagnosticsNode diagnostic) {
...
+      case info:
+      default:
+        return SimpleTextAttributes.REGULAR_ATTRIBUTES;
+    }
+  }
+
+  private static class PropertyNameRenderer extends ColoredTableCellRenderer {
+    @Override
+    protected void customizeCellRenderer(JTable table, @Nullable Object value, boolean selected, boolean hasFocus, int row, int column) {
+      if (value == null) return;
```
pq: Might it be safer to replace this with:

    if (!(value instance of DiagnosticNode)) return;

?

jacob314: I'd rather throw if we get the wrong type of value rather than quietly filtering them out.

src/io/flutter/view/InspectorPanel.java
jacob314: Good idea. Color.ORANGE works fine for both Dracula and the Default theme. Added.

Issue #1572, opened 2017-12-06
Plugin example tests not recognized as Flutter tests.


pq:
Plugin example tests live in `example/test` and are not recognized by our test discovery logic.

![image](https://user-images.githubusercontent.com/67586/33683722-c198dc7a-da80-11e7-892c-44d39e01c19f.png)

PR #1573, opened 2017-12-06
Fix test discovery for plugin example tests (#1572).

pq:
image

🤘 

@stevemessick @devoncarew

### DIFF
--- a/src/io/flutter/FlutterUtils.java
+++ b/src/io/flutter/FlutterUtils.java
@@ -94,7 +94,7 @@ public class FlutterUtils {
     if (candidate == null) return false;
 
     final String relativePath = root.getRelativePath(candidate);
-    return relativePath != null && relativePath.startsWith("test/");
+    return relativePath != null && (relativePath.startsWith("test/") || relativePath.startsWith("example/test/"));
   }
 
   @Nullable
--- a/src/io/flutter/run/test/TestConfigProducer.java
+++ b/src/io/flutter/run/test/TestConfigProducer.java
@@ -89,16 +89,10 @@ public class TestConfigProducer extends RunConfigurationProducer<TestConfig> {
   }
 
   private VirtualFile verifyFlutterTestFile(TestConfig config, ConfigurationContext context, DartFile file) {
-    final PubRoot root = PubRoot.forPsiFile(file);
-    if (root == null) return null;
-
     final VirtualFile candidate = FlutterRunConfigurationProducer.getFlutterEntryFile(context, false, false);
     if (candidate == null) return null;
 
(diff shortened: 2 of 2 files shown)


brianegan:
Could this logic be widened a bit further even? I have a few examples in the my example folder, and I'm not sure this would work for that case? E.g:

`/example/redux/test`, `/example/vanilla/test`, etc

pq:
@brianegan : are these projects in github?  Can you send me a pointer?

Also note that this just gets us part of the way there.  See: #1575

devoncarew:
It _should_ work if `/example/vanilla` has it's own pubspec.

brianegan:
Hey hey -- yep! My repo is here: https://github.com/brianegan/flutter_architecture_samples

If you open the `example/inherited_widgets/test/app_state_test.dart` file, for example, and try to run it, I always get this same type of error:

```
Failed to load "/Users/phillywiggins/lab/flutter_architecture_samples/example/inherited_widget/test/app_state_test.dart":
Unable to spawn isolate: The built-in library 'dart:ui' is not available on the stand-alone VM.
'package:flutter_test/src/binding.dart': error: line 7 pos 1: library handler failed
import 'dart:ui' as ui;
```

I assumed this change would fix that, but if this is unrelated / out of scope please let me know and I'll create a separate issue.

src/io/flutter/FlutterUtils.java
```diff
@@ -94,7 +94,7 @@ public static boolean isInTestDir(DartFile file) {
     if (candidate == null) return false;
 
     final String relativePath = root.getRelativePath(candidate);
-    return relativePath != null && relativePath.startsWith("test/");
+    return relativePath != null && (relativePath.startsWith("test/") || relativePath.startsWith("example/test/"));
```
devoncarew: We may want to tweak this check to:
- find the closest pubspec.yaml to `file`
- validate that `file` is within a `test` dir that's a sibling of the pubspec

devoncarew: It looks like the previous version of `verifyFlutterTestFile()` was doing something very similar to that (looking for the closest containing pubspec.yaml making sure the relative path starts with `test/`), so I'm guess I'm a bit confused about the best fix.

Probably just best to figure out how  to support `test/`, `example/test/`, and `example/foo/test` in a relatively generic way...

pq: Absolutely.  I've been thinking some more about this and agree that this would be a more robust approach. 👍

pq:
Thanks for all of the comments!

I've got a more robust method in mind (see #1583) but will land this in the meantime since it at least fixes #1572.

Expect a more complete fix soon! 👍

PR #1578, opened 2017-12-07
Cleanup code to display colors nicely so that it depends on getting the

jacob314:
actual Dart properties of the Color object instead of depending on parsing
the toString output for Color objects.

The code to get the Dart properties is ugly. Unfortunately we can't just use the Dart fields which would be more cleanly exposed through the Observatory protocol because many common objects we care about use properties not fields for the relevant data. If it becomes a performance bottlekneck (unlikely given this is just for properties) we could get a built in observatory protocol method to access a specified list of Dart properties for an object.

One valid design question is why these properties aren't just specified in the DiagnosticsNode protocol. The answer is that it could be but there are some nice advantages of this approach.  With this method, we can iterate on visual display of property values pretty much arbitrarily from the intellij plugin without having to worry about waiting for matching changes to the flutter repo or version skew issues.

### DIFF
--- a/src/io/flutter/inspector/InspectorService.java
+++ b/src/io/flutter/inspector/InspectorService.java
@@ -118,7 +118,8 @@ public class InspectorService implements Disposable {
 
   /**
    * Returns a CompletableFuture with a Map of property names to Observatory
-   * InstanceRef objects.
+   * InstanceRef objects. This method is shorthand for individually evaluating
+   * each of the getters specified by property names.
    *
    * It would be nice if the Observatory protocol provided a built in method
    * to get InstanceRef objects for a list of properties but this is
--- a/src/io/flutter/inspector/InspectorService.java
+++ b/src/io/flutter/inspector/InspectorService.java
@@ -5,6 +5,7 @@
  */
 package io.flutter.inspector;
 
+import com.google.common.base.Joiner;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonParser;
@@ -16,15 +17,9 @@ import io.flutter.run.FlutterDebugProcess;
 import io.flutter.utils.VmServiceListenerAdapter;
 import org.dartlang.vm.service.VmService;
 import org.dartlang.vm.service.VmServiceListener;
-import org.dartlang.vm.service.element.Event;
-import org.dartlang.vm.service.element.EventKind;
-import org.dartlang.vm.service.element.Instance;
-import org.dartlang.vm.service.element.InstanceRef;
+import org.dartlang.vm.service.element.*;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.CompletableFuture;
 
 /**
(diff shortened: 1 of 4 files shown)


src/io/flutter/inspector/InspectorService.java
```diff
@@ -121,6 +116,55 @@ public void addClient(InspectorServiceClient client) {
...
+      final String objectName = "that";
+      for (String propertyName : propertyNames) {
+        propertyAccessors.add(objectName + "." + propertyName);
+      }
+      sb.append("<Object>[");
+      sb.append(Joiner.on(',').join(propertyAccessors));
+      sb.append("]");
+      final Map<String, String> scope = new HashMap<>();
+      scope.put(objectName, instanceRef.getId());
+      return getInstance(inspectorLibrary.eval(sb.toString(), scope)).thenApplyAsync(
```
devoncarew: Ah, this is shorthand for individually eval'ing each getter?

src/io/flutter/view/InspectorPanel.java
```diff
@@ -538,16 +543,39 @@ public void showProperties(DiagnosticsNode diagnostic) {
           return;
         }
 
-        properties.sort(Comparator.comparing(DiagnosticsNode::getName));
+        if (properties.size() == 0) {
+          // TODO(jacobr): display message that there are no properties
+          // and that no errors occurred.
```
devoncarew: IntelliJ does have a convention for this - just putting a label (like "no properties" or similar) centered in the panel:

screen shot 2017-12-07 at 7 11 23 am

src/io/flutter/view/InspectorPanel.java
```diff
@@ -538,16 +543,39 @@ public void showProperties(DiagnosticsNode diagnostic) {
...
-          }
+          futures[i] = property.getValueProperties();
+          ++i;
         }
-        model.setRoot(root);
+        whenCompleteUiThread(CompletableFuture.allOf(futures), (Void ignored, Throwable errorGettingInstances) -> {
+          if (errorGettingInstances != null) {
+            // TODO(jacobr): show error message explaining properties could not
+            // be loaded.
+            LOG.error(errorGettingInstances);
```
devoncarew: Could also use the centered display for this (w/ different text)?

src/io/flutter/view/InspectorPanel.java
```diff
@@ -608,29 +642,25 @@ protected void customizeCellRenderer(JTable table, @Nullable Object value, boole
...
-          try {
-            Color color = Color.decode("0X" + opaqueColor);
-            int alpha = Integer.parseUnsignedInt(alphaString, 16);
-            color = new Color(color.getRed(), color.getGreen(), color.getBlue(), alpha);
+      final CompletableFuture<Map<String, InstanceRef>> propertiesFuture = node.getValueProperties();
+      if (propertiesFuture.isDone() && !propertiesFuture.isCompletedExceptionally()) {
+        final Map<String, InstanceRef> properties = propertiesFuture.getNow(null);
+        switch (node.getPropertyType()) {
+          case "Color": {
+            int alpha = getIntProperty(properties, "alpha");;
```
devoncarew: `final`s?

jacob314:
One other note is I removed sorting by default as there are a number of
cases where the default order is curated. For example, height and width
should show up next to each other. I will add in sorting by clicking on the
names column or something similar consistent with intellij defaults.

On Dec 7, 2017 7:13 AM, "Devon Carew" <notifications@github.com> wrote:

*@devoncarew* approved this pull request.
------------------------------

In src/io/flutter/inspector/InspectorService.java
<https://github.com/flutter/flutter-intellij/pull/1578#discussion_r155547684>
:

> +   */
+  public CompletableFuture<Map<String, InstanceRef>> getDartObjectProperties(
+      InspectorInstanceRef inspectorInstanceRef, final String[]
propertyNames) {
+    return toObservatoryInstanceRef(inspectorInstanceRef).thenComposeAsync((InstanceRef
instanceRef) -> {
+      final StringBuilder sb = new StringBuilder();
+      final List<String> propertyAccessors = new ArrayList<>();
+      final String objectName = "that";
+      for (String propertyName : propertyNames) {
+        propertyAccessors.add(objectName + "." + propertyName);
+      }
+      sb.append("<Object>[");
+      sb.append(Joiner.on(',').join(propertyAccessors));
+      sb.append("]");
+      final Map<String, String> scope = new HashMap<>();
+      scope.put(objectName, instanceRef.getId());
+      return getInstance(inspectorLibrary.eval(sb.toString(),
scope)).thenApplyAsync(

Ah, this is shorthand for individually eval'ing each getter?
------------------------------

In src/io/flutter/view/InspectorPanel.java
<https://github.com/flutter/flutter-intellij/pull/1578#discussion_r155548256>
:

> @@ -538,16 +543,39 @@ public void showProperties(DiagnosticsNode diagnostic) {
           return;
         }

-        properties.sort(Comparator.comparing(DiagnosticsNode::getName));
+        if (properties.size() == 0) {
+          // TODO(jacobr): display message that there are no properties
+          // and that no errors occurred.

IntelliJ does have a convention for this - just putting a label (like "no
properties" or similar) centered in the panel:

[image: screen shot 2017-12-07 at 7 11 23 am]
<https://user-images.githubusercontent.com/1269969/33722112-f22b1a7c-db1d-11e7-9c19-e8c4b3ee75da.png>
------------------------------

In src/io/flutter/view/InspectorPanel.java
<https://github.com/flutter/flutter-intellij/pull/1578#discussion_r155548498>
:

>          }
-        model.setRoot(root);
+        whenCompleteUiThread(CompletableFuture.allOf(futures), (Void
ignored, Throwable errorGettingInstances) -> {
+          if (errorGettingInstances != null) {
+            // TODO(jacobr): show error message explaining properties could not
+            // be loaded.
+            LOG.error(errorGettingInstances);

Could also use the centered display for this (w/ different text)?
------------------------------

In src/io/flutter/view/InspectorPanel.java
<https://github.com/flutter/flutter-intellij/pull/1578#discussion_r155548660>
:

> -        final String colorValue = node.getDescription();
-        final Matcher matcher = colorPattern.matcher(colorValue);
-        if (matcher.matches()) {
-          final String colorWithAlpha = matcher.group(1);
-          final String alphaString = colorWithAlpha.substring(0, 2);
-          final String opaqueColor = colorWithAlpha.substring(2);
-          try {
-            Color color = Color.decode("0X" + opaqueColor);
-            int alpha = Integer.parseUnsignedInt(alphaString, 16);
-            color = new Color(color.getRed(), color.getGreen(),
color.getBlue(), alpha);
+      final CompletableFuture<Map<String, InstanceRef>>
propertiesFuture = node.getValueProperties();
+      if (propertiesFuture.isDone() &&
!propertiesFuture.isCompletedExceptionally()) {
+        final Map<String, InstanceRef> properties =
propertiesFuture.getNow(null);
+        switch (node.getPropertyType()) {
+          case "Color": {
+            int alpha = getIntProperty(properties, "alpha");;

finals?

—
You are receiving this because you authored the thread.

Reply to this email directly, view it on GitHub
<https://github.com/flutter/flutter-intellij/pull/1578#pullrequestreview-81884088>,
or mute the thread

devoncarew:
> removed sorting by default as there are a number of cases where the default order is curated

👍

src/io/flutter/inspector/InspectorService.java
jacob314: Exactly. Added that to the comment.

src/io/flutter/view/InspectorPanel.java

jacob314:
Added error messages for the various failure states. I expect to iterate on the error messages.
Not sure what naming scheme we should use for FlutterBundle.properties
For now I put them as 
app.inspector.some_message_name
Happy to rename if there is a different scheme I should be using.

src/io/flutter/view/InspectorPanel.java

jacob314:
Added error messages for the various failure states. I expect to iterate on the error messages.
Not sure what naming scheme we should use for FlutterBundle.properties
For now I put them as 
app.inspector.some_message_name
Happy to rename if there is a different scheme I should be using.

PR #1580, opened 2017-12-07
Change version to NN.M.I

stevemessick:
@devoncarew @pq 

Verified that it produces, in the appropriate files:
```xml
<version>20.0.1</version>
<version>20.0.2</version>
<version>20.0.3</version>
```


pq:
🤘

Issue #1575, opened 2017-12-06
Can't run plugin example flutter tests (wrong CWD).


pq:
The rub is we run `flutter test` from the root of the project (e.g., `project/`) which contains a pubspec without the flutter_test dev dependency.

![image](https://user-images.githubusercontent.com/67586/33687293-96d0a80e-da8c-11e7-833c-29e37374fd86.png)

Proposed solution: have the run configuration do the right thing and run flutter test from within `project/example`.

Issue #1583, opened 2017-12-08
Make test discovery more robust.


pq:
Following up from, https://github.com/flutter/flutter-intellij/pull/1573, we should be (even) smarter about test directory identification so that tests in projects with alternative directory structures are identifiable.  As an example, see:

https://github.com/brianegan/flutter_architecture_samples

PR #1585, opened 2017-12-08
Improve pub root detection for flutter tests (#1575).

pq:
Fixes: #1575

@stevemessick @devoncarew

### DIFF
--- a/src/io/flutter/pub/PubRoot.java
+++ b/src/io/flutter/pub/PubRoot.java
@@ -57,6 +57,20 @@ public class PubRoot {
   }
 
   /**
+   * Returns the first containing pub root for a given file.
+   */
+  @Nullable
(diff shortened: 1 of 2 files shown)


pq:
FWIW: similar logic will make it's way into a fix for #1583.

src/io/flutter/pub/PubRoot.java
```diff
@@ -56,6 +56,20 @@ private PubRoot(@NotNull VirtualFile root, @NotNull VirtualFile pubspec, @Nullab
     this.lib = lib;
   }
 
+  /**
+   * Returns the first containing pub root for a given file.
```
stevemessick: I don't understand this comment. Is a word missing, perhaps?

pq: Yes.  Very awkward.  Will fix.  Thanks!

PR #1586, opened 2017-12-08
Fix discovery of tests in example subdirs (#1583)

pq:
Fixes: #1583.

@stevemessick @devoncarew 

FYI @brianegan

image

And thanks to #1575, they can even run too! 👍 

image

### DIFF
--- a/src/io/flutter/FlutterUtils.java
+++ b/src/io/flutter/FlutterUtils.java
@@ -84,8 +84,9 @@ public class FlutterUtils {
     return module != null && FlutterModuleUtils.usesFlutter(module);
   }
 
-  public static boolean isInTestDir(DartFile file) {
-    final PubRoot root = PubRoot.forPsiFile(file);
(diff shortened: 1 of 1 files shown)


pq:
> This probably fixes a bunch of issues related to having an example project that is part of a plugin project. Or many examples.

I'm hoping so.  It appears to solve at least the basic case (and the slightly less basic case of nested example projects).

src/io/flutter/FlutterUtils.java
```diff
@@ -94,7 +95,7 @@ public static boolean isInTestDir(DartFile file) {
     if (candidate == null) return false;
 
     final String relativePath = root.getRelativePath(candidate);
-    return relativePath != null && (relativePath.startsWith("test/") || relativePath.startsWith("example/test/"));
+    return relativePath != null && (relativePath.startsWith("test/"));
```
devoncarew: 🍾 woohoo

brianegan:
Cool, thanks so much :)

Issue #1571, opened 2017-12-06
Missing hyperlink for local file in plugin run console output.


pq:
![image](https://user-images.githubusercontent.com/67586/33645998-8c5d0ee2-da02-11e7-8e31-38045977ee4c.png)

`example/lib/main.dart` should be linked.

PR #1588, opened 2017-12-11
Improve console hyperlinking for local files.

pq:
More hyperlinks!

* adds links to files in restart messages
image

* adds links to stack trace path elements
image

Fixes: #1571

cc @stevemessick @devoncarew 

￼￼
￼

### DIFF
--- a/src/io/flutter/console/FlutterConsoleFilter.java
+++ b/src/io/flutter/console/FlutterConsoleFilter.java
@@ -72,7 +72,7 @@ public class FlutterConsoleFilter implements Filter {
   @Nullable
   public VirtualFile fileAtPath(@NotNull String pathPart) {
 
-    // lib/main.dart:6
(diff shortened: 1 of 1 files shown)


src/io/flutter/console/FlutterConsoleFilter.java
```diff
@@ -67,48 +68,76 @@ public FlutterConsoleFilter(@NotNull Module module) {
     this.module = module;
   }
 
+  @VisibleForTesting
+  @Nullable
+  public VirtualFile fileAtPath(@NotNull String pathPart) {
+
+    // lib/main.dart:6
```
stevemessick: I do not understand this comment.

Oh! that's an example. The quotes on line 134 helped. See also line 115.

pq: 😄   

I'll add more quotes.

PR #1610, opened 2018-01-04
Reduce inspector tree indent.

pq:
Gives us, for example, a tree like this:

image

and

image

compared to the current default:

image

Happy to nudge out a bit if this looks too tight.

cc @jacob314 @devoncarew

### DIFF
--- a/src/io/flutter/view/InspectorPanel.java
+++ b/src/io/flutter/view/InspectorPanel.java
@@ -435,9 +435,9 @@ public class InspectorPanel extends JPanel implements Disposable, InspectorServi
       setRootVisible(false);
       registerShortcuts();
 
-      // Decrease indent.
-      final BasicTreeUI treeUI = (BasicTreeUI) getUI();
-      treeUI.setRightChildIndent(3);
+      // Decrease indent, scaled for different display types.
+      final BasicTreeUI ui = (BasicTreeUI) getUI();
+      ui.setRightChildIndent(JBUI.scale(3));
     }
 
     void registerShortcuts() {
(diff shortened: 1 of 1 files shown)


src/io/flutter/view/InspectorPanel.java
```diff
@@ -433,6 +434,10 @@ private MyTree(final DefaultMutableTreeNode treemodel) {
 
       setRootVisible(false);
       registerShortcuts();
+
+      // Decrease indent.
+      final BasicTreeUI treeUI = (BasicTreeUI) getUI();
+      treeUI.setRightChildIndent(3);
```
devoncarew: Good find!

Perhaps use something like `JBUI.scale()`? That takes into account HD vs non-HD screens.

devoncarew:
Could you also attach a screen shot w/ a node with multiple children? I just want to make sure that the indent we choose doesn't make it too difficult to visually line up which node is a child of what parent.

jacob314:
Lgtm! Excellent! Drawing lines between parents and children might help make the parent child relationships unambiguous with the smaller indent.

src/io/flutter/view/InspectorPanel.java
pq: New screenshot w/ children added.

jacob314:
Looks fine with the children added. I think this is definitely the right tradeoff given how deeply nested trees are.

src/io/flutter/view/InspectorPanel.java
pq: Good call on the scaling.  Fixed.

PR #1612, opened 2018-01-04
Update FlutterViewCondition to be bazel project aware.

jwren:


src/io/flutter/view/FlutterViewCondition.java
```diff
@@ -12,6 +12,6 @@
 public class FlutterViewCondition implements Condition<Project> {
   @Override
   public boolean value(Project project) {
```
devoncarew: Do you happen to know, does `Condition` allow for passing null projects? I don't know if it does or doesn't, but if it doesn't, we may want to add a `@NotNull` annotation on the param.

jwren: I checked, it does not explicitly have the `@NotNull`, which is why I added the check for null.

PR #1616, opened 2018-01-07
remove the memory dashboard reference; add an open obs. action

devoncarew:
- remove specific references to the timeline and memory views (the memory view was not complete)
- add a reference in the flutter view to the observatory; this mirrors what people see in the run and debug views

@pq, @jacob314, /cc @mit-mit


pq:
FWIW: kicked the failing travis build; fingers crossed!

PR #1625, opened 2018-01-10
show flutter material icons in the inspector

devoncarew:
- refactor the code to create flutter material icons out of the editor annotations and into a common class
- show flutter material icons in the inspector

@jacob314 

screen shot 2018-01-10 at 8 56 25 am

screen shot 2018-01-10 at 8 56 43 am

### DIFF
--- a/src/io/flutter/view/InspectorPanel.java
+++ b/src/io/flutter/view/InspectorPanel.java
@@ -23,6 +23,7 @@ import com.intellij.util.ui.tree.TreeUtil;
 import com.intellij.xdebugger.impl.ui.DebuggerUIUtil;
 import com.intellij.xdebugger.impl.ui.tree.nodes.XValueNodeImpl;
 import io.flutter.FlutterBundle;
+import io.flutter.editor.FlutterMaterialIcons;
 import io.flutter.inspector.*;
 import io.flutter.run.daemon.FlutterApp;
 import io.flutter.utils.ColorIconMaker;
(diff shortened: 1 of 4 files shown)


src/io/flutter/view/InspectorPanel.java
```diff
@@ -645,30 +646,59 @@ protected void customizeCellRenderer(JTable table, @Nullable Object value, boole
...
               append(String.format("#%02x%02x%02x%02x", alpha, red, green, blue), textAttributes);
             }
+
+            appendDescription = false;
+
+            break;
+          }
+
+          case "IconData": {
+            final String valueToString = node.getStringMember("valueToString");
```
jacob314: Instead of using valueToString
a more robust option is to use
getIntProperty(properties, "codePoint") 
as in the "Color" case above.
You'll need to repeat yourself in DiagnosticsNode.java indicating that you want the "codePoint" Dart property but you should just be able to search for "Color" to see what you need to copy.
Should be more robust than the string parsing. Feel free to also checkin like this and add a follow up CL that using getIntProperty or I can add that CL.

devoncarew: I don't think I saw that property (`codePoint`) when debugging, but I'll double check.

devoncarew:
Updated to use `codePoint`; ptal.

devoncarew:
Actually, as a follow-up to this, we should be able to show the material name for the icon in the property view - so, not just the plus icon, but the material name `'plus'`.

PR #1640, opened 2018-01-18
Inspector stability improvements.

jacob314:
Fix multiple race conditions occurring when the inspector is executed
in run mode.
Fix case where isolate ids were not available and fix case where the inspector
is called after the first frame has already rendered.
Add ability to perform capability detection to determine whether the
inspector supports a specific service method.
Fix null property type case.

Combined with a package:flutter fix, this fixes all pending inspector crasher bugs.

### DIFF
--- a/src/io/flutter/inspector/EvalOnDartLibrary.java
+++ b/src/io/flutter/inspector/EvalOnDartLibrary.java
@@ -41,7 +41,7 @@ public class EvalOnDartLibrary implements Disposable {
     this.myRequestsScheduler = new Alarm(Alarm.ThreadToUse.POOLED_THREAD, this);
     libraryRef = new CompletableFuture<>();
 
-    if (isolates.size() != 1) {
+    if (isolates.isEmpty()) {
       // Due to a race condition that would take fixing the Dart plugin to fix,
       // it is possible the debugProcess doesn't yet show any isolate infos
       // even though the isolates have started so we get the isolate info
@@ -50,9 +50,9 @@ public class EvalOnDartLibrary implements Disposable {
         @Override
         public void received(VM vm) {
           final ElementList<IsolateRef> isolatesList = vm.getIsolates();
-          if (isolatesList.size() != 1) {
-            libraryRef.completeExceptionally(new RuntimeException("Unexpected number of isolates:" + isolatesList.size()));
-          }
+          // If Flutter applications support multiple isolates we need to add
+          // logic to determine which isolate is running Flutter UI code.
+          assert(isolatesList.size() == 1);
           isolateId = isolatesList.get(0).getId();
           initialize();
         }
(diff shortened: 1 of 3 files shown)


src/io/flutter/view/InspectorPanel.java
```diff
@@ -653,40 +658,43 @@ protected void customizeCellRenderer(JTable table, @Nullable Object value, boole
...
-            this.setIcon(colorIconMaker.getCustomIcon(color));
-            if (alpha == 255) {
-              append(String.format("#%02x%02x%02x", red, green, blue), textAttributes);
-            }
-            else {
-              append(String.format("#%02x%02x%02x%02x", alpha, red, green, blue), textAttributes);
-            }
+        final String propertyType = node.getPropertyType();
+        if (propertyType != null) {
+          switch (propertyType) {
```
jacob314: this diff is just increasing indent by 2 and adding the propertyType != null check to avoid throwing exceptions for property objects that aren't really properties (e.g. new DiagnosticNode.message()

src/io/flutter/inspector/InspectorService.java
```diff
@@ -229,6 +230,56 @@ public void addClient(InspectorServiceClient client) {
...
   }
 
+  /**
+   * If the widget tree is not ready, the application should wait for the next
+   * Flutter.Frame event before attempting to display the widget tree. If the
+   * application is ready, the next Flutter.Frame event may never come as no
+   * new frames will be triggered to draw unless something changes in the UI.
+   */
+  public CompletableFuture<Boolean> isWidgetTreeReady() {
+    return hasServiceMethod("isWidgetTreeReady").<Boolean>thenComposeAsync((Boolean hasMethod) -> {
```
devoncarew: Can you add a TODO: to delete this code at some point? I think we can safely remove after one or two revs of the Flutter `alpha` channel.

devoncarew: (the `hasServiceMethod("isWidgetTreeReady")` code)

jacob314: done.

src/io/flutter/inspector/EvalOnDartLibrary.java
```diff
@@ -30,27 +27,50 @@
...
+    if (isolates.size() != 1) {
+      // Due to a race condition that would take fixing the Dart plugin to fix,
+      // it is possible the debugProcess doesn't yet show any isolate infos
+      // even though the isolates have started so we get the isolate info
+      // ourselves in that case.
+      vmService.getVM(new VMConsumer() {
+        @Override
+        public void received(VM vm) {
+          final ElementList<IsolateRef> isolatesList = vm.getIsolates();
+          if (isolatesList.size() != 1) {
```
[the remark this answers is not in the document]
devoncarew: This seems like an area we should keep an eye on, in the event that our assumptions change.

PR #1644, opened 2018-01-18
Add DartAnalysisServerServiceEx that wraps the Dart Analysis Service.

scheglov:
…and allows sending arbitrary requests and listen for all responses.


src/io/flutter/dart/DartAnalysisServerServiceEx.java
```diff
@@ -0,0 +1,145 @@
...
+      }
+    }
+
+    final ResponseStream serverResponseStream =
+      ReflectionUtil.getField(analysisServer.getClass(), analysisServer, ResponseStream.class, "responseStream");
+    if (serverResponseStream == null) {
+      return null;
+    }
+
+    Thread responseReaderThread = null;
```
devoncarew: Can you describe what this block of code is doing in a comment? I think it'll be useful for other people through the code.

Issue #1646, opened 2018-01-19
Intellij reported that flutter plugin crashed.


aam:
Please file this bug report at https://github.com/flutter/flutter-intellij/issues/new.

---

## What happened

Crash as soon as IntelliJ restarted after an update to 21.0.3.

## Version information

IntelliJ IDEA `2017.3.3` • Flutter plugin `io.flutter 21.0.3` • Dart plugin `173.4301.22`

No Flutter sdk configured.

## Exception

null

```
java.lang.NullPointerException
	at io.flutter.sdk.FlutterSdkVersion.isSupported(FlutterSdkVersion.java:74)
	at io.flutter.inspections.SdkConfigurationNotificationProvider.createNotificationPanel(SdkConfigurationNotificationProvider.java:76)
	at io.flutter.inspections.SdkConfigurationNotificationProvider.createNotificationPanel(SdkConfigurationNotificationProvider.java:30)
	at com.intellij.ui.EditorNotificationsImpl$3.performInReadAction(EditorNotificationsImpl.java:148)
	at com.intellij.openapi.progress.util.ReadTask.a(ReadTask.java:66)
	at com.intellij.openapi.application.ReadAction.compute(ReadAction.java:47)
	at com.intellij.openapi.progress.util.ReadTask.runBackgroundProcess(ReadTask.java:66)
	at com.intellij.openapi.progress.util.ProgressIndicatorUtils.a(ProgressIndicatorUtils.java:238)
	at com.intellij.openapi.progress.impl.CoreProgressManager.a(CoreProgressManager.java:173)
	at com.intellij.openapi.progress.impl.CoreProgressManager.a(CoreProgressManager.java:157)
	at com.intellij.openapi.progress.impl.CoreProgressManager.a(CoreProgressManager.java:543)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:488)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:94)
	at com.intellij.openapi.progress.impl.CoreProgressManager.runProcess(CoreProgressManager.java:144)
	at com.intellij.openapi.progress.impl.CoreProgressManager.runProcess(CoreProgressManager.java:173)
	at com.intellij.openapi.progress.util.ProgressIndicatorUtils.a(ProgressIndicatorUtils.java:236)
[...]

PR #1651, opened 2018-01-22
handle updates to the flutter version file

devoncarew:
The flutter version file was renamed from `VERSION` to `version` (and the contents changed slightly). We received a crash report in https://github.com/flutter/flutter-intellij/issues/1646; the existing code should have been hardened against the name change - I don't see a clear reason for the crash - but this PR does update to pull from the new file name.

Also, this PR removes several (now pretty old) version numbers.

@pq @jwren

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkVersion.java
+++ b/src/io/flutter/sdk/FlutterSdkVersion.java
@@ -66,12 +66,11 @@ public class FlutterSdkVersion {
   }
 
   public boolean flutterTestSupportsMachineMode() {
-    return version.compareTo(MIN_SUPPORTED_SDK.version) >= 0;
+    return isMinRecommendedSupported();
   }
 
-  @SuppressWarnings("BooleanMethodIsAlwaysInverted")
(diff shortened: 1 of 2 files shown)


src/io/flutter/sdk/FlutterSdkVersion.java
```diff
@@ -48,62 +27,55 @@ private FlutterSdkVersion(@NotNull String versionString) {
...
+      return null;
+    }
+    catch (IOException e) {
+      return null;
+    }
   }
 
-  public boolean isSupported() {
+  public boolean isMinRecommendedSupported() {
     return version.compareTo(MIN_SUPPORTED_SDK.version) >= 0;
```
jwren: If these three are implemented in the same way, the other two methods, flutterTestSupportsMachineMode() and  flutterTestSupportsFiltering() should just return the result of this one, instead of C&P.

devoncarew: 👍

PR #1656, opened 2018-01-23
Add FlutterDartAnalysisServer for Flutter specific requests and notifications.

scheglov:


thirdPartySrc/analysisServer/org/dartlang/analysis/server/protocol/FlutterOutline.java
```diff
@@ -0,0 +1,322 @@
+/*
+ * Copyright (c) 2015, the Dart project authors.
```
stevemessick: Instead of adding another top-level directory, could thirdPartySrc/analysisServer be moved into third_party? It also needs to have an Intellij "source" directory defined (see third_party/intellij-plugins-dart/src) in the module definition. That will correct the build failure.

The header comment looks like it should be updated -- not sure on the rules, but we're not doing Eclipse now. That applies to all files in thirdPartySrc.

scheglov:
PTAL

stevemessick:
Looks like we need to remove build.xml soon since it is no longer used.

src/io/flutter/dart/FlutterDartAnalysisServer.java
```diff
@@ -0,0 +1,90 @@
...
+import org.dartlang.analysis.server.protocol.FlutterOutline;
+import org.dartlang.analysis.server.protocol.FlutterService;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.*;
+
+public class FlutterDartAnalysisServer {
+  private static final String FLUTTER_NOTIFICATION_OUTLINE = "flutter.outline";
+
+  private final DartAnalysisServerServiceEx dartServiceEx;
```
stevemessick: @NotNull ?

tool/plugin/plugin.iml
```diff
@@ -11,7 +11,5 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
-    <orderEntry type="library" name="Dart SDK" level="project" />
```
stevemessick: I don't know if these two lines are required. This is a Dart module so it makes sense to have them. Might be safest to leave them.

src/io/flutter/dart/FlutterDartAnalysisServer.java
scheglov: Done.

tool/plugin/plugin.iml
scheglov: Reverted.

IntelliJ removes it for me.
I have Dart plugin installed, and I use `2017.2` version of IntelliJ.
So, I don't know why it remove these items.
Do I to do something else than specified in https://github.com/scheglov/flutter-intellij/blob/master/CONTRIBUTING.md?

devoncarew: `2017.2` or `2017.3`?

devoncarew:
Source in `third_party` is just there for copyright reasons; since we're generating the `org/dartlang/analysis/server/protocol/` files, we can just have in the main `src` directory.

(Sorry for the late review)

tool/plugin/plugin.iml
scheglov: I have `2017.2.6` installed.
```
IntelliJ IDEA 2017.2.6
Build #IC-172.4574.11, built on November 13, 2017
JRE: 1.8.0_152-release-915-b12 x86_64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Mac OS X 10.13.2
```

scheglov:
See https://github.com/flutter/flutter-intellij/pull/1664

PR #1668, opened 2018-01-26
some misc build system changes

devoncarew:
some misc build system changes:
- make the `abuild` command hidden (when you run `bin/plugin -h`)
- change the directory names the release artifacts are written to from `artifacts/release_21.0` to `artifacts/release_21` (open use the major version number)

@stevemessick


devoncarew:
> Would have been easier to remove the abuild command and build.xml, but there's no hurry.

I've not been using abuild or build.xml for the recent releases (`bin/plugin build` has been working great); sounds like we can remove the old build system whenever convenient.

PR #1676, opened 2018-01-29
misc inspector work

devoncarew:
- restore inspector splitter position between IntelliJ sessions
- show text `'Widget tree for the running app'` when there is no content in the widget / element tree view. This is to help the user know that an app must be running for content to display
- rename the 'Toggle Inspect Mode' toolbar button to 'Toggle Select Widget Mode'. I think the new name makes more sense now that the view has been renamed to 'Flutter Inspector'
- reorder the toolbar icons so the inspect mode action comes first

screen shot 2018-01-27 at 10 39 54 pm

screen shot 2018-01-28 at 9 17 53 am

@jacob314 @pq

### DIFF
--- a/src/io/flutter/view/InspectorPanel.java
+++ b/src/io/flutter/view/InspectorPanel.java
@@ -42,6 +42,8 @@ import javax.swing.tree.DefaultTreeModel;
 import javax.swing.tree.TreeNode;
 import javax.swing.tree.TreePath;
 import java.awt.*;
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
 import java.util.ArrayList;
 import java.util.Map;
 import java.util.concurrent.CancellationException;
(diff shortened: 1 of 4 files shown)


src/io/flutter/view/InspectorPanel.java
```diff
@@ -89,15 +91,29 @@ public InspectorPanel(FlutterView flutterView,
...
-    treeSplitter.setProportion(0.8f);
+    treeSplitter.setProportion(flutterView.getState().getSplitterProportion());
+    flutterView.getState().addListener(e -> {
+      final float newProportion = flutterView.getState().getSplitterProportion();
+      if (treeSplitter.getProportion() != newProportion) {
+        treeSplitter.setProportion(newProportion);
+      }
+    });
+    //noinspection Convert2Lambda
+    treeSplitter.addPropertyChangeListener("proportion", new PropertyChangeListener() {
```
jacob314: nit: is there a reason this isn't a lambda?

devoncarew: The line was just a bit too long and harder to read.

PR #1680, opened 2018-01-29
Update building.md

devoncarew:
@stevemessick; we should update (or delete) this doc.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -1,3 +1,5 @@
+_Note: this doc is out of date_
+
 ## Building the plugin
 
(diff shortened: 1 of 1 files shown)


docs/building.md
```diff
@@ -1,3 +1,5 @@
+_Note: this doc is out of date_
+
```
stevemessick: But the gsutil info is still useful, and may not be recorded elsewhere yet.

devoncarew: We should probably just update to reference `bin/plugin`.

Issue #1558, opened 2017-12-03
inspector icons don't include HD variants


devoncarew:
The icons for nodes in the inspector look pixelated on HD screens - we should add HD variants.

/cc @pq @jacob314

pq:
As per our conversation, opting to disable the (non HD) category widgets in favor altogether until we have a better, more consistent set.

jacob314:
I'd love to be able to demo with some icons at IO even if they are not ideal. I think the icons add some nice structure to the tree view.

PR #1682, opened 2018-01-30
HD Inspector Widgets

pq:
Reenables inspector category widgets and adds 2x variants.

I'm not sure if anything else needs to be done or if they'll magically get loaded...

Fixes: #1558

cc @alexander-doroshko @devoncarew ?


devoncarew:
They'll magically get loaded (you should be able to verify on an HD screen).

I don't think the current set of icons has any unifying theme; they look like they were selected from a few different icon sets. On the whole I would choose to use the auto-generated icons (the ones with a single letter in their image) as they all look consistent, but will defer to you and @jacob314 for inspector styling.

pq:
I'm not wild about these widgets but agree w/ @jacob314 that they improve understandability (at least that's what my gut says).  Let's land and iterate.

Maybe @InMatrix can advise?

Issue #1373, opened 2017-10-05
Open in xCode action for a plugin project doesn't work


mit-mit:
## Steps to Reproduce

1. Start IDEA with the M18 RC plugin installed
1. File > New Project, select Plugin type, and Next
1. After IDEA has initialised, right-click the root node in the project view and select Flutter > Open in xCode

Actual result: xCode starts, but the plugin code isn't available

Expected result: xCode opens, the plugincode is available for running, and the Run button starts the plugin example app as per:
https://flutter.io/developing-packages/#step-2c-add-ios-platform-code-hmswift

mit-mit:
I think it's opening `./plugin-name/Runner.xcodeproj`? It should be opening `./plugin-name/example/ios/Runner.xcworkspace`.

mit-mit:
cc @devoncarew

mit-mit:
Confirmed after feedback from @stevemessick that this works if you do a Run in IDEA prior to opening in xCode. Removing M18 blocking status.

stevemessick:
AI: Figure out a good UX to ensure a build is done prior to launching the other IDE. You may need to do a build prior to every launch, but that is time-consuming and it would be nice to not have to do it unless needed.

pq:
@mit-mit: #1527 updates our logic to open `Runner.xcworkspace`.  I believe this should be fixed?

mit-mit:
No, it still fails if you open xCode prior to doing a built in IJ:

```
/Users/mit/tmp/untitled4/example/ios/Runner/GeneratedPluginRegistrant.m:6:9: 'untitled4/Untitled4Plugin.h' file not found
```

I think the best solution might be to do a build in IJ as part of project creation?

pq:
Ah, OK.  I like the idea of volunteering to run a build before opening.

What's the best resource to look for as a sentinel?

/cc @cbracken

pq:
@cbracken, @szakarias  : could you advise on a sentinel resource to look for to indicate that an xcode build should (not) be performed?

pq:
@cbracken: any ideas on this one?

cbracken:
The ideal would be to always perform the build but Flutter should be smart enough to turn that into a no-op if everything that needs building is already built. I suspect that that's not yet feasible though it seems like something we should open a bug for.

[...]

PR #1694, opened 2018-01-30
Smarts to run `flutter build` before trying open Xcode (#1373).

pq:
Checks for `build/ios` and in it's absence kicks off a `flutter build --debug`, showing progress in the status line.

image

Otherwise, simply opens.

Fixes: #1373

@devoncarew

### DIFF
--- a/src/io/flutter/actions/OpenInXcodeAction.java
+++ b/src/io/flutter/actions/OpenInXcodeAction.java
@@ -97,12 +97,18 @@ public class OpenInXcodeAction extends AnAction {
       sdk.flutterBuild(pubRoot, "ios", "--debug").start(null, new ProcessAdapter() {
         @Override
         public void processTerminated(@NotNull ProcessEvent event) {
-          super.processTerminated(event);
           progressHelper.done();
+
+          if (event.getExitCode() != 0) {
+            FlutterMessages.showError("Error Opening Xcode", "`flutter build` returned: " + event.getExitCode());
+            return;
+          }
+
           openWithXcode(file.getPath());
         }
       });
-    } else {
+    }
+    else {
       openWithXcode(file.getPath());
     }
   }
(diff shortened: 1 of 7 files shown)


src/io/flutter/actions/OpenInXcodeAction.java
```diff
@@ -72,7 +76,43 @@ private VirtualFile findProjectFile(@Nullable AnActionEvent e) {
...
+
+    // Trigger an iOS build if necessary.
+    if (!hasBeenBuilt(pubRoot)) {
+      final ProgressHelper progressHelper = new ProgressHelper(project);
+      progressHelper.start("Building for iOS");
+
+      sdk.flutterBuild(pubRoot, "ios", "--debug").start(null, new ProcessAdapter() {
+        @Override
+        public void processTerminated(@NotNull ProcessEvent event) {
+          super.processTerminated(event);
```
devoncarew: Do we need to call super here?

src/io/flutter/actions/OpenInXcodeAction.java
```diff
@@ -72,7 +76,43 @@ private VirtualFile findProjectFile(@Nullable AnActionEvent e) {
...
+          progressHelper.done();
+          openWithXcode(file.getPath());
```
devoncarew: We should probably check `event.getExitCode()` here, and display an error if there was a build failure (else open Xcode).

src/io/flutter/actions/OpenInXcodeAction.java
pq: Nope.  Good catch.

src/io/flutter/actions/OpenInXcodeAction.java
pq: 👍

PR #1696, opened 2018-01-30
Add speed search to the Preview view.

scheglov:


src/io/flutter/preview/PreviewView.java
```diff
@@ -264,6 +279,42 @@ Icon getIcon() {
     return icon;
   }
 
+  /**
+   * Return the string that is suitable for speed search. It has every name part separted so that we search only inside individual name
+   * parts, but not in their accidential concatenation.
+   */
+  @NotNull
+  String getSpeedSearchString() {
```
devoncarew: private?

scheglov: In Java private is stronger than in Dart, it prevents access even for classes in the same file. So, default visibility.

scheglov:
I opened a new issue https://github.com/flutter/flutter-intellij/issues/1697 for speed search in Flutter Inspector.

pq:
Late to the party (sorry) but 🎉 🎉 🎉 !

PR #1704, opened 2018-01-31
various ui tweaks in the preview view

devoncarew:
- add expand all and collapse all actions in the top (tool window) toolbar
- remove the view condition class for the Inspector view, and simplify the one for the Preview view; neither require a flutter project in order to be listed as available. I think this will make them less error prone wrt our project type detection logic
- disable the 'basics' and 'cupertino' categories for widget icons. These used the same base icon as the synthesized icon. In trees, 'basics' looked like the other icons except without text in the center of the icon - this made it look like the text had been forgotten. The 'cupertino' icon had a 'c' in the center of the icon, but the cupertino widgets didn't stop w/ the letter c; this is a strange juxtaposition next to the synthesized icons, which do use the class.png icon w/ the first letter in the center; /cc @pq and @jacob314 
- remove the first level node from the tree - `<unit>`
- have a single click jump from the outline to the source (instead of a double click)
- slightly simplify how we display node properties; from `[foo: bar] [foo2: baz]` to `foo: bar, foo2: baz`

screen shot 2018-01-30 at 5 32 42 pm

screen shot 2018-01-30 at 5 32 03 pm

@scheglov


src/io/flutter/preview/PreviewView.java
```diff
@@ -384,20 +400,33 @@ public void customizeCellRenderer(
...
     final List<FlutterOutlineAttribute> attributes = outline.getAttributes();
     if (attributes != null) {
-      for (FlutterOutlineAttribute attribute : attributes) {
+      if (attributes.size() == 1) {
+        final FlutterOutlineAttribute attribute = attributes.get(0);
         append(" ");
-        append("[", SimpleTextAttributes.GRAYED_ATTRIBUTES);
-        appendSearch(attribute.getName(), SimpleTextAttributes.GRAYED_ATTRIBUTES);
-        append(": ", SimpleTextAttributes.GRAYED_ATTRIBUTES);
         appendSearch(attribute.getLabel(), SimpleTextAttributes.GRAYED_ATTRIBUTES);
```
scheglov: It looks good for `Text` and `Tab`.
But there might be cases when it is not obvious which attribute is set, e.g. `appBar: AppBar false`. What is `false`?

src/io/flutter/preview/PreviewView.java
```diff
@@ -55,6 +58,9 @@
 import java.util.List;
 import java.util.Objects;
 
+import static com.intellij.icons.AllIcons.General.CollapseAll;
```
scheglov: It does not save us much, and we have space on the lines that use icons, we could just import `AllIcons`.

devoncarew: I used the organize imports, but will switch to importing allicons, re-organize, and see if it's stable.

src/io/flutter/preview/PreviewView.java
devoncarew: I agree that it doesn't work well with just `bool`s. I think it's terser and still understandable with things like `CrossAxisAlignment.stretch`.

Let me try just eliding `data: ` and see how that looks.

devoncarew: I simplified this to always eliding `data:`, and only eliding `text:` when it's the only attribute.

pq:
Awesome! 🎉

jacob314:
This looks great. I'd like to update the widget tree in the inspector to
match this style as much as possible. The inspector now has sufficient
information to display this same information about properties defined
directly by widget constructors.

PR #1705, opened 2018-01-31
Pull the existing or new Analysis Server service.

scheglov:
So, now it is safe to subscribe for Flutter notifications even when Analysis Server has not yet started. For example when the view is open on start.

And subscribe to the currently selected file during opening of the view.

### DIFF
--- a/src/io/flutter/dart/FlutterDartAnalysisServer.java
+++ b/src/io/flutter/dart/FlutterDartAnalysisServer.java
@@ -6,9 +6,9 @@
 package io.flutter.dart;
 
 import com.google.common.collect.Lists;
-import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import com.intellij.concurrency.JobScheduler;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.project.Project;
 import com.jetbrains.lang.dart.analyzer.DartAnalysisServerService;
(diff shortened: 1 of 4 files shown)


src/io/flutter/preview/PreviewView.java
```diff
@@ -91,24 +84,22 @@ public void outlineUpdated(@NotNull String filePath, @NotNull FlutterOutline out
 
   public PreviewView(@NotNull Project project) {
     this.project = project;
-    final DartAnalysisServerService analysisService = DartPlugin.getInstance().getAnalysisService(project);
-    final DartAnalysisServerServiceEx analysisServiceEx = DartAnalysisServerServiceEx.get(analysisService);
-    flutterAnalysisServer = new FlutterDartAnalysisServer(analysisServiceEx);
+    flutterAnalysisServer = FlutterDartAnalysisServer.getInstance(project);
 
+    // Show preview for the file selected when the view is being open.
```
devoncarew: `... being opened.`

src/io/flutter/dart/FlutterDartAnalysisServer.java
```diff
@@ -88,4 +107,25 @@ private void processNotification(JsonObject response) {
...
+
+  private class RefreshDartAnalysisServerServiceExThread extends Thread {
+    RefreshDartAnalysisServerServiceExThread() {
+      setDaemon(true);
+    }
+
+    @Override
+    public void run() {
+      //noinspection InfiniteLoopStatement
+      while (true) {
```
devoncarew: I think we should prefer using something like JobScheduler.getScheduler().scheduleWithFixedDelay() as it may play better with the IntelliJ APIs.

https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/sdk/FlutterSdkManager.java#L45

src/io/flutter/preview/PreviewView.java
scheglov: Fixed.

src/io/flutter/dart/FlutterDartAnalysisServer.java
scheglov: Ah, this is much better. Thanks.

PR #1710, opened 2018-02-01
Navigate from source to Preview view.

scheglov:
Scrolling the node of the enclosing top-level declaration to the top of the tree has one not quite ideal consequence - the tree jumps us you click it.


devoncarew:
If I understand that, it sounds like we only want to scroll the top-level to the top if the user interacted with the editor, but not do that when they interacted with the preview view.

devoncarew:
Looks like there's a failure when compiling against 2017.1 (https://travis-ci.org/flutter/flutter-intellij/jobs/335895636#L1011). Perhaps there an adapter class we can use there (instead of implementing the interface)?

PR #1721, opened 2018-02-01
Live update of property values triggered each time a flutter frame is drawn

jacob314:
For the full benefit a small matching change is required on the flutter side as there is some caching of property values that gets in the way.

Tests for the rate limiter code are a bit clumsy as I didn't see an obvious way to plumb through mock time objects. This rate limiter has the advantage over more simple approaches of ensuring the first update happens immediately and only subsequent updates are rate limited to avoid overtaxing the device. This means that after a hot reload you will see new properties immediately but you won't try to get property value updates at 60fps while scrolling.
Integration with completable future is also quite nice so we won't schedule a new update until at least after the previous update is complete ensuring we avoid denial of service level attacks on a slow device running a complex app.

### DIFF
--- a/src/io/flutter/inspector/DiagnosticsNode.java
+++ b/src/io/flutter/inspector/DiagnosticsNode.java
@@ -20,6 +20,7 @@ import javax.swing.*;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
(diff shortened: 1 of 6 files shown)


src/io/flutter/inspector/DiagnosticsNode.java
```diff
@@ -620,4 +616,33 @@ private static Icon getCustomIcon(String text) {
...
+   * properties for the DiagnosticsNode objects are identical. The valueId
+   * field may change even for properties that have not changed because in
+   * some cases such as the 'created' property for an element, the property
+   * value is created dynamically each time 'getProperties' is called.
+   */
+  public boolean identicalDisplay(DiagnosticsNode node) {
+    if (node == null) {
+      return false;
+    }
+    Set<String> keys = json.keySet();
```
devoncarew: @jacob314, this API doesn't exist in 2017.1 and 2017.2; see https://travis-ci.org/flutter/flutter-intellij/builds/336668158?utm_source=github_status&utm_medium=notification

The build was broken for pub reasons, so it masked this additional failure, but we should re-run travis in those cases to get to green builds :)

PR #1722, opened 2018-02-02
Track creation locations.

jacob314:
This CL is fine to land before the corresponding changes in Flutter or
Flutter Engine as without those changes source locations just show
up as missing. The source location menu item shows up as disabled
for widgets or render objects without a source location.
Note that RenderObjects will not have source locations even after the
Flutter change as there isn't a clear benefit in tracking them.

### DIFF
--- a/src/io/flutter/inspector/DiagnosticsNode.java
+++ b/src/io/flutter/inspector/DiagnosticsNode.java
@@ -50,6 +50,8 @@ import static io.flutter.sdk.FlutterSettingsConfigurable.WIDGET_FILTERING_ENABLE
 public class DiagnosticsNode {
   private static final CustomIconMaker iconMaker = new CustomIconMaker();
 
+  private Location location;
+
   public DiagnosticsNode(JsonObject json, InspectorService inspectorService) {
     this.inspectorService = inspectorService;
     this.json = json;
@@ -364,10 +366,14 @@ public class DiagnosticsNode {
   }
 
   public Location getCreationLocation() {
+    if (location != null) {
+      return location;
+    }
     if (!hasCreationLocation()) {
       return null;
     }
-    return new Location(json.getAsJsonObject("creationLocation"), null);
+    location = new Location(json.getAsJsonObject("creationLocation"), null);
+    return location;
   }
 
   /**
--- a/src/io/flutter/inspector/DiagnosticsNode.java
+++ b/src/io/flutter/inspector/DiagnosticsNode.java
@@ -359,6 +359,17 @@ public class DiagnosticsNode {
     return json.has("exception");
   }
 
+  public boolean hasCreationLocation() {
+    return json.has("creationLocation");
+  }
+
+  public Location getCreationLocation() {
+    if (!hasCreationLocation()) {
+      return null;
+    }
+    return new Location(json.getAsJsonObject("creationLocation"), null);
+  }
+
   /**
    * String representation of the type of the property [value].
    * <p>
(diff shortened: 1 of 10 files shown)


src/io/flutter/view/InspectorPanel.java
```diff
@@ -163,6 +168,14 @@ private InspectorService getInspectorService() {
   }
 
   void setActivate(boolean enabled) {
+    // TODO(jacobr): where is the right place to be caching the pub roots.
```
jacob314: I'd like to cache the pubRoots somewhere as I do use the pub roots every  time a node is styled to determine whether it should be bold or not. If there is an event I can listen for when pub roots change that is also reasonable. I assume pub roots can't change on a hot reload.

src/io/flutter/inspector/DiagnosticsNode.java
```diff
@@ -358,6 +358,17 @@ boolean hasException() {
...
 
+  public boolean hasCreationLocation() {
+    return json.has("creationLocation");
+  }
+
+  public Location getCreationLocation() {
+    if (!hasCreationLocation()) {
+      return null;
+    }
+    return new Location(json.getAsJsonObject("creationLocation"), null);
```
devoncarew: Not sure how frequently this is called; consider caching the `Location` instance in a private field?

src/io/flutter/inspector/Location.java
```diff
@@ -0,0 +1,77 @@
...
+import com.intellij.xdebugger.impl.XSourcePositionImpl;
+import io.flutter.utils.JsonUtils;
+
+import java.util.ArrayList;
+
+public class Location {
+  final JsonObject json;
+  Location parent;
+
+  public Location(JsonObject json, Location parent) {
```
devoncarew: we're not using the `parent` param here?

src/io/flutter/inspector/Location.java
```diff
@@ -0,0 +1,77 @@
...
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.xdebugger.XSourcePosition;
+import com.intellij.xdebugger.impl.XSourcePositionImpl;
+import io.flutter.utils.JsonUtils;
+
+import java.util.ArrayList;
+
+public class Location {
+  final JsonObject json;
+  Location parent;
```
devoncarew: `parent` is never assigned to

src/io/flutter/view/InspectorPanel.java
devoncarew: Perhaps in the `FlutterApp`? Or at least something that's per-debug session. I think it's reasonable to assume that they won't change during a debug session.

devoncarew: What happens if we have two sessions running at the same time? Does the inspector become unhappy?

You can test this by launching an app against one device and then a 2nd.

jacob314: Cached on the FlutterApp for now. Good point about having two session running at the same time. We definitely don't handle that case currently.

src/io/flutter/inspector/DiagnosticsNode.java
jacob314: Went ahead and cached the Location object.

src/io/flutter/inspector/Location.java
jacob314: the usage isn't obvious but we are on line 29. the behavior is the file from the parent is used if a fileName is not specified as is the case for parameter locations.
Notice that on line 71 we pass in parent as "this".

jacob314:
@devoncarew  you are right that we handle multiple sessions at the same time badly. I'll send a follow up Cl that handles that in general. Based on offline discussion, the simplest solution is to ensure we only show the inspector for the first running session.

PR #1723, opened 2018-02-02
handle tree selection event

devoncarew:
- handle the general tree selection event (sync the editor position with tree selection changes, like those from key events)
- elide `icon` param names - this makes `Icon` nodes look better
- don't show the parent association label; this is generally `child:` or `children:`. There's a little less noise in the tree w/o the labels, and the tree looks a bit more like a hierarchical representation of the widgets. @scheglov, I'm happy to discuss more in the office if you have concerns about this change

screen shot 2018-02-01 at 7 37 34 pm

screen shot 2018-02-01 at 7 35 49 pm

@scheglov

### DIFF
--- a/src/io/flutter/preview/PreviewView.java
+++ b/src/io/flutter/preview/PreviewView.java
@@ -476,11 +476,6 @@ class OutlineObject {
     else {
       builder.append(outline.getClassName());
     }
-    if (outline.getParentAssociationLabel() != null) {
-      builder.append('|');
-      builder.append(outline.getParentAssociationLabel());
-      builder.append(": ");
-    }
     if (outline.getVariableName() != null) {
       builder.append('|');
       builder.append(outline.getVariableName());
--- a/src/io/flutter/preview/PreviewView.java
+++ b/src/io/flutter/preview/PreviewView.java
@@ -15,6 +15,7 @@ import com.intellij.openapi.actionSystem.ActionManager;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.actionSystem.DefaultActionGroup;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.components.PersistentStateComponent;
 import com.intellij.openapi.components.Storage;
 import com.intellij.openapi.editor.Caret;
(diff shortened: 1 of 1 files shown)


src/io/flutter/preview/PreviewView.java
```diff
@@ -249,34 +250,45 @@ protected String getElementText(Object element) {
...
+            selectPath(selectionPath, true);
           }
         }
       }
     });
 
+    tree.addTreeSelectionListener(e -> {
+      final TreePath selectionPath = e.getNewLeadSelectionPath();
+      if (selectionPath != null) {
+        ApplicationManager.getApplication().invokeLater(() -> selectPath(selectionPath, false));
```
scheglov: Isn't listener already executed in the UI thread?
Why do we want `invokeLater`?

src/io/flutter/preview/PreviewView.java
```diff
@@ -545,13 +556,8 @@ public void customizeCellRenderer(
       setIcon(icon);
     }
 
-    // Render the parent/child association.
-    if (outline.getParentAssociationLabel() != null) {
-      appendSearch(outline.getParentAssociationLabel() + ": ", SimpleTextAttributes.REGULAR_ATTRIBUTES);
```
scheglov: If we don't show it, we might also exclude it from `getSpeedSearchString`.

src/io/flutter/preview/PreviewView.java
```diff
@@ -545,13 +556,8 @@ public void customizeCellRenderer(
       setIcon(icon);
     }
 
-    // Render the parent/child association.
```
scheglov: You're right, most associations are just `child:` (I think we don't even show `children:` exactly because it is not interesting for `Row` or `Column`). But some widgets might have interesting associations, like `Scaffold` (`appBar` or `body`) or `AppBar` (`title` or `bottom`). In WindowBuilder we would display these "places" in the components tree to allow uses to drop new components there. We don't have any DnD yet, so maybe not very useful. We can discuss tomorrow.

src/io/flutter/preview/PreviewView.java
devoncarew: I was seeing exceptions without this:

```
write-unsafe context! Model changes are allowed from write-safe contexts only. Please ensure you're using invokeLater/invokeAndWait with a correct modality state (not "any"). See TransactionGuard documentation for details.
```

src/io/flutter/preview/PreviewView.java
devoncarew: makes sense; done

PR #1724, opened 2018-02-02
Unpack BlockingRequestSink to get the actual ByteRequestSink.

scheglov:
This solves a problem with restarting the analysis server, or the case when the view is open when you start IDE.

### DIFF
--- a/src/io/flutter/dart/DartAnalysisServerServiceEx.java
+++ b/src/io/flutter/dart/DartAnalysisServerServiceEx.java
@@ -19,6 +19,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.WeakHashMap;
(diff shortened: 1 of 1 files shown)


src/io/flutter/dart/DartAnalysisServerServiceEx.java
```diff
@@ -35,7 +36,12 @@
   public DartAnalysisServerServiceEx(DartAnalysisServerService base, AnalysisServer analysisServer) {
     this.base = base;
     this.analysisServer = analysisServer;
-    this.requestSink = ReflectionUtil.getField(analysisServer.getClass(), analysisServer, RequestSink.class, "requestSink");
+
+    RequestSink requestSink = ReflectionUtil.getField(analysisServer.getClass(), analysisServer, RequestSink.class, "requestSink");
```
devoncarew: I assume this use of reflection will go away with upstream changes to the Dart plugin? If so, can you add a TODO: / note here?

PR #1730, opened 2018-02-02
Don't move caret in the editor when the editor is focused.

scheglov:


src/io/flutter/preview/PreviewView.java
```diff
@@ -311,7 +311,7 @@ private void handleTreeSelectionEvent(TreeSelectionEvent e) {
   private void selectPath(TreePath selectionPath, boolean focusEditor) {
     final FlutterOutline outline = getOutlineOfPath(selectionPath);
     final int offset = outline.getDartElement() != null ? outline.getDartElement().getLocation().getOffset() : outline.getOffset();
-    if (currentFile != null) {
+    if (currentFile != null && !isEditorFocused()) {
```
devoncarew: How do we know the caret is already positioned where we want it?

scheglov:
Focus turned out to be not an ideal solution. When the tree is focused, and you click in the editor, you get the caret change even first, and focus is set later. So, we don't know that the editor is "being focused".

Removing/adding listeners does the trick.

PR #1736, opened 2018-02-04
Add preview area to the Preview view

devoncarew:
- add a splitter to the Preview view, and, when a file with a Flutter build() method is active, shows a JPanel there with the name of the currently selected Widget
- have a compile time flag to enable the panel; the panel doesn't do anything useful currently
- remember the splitter position in the view's persisted state
- render build() methods in bold, in order to highlight them in the UI

@scheglov, in the future, we use this to show a live preview of the current widget. That would likely be a ways out; we can choose to land the parts of this PR that are immediately useful, hold off on the PR, or commit it w/ the understanding that the panel might be refactored away if we don't go this direction.

### DIFF
--- a/src/io/flutter/preview/PreviewView.java
+++ b/src/io/flutter/preview/PreviewView.java
@@ -368,11 +368,10 @@ public class PreviewView implements PersistentStateComponent<PreviewViewState>, 
       }
 
       final FlutterOutline flutterOutline = outlineElement.outline;
+      final Element element = flutterOutline.getDartElement();
 
-      if (flutterOutline.getDartElement() != null) {
-        if (ModelUtils.isBuildMethod(flutterOutline.getDartElement())) {
-          return flutterOutline.getDartElement();
-        }
+      if (element != null && ModelUtils.isBuildMethod(element)) {
+        return element;
       }
     }
 
--- a/src/io/flutter/preview/PreviewView.java
+++ b/src/io/flutter/preview/PreviewView.java
@@ -52,6 +52,7 @@ import io.flutter.inspector.FlutterWidget;
 import io.flutter.utils.CustomIconMaker;
 import org.dartlang.analysis.server.protocol.*;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
 import javax.swing.event.TreeSelectionEvent;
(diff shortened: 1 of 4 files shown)


devoncarew:
screen shot 2018-02-03 at 5 41 19 pm

src/io/flutter/preview/PreviewView.java
```diff
@@ -322,6 +332,51 @@ private void selectPath(TreePath selectionPath, boolean focusEditor) {
...
       }
     }
+
+    if (SHOW_PREVIEW_AREA) {
+      final Element buildMethodElement = getBuildMethodElement(selectionPath);
+      previewAreaPanel.updatePreviewElement(getElementParentFor(buildMethodElement), buildMethodElement);
+    }
+  }
+
+  // TODO: Add parent relationship info to FlutterOutline instead of this O(n^2) traversal.
```
scheglov: Other alternatives.

We could remember the top-level FlutterOutline, then we would need recursive visit, but O(n).

Or we could fill childToParentMap after receiving the top-level outline and rendering it into the tree.

src/io/flutter/preview/PreviewView.java
```diff
@@ -322,6 +332,51 @@ private void selectPath(TreePath selectionPath, boolean focusEditor) {
...
+    for (Object n : path.getPath()) {
+      final DefaultMutableTreeNode node = (DefaultMutableTreeNode)n;
+      final OutlineObject outlineElement = (OutlineObject)node.getUserObject();
+      if (outlineElement == null) {
+        continue;
+      }
+
+      final FlutterOutline flutterOutline = outlineElement.outline;
+
+      if (flutterOutline.getDartElement() != null) {
```
scheglov: We repeat `flutterOutline.getDartElement()` three times :)

src/io/flutter/preview/PreviewView.java
devoncarew: We could populate a `parent` field when re-serializing a FlutterOutline?

scheglov: `FlutterOutline` is code generated.
We still could tweak the generator though.

src/io/flutter/preview/PreviewView.java
devoncarew: done

src/io/flutter/preview/PreviewView.java
devoncarew: Gotcha.

I don't think this will present a performance issue in practice, but would be nice cleanup at some point in the future.

PR #1739, opened 2018-02-05
Render generic label for outline items.

scheglov:
These labels are long though.
We could shorten them on the client, e.g. leave first and last 50 characters.
Or we could shorten them on the server, e.g. omit long arguments.

image

### DIFF
--- a/src/io/flutter/preview/PreviewView.java
+++ b/src/io/flutter/preview/PreviewView.java
@@ -707,14 +707,14 @@ class OutlineTreeCellRenderer extends ColoredTreeCellRenderer {
     // Render the variable.
     if (outline.getVariableName() != null) {
       append(" ");
-      appendSearch(outline.getVariableName(), SimpleTextAttributes.REGULAR_ATTRIBUTES);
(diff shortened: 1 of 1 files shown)


src/io/flutter/preview/PreviewView.java
```diff
@@ -708,6 +708,13 @@ public void customizeCellRenderer(
       appendSearch(outline.getVariableName(), SimpleTextAttributes.REGULAR_ATTRIBUTES);
     }
 
+    // Render a generic label.
+    if (outline.getKind().equals(FlutterOutlineKind.GENERIC) && outline.getLabel() != null) {
```
devoncarew: I think the previous `Widget` label worked pretty well here. I do think that this label will end up being too long.

What about `Widget` _buildListForSessions( ... )` or similar? So the actual method called is part of the suffix text of the node, and is partially elided for long items, either in IntelliJ or on the server.

scheglov: Done in Analysis Server.
So, in the plugin we can continue rendering the whole label as is.

image

scheglov:
image

PR #1743, opened 2018-02-06
Add icons and actions for wrapping into Column and Row.

scheglov:
image


src/io/flutter/preview/PreviewView.java
```diff
@@ -186,6 +185,8 @@ public void initToolWindow(@NotNull ToolWindow toolWindow) {
 
     toolbarGroup.add(new QuickAssistAction(FlutterIcons.Center, "Center widget"));
     toolbarGroup.add(new QuickAssistAction(FlutterIcons.Padding, "Add widget padding"));
+    toolbarGroup.add(new QuickAssistAction(FlutterIcons.Column, "Put into new Column"));
```
devoncarew: It looks like we're using the user visible name of the refactoring as the unique id. Is there a more stable id we could use? `flutter-center-widget` or something? I'm guessing we haven't had a need for such until now.

It would be nice if the id and the user visible name could be separate. One benefit would be that we could adjust the names in the analysis server, w/o that being a breaking change for clients.

scheglov: @bwilkerson Maybe we can discuss this tomorrow.

PR #1749, opened 2018-02-07
restore framework toggles after a restart

devoncarew:
- after a full restart, restore any framework toggles that the user had set (debug draw, fps overlay, ...). We choose to unset the inspector toggle instead of trying to restore; I think this will mostly match user expectations.

Since we're currently the only client toggling these settings, we don't need events from the framework in order to keep these settings up to date. We should add such long term however.

@jacob314


devoncarew:
This PR (or a follow-on) may also need to do some work to restore toggles between application runs (or, un-set them, as appropriate for the individual toggle).

devoncarew:
Updated the PR to also handle restoring settings on app launches (if the settings had been toggled on previous runs).

src/io/flutter/view/FlutterView.java
```diff
@@ -336,72 +425,141 @@ public void actionPerformed(AnActionEvent event) {
...
+  }
 }
 
-class RepaintRainbowAction extends AbstractToggleableAction {
+class RepaintRainbowAction extends FlutterViewToggleableAction {
   RepaintRainbowAction(@NotNull FlutterView view) {
     super(view, "Enable Repaint Rainbow");
   }
 
   protected void perform(AnActionEvent event) {
```
scheglov: @Nullable ?

src/io/flutter/view/FlutterView.java
```diff
@@ -336,72 +425,141 @@ public void actionPerformed(AnActionEvent event) {
...
 }
 
-class HideSlowBannerAction extends AbstractToggleableAction {
+class HideSlowBannerAction extends FlutterViewToggleableAction {
   HideSlowBannerAction(@NotNull FlutterView view) {
     super(view, "Hide Slow Mode Banner");
   }
 
   @Override
   protected void perform(AnActionEvent event) {
```
scheglov: @Nullable  ?

src/io/flutter/view/FlutterView.java
```diff
@@ -336,72 +425,141 @@ public void actionPerformed(AnActionEvent event) {
...
 }
 
-class ShowPaintBaselinesAction extends AbstractToggleableAction {
+class ShowPaintBaselinesAction extends FlutterViewToggleableAction {
   ShowPaintBaselinesAction(@NotNull FlutterView view) {
     super(view, "Show Paint Baselines");
   }
 
   @Override
   protected void perform(AnActionEvent event) {
```
scheglov: @Nullable  ?

src/io/flutter/view/FlutterView.java
devoncarew: done

src/io/flutter/view/FlutterView.java
devoncarew: done

src/io/flutter/view/FlutterView.java
devoncarew: done

PR #1753, opened 2018-02-08
no longer set the initial value for full file dartfmt

devoncarew:
The Flutter plugin used to set the initial value of full-file dartfmt to true; this is now (since 2017.3) the default setting, so we no longer need to do this.

@alexander-doroshko


stevemessick:
Does this work for Android Studio?

devoncarew:
I _believe_ that the option was introduced in 2017.2, and was made on by default in 2017.3. The Flutter plugin had been setting the option to on once at first run; this should no long be necessary in 2017.3.

> Does this work for Android Studio?

So, for AS 3.0 (based on 2017.1), this setting won't be available (setting the value to true should be a no-op). For AS 3.1 (based on 2017.3), this setting should be available and on by default.

PR #1755, opened 2018-02-08
Don't use 'org.apache.commons' because it is not exported from the Dart plugin.

scheglov:


devoncarew:
Can you add validation for this in https://github.com/flutter/flutter-intellij/blob/master/tool/plugin/lib/lint.dart? That we don't re-introduce code with these imports?

scheglov:
I updated https://github.com/flutter/flutter-intellij/blob/master/tool/plugin/lib/lint.dart.

devoncarew:
👍

PR #1759, opened 2018-02-09
Add action for 'Replace with children' assist.

scheglov:


devoncarew:
In terms of the icon, you may want the same icon, with the arrow pointing left (since it's overwriting the node with the children, and the children will all move left one in the tree). You could also look for an icon that had a connotation of deleting something (without deleting being the main action for the icon).

PR #1764, opened 2018-02-13
Rename 'Replace with children' to 'Remove widget'.

scheglov:
See also https://dart-review.googlesource.com/c/sdk/+/40961 for the corresponding Analysis Server change.


src/io/flutter/dart/DartAnalysisServerServiceEx.java
```diff
@@ -93,6 +93,7 @@ public void lastRequestProcessed() {
     public JsonObject take() throws Exception {
       final JsonObject json = base.take();
       synchronized (listeners) {
+        System.out.println(json);
```
devoncarew: remove?

scheglov: Ouch! Done. Thanks.

PR #1766, opened 2018-02-13
Use the FlutterBundle.message(...) instead of the same hard-coded string.

jwren:
This is a cleanup nit, the string name does not change with this PR.


jwren:
@pq @devoncarew

Issue #1716, opened 2018-02-01
tracking analytics for the refactorings used from Flutter Preview


devoncarew:
We want to make sure we're tracking analytics for the refactorings used from Flutter Preview.

PR #1769, opened 2018-02-13
Add analytics for opening the view, executing actions, and navigation.

scheglov:
closes https://github.com/flutter/flutter-intellij/issues/1716

### DIFF
--- a/src/io/flutter/preview/PreviewView.java
+++ b/src/io/flutter/preview/PreviewView.java
@@ -275,7 +275,7 @@ public class PreviewView implements PersistentStateComponent<PreviewViewState>, 
     updateActionsForOutlines(selectedOutlines);
   }
 
-  private void selectPath(TreePath selectionPath, boolean requestFocus) {
+  private void selectPath(TreePath selectionPath, boolean focusEditor) {
     final FlutterOutline outline = getOutlineOfPath(selectionPath);
     if (outline == null) {
       return;
@@ -287,7 +287,7 @@ public class PreviewView implements PersistentStateComponent<PreviewViewState>, 
     if (currentFile != null) {
       currentEditor.getCaretModel().removeCaretListener(caretListener);
       try {
-        new OpenFileDescriptor(project, currentFile, offset).navigate(requestFocus);
+        new OpenFileDescriptor(project, currentFile, offset).navigate(focusEditor);
       }
       finally {
         currentEditor.getCaretModel().addCaretListener(caretListener);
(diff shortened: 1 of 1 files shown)


src/io/flutter/preview/PreviewView.java
```diff
@@ -172,16 +173,18 @@ public void loadState(PreviewViewState state) {
   }
 
   public void initToolWindow(@NotNull ToolWindow toolWindow) {
+    sendAnalyticEvent("initToolWindow");
```
devoncarew: I think we're already getting this event from listening to general view changes; I think we can avoid sending this specific one in order to match what we do for other view (unless we think this adds some specific value).

src/io/flutter/preview/PreviewView.java
```diff
@@ -616,14 +628,22 @@ private void scrollTreeToNodeOnTop(TreeNode node) {
     }
   }
 
+  private void sendAnalyticEvent(@NotNull String name) {
+    FlutterInitializer.getAnalytics().sendEvent("flutterPreview", name);
```
devoncarew: `flutterPreview` ==> `preview`?

src/io/flutter/preview/PreviewView.java
```diff
@@ -616,14 +628,22 @@ private void scrollTreeToNodeOnTop(TreeNode node) {
...
+
+    QuickAssistAction(@NotNull String id, Icon icon, String assistMessage) {
       super(assistMessage, null, icon);
+      this.id = id;
       messageToActionMap.put(assistMessage, this);
     }
 
     @Override
     public void actionPerformed(AnActionEvent e) {
+      sendAnalyticEvent(id);
```
devoncarew: 👍

src/io/flutter/preview/PreviewView.java
```diff
@@ -291,6 +294,15 @@ private void selectPath(TreePath selectionPath, boolean focusEditor) {
       return;
     }
 
+    if (focusEditor) {
```
devoncarew: rename both to 'jumpToSource'? 'showSource'?

src/io/flutter/preview/PreviewView.java
scheglov: OK, removing this.

src/io/flutter/preview/PreviewView.java
scheglov: I was not sure if we get enough information to disambiguate it from other possible previews.
Fixed.

src/io/flutter/preview/PreviewView.java
scheglov: Not sure which "both".
I speculatively changed the name of the parameter and the name of the analytics event.
PTAL

devoncarew: Sorry, I meant both `jumpToSourceDartOutline` and `jumpToSourceFlutterOutline` could just be simplified to `jumpToSource` - that we don't need to differentiate between the source element type.

Issue #1761, opened 2018-02-11
Special characters in Flutter project description break project


matthew-carroll:
## Steps to Reproduce

1. Launch the new project wizard
2. Enter a description for the project that includes a colon ":"
3. Create the project.

## Result

Project fails to open in IDEA correctly.  Some kind of error dialog pops up.  I was able to rectify the situation by opening the YAML file and removing the colon from the description and then relaunching the IDE.

## Version info

[✓] Flutter (on Mac OS X 10.12.6 16G29, locale en-US, channel alpha)
    • Flutter version 0.0.21 at [redacted]
    • Framework revision 2e449f06f0 (12 days ago), 2018-01-29 14:26:51 -0800
    • Engine revision 6921873c71
    • Tools Dart version 2.0.0-dev.16.0
    • Engine Dart version 2.0.0-edge.da1f52592ef73fe3afa485385cb995b9aec0181a

[✓] Android toolchain - develop for Android devices (Android SDK 26.0.1)
    • Android SDK at [redacted]
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-26, build-tools 26.0.1
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_112-release-b06)

[✓] iOS toolchain - develop for iOS devices (Xcode 9.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 9.1, Build version 9B55
    • ios-deploy 1.9.2
    • CocoaPods version 1.3.1

[✓] Android Studio
    • Android Studio at /Applications/Android Studio 3.0 Preview.app/Contents
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-915-b01)

[✓] Android Studio
    • Android Studio at /Applications/Android Studio 3.1 Preview.app/Contents
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1012-b01)

[✓] Android Studio (version 2.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Java version OpenJDK Runtime Environment (build 1.8.0_112-release-b06)

[✓] IntelliJ IDEA Community Edition (version 2017.3.4)
    • Flutter plugin version 21.2.3
    • Dart plugin version 173.4548.30

[✓] Connected devices
[...]

PR #1773, opened 2018-02-14
Validate descriptions

pq:
Fixes: #1761.

image

Not pretty or ideal but consistent with the name validation UX and is another nudge to keep folks on the rails.

@stevemessick @devoncarew

### DIFF
--- a/src/io/flutter/module/FlutterModuleBuilder.java
+++ b/src/io/flutter/module/FlutterModuleBuilder.java
@@ -137,6 +137,14 @@ public class FlutterModuleBuilder extends ModuleBuilder {
     return flutter;
   }
 
+  private static String validateSettings(FlutterCreateAdditionalSettings settings) {
+    final String description = settings.getDescription();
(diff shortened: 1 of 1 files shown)


src/io/flutter/module/FlutterModuleBuilder.java
```diff
@@ -137,6 +137,14 @@ public Module commitModule(@NotNull Project project, @Nullable ModifiableModuleM
     return flutter;
   }
 
+  private static String validateSettings(FlutterCreateAdditionalSettings settings) {
+    final String description = settings.getDescription();
+    if (description != null && description.contains(": ")) {
```
devoncarew: Would just a `':'` be an error? If so, perhaps shorten from checking for `': '`.

pq: Nope.  Steve pointed out that it needs the trailing space to be (mis)treated as a delimiter.

Issue #1697, opened 2018-01-30
Add speed search for the Inspector view.


scheglov:
See https://github.com/flutter/flutter-intellij/pull/1696 for how it is done for Preview.

PR #1776, opened 2018-02-15
Basic speed search for the Inspector Tree.

pq:
Fixes: #1697

image

This gets us simple search using Widget `toString()`s.  The Preview View takes it further (supporting path searches and more).  If we want to do that, we should consider adding something like a `getSpeedSearch()` method to `DiagnosticNode`.  Happy to follow-up some refinements if there's interest!

@jacob314 @scheglov

### DIFF
--- a/src/io/flutter/view/InspectorPanel.java
+++ b/src/io/flutter/view/InspectorPanel.java
@@ -426,6 +426,20 @@ public class InspectorPanel extends JPanel implements Disposable, InspectorServi
     TreeUtil.installActions(tree);
 
     PopupHandler.installUnknownPopupHandler(tree, createTreePopupActions(), ActionManager.getInstance());
+
+    new TreeSpeedSearch(tree) {
+      @Override
+      protected String getElementText(Object element) {
+        final TreePath path = (TreePath)element;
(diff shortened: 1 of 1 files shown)


scheglov:
LGTM

BTW, I did not get highlighting matching substrings without special code in our `ColoredTreeCellRenderer` implementation. Does it work for you in the inspector?

pq:
@scheglov : good catch.  TBH I didn't even notice the rendering.  Looks much better with it, so thanks!

pq:
EDIT: the rendering bits aren't quite ready to land...  I need to make a few tweaks.

src/io/flutter/view/InspectorPanel.java
```diff
@@ -782,6 +804,10 @@ public void customizeCellRenderer(
         setIcon(icon);
       }
     }
+
+    private void appendSearch(@NotNull String text, @NotNull SimpleTextAttributes attributes) {
```
jacob314: nit: maybe just keep calling this append as it is still just appending text but now it is appending text in a way that is compatible with search.

pq: `append` is final in super but I agree so made it `appendText` 👍

Issue #1754, opened 2018-02-08
upgrade the bots to start compiling against 2018.1


devoncarew:
We should start building and compiling against 2018.1, even if we don't yet ship support for that release.

PR #1777, opened 2018-02-15
support 2018.1; remove 2017.2

devoncarew:
- support 2018.1 EAP
- remove 2017.2 - I don't believe we need to support this any longer (we still need to support 2017.1, for Android Studio)
- fix https://github.com/flutter/flutter-intellij/issues/1754

Also, remove the rest of the old ant based build system.

@stevemessick 

Steve, I'd definitely appreciate a close eye on the version range changes, to make sure it all looks good.


devoncarew:
(this needs a bit more work)

devoncarew:
This has been updated to run `bin/plugin build` on travis, on 3 separate bots, and to allow individual builds per platform, instead of one for all supported platforms, to better allow sharding them into bots.

devoncarew:
And, a green build :) https://travis-ci.org/flutter/flutter-intellij/builds/341760226?utm_source=github_status&utm_medium=notification

tool/plugin/lib/plugin.dart
```diff
@@ -515,6 +480,9 @@ class BuildCommand extends ProductCommand {
   final BuildCommandRunner runner;
 
   BuildCommand(this.runner) : super('build') {
+    argParser.addOption('only-version',
```
stevemessick: This could also be useful for debugging. Adding an 'abbr' would be handy, but not urgent.

tool/travis.sh
```diff
@@ -19,20 +19,5 @@ echo "pub get"
...
-  fi
-else
-  if [ -z "$DART_PLUGIN_VERSION" ]
-  then
-    ant build -Didea.product=$IDEA_PRODUCT -Didea.version=$IDEA_VERSION
-  else
-    ant build -Didea.product=$IDEA_PRODUCT -Didea.version=$IDEA_VERSION -Ddart.plugin.version=$DART_PLUGIN_VERSION
-  fi
-fi
+# Run the build.
```
stevemessick: Nice simplification :)

tool/plugin/lib/plugin.dart
devoncarew: sounds good, will add an abbreviation

devoncarew: (`-o`)

stevemessick: LGTM

PR #1778, opened 2018-02-15
Inspector support for multiple running applications.

jacob314:
Now each running app gets its own inspector tabs and clicking the toolbar buttons in the tab only impact that app.
Previously, the toolbar buttons would only impact the last running app leaving the other apps surprisingly unresponsive.
The design is now that inspector tabs close automatically when the apps stop making the lifecycles a bit easier resulting in removing some of the stopped events.

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -18,7 +18,6 @@ import com.intellij.openapi.components.PersistentStateComponent;
 import com.intellij.openapi.components.Storage;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.SimpleToolWindowPanel;
-import com.intellij.openapi.util.Computable;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.wm.ToolWindow;
 import com.intellij.openapi.wm.ToolWindowManager;
@@ -39,10 +38,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 // TODO(devoncarew): Display an fps graph.
 
@@ -52,7 +48,17 @@ import java.util.Map;
   name = "FlutterView",
   storages = {@Storage("$WORKSPACE_FILE$")}
 )
+
+
 public class FlutterView implements PersistentStateComponent<FlutterViewState>, Disposable {
+
+  private static class PerAppState {
+    ArrayList<FlutterViewAction> flutterViewActions = new ArrayList<>();
+    ArrayList<InspectorPanel> inspectorPanels = new ArrayList<>();
+    ArrayList<Content> contents = new ArrayList<>();
+    boolean sendRestartNotificationOnNextFrame = false;
+  }
+
   public static final String TOOL_WINDOW_ID = "Flutter Inspector";
 
   @NotNull
@@ -63,12 +69,7 @@ public class FlutterView implements PersistentStateComponent<FlutterViewState>, 
 
   private String restoreToolWindowId;
 
-  @Nullable
-  FlutterApp app;
-
-  private final List<FlutterViewAction> flutterViewActions = new ArrayList<>();
-
-  private final ArrayList<InspectorPanel> inspectorPanels = new ArrayList<>();
+  private final Map<FlutterApp, PerAppState> perAppViewState = new HashMap<>();
 
   public FlutterView(@NotNull Project project) {
     myProject = project;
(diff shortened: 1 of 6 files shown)


src/io/flutter/view/FlutterViewFactory.java
```diff
@@ -33,9 +33,5 @@ private static void initFlutterView(@NotNull Project project, FlutterViewMessage
 
   @Override
   public void createToolWindowContent(@NotNull Project project, @NotNull ToolWindow toolWindow) {
-    //noinspection CodeBlock2Expr
-    DumbService.getInstance(project).runWhenSmart(() -> {
-      (ServiceManager.getService(project, FlutterView.class)).initToolWindow(toolWindow);
-    });
```
pq: I'm missing some context but I'm surprised that this is no longer needed?

src/io/flutter/view/FlutterViewAction.java
```diff
@@ -7,37 +7,35 @@
...
     super(text, description, icon);
 
-    this.view = view;
+    this.app = app;
   }
 
   @Override
   public void update(@NotNull AnActionEvent e) {
-    e.getPresentation().setEnabled(view.getFlutterApp() != null);
+    e.getPresentation().setEnabled(app != null && app.isSessionActive());
```
pq: I think app is sure to not be null here?

src/io/flutter/view/FlutterViewToggleableAction.java
```diff
@@ -19,17 +19,17 @@
...
-  FlutterViewToggleableAction(@NotNull FlutterView view, @Nullable String text, @Nullable String description, @Nullable Icon icon) {
-    super(view, text, description, icon);
+  FlutterViewToggleableAction(@NotNull FlutterApp app, @Nullable String text, @Nullable String description, @Nullable Icon icon) {
+    super(app, text, description, icon);
   }
 
   @Override
   public final void update(@NotNull AnActionEvent e) {
-    final boolean hasFlutterApp = view.getFlutterApp() != null;
+    final boolean hasFlutterApp = app != null && app.isSessionActive();
```
pq: `app` is `@NotNull` I think?

src/io/flutter/view/FlutterViewToggleableAction.java
```diff
@@ -42,7 +42,7 @@ public final void update(@NotNull AnActionEvent e) {
 
   @Override
   public void actionPerformed(AnActionEvent event) {
-    if (view.getFlutterApp() == null) {
+    if (app == null) {
```
pq: And here, I think we can get rid of this test since `app` should never be null?

src/io/flutter/view/InspectorPanel.java
```diff
@@ -88,8 +87,8 @@
   // with the Dracula and the default themes.
```
pq: => Darcula

devoncarew:
Hey @jacob314, I patched this in locally;

screen shot 2018-02-15 at 4 54 12 pm

screen shot 2018-02-15 at 5 03 07 pm

The UI for this is a bit confusing. It's not clear exactly why there are 4 tabs (widgets, widgets, render tree, render tree), or which ones are for which apps.

I don't think we can open multiple tool windows for the inspector - if we could, that might be ideal (that was what I meant when we discussed over VC).

The Debug view uses one tool window, and I believe one tab per running app. That won't work great for the inspector, as it already uses tabs for the different DOM trees.

If we can't wrangle a UI for this that we're happy with, perhaps more of a fast fail solution? Not showing an inspector for a 2nd launch? Or, we might be able to swap the content out for whatever app happens to to be in the foreground.

jacob314:
I agree the multiple tabs per app are confusing. What do you think about tweaking the UI so instead there is a menu item on the inspector toolbar  that lets you switch between the Widget and Render Tree?
That design would also make sense in the future if we add the Semantics Tree and Layer Tree. I think it would also make it less surprising that that the trees are linked relative to separate tabs.

After having a chance to play around with it, I think supporting multiple apps correctly actually does have a real benefit. Having multiple apps open would have improved a couple of the examples at Dart Conf as you can have one object of interest selected on one device and the other selected on the other making it much easier to go back and forth to see differences.

src/io/flutter/view/FlutterViewFactory.java
jacob314: I'm missing the context on this method so I don't quite understand what is going on here.

We don't have an initToolWindow method anymore as the tool window is blank until an app starts.
I can't figure out how to locally trigger the plugin without this tool window enabled when an app starts. Maybe I'm doing something wrong. If there is a way to reach that case, then we should be initializing the tool window with all currently running applications. I'm not clear how you would get that list of running applications though.

src/io/flutter/view/InspectorPanel.java

src/io/flutter/view/FlutterViewToggleableAction.java
jacob314: yep. removed these legacy checks.

src/io/flutter/view/FlutterViewToggleableAction.java
jacob314: done

src/io/flutter/view/FlutterViewAction.java
jacob314: done

devoncarew:
> I agree the multiple tabs per app are confusing. What do you think about tweaking the UI so instead there is a menu item on the inspector toolbar that lets you switch between the Widget and Render Tree?
> That design would also make sense in the future if we add the Semantics Tree and Layer Tree. I think it would also make it less surprising that that the trees are linked relative to separate tabs.

So,
- I like having separate 'widget' and 'rendering' tabs, as things are very clear and discoverable to users
- that doesn't scale great when we have 4 tress to display (semantics and layers)
- switching trees is a much more common use case than switching running apps, so we'd want to optimize for that

and, the current ui for multiple apps and trees doesn't tell the user which tab is associated with which app (they're just named 'widgets')

> After having a chance to play around with it, I think supporting multiple apps correctly actually does have a real benefit. Having multiple apps open would have improved a couple of the examples at Dart Conf as you can have one object of interest selected on one device and the other selected on the other making it much easier to go back and forth to see differences.

Here's one proposal:
- we keep the different types of tree as tabs for now
- for multiple apps, we provide a combo box pulldown - above all the content - that's visible when there is more than one app running. It would allow you to switch between the app content that's displayed in the view

Having it be large - say, full width in the UI - means that when multiple apps are running the way to switch apps is discoverable. A toolbar button wouldn't be as easy to locate

devoncarew:
From an off-line discussion we can land this as is. We'll look to adjust the UX for supporting multiple apps in future work; this will likely take the form of a view w/ multiple tabs per running app (similar to the debug view), and some additional mechanism to select the tree to view (widget, render, ...).

src/io/flutter/view/FlutterViewFactory.java
jacob314: I added this back so I could call
toolWindow.setToHideOnEmptyContent(true)
to the initToolWindow code.
So the flutter inspector isn't shown when there is no active debug session.

src/io/flutter/view/FlutterView.java
```diff
@@ -89,61 +90,71 @@ public void loadState(FlutterViewState state) {
     this.state.copyFrom(state);
   }
 
-  public void initToolWindow(@NotNull ToolWindow toolWindow) {
-    final ContentFactory contentFactory = ContentFactory.SERVICE.getInstance();
+  public void initToolWindow(ToolWindow window) {
+    window.setToHideOnEmptyContent(true);
```
devoncarew: Interesting! I didn't know about this method. I wonder how it determines whether the view is empty or not?

jacob314: Empty content means that the contents (list of tabs) is empty. We now remove tabs when each flutter application is stopped matching the behavior of the debugger pane for better or worse.

PR #1784, opened 2018-02-19
update when we use the flutter icon for modules

devoncarew:
- update when we use the flutter icon for modules

I noticed in some situations the icons was not showing for flutter modules, and was instead showing for non-flutter ones. Before this change:

screen shot 2018-02-18 at 7 13 24 pm

(perf_tool is a web module, and conference_app is a flutter one); after this change:

screen shot 2018-02-18 at 7 22 41 pm

@pq

### DIFF
--- a/src/io/flutter/project/FlutterIconProvider.java
+++ b/src/io/flutter/project/FlutterIconProvider.java
@@ -40,9 +40,9 @@ public class FlutterIconProvider extends IconProvider {
       final VirtualFile file = ((PsiDirectory)element).getVirtualFile();
       if (!file.isInLocalFileSystem()) return null;
 
-      // Project root.
-      final VirtualFile baseDir = project.getBaseDir();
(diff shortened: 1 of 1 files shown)


pq:
FWIW I don't think we're entirely out of the woods here.

image

(Testing w/ M22.)

One guess is that there's a race condition w/ the Dart Plugin and sometimes (as in the screenshot above) Dart wins...  Maybe this is something we can fix by adjusting priority for the IconProvider?

  <iconProvider implementation="io.flutter.project.FlutterIconProvider" order="first"/>

FWIW: I did adjust order but it didn't seem to do it so there may be more going on...

cc @alexander-doroshko for ideas...

alexander-doroshko:
Which icon do you mean? Dart plugin is not trying to change the content root icon at all.

pq:
This one:

image

If the Dart Plugin is doing nothing, then this is even more curious!

Issue #1717, opened 2018-02-01
Track analytics to understand how the inspector is being used


jacob314:
At minimum track:
* Number of times the inspect button is selected
* Number of times a widget is inspected on device
* Number of times the render object view is opened.
* Number of times jump to source location is called?

Optional:
* Track stats about expanding and collapsing widgets in tree
* Number of times jump to type definition is called.
* Track stats on scrolling within the view?

PR #1788, opened 2018-02-20
Inspector usage analytics (#1717)

pq:
Adds analytics for:

* inspect button selections
* jump to source actions
* jump to type source actions
* render tree activations.

Fixes: #1717

@jacob314 @devoncarew


src/io/flutter/inspector/JumpToSourceAction.java
```diff
@@ -8,8 +8,13 @@
 import com.intellij.xdebugger.XSourcePosition;
 import com.intellij.xdebugger.frame.XNavigatable;
 import com.intellij.xdebugger.frame.XValue;
+import org.jetbrains.annotations.NotNull;
 
 public class JumpToSourceAction extends JumpToSourceActionBase {
+  public JumpToSourceAction() {
+    super(InspectorActions.JUMP_TO_SOURCE);
```
pq: @devoncarew : what's the preferred format for these IDs?  Would "jumpToSource" be more consistent?

pq: FWIW, updated...

src/io/flutter/inspector/JumpToSourceActionBase.java
```diff
@@ -12,18 +12,34 @@
...
+  }
+
   @Override
   protected void perform(final DefaultMutableTreeNode node, final DiagnosticsNode diagnosticsNode, final AnActionEvent e) {
     final Project project = e.getProject();
     if (project == null) {
       return;
     }
+
+    FlutterInitializer.getAnalytics().sendEvent("inspector", id);
```
devoncarew: 👍 looks good

src/io/flutter/view/FlutterView.java
```diff
@@ -104,8 +103,26 @@ public void initToolWindow(@NotNull ToolWindow toolWindow) {
...
+  private static void listenForRenderTreeActivations(@NotNull ToolWindow toolWindow) {
+    final ContentManager contentManager = toolWindow.getContentManager();
+    contentManager.addContentManagerListener(new ContentManagerAdapter(){
+      @Override
+      public void selectionChanged(ContentManagerEvent event) {
+        final ContentManagerEvent.ContentOperation operation = event.getOperation();
+        if (operation == ContentManagerEvent.ContentOperation.add) {
+          final String name = event.getContent().getTabName();
+          if (Objects.equals(name, RENDER_TREE_LABEL)) {
+            FlutterInitializer.getAnalytics().sendEvent("inspector", "renderTreeSelected");
```
devoncarew: We're tracking what tree we're in here - or when they switch. Should we send for both switching to the render tree, and switching to the widget one?

We also track what screens people are in, so 'flutter inspector', that _could_ map well here? It may be difficult to determine that the screen should be 'flutter inspector / widgets'.

pq: Sure.  Will update! 👍

PR #1792, opened 2018-02-20
Compile against 3.1 beta 3

stevemessick:
@devoncarew @pq

Fix compile errors to run in debug mode.
Compile cleanly with AS 3.1 beta 3.
Upload binaries for plugin build cmd.
Do NOT add new target to travis because it isn't tested yet at all.
If you want to test it you should be able to make a local copy via:
```bash
./bin/plugin build -o3.1
```


devoncarew:
Very cool! Looks like it compiles cleanly against 3.0 as well.

PR #1793, opened 2018-02-20
Highlight Outline tree background for Flutter elements.

scheglov:
This PR is mostly to demo that we can change background.
I'm not sure that we want to, or that the color is right :-)
image

### DIFF
--- a/src/io/flutter/preview/PreviewView.java
+++ b/src/io/flutter/preview/PreviewView.java
@@ -27,10 +27,7 @@ import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.openapi.wm.ToolWindow;
 import com.intellij.openapi.wm.ex.ToolWindowEx;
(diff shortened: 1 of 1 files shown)


devoncarew:
Cool! I think we also have the ability to prefix more than one icon to a node, so we could have an icon that meant 'flutter widget', in addition to show a icon per widget type.

devoncarew:
This one for instance: https://github.com/flutter/flutter-intellij/blob/master/resources/icons/flutter%402x.png.

devoncarew:
I think we've gone a long way here w/ the flutter badge icons.

If we do want to land something like there, there are some theme colors we could use. The Project view has a different background color for items that are excluded from the project (separate from the selected item background color):

screen shot 2018-02-23 at 2 37 46 pm

And the tree view there also shows items that git thinks are modified with a different text foreground color.

I think using the background color might be a it more than we need, but we could experiment w/ it - land it and back it out if we think it's too busy.

scheglov:
Closing for now.

PR #1797, opened 2018-02-20
Flutter View Action analytics.

pq:
Follow-up from: https://github.com/flutter/flutter-intellij/pull/1788#pullrequestreview-97972127.

Instruments all flutter view actions (not just toggleable ones).

@devoncarew

### DIFF
--- a/src/io/flutter/view/FlutterViewAction.java
+++ b/src/io/flutter/view/FlutterViewAction.java
@@ -7,6 +7,7 @@ package io.flutter.view;
 
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.project.DumbAwareAction;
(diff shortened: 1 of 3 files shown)


src/io/flutter/view/FlutterViewAction.java
```diff
@@ -28,6 +29,14 @@
     this.app = app;
   }
 
+  @Override
+  public void actionPerformed(AnActionEvent e) {
```
devoncarew: `final`?

devoncarew: Nevermind, I think that would break `FlutterViewToggleableAction`.

pq: Exactly right.  (I tried that! 😄 )

PR #1785, opened 2018-02-19
explore building for AS 3.1

devoncarew:
- explore building for AS 3.1 (not for landing)

This PR updates our 2017.3 build to build against Android Studio Beta 1. @stevemessick, this doesn't try to address the compile issues, mostly just flush out where we are.

PR #1799, opened 2018-02-20
compile against AS 3.1 when building 2017.3

devoncarew:
- compile against AS 3.1 when building 2017.3 (follow up from https://github.com/flutter/flutter-intellij/pull/1785)
- rename the `3.0` artifact directory to `2017.1` (for similarity with the 2017.3 and 2018.1 directories)

w/ this change, we now build:
- 2017.1 (IntelliJ and Android Studio)
- 2017.3 (IntelliJ and Android Studio)
- 2018.1 (IntelliJ)

@stevemessick

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,6 @@ script: ./tool/travis.sh
 # Testing product matrix; the values should match the version field in the
 # product-matrix.json file.
 env:
-  - IDEA_VERSION=2017.1
-  - IDEA_VERSION=2017.3
+  - IDEA_VERSION=3.0
+  - IDEA_VERSION=3.1
   - IDEA_VERSION=2018.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,6 @@ script: ./tool/travis.sh
 # Testing product matrix; the values should match the version field in the
 # product-matrix.json file.
 env:
-  - IDEA_VERSION=3.0
+  - IDEA_VERSION=2017.1
   - IDEA_VERSION=2017.3
   - IDEA_VERSION=2018.1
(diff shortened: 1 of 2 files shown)


stevemessick:
Seems to me there are two problems here:
1. AS 3.1 works with Dart plugin 173.4301.22
2. Putting both AS and IJ builds into the same directory breaks `plugin build` since the plugin files all have the same name

devoncarew:
I'll change this to `173.4301.22`.

Re: the 2nd change, we always build something called `flutter-intellij.zip`, whether that binary support intellij or intellij and android studio. The only practical effect of this change is that in the artifacts directory, the subdirs are named `2017.3` instead of `3.1`. I think that makes it a bit clearer what api version the binary targets, but happy to revert this.

devoncarew:
(reverted the change for the dart plugin version)

stevemessick:
Forgot to mention it also breaks `plugin deploy` which we should start using someday. Both `plugin build` and `plugin deploy` expect to have all the artifacts available at the same time. The (former) 3.0 builds did not include the fluter-studio plugin, unless I'm misremembering.

stevemessick:
I guess I don't understand the intent of these changes. Removing 3.0, 3.1 removes clarity for me.

devoncarew:
I'll revert the change from 3.0 => 2017.1. But I think that will want one 2017.3 compatible artifact, instead of one in `2017.3/flutter-intellij.zip` which is compatible with 2017.3 and supports IntelliJ, and one in `3.1/flutter-intellij.zip` which is compatible with 2017.3 and supports IntelliJ and android studio. We'll only really be able to upload one to the plugin store, and I think we want that one to be the one built for IntelliJ and android stdio.

stevemessick:
Now I understand. I'm fine with just using the IntelliJ nomenclature if you prefer. Sorry if that means you had to do an unnecessary push.

PR #1811, opened 2018-02-22
make the flutter icon in the outline view more subdued

devoncarew:
- make the flutter icon in the outline view more subdued

This is similar to what the IntelliJ structure view does for Java elements.

screen shot 2018-02-22 at 12 52 08 pm

@scheglov


src/io/flutter/preview/PreviewView.java
```diff
@@ -766,14 +766,17 @@ Icon getFlutterDecoratedIcon() {
...
     if (decorated == null) {
-      final Icon prefix = FlutterIcons.Flutter;
+      final Icon badgeIcon = FlutterIcons.Flutter_badge;
 
       decorated = new LayeredIcon(2);
-      decorated.setIcon(prefix, 0, 0, 0);
-      decorated.setIcon(icon, 1, prefix.getIconWidth(), 0);
+      decorated.setIcon(icon, 0, 0, 0);
+      decorated.setIcon(badgeIcon, 1,
+                        icon.getIconWidth() + (icon.getIconWidth() - badgeIcon.getIconWidth()) / 2,
```
scheglov: What `(icon.getIconWidth() - badgeIcon.getIconWidth()) / 2` does?
It looks like we need a whitespace between these two items, but the way we compute is peculiar :-)

devoncarew: It's a 8x8 icon centered in a 16x16 area (as if there were two 16x16 icons right next to each other).

scheglov: Ah, makes sense.
Thanks.

devoncarew:
some slight additional tweaks to the look:

screen shot 2018-02-22 at 8 48 18 pm

PR #1813, opened 2018-02-23
Make the Flutter Inspector use two levels of nested tabs.

jacob314:
This makes the multiple app case look much better compared to having a
confusing single list of Widgets and Render Tree tabs for all the running
apps.

Any ideas on the right way to add a separator between the words Flutter Inspector and the device name?
I tried to add the Flutter icon but it doesn't show unless the icon is clicked so doesn't help.

Here I want a separator between the words
`Flutter Inspector` and `Pixel 2 XL`
image

The multi tab case works well. The problem is the IntelliJ UI is clever and doesn't show a single tab like a tab which looks bad as I can't see how to give the content an icon... Probably there is something obvious I'm just missing or maybe someone has a better idea than showing redundant Flutter icons.
image

Note that if you had multiple devices with the same name I am able to reuse existing code to give unique names for the tabs.

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -45,14 +45,9 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 
 // TODO(devoncarew): Display an fps graph.
-
-// TODO(devoncarew): Ensure all actions in this class send analytics.
 
 @com.intellij.openapi.components.State(
   name = "FlutterView",
@@ -156,7 +151,7 @@ public class FlutterView implements PersistentStateComponent<FlutterViewState>, 
     final ContentFactory contentFactory = ContentFactory.SERVICE.getInstance();
     final SimpleToolWindowPanel toolWindowPanel = new SimpleToolWindowPanel(true);
     final JBRunnerTabs tabs = new JBRunnerTabs(myProject, ActionManager.getInstance(), null, this);
-    ArrayList<FlutterDevice> existingDevices = new ArrayList<>();
+    final List<FlutterDevice> existingDevices = new ArrayList<>();
     for (FlutterApp otherApp : perAppViewState.keySet()) {
       existingDevices.add(otherApp.device());
     }
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -7,6 +7,7 @@ package io.flutter.view;
 
 import com.intellij.execution.runners.ExecutionUtil;
 import com.intellij.execution.ui.ConsoleViewContentType;
+import com.intellij.execution.ui.layout.impl.JBRunnerTabs;
 import com.intellij.icons.AllIcons;
 import com.intellij.ide.BrowserUtil;
 import com.intellij.ide.browsers.BrowserLauncher;
@@ -25,6 +26,11 @@ import com.intellij.openapi.wm.ToolWindowManager;
 import com.intellij.openapi.wm.ex.ToolWindowEx;
 import com.intellij.openapi.wm.ex.ToolWindowManagerEx;
 import com.intellij.ui.content.*;
+import com.intellij.ui.*;
+import com.intellij.ui.content.Content;
+import com.intellij.ui.content.ContentFactory;
+import com.intellij.ui.content.ContentManager;
+import com.intellij.ui.tabs.TabInfo;
 import icons.FlutterIcons;
 import io.flutter.FlutterBundle;
 import io.flutter.FlutterInitializer;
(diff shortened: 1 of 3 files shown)


pq:
> Any ideas on the right way to add a separator between the words Flutter Inspector and the device name?

Not sure.  Maybe @alexander-doroshko has ideas?

src/io/flutter/view/FlutterView.java
```diff
@@ -1,671 +1,687 @@
...
+
+  import javax.swing.*;
+  import java.util.ArrayList;
+  import java.util.HashMap;
+  import java.util.Map;
+  import java.util.Objects;
+
+  // TODO(devoncarew): Display an fps graph.
+
+  // TODO(devoncarew): Ensure all actions in this class send analytics.
```
devoncarew: note: we can delete this todo now

src/io/flutter/view/FlutterView.java
```diff
@@ -1,671 +1,687 @@
...
-      content.setCloseable(true);
-      final SimpleToolWindowPanel windowPanel = new SimpleToolWindowPanel(true, true);
-      content.setComponent(windowPanel);
-
-      final InspectorPanel inspectorPanel = new InspectorPanel(this, flutterApp, flutterApp::isSessionActive, treeType);
-      windowPanel.setContent(inspectorPanel);
-      windowPanel.setToolbar(ActionManager.getInstance().createActionToolbar("FlutterViewToolbar", toolbarGroup, true).getComponent());
+      final SimpleToolWindowPanel toolWindowPanel = new SimpleToolWindowPanel(true);
+      final JBRunnerTabs tabs = new JBRunnerTabs(myProject, ActionManager.getInstance(), null, this);
+      ArrayList<FlutterDevice> existingDevices = new ArrayList<>();
```
devoncarew: I think we should prefer the interfaces for the LHS (`List`).

devoncarew:
I patched this in - this looks great! The tabs make sense - both for multiple devices, and for the different trees in the devices. Toolbar position looks good.

screen shot 2018-02-26 at 4 40 49 pm

screen shot 2018-02-26 at 4 41 36 pm

The tool window title gets long with multiple running apps, but I don't think that's necessarily something worth addressing. At some point we'll remove the feedback button, which will give the title area more room.

The Widgets+RenderTree+Toolbar area did get pretty wide - at one width for me the toolbar overflowed a bit. Again, not a huge issue.

The state of the toolbar icons made sense for each running app - I could toggle the perf. overlay on for one app, that icon would stay toggled, and toggle it independently for the other running app.

We may want to look at moving the toolbar to above the Widget+RenderTree tabs. That would help a bit with signifying containment, and with some horiz. space issues. It would give us a bit more room to add more tabs for other trees, like the semantic one.

alexander-doroshko:
>> Any ideas on the right way to add a separator between the words Flutter Inspector and the device name?

> Not sure. Maybe @alexander-doroshko has ideas?

I don't know without digging. I can only say that in 2018.1 tool windows headers are tuned a bit so you may also want to check how it looks there.

jacob314:
Pushed a new branch that cleans up the diff. I had a fix for the busted indentation but I'd failed to push it properly to my branch.

src/io/flutter/view/FlutterView.java
jacob314: removed this TODO but not the one above it.

jacob314:
To some degree I think the length of the tool window headers is a good thing as it pushes users to make sure the window is wide enough to be usable navigating around the large widget trees.

I'm torn about moving the toolbar above the existing tabs. Keeping the toolbar with the tabs feels a bit more consistent with other IntelliJ tool windows and conserves vertical space. 
Particularly when I have the inspector at the bottom of the window next to the console which I think is the best place for it generally, I feel like vertical space is at a premium. What do you think of having this be the default location for the inspector window?
image

src/io/flutter/view/FlutterView.java
jacob314: done

devoncarew:
> What do you think of having this be the default location for the inspector window?

Possibly? It would then fight for space with the console view - something developers would also want to see when running their app.

In that pic above the console area is sized a bit higher than user's might normally have it.

If we do want to go more with a horizontal orientation, the properties area could go on the right hand side.

And - if we changed the view to be displayed on the bottom - we could duplicate the `Toggle Select Widget Mode` toolbar action into the debug and run views, as alternative ways to access the inspector view.

devoncarew:
> Keeping the toolbar with the tabs feels a bit more consistent with other IntelliJ tool windows

👍, and I like the idea of experimenting with a horizontal layout for the view.

devoncarew:
There's also the possibility of adding this as a new tab in the debugger (like the idea.log tab below):

screen shot 2018-02-27 at 9 21 16 am

But I think that's a net loss in terms of discoverability.

jacob314:
I agree this should be a sister tab of the debugger not a tab under the debugger as to me the debugger in IntelliJ implies you are stopped on a breakpoint while the inspector is best if you are not stopped as a breakpoint.
I think adding a `Toggle Select Widget Mode` action to the debug and run views is a good idea. Filed a bug to track that as part of making the default docking the inspector view vertically. I'm not sure it is a good idea yet but I think it is promising enough to track. Agreed it requires making the properties go on the right hand side to be useful.

jacob314:
The tabs in Intellij 2018 are really subtle so I think having icons to break up the tabs would definitely be a win even with multiple apps open.
image

PR #1814, opened 2018-02-23
tweak the outline view icons

devoncarew:
- tweak the icons used in the outline view; move from 24x24 icons to 32x32, and make the color not grey (use `#5ab4f0`), do have greater contrast between the disabled and enabled states

screen shot 2018-02-22 at 5 51 13 pm

@scheglov


devoncarew:
I'm landing this, but another option here is to explore 28x28 icons (w/ HD, 56x56), and two variants of a gray color instead of the blue - one gray for the normal theme, and on gray for use with the darkula theme.

PR #1815, opened 2018-02-23
additional release version validation

devoncarew:
- additional release version validation

This validates that the release version matches something like `00.0` (major.minor). Steve, I'm adding this because I passed in a version of `22` recently, and almost published 22.1, 22.2, and 22.3 instead of 22.0.1, 22.0.2, and 22.0.3.

@stevemessick


devoncarew:
Looks like I have a travis failure...

devoncarew:
(fixed)

devoncarew:
> Do we need to add more tests for this?
> I think we ought to have tests for -o, too.

I can do in this PR - it's not critical to land soon, but will be useful for the next time we need to build new plugin artifacts.

devoncarew:
This has been updated quite a bit. Added:
- added a new bot configuration; this analyzes the dart code in the repo, and runs the plugin tests on travis
- fixed some tests which had regressed w/ the additional of the 2018.1 build; fixed some tests that had async issues
- added tests for `--only-version`, and for the release number validation

devoncarew:
It looks like running `dart test/plugin_test.dart` on the DART_BOT bot actually performs the build - it doesn't run the tests. I have no idea why it would be doing that.

devoncarew:
If I try running `pub run test` instead, I see:

```
00:00 +0: All tests skipped.                      
```

devoncarew:
Fixed :) The tests now run: https://travis-ci.org/flutter/flutter-intellij/jobs/346039461#L668

tool/plugin/test/plugin_test.dart
```diff
@@ -178,6 +218,8 @@ BuildCommandRunner makeTestRunner() {
 class TestBuildCommand extends BuildCommand {
   TestBuildCommand(runner) : super(runner);
 
+  bool get isTesting => true;
```
stevemessick: We probably should've defined a mixin for testing. Something to keep in mind for later.

PR #1816, opened 2018-02-23
Show assist actions also in Outline popup.

scheglov:
image

### DIFF
--- a/src/io/flutter/preview/PreviewView.java
+++ b/src/io/flutter/preview/PreviewView.java
@@ -7,6 +7,7 @@ package io.flutter.preview;
 
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.common.util.concurrent.Uninterruptibles;
 import com.intellij.icons.AllIcons;
 import com.intellij.ide.BrowserUtil;
 import com.intellij.ide.CommonActionsManager;
@@ -64,6 +65,7 @@ import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 import java.util.*;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 @com.intellij.openapi.components.State(
   name = "FlutterPreviewView",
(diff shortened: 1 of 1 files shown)


devoncarew:
Awesome!

I think we can omit items from the menu instead of showing them as disabled (@InMatrix to confirm).

If we're going to show icons there, we should probably use more std 16x16 ones. I'm busy fiddling w/ the icons for the toolbar, so having 16x16 icons for this PR isn't necessary.

scheglov:
PTAL
image

src/io/flutter/preview/PreviewView.java
```diff
@@ -285,6 +311,40 @@ public void propertyChange(PropertyChangeEvent evt) {
     contentManager.setSelectedContent(content);
   }
 
+  private ActionGroup createTreePopupActions() {
+    // The corresponding tree item has just been selected.
+    // Wait short time for receiving assists from the server.
+    for (int i = 0; i < 20 && actionToChangeMap.isEmpty(); i++) {
```
devoncarew: This seems practical, but I wonder if there are any other ways of doing this. Perhaps an IntelliJ utility API?

@jwren, who I believe has used an API to see it there are any event interrupts?

src/io/flutter/preview/PreviewView.java
```diff
@@ -205,6 +222,15 @@ public void initToolWindow(@NotNull ToolWindow toolWindow) {
     tree.setCellRenderer(new OutlineTreeCellRenderer());
     tree.expandAll();
 
+    tree.addMouseListener(new PopupHandler() {
```
devoncarew: Can we scope this to just items we know might have contributions? Nodes that represent widgets?

src/io/flutter/preview/PreviewView.java
```diff
@@ -205,6 +222,15 @@ public void initToolWindow(@NotNull ToolWindow toolWindow) {
...
+      public void invokePopup(Component comp, int x, int y) {
+        final ActionManager actionManager = ActionManager.getInstance();
+        final ActionGroup group = createTreePopupActions();
+        final ActionPopupMenu popupMenu = actionManager.createActionPopupMenu(ActionPlaces.UNKNOWN, group);
+        popupMenu.getComponent().show(comp, x, y);
```
devoncarew: If the group is empty, does anything show? Should we just not call toe show() method in those cases?

src/io/flutter/preview/PreviewView.java
scheglov: Unless there is anything asynchronous for popup menus (and I have not seen this anywhere yet), we have to wait with some timeout. There are different ways to implement this waiting, this is just one of them.

devoncarew: Makes sense, I just wasn't sure whether there was some prior art in the IntelliJ APIs for helping us here. lgtm as is

src/io/flutter/preview/PreviewView.java
scheglov: Done.

src/io/flutter/preview/PreviewView.java
scheglov: Yes, we did show.
I'm fixing this.

InMatrix:
This menu looks great!

PR #1818, opened 2018-02-23
Add OutlineComponent that provides PlatformDataKeys.FILE_EDITOR for undo

scheglov:
So, the Undo action works in `Flutter Outline`.


src/io/flutter/preview/PreviewView.java
```diff
@@ -715,6 +717,23 @@ public void update(AnActionEvent e) {
   }
 }
 
+class OutlineComponent extends SimpleToolWindowPanel {
```
devoncarew: Can you add a brief note about why we're subclassing (in order to proxy through undo/redo to the current active editor)?

PR #1825, opened 2018-02-25
Add 'Extract method' action.

scheglov:


src/io/flutter/preview/PreviewView.java
```diff
@@ -206,6 +206,8 @@ public void initToolWindow(@NotNull ToolWindow toolWindow) {
     toolbarGroup.add(actionMoveDown);
     toolbarGroup.addSeparator();
     toolbarGroup.add(actionRemove);
+    toolbarGroup.addSeparator();
+    toolbarGroup.add(actionExtractMethod);
```
devoncarew: Perhaps group at the end of the move up / down section?

src/io/flutter/preview/PreviewView.java
```diff
@@ -357,6 +359,11 @@ public void invokePopup(Component comp, int x, int y) {
           hasAction = true;
           group.add(actionRemove);
         }
+        group.addSeparator();
+        if (actionExtractMethod.isEnabled()) {
```
devoncarew: If we change the location in the toolbar where this is contributed, we should update here as well.

src/io/flutter/preview/PreviewView.java
```diff
@@ -783,6 +790,59 @@ boolean isEnabled() {
...
+    public void actionPerformed(AnActionEvent e) {
+      final AnAction action = ActionManager.getInstance().getAction("ExtractMethod");
+      if (action != null) {
+        final FlutterOutline outline = getWidgetOutline();
+        if (outline != null) {
+          TransactionGuard.submitTransaction(project, () -> {
+            // Ideally we don't need this - just caret at the beginning should be enough.
+            // Unfortunately this was implemented only recently.
+            // So, we have to select the widget range.
+            final int offset = outline.getOffset();
```
devoncarew: Does this need to be updated for windows `/r/n`?

Issue #1807, opened 2018-02-22
Migrate project to gradle warning


mit-mit:
With IntelliJ IDEA 2017.3.4 (Community Edition) on Windows, and a very recent flutter:
```
Flutter 0.1.5-pre.9 • channel master • git@github.com:flutter/flutter.git
Framework • revision 21c514fc56 (29 hours ago) • 2018-02-22 10:11:09 +0100
```

1. Create new Flutter project
1. Wait a bit

![image](https://user-images.githubusercontent.com/13644170/36539904-a356243a-17d8-11e8-816f-7acdc39f76db.png)

devoncarew:
This _is_ an issue, but is not a recent regression.

pq:
👍 This drives me nuts too but is not new...

cc @alexander-doroshko

stevemessick:
Somehow, I don't know how, Android Studio eliminates that prompt because it puts this line in .idea/.workspace.xml during project creation:
```xml
<component name="PropertiesComponent">
  ...
  <property name="show.migrate.to.gradle.popup" value="false" />
  ...
</component>
```
I think deleting all files prior to `flutter create` was a necessary but not sufficient condition to make that happen, but I'm not sure.

devoncarew:
I'm looking into a way to suppress this programmatically at project creation and project open.

vjbharani:
hello ,is the app supposed to open with this warning on android? because mine isin't opening.
is there any solution?

PR #1826, opened 2018-02-25
move to using JAVA_MODULE for our module types

devoncarew:
- if we see a WEB_MODULE module type on project open, convert it to a JAVA_MODULE; display a notification when we do this
- no long check to try and convert the older FLUTTER_MODULE_TYPE module types
- when opening or creating projects, disable the `show.migrate.to.gradle.popup` toast notification (fix https://github.com/flutter/flutter-intellij/issues/1807)

Changing the module type means that we get a bit more (automatic UI) in the project wizard, in the page where we ask people for the path to the flutter sdk. They'll now also see a combo box for selecting the Java SDK to use for this project. From experimenting, it's ok if this field is not populating - we can still create Flutter projects that work.

screen shot 2018-02-24 at 12 02 37 pm

screen shot 2018-02-23 at 8 18 22 am

@stevemessick for review; @jwren for FYI re: the module type change

### DIFF
--- a/src/io/flutter/FlutterInitializer.java
+++ b/src/io/flutter/FlutterInitializer.java
@@ -110,9 +110,17 @@ public class FlutterInitializer implements StartupActivity {
     // Convert all modules of deprecated type FlutterModuleType.
     if (FlutterModuleUtils.convertFromDeprecatedModuleType(project)) {
       // If any modules were converted over, create a notification
-      FlutterMessages.showInfo(FlutterBundle.message("flutter.initializer.module.converted.title"),
-                               FlutterBundle.message("flutter.initializer.module.converted.content"));
+      FlutterMessages.showInfo(
+        FlutterBundle.message("flutter.initializer.module.converted.title"),
+        "Converted from '" +
+        FlutterModuleUtils.DEPRECATED_FLUTTER_MODULE_TYPE_ID +
+        "' to '" +
+        FlutterModuleUtils.getModuleTypeIDForFlutter() +
+        "'.");
     }
+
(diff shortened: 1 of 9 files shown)


devoncarew:
(companion PR to https://github.com/flutter/flutter/pull/14867)

src/io/flutter/FlutterInitializer.java
```diff
@@ -110,10 +110,18 @@ public void runActivity(@NotNull Project project) {
     // Convert all modules of deprecated type FlutterModuleType.
     if (FlutterModuleUtils.convertFromDeprecatedModuleType(project)) {
```
jwren: I think that as-is, all WEB_MODULE projects are going to be converted to JAVA_MODULE projects, regardless if the project is Flutter.

devoncarew: oops!

I'll change this to == WEB_MODULE and has a dependency on flutter in the pubspec. I think that should be a good heuristic.

devoncarew: (done)

src/io/flutter/FlutterUtils.java
```diff
@@ -72,6 +73,20 @@ public static boolean isAndroidStudio() {
     return StringUtil.equals(PlatformUtils.getPlatformPrefix(), "AndroidStudio");
   }
 
+  public static void disableGradleProjectMigrationNotification(@NotNull Project project) {
+    // TODO(devoncarew): In Android Studio, do we want to show this notification?
```
stevemessick: I think you can delete the code that was commented-out.

PR #1840, opened 2018-02-27
fix an exception

devoncarew:
- fix an exception discovered via analytics
- do some very slight compression of stack traces to fit a few more stackes into the google analytics limit

@stevemessick

### DIFF
--- a/src/io/flutter/analytics/Analytics.java
+++ b/src/io/flutter/analytics/Analytics.java
@@ -105,6 +105,7 @@ public class Analytics {
     }
 
     String description = stringWriter.toString().trim();
+    description = description.replaceAll("com.intellij.openapi.", "c.i.o.");
     if (description.length() > maxExceptionLength) {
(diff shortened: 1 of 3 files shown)


src/io/flutter/analytics/Analytics.java
```diff
@@ -105,6 +105,7 @@ public void sendException(Throwable throwable, boolean isFatal) {
     }
 
     String description = stringWriter.toString().trim();
+    description = description.replaceAll("com.intellij.openapi.", "c.i.o.");
```
stevemessick: Did you compare the net change of
com.intellij => c.i
vs
com.intellij.openapi => c.i.o
? 
I see a lot of com.intellij and wonder if a simpler transform might give greater savings.

devoncarew: The specific stack trace I saw had most things prefixed by `com.intellij.openapi.`, but I'm happy to check for the longer, than the shorter.

PR #1848, opened 2018-02-27
notify the user if the flutter daemon dies

devoncarew:
- unify some of our IntelliJ notifications under the name 'Flutter Messages' (visible in the notifications prefs)
- remove a log.error() when the flutter daemon terminates (instead, choose to notify the user)

This addresses an exception seen in analytics (instead of sending an exception to analytics, show a notification to the user).


devoncarew:
And for posterity:

screen shot 2018-02-27 at 2 43 19 pm

PR #1849, opened 2018-02-28
remove the experimental flag for the outline view

devoncarew:
- remove the experimental flag for the outline view

This is in anticipation of the view being released out from behind a flag next release.

@scheglov


resources/META-INF/plugin.xml
```diff
@@ -604,8 +604,7 @@ miscellaneous:
     <errorHandler implementation="io.flutter.FlutterErrorReportSubmitter"/>
 
     <toolWindow id="Flutter Outline" anchor="right" icon="FlutterIcons.Flutter_13"
-                factoryClass="io.flutter.preview.PreviewViewFactory"
-                conditionClass="io.flutter.preview.PreviewViewCondition"/>
```
scheglov: We used to check that the project is a Flutter project.
Did we stop doing this check?

devoncarew: Yes, we dropped it to eliminate cases where it should show up but didn't. IntelliJ only checks once at project open. People who have the plugin installed, but aren't doing flutter development at that moment, can just ignore this view.

PR #1857, opened 2018-02-28
Template files

devoncarew:
- regenerate the xml templates files
- rename templates files from foo.xml_template to foo_template.xml (to make it easier to edit the file types)
- put 'do not edit' headers at the top of the generated xml files
- regenerate the template files on the bots and fail the build if there were any diffs (a non-template files was editing, or a template was editing and `./bin/plugin generate` not re-run)

@stevemessick


devoncarew:
> I think file editing could have been improved in other ways, such as how flutter does it. But that's not a deal breaker.

Happy to have this swizzled around in the future; just - while putting this PR together - I found it useful to have the `.yml` and `.xml` extensions on the templates.

PR #1858, opened 2018-02-28
Fix Open Module in AS

### LINKED ISSUE #1820: open in android studio open foo_bar_android.iml in a text editor
When I select 'Open in Android Studio' on the android/ sub-project in IntelliJ, it opens the entire project (`foo_bar/`) in Android Studio.

screen shot 2018-02-23 at 3 48 08 pm

When I then select 'Open in Android Studio' on the android/ sub-project in that new Android Studio window, it opens the `foo_bar_android.iml` file in a text editor.

(I do see that the Java code is resolved in the Android studio instance - I can jump to declaration for the PluginRegistry classes)

/cc @stevemessick

stevemessick:
@devoncarew @pq 

I think Open in Android Studio has been broken for a while. Attempting to open the .iml file as a project does not work they way you'd expect. The parent directory gets opened, not the module defined by the .iml file.

Making this work from within Android Studio requires replacing the action object with one that does not try to spawn a new Android Studio.

Fixes #1820

### DIFF
--- a/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
+++ b/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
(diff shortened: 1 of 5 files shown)


pq:
Needless to say, cool to see this fixed! 👍

Issue #1847, opened 2018-02-27
Show property description and link to API docs in the inspector


InMatrix:
In a UX review, I found users might not be familiar with some widget properties and have trouble figuring out what property might be related to the problem they're trying to fix. It would be helpful to at least show the property description in the inspector and allow the user to view its API doc quickly.

A mock:

![image](https://user-images.githubusercontent.com/348942/36749747-8768fd7a-1bb0-11e8-9cab-29c05c67d95d.png)

PR #1859, opened 2018-02-28
Present names for user-defined properties in BOLD.

pq:
A baby-step along the way to doc hovers but useful in and of itself. 🎁 

image

See: #1847 

Depends on: https://github.com/flutter/flutter/pull/14993

@jacob314


src/io/flutter/inspector/DiagnosticsNode.java
```diff
@@ -362,7 +359,7 @@ boolean hasException() {
   }
 
   public boolean hasCreationLocation() {
-    return json.has("creationLocation");
+    return location != null || json.has("creationLocation");
```
jacob314: good catch :)

PR #1863, opened 2018-03-01
Fix Open Module in AS

stevemessick:
This replace a previous PR after fixing a merge problem.

@devoncarew @pq 

I think Open in Android Studio has been broken for a while. Attempting to open the .iml file as a project does not work they way you'd expect. The parent directory gets opened, not the modules defined by the .iml file.

Making this work from within Android Studio requires replacing the action object with one that does not try to spawn a new Android Studio.

### DIFF
--- a/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
+++ b/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter;
+
+import com.intellij.openapi.actionSystem.ActionManager;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.project.Project;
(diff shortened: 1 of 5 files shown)


flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
```diff
@@ -0,0 +1,32 @@
...
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.startup.StartupActivity;
+import io.flutter.actions.OpenAndroidModule;
+import org.jetbrains.annotations.NotNull;
+
+public class FlutterStudioStartupActivity implements StartupActivity {
+  @Override
+  public void runActivity(@NotNull Project project) {
+    replaceAction("flutter.androidstudio.open", new OpenAndroidModule());
```
devoncarew: Can you add some brief in-lines docs here, to say that when in Android Studio, we replace the 'open in android studio' action with one that opens a new window instead of spawn a new IDE process?

stevemessick: Will do!

PR #1871, opened 2018-03-01
Track the pub root directories on the device so that Flutter code knows which widgets were actually created on the device.

jacob314:
This CL doesn't do much on its own but will enable efficient filtering on the device of widgets based on their creation locations.

Depends on 
https://github.com/flutter/flutter/pull/15041

### DIFF
--- a/src/io/flutter/inspector/InspectorService.java
+++ b/src/io/flutter/inspector/InspectorService.java
@@ -6,9 +6,7 @@
 package io.flutter.inspector;
 
 import com.google.common.base.Joiner;
-import com.google.gson.JsonArray;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonParser;
+import com.google.gson.*;
 import com.intellij.openapi.Disposable;
 import com.intellij.openapi.application.ApplicationManager;
 import com.jetbrains.lang.dart.ide.runner.server.vmService.VmServiceConsumers;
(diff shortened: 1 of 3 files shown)


src/io/flutter/inspector/InspectorService.java
```diff
@@ -112,6 +110,26 @@ public void addClient(InspectorServiceClient client) {
...
+    // TODO(jacobr): remove call to hasServiceMethod("setPubRootDirectories") after
+    // the `setPubRootDirectories` method has been in two revs of the Flutter Alpha
+    // channel. The feature is expected to have landed in the Flutter dev
+    // chanel on March 2, 2018.
+
+    return hasServiceMethod("setPubRootDirectories").thenComposeAsync((Boolean hasMethod) -> {
+      if (!hasMethod) {
+        return null;
+      }
+      JsonArray jsonArray = new JsonArray(rootDirectories.size());
```
pq: It looks like this is breaking some build configs on travis:

```
     [javac2] /home/travis/build/flutter/flutter-intellij/src/io/flutter/inspector/InspectorService.java:123: error: constructor JsonArray in class JsonArray cannot be applied to given types;
     [javac2]       JsonArray jsonArray = new JsonArray(rootDirectories.size());
     [javac2]                             ^
     [javac2]   required: no arguments
```

I suppose we're getting different versions of the gson bits?

Maybe just use the default constructor here to work around?

src/io/flutter/inspector/InspectorService.java
```diff
@@ -120,7 +120,7 @@ public void addClient(InspectorServiceClient client) {
       if (!hasMethod) {
         return null;
       }
-      JsonArray jsonArray = new JsonArray(rootDirectories.size());
+      JsonArray jsonArray = new JsonArray();
```
pq: 👍

Hixie:
@jacob314 is this ready to land?

Issue #1205, opened 2017-07-12
FR: dartfmt on save support


DaveShuckerow:
I've been using Atom for Flutter development (even without the Flutter plugin) for several months.

My biggest reason for staying is that Atom will run dartfmt when I save.  I haven't had to think about formatting my code since I started using it, and that's a fantastic savings in engineering productivity.

Can we add support to run dartfmt on save to IntelliJ?

alexander-doroshko:
True IntelliJ users rarely perform Save action, but they do frequently use Reformat Code action (Ctrl+Alt+L or Cmd+Alt+L). So in IntelliJ we tied a call to the `dartfmt` tool to its standard and well-known formatting action. This is a new feature of version 2017.2 that has Release Candidate status at the moment and is expected to be released very soon. You may try it [right now]( https://www.jetbrains.com/idea/nextversion/).

mit-mit:
You could also do this with a Macro:

Edit > Macros > Start Macro Recording
Code > Reformat with Dart Style
File > Save All
Edit > Macros > Stop Macro Recording
Name the macro, and hit OK

Next, open Preferences, locate Keymap, search for the macro by the name you gave it, right-click, and assign cmd-s to it.

AskYous:
Thanks @mit-mit!

timokz:
Thanks a lot! Should have thought of that.

PR #1876, opened 2018-03-02
implement format on save

devoncarew:
- implement format on save
- also, implement organize imports on save
- fix https://github.com/flutter/flutter-intellij/issues/1205

screen shot 2018-03-02 at 3 12 12 pm

This might be a bit twitchy, so I added it in the experimental section.

@scheglov

### DIFF
--- a/src/io/flutter/editor/FlutterSaveActionsManager.java
+++ b/src/io/flutter/editor/FlutterSaveActionsManager.java
@@ -1,0 +1,195 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.editor;
+
+import com.intellij.AppTopics;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.command.WriteCommandAction;
+import com.intellij.openapi.components.ServiceManager;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.editor.Document;
+import com.intellij.openapi.fileEditor.FileDocumentManager;
+import com.intellij.openapi.fileEditor.FileDocumentManagerAdapter;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleManager;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.roots.ModuleRootManager;
+import com.intellij.openapi.util.text.StringUtil;
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.psi.PsiDocumentManager;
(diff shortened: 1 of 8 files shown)


devoncarew:
This does send several calls to the analysis server, which we may find ways to optimize in the future (organize imports, format, then - because the we modified the file contents - another round of organize imports and format).

src/io/flutter/editor/FlutterSaveActionsManager.java
```diff
@@ -0,0 +1,195 @@
...
+
+        if (reSave || didFormat) {
+          FileDocumentManager.getInstance().saveDocument(document);
+        }
+      }
+    }.execute());
+  }
+
+  private @Nullable
+  Module getModuleForPath(@NotNull VirtualFile file) {
```
scheglov: getModuleForFile?

src/io/flutter/editor/FlutterSaveActionsManager.java
```diff
@@ -0,0 +1,195 @@
...
+    if (PsiErrorElementUtil.hasErrors(myProject, psiFile.getVirtualFile())) {
+      return;
+    }
+
+    DartAnalysisServerService.getInstance(myProject).serverReadyForRequest(myProject);
+
+    if (settings.isOrganizeImportsOnSaveKey()) {
+      performOrganizeImportsOnSave(document, file);
+    }
+    else {
```
scheglov: So, we do either organize import, or format?
Maybe I don't understand IntelliJ UI format, but it says `JCheckBox`, so both can be checked?

scheglov: Ah, I see.
We do both when organize imports.

src/io/flutter/editor/FlutterSaveActionsManager.java
alexander-doroshko: Go To Symbol -> `findModuleForFile` will show you an existing helper method that works better :)

devoncarew: Updated! Thanks :)

src/io/flutter/editor/FlutterSaveActionsManager.java
devoncarew: I renamed the methods to make this clearer.

devoncarew:
From testing:
- we do correctly ignore dart files that are not part of the project
- it is twitchy with very fast edits and saves (less than 100ms between the initial modification and the cms-s); not sure why - I tried adding some logic to use the document modification stamps, but this didn't improve the race condition

filiph:
Where does one find this? I'm trying to search in IntelliJ setting but nothing finds this option.

devoncarew:
It's in the Flutter plugin settings (`Languages & Frameworks` > `flutter`).

PR #1882, opened 2018-03-05
Initial implementation of rendering in Flutter Outline.

scheglov:
@jacob314


src/io/flutter/preview/PreviewArea.java
```diff
@@ -0,0 +1,167 @@
...
+// TODO: we want to preview anything in a state, stateful, or stateless class (not
+//       just things contained in a build method)
+
+// TODO: we should be bolding stateful and stateless (and state) classes, not build() methods
+//       or, show all elements of these classes with some additional emphasis (italic? background color?)
+
+public class PreviewArea {
+  public static int BORDER_WITH = 5;
+
+  @SuppressWarnings("UseJBColor")
```
pq: Maybe pastels are OK but in general, how does this do with different themes?  We might consider going to `JBColor` if we want the colors to be theme aware...  Dunno.

src/io/flutter/preview/PreviewView.java
```diff
@@ -145,10 +156,41 @@ public void caretRemoved(CaretEvent e) {
 
   private final TreeSelectionListener treeSelectionListener = this::handleTreeSelectionEvent;
 
+  final RenderHelper.Listener renderListener = new RenderHelper.Listener() {
+    @Override
+    public void noWidget() {
+      ApplicationManager.getApplication().invokeLater(() -> {
+        if (previewArea != null) {
+          previewArea.clear();
```
pq: Man, makes me wish for `preview?.clear()` ;)

src/io/flutter/preview/RenderHelper.java
```diff
@@ -0,0 +1,282 @@
...
+                RenderHelper.Listener listener) {
+    this.testerPath = testerPath;
+    this.packages = packages;
+    this.file = file;
+    this.codeToRender = codeToRender;
+    this.widgetClass = widgetClass;
+    this.widgetConstructor = widgetConstructor;
+    this.listener = listener;
+  }
+}
```
pq: Add a trailing newline?

src/io/flutter/preview/render_server_template.txt
```diff
@@ -0,0 +1,192 @@
...
+    io.exit(0);
+  } catch (e, st) {
+    var json = {'exception': '$e', 'stackTrace': '$st'};
+    io.stdout.writeln(JSON.encode(json));
+    io.exit(1);
+  }
+}
+
+Iterable<Element> collectAllElementsFrom(Element rootElement) {
+  return new CachingIterable<Element>(
```
pq: => ... ?

src/io/flutter/preview/render_server_template.txt
```diff
@@ -0,0 +1,192 @@
...
+    json['$id'] = {
+      'globalBounds': {
+        'left': globalBounds.left.truncate(),
+        'top': globalBounds.top.truncate(),
+        'width': globalBounds.width.truncate(),
+        'height': globalBounds.height.truncate(),
+      }
+    };
+
+    for (int i = 0; i < children.length; i++) {
```
jacob314: Nit:
for (var child in children) {
  child.appendToJson(json);
}
of 
children.forEach
is a bit more idiomatic dart.

src/io/flutter/preview/render_server_template.txt
```diff
@@ -0,0 +1,192 @@
...
+      children[i].appendToJson(json);
+    }
+  }
+
+  static ElementInfo buildHierarchy(
+      Map<Widget, Element> elementsMap, Map<int, Widget> registeredWidgets) {
```
jacob314: storing a map fro Widget to Element will cause some challenges for edge cases with const Widget classes and other cases where the same widget may be instantiated multiple places in the Element tree.

src/io/flutter/preview/PreviewArea.java
scheglov: I don't know any existing code that uses large colorer panels, so there is nothing to copy colors from. I will leave it as is for now, maybe someone will want to experiment with colors for `Darcula`. Or we might get actual rendering (which might be also not dark).

src/io/flutter/preview/RenderHelper.java
scheglov: Done.

src/io/flutter/preview/render_server_template.txt
scheglov: Done.

src/io/flutter/preview/render_server_template.txt
scheglov: ACK
I needed index in one of the previous incarnations of this code.
But not anymore.

src/io/flutter/preview/render_server_template.txt
scheglov: Yes, that's correct.
We need a way to fix this.

For now we could replace `const X()` with `new X()` while rewriting code to render. This will not work for widgets created as constants externally, outside the instrumented library. Probably rare?

src/io/flutter/preview/render_server_template.txt
```diff
@@ -0,0 +1,192 @@
...
+      var id = elementToId[element];
+      if (id == null) {
+        return null;
+      }
+
+      var info = elementToInfoMap[element];
+      if (info == null) {
+        Element ancestor = getInterestingAncestor(element);
+        ElementInfo ancestorInfo = createInfoForElement(ancestor);
+        ui.Rect globalBounds = getGlobalBounds(element);
```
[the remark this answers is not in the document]
scheglov: I will add TODO and read/experiment later.

src/io/flutter/preview/render_server_template.txt
```diff
@@ -0,0 +1,192 @@
...
+  Widget build(BuildContext context) {
+    return new MaterialApp(
+      title: 'Flutter Demo',
+      theme: new ThemeData(
+        primarySwatch: Colors.blue,
+      ),
+      home: new Scaffold(
+        body: new Center(
+          child: new SizedBox(
+            width: 350.0,
```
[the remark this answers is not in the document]
scheglov: Yeah, size is an interesting question.

This makes sense it we design a top-level UI.

I will update this eventually to use the size of the preview panel in `Flutter Outline`. Most probably it will be even further from a typical phone, but we will use space more efficiently. Probably OK if we design not a top-level widget.

Ideally we need the whole editor tab space, with ability to resize widget as we want to. Maybe, if we will grow to a full WYSIWYG designer.

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -373,6 +374,24 @@ public String getDartSdkPath() {
...
+    final String platformString;
+    if (SystemInfo.isMac) {
+      platformString = "darwin-x64";
+    } else if (SystemInfo.isLinux) {
+      platformString = "linux-x64";
+    } else {
+      return null;
+    }
+
+    final String relativePath = "bin/cache/artifacts/engine/" + platformString + "/flutter_tester";
```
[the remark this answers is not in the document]
scheglov: Hm... If `flutter_tester` is basically a Dart VM, it should be possible just to give it a Dart file to execute, and it will care about finding any platform dills itself, shouldn't it?

Or we need a special way of running by first compiling code outside of VM into kernel, because we want to massage this kernel?

PR #1883, opened 2018-03-06
Use the bottom area size as the size for the widget to render.

scheglov:
image

### DIFF
--- a/src/io/flutter/preview/RenderHelper.java
+++ b/src/io/flutter/preview/RenderHelper.java
@@ -38,6 +38,9 @@ public class RenderHelper {
 
   private FlutterOutline myWidgetOutline;
 
+  private int myWidth = 0;
+  private int myHeight = 0;
+
   public RenderHelper(@NotNull Project project, @NotNull Listener listener) {
     myProject = project;
     myListener = listener;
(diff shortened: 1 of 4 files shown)


src/io/flutter/preview/RenderHelper.java
```diff
@@ -146,6 +163,8 @@ private static void render(@NotNull RenderRequest request) {
       template = template.replace("// TODO: import library to render", "import '" + toRenderFile.toURI() + "';");
       template = template.replace("new Container(); // TODO: create widget", widgetCreation);
       template = template.replace("{}; // TODO: use flutterDesignerWidgets", "flutterDesignerWidgets;");
+      template = template.replace("350.0 /*TODO: width*/", request.width + ".0");
```
jacob314: I would recommend having the template have an entry something like
__WINDOW_WIDTH_UID1239394032__
instead of matching on something that doesn't look like a template hole as far as a developer unfamiliar with the code is concerned.
so that filling out the holes in the template is more robust and less surprising.

src/io/flutter/preview/render_server_template.txt
```diff
@@ -181,8 +181,8 @@ class _DesignerApp extends StatelessWidget {
       home: new Scaffold(
         body: new Center(
           child: new SizedBox(
-            width: 350.0,
-            height: 400.0,
+            width: 350.0 /*TODO: width*/,
```
jacob314: this code would become
width: __WINDOW_WIDTH__,
or something along those lines with my suggested change.
/*TODO: width*/ makes it look like this a TODO suggesting that we should pass in the width not
the way you are actually passing in the width.

scheglov:
I documented how we use `TODO` comments.

devoncarew:
Very cool!

For some future work, we may want to consider a small toolbar over the bottom area (the live preview area), with a few different aspect ratio buttons. Something like a full width button (what this PR implements), and one that has a phone aspect ratio.

jacob314:
It would be nice if a string other than TODO was used. Todo implies something very different.
something like TEMPLATE_HOLE or TEMPLATE_VALUE would make it much clearer to me as a reader of the code that this wasn't a TODO.

PR #1890, opened 2018-03-07
Property description tooltips (#1847)

pq:
Fixes: #1847

mar-07-2018 10-32-27

TODO:
* consider hovers for values
* consider rich navigation hovers (w/ styling and navigable docs) 

@jacob314

### DIFF
--- a/src/io/flutter/inspector/DiagnosticsNode.java
+++ b/src/io/flutter/inspector/DiagnosticsNode.java
@@ -10,8 +10,16 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonNull;
 import com.google.gson.JsonObject;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.project.ProjectUtil;
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.xdebugger.XSourcePosition;
+import com.jetbrains.lang.dart.analyzer.DartAnalysisServerService;
+import com.jetbrains.lang.dart.ide.runner.server.vmService.frame.DartVmServiceValue;
+import io.flutter.run.FlutterDebugProcess;
 import io.flutter.utils.CustomIconMaker;
 import io.flutter.utils.JsonUtils;
+import org.dartlang.analysis.server.protocol.HoverInformation;
 import org.dartlang.vm.service.element.InstanceRef;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -49,6 +57,8 @@ public class DiagnosticsNode {
 
   private Location location;
   private DiagnosticsNode parent;
+
+  private CompletableFuture<String> propertyDocFuture;
 
   public DiagnosticsNode(JsonObject json, InspectorService inspectorService) {
     this.inspectorService = inspectorService;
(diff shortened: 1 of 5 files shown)


src/io/flutter/inspector/DiagnosticsNode.java
```diff
@@ -612,29 +622,73 @@ public InspectorInstanceRef getDartDiagnosticRef() {
...
         final ArrayList<Location> parameterLocations = location.getParameterLocations();
         if (parameterLocations != null) {
-            final Map<String, Location> names = new HashMap<>();
-            for (Location pl : parameterLocations) {
-              final String name = pl.getName();
-              if (name != null) {
-                names.put(name, pl);
-              }
+          final Map<String, Location> names = new HashMap<>();
+          for (Location pl : parameterLocations) {
```
jacob314: nit: pl --> location

src/io/flutter/inspector/DiagnosticsNode.java
```diff
@@ -612,29 +622,73 @@ public InspectorInstanceRef getDartDiagnosticRef() {
...
+    if (propertyDocFuture == null) {
+      propertyDocFuture = createPropertyDocFurure();
+    }
+    return propertyDocFuture;
+  }
+
+  private CompletableFuture<String> createPropertyDocFurure() {
+    final DiagnosticsNode parent = getParent();
+    if (parent != null) {
+      final InspectorService inspectorService = parent.getInspectorService();
```
jacob314: nit: just use the inspectorService field of this Diagnostic directly instead of calling
parent.getInspectorService()

src/io/flutter/inspector/DiagnosticsNode.java
```diff
@@ -612,29 +622,73 @@ public InspectorInstanceRef getDartDiagnosticRef() {
...
+      final CompletableFuture<DartVmServiceValue> valueFuture =
```
jacob314: any reason valueFuture is separate instead of just in the chain of thenCompaseAsync calls?

src/io/flutter/inspector/DiagnosticsNode.java
```diff
@@ -612,29 +622,73 @@ public InspectorInstanceRef getDartDiagnosticRef() {
...
+        });
+      });
+    }
+
+    return CompletableFuture.completedFuture("Unable to find property source");
+  }
+
+  @Nullable
+  private Project getProject(@NotNull VirtualFile file) {
+    final FlutterDebugProcess process = inspectorService.getDebugProcess();
```
jacob314: nit: the debugProcess seems like a surprising place to be getting the app from. Perhaps have inspectorService expose
getApp() directly?

src/io/flutter/inspector/InspectorService.java
```diff
@@ -60,6 +61,30 @@ public void connectionClosed() {
...
+  }
+
+  public CompletableFuture<XSourcePosition> getPropertyLocation(InstanceRef instanceRef, String name) {
+    return getInstance(instanceRef).thenComposeAsync((Instance instance) -> getPropertyLocationHelper(instance.getClassRef(), name));
+  }
+
+  public CompletableFuture<XSourcePosition> getPropertyLocationHelper(ClassRef classRef, String name) {
+    return inspectorLibrary.getClass(classRef).thenComposeAsync((ClassObj clazz) -> {
+      for (FuncRef f : clazz.getFunctions()) {
+        if (f.getName().equals(name)) {
```
jacob314: Add a TODO to handle checking for private properties that match the name.
Eg. if name == 'textAlignment' 
we should check if the class has a 
_textAlignment property.

src/io/flutter/run/FlutterDebugProcess.java
```diff
@@ -53,6 +53,11 @@ public FlutterDebugProcess(@NotNull FlutterApp app,
     this.app = app;
   }
 
+  @NotNull
```
jacob314: btw when we create the InspectorService we probably have an app available. We could just flow the dependency in that way.

src/io/flutter/view/InspectorPanel.java
```diff
@@ -495,6 +496,7 @@ void registerShortcuts() {
       DebuggerUIUtil.registerActionOnComponent(InspectorActions.JUMP_TO_TYPE_SOURCE, this, this);
     }
 
+    @SuppressWarnings("EmptyMethod")
```
jacob314: noob question. this method doesn't seem empty?

src/io/flutter/view/InspectorPanel.java
```diff
@@ -632,9 +636,8 @@ private void showPropertiesHelper(ArrayList<DiagnosticsNode> properties) {
         getEmptyText().setText(FlutterBundle.message("app.inspector.no_properties"));
         return;
       }
-      CompletableFuture<Void> loaded = loadPropertyMetadata(properties);
-
-      whenCompleteUiThread(loaded, (Void ignored, Throwable errorGettingInstances) -> {
+      final CompletableFuture<Void> loaded = loadPropertyMetadata(properties);
+      whenCompleteUiThread(loadPropertyMetadata(properties), (Void ignored, Throwable errorGettingInstances) -> {
```
jacob314: loadPropertymetadata is now getting called twice.
I assume that is a bug?

src/io/flutter/view/InspectorPanel.java
```diff
@@ -748,8 +751,23 @@ protected void customizeCellRenderer(JTable table, @Nullable Object value, boole
...
+      //  * consider rich navigation hovers (w/ styling and navigable docs)
+      final CompletableFuture<String> propertyDoc = node.getPropertyDoc();
+      final String doc = propertyDoc.getNow(null);
+      if (doc != null) {
+        setToolTipText(doc);
+      }
+      else {
+        AsyncUtils.whenCompleteUiThread(propertyDoc, (String tooltip, Throwable th) -> {
+          if (th != null) {
+            LOG.warn(th);
```
jacob314: LOG.error not warn?

src/io/flutter/view/InspectorPanel.java
```diff
@@ -748,8 +751,23 @@ protected void customizeCellRenderer(JTable table, @Nullable Object value, boole
...
+      // TODO (pq):
+      //  * consider tooltips for values
+      //  * consider rich navigation hovers (w/ styling and navigable docs)
+      final CompletableFuture<String> propertyDoc = node.getPropertyDoc();
+      final String doc = propertyDoc.getNow(null);
+      if (doc != null) {
+        setToolTipText(doc);
+      }
+      else {
+        AsyncUtils.whenCompleteUiThread(propertyDoc, (String tooltip, Throwable th) -> {
```
jacob314: I think you want to call 
setToolTipText(null) 
or setToolTipText("")
right before awaiting the future or you could show stale tooltips for cases where the inspectorService is overloaded with requests

src/io/flutter/inspector/DiagnosticsNode.java
pq: 👍

src/io/flutter/inspector/DiagnosticsNode.java
pq: 👍

src/io/flutter/inspector/DiagnosticsNode.java
pq: My temporary sanity while debugging? 😉 

Updated.

src/io/flutter/inspector/DiagnosticsNode.java
pq: Sounds good.  Updated.

src/io/flutter/inspector/InspectorService.java
pq: Good catch!  Done.

src/io/flutter/run/FlutterDebugProcess.java
pq: Good point.  I need to do some reconsidering when we do more source navigation stuff.  Will revisit.

src/io/flutter/view/InspectorPanel.java
pq: Right.  IDEA flags it though.  This gets the inspection to shut up.

src/io/flutter/view/InspectorPanel.java
pq: Whoops.  Good catch.  Fixed.

src/io/flutter/view/InspectorPanel.java
pq: 👍

src/io/flutter/view/InspectorPanel.java
pq: Good idea!

jacob314:
lgtm

Issue #1889, opened 2018-03-07
Make "Import" work


stevemessick:
The "Import Project/Module" function does not work with Flutter projects. This needs to be fixed for both IntelliJ and Android Studio.

devoncarew:
+1

It looks like the android module portion is recognized first, and hinders the import of the dart portion of the project. Perhaps we need to add some type of Flutter module recognizer?

devoncarew:
Possibly we need to implement a `com.intellij.projectImport.ProjectImportProvider`?

pq:
After some exploring with @stevemessick, I think I've got a good bead on this.  Optimistically scheduling.

matejthetree:
please check this out too 
https://github.com/apptreesoftware/letsvote#intellj-multi-module-setup

pq:
Oh wow.  That's complicated.  Thanks for the heads-up @matejthetree!

devoncarew:
From some user feedback, people are having trouble importing Flutter projects into Android Studio; making this easier would be great.

pq:
Quick update: I made some good progress on this today but not close enough to try and land for M23.  Deferring but treating as high priority for M24.

PR #1892, opened 2018-03-07
Intercept and redirect Project import actions (#1889).

pq:
Fixes: #1889.

In the case where you attempt to import a directory that defines a Flutter root, it works just like a project open (which is to say "it just works").  Otherwise, you're injected into the standard import flow where you're greeted with a wizard like this:

image

Incidentally, this wizard highlights *why* we want to do something like this (and why taking on a little technical risk may well be worth it).  A common issue (encountered for example just this week by @Sfshaza) is that folks don't know what to do when confronted with this wizard.  Not only is that confusing, if they do the wrong thing, they are left with a project that won't work.

(FWIW still a WIP and more invasive than I'd hoped...  For now, just food for thought.)

/cc @devoncarew @stevemessick

### DIFF
--- a/src/io/flutter/FlutterInitialConfigurator.java
+++ b/src/io/flutter/FlutterInitialConfigurator.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter;
+
+import com.intellij.icons.AllIcons;
+import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.openapi.actionSystem.ActionManager;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.fileTypes.FileTypeManager;
+import com.intellij.util.messages.MessageBus;
+import io.flutter.actions.FlutterImportProjectAction;
+
+public class FlutterInitialConfigurator {
+  @SuppressWarnings("UnusedParameters")
+  public FlutterInitialConfigurator(final MessageBus bus,
+                                    final PropertiesComponent propertiesComponent,
+                                    final FileTypeManager fileTypeManager) {
+    // Intercept and redirect Import actions.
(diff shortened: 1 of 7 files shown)


stevemessick:
What would be surprising is to find something is less invasive than expected!

src/io/flutter/project/FlutterProjectOpenProcessor.java
```diff
@@ -45,9 +46,7 @@ public Icon getIcon() {
 
   @Override
   public boolean canOpenProject(@Nullable VirtualFile file) {
-    if (file == null) return false;
-    final PubRoot root = PubRoot.forDirectory(file);
-    return root != null && root.declaresFlutter();
+    return FlutterUtils.isFlutterProjectRoot(file);
```
stevemessick: Good catch :)

stevemessick:
We need to extend the plugin test script to exercise the non-Flutter import case. We should have a new step to import a project that is not a Flutter project and ensure the UX does not change.

src/io/flutter/FlutterInitialConfigurator.java
```diff
@@ -0,0 +1,37 @@
...
+import com.intellij.util.messages.MessageBus;
+import io.flutter.actions.FlutterImportProjectAction;
+
+public class FlutterInitialConfigurator {
+  @SuppressWarnings("UnusedParameters")
+  public FlutterInitialConfigurator(final MessageBus bus,
+                                    final PropertiesComponent propertiesComponent,
+                                    final FileTypeManager fileTypeManager) {
+    // Intercept and redirect Import actions.
+    if (!FlutterUtils.isAndroidStudio()) {
```
stevemessick: TODO(messick) Ensure something is done in AS to handle project import properly.
Either remove the guard clause (that I asked @pq to add) if there is no difference in the AS initialization code, or add equivalent logic to the Flutter initializer.

stevemessick:
Does this work correctly if a second project window is opened when a project window is already open?

devoncarew:
Just as a note, we could explore the `ProjectImportProvider` extension point.

PR #1903, opened 2018-03-09
Select the widget under the caret after rendering is done

scheglov:
Also performs lock/unlock when rendering starts/ends, so that changing text does not blink selection.

### DIFF
--- a/src/io/flutter/preview/PreviewArea.java
+++ b/src/io/flutter/preview/PreviewArea.java
@@ -107,7 +107,7 @@ public class PreviewArea {
   public void clear() {
     primaryLayer.removeAll();
     primaryLayer.setLayout(new BorderLayout());
-    primaryLayer.add(new JLabel("Nothing to show", SwingConstants.CENTER), BorderLayout.CENTER);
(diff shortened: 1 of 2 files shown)


scheglov:
ping?

src/io/flutter/preview/PreviewArea.java
```diff
@@ -62,44 +66,75 @@
...
   }
 
   public void clear() {
-    panel.removeAll();
-    panel.setLayout(new BorderLayout());
-    panel.add(new JLabel("Nothing to show", SwingConstants.CENTER), BorderLayout.CENTER);
-    panel.repaint();
+    primaryLayer.removeAll();
+    primaryLayer.setLayout(new BorderLayout());
+    primaryLayer.add(new JLabel("Nothing to show", SwingConstants.CENTER), BorderLayout.CENTER);
```
devoncarew: Perhaps have text similar to:

screen shot 2018-03-12 at 10 54 32 am

Although, longer term, we'd want to iterate to help the user understand why there was no preview, and the steps they could take to make one available.

pq: +1

PR #1908, opened 2018-03-12
Index less

stevemessick:
@pq @devoncarew 

For Flutter projects created with Android Studio, project creation should be quiet:
- Android framework is not detected, so there is no notification to configure it
- Constant 2-sec refresh is disabled
- "Migrate to Gradle" is not suggested

Opening a Flutter project created in IntelliJ:
- Constant 2-sec refresh is disabled
- "Migrate to Gradle" is suggested only first time the first time a project is opened -- `FlutterUtils.disableGradleProjectMigrationNotification()` is not run until after the notification is displayed, but does work for subsequent openings

That last item needs further work. The user won't see the prompt twice, even if no action is taken, but they should not even see it once. I think it is a problem that needs to be solved in IntelliJ.

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -23,7 +23,9 @@
 
   <!-- Contributes Android Studio-specific features and implementations. -->
   <!--suppress PluginXmlValidity -->
-  <depends optional="true" config-file="studio-contribs.xml">com.intellij.modules.androidstudio</depends>
+  <!-- depends optional="true" config-file="studio-contribs.xml">com.android.tools.apk</depends -->
+  <!-- note that com.android.tools.apk also implies com.intellij.modules.androidstudio but it is not available in sources -->
+  <depends optional="true" config-file="studio-contribs.xml">com.intellij.modules.java</depends>> <!-- androidstudio</depends -->
 
   <change-notes>
     <![CDATA[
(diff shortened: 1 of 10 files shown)


resources/META-INF/plugin.xml
```diff
@@ -615,7 +615,7 @@ miscellaneous:
     <projectService serviceInterface="io.flutter.view.FlutterView"
                     serviceImplementation="io.flutter.view.FlutterView"
                     overrides="false"/>
-    <projectOpenProcessor implementation="io.flutter.project.FlutterProjectOpenProcessor" order="first"/>
+    <projectOpenProcessor id="flutter" implementation="io.flutter.project.FlutterProjectOpenProcessor" order="first"/>
```
devoncarew: Is this id unique enough? I'm not sure what other ids in this space would look like.

resources/META-INF/studio-contribs.xml
```diff
@@ -15,6 +15,7 @@
   <extensions defaultExtensionNs="com.intellij">
     <androidStudioInitializer implementation="io.flutter.FlutterStudioInitializer"/>
     <postStartupActivity implementation="io.flutter.FlutterStudioStartupActivity"/>
+    <projectOpenProcessor implementation="io.flutter.project.FlutterStudioProjectOpenProcessor" order="after flutter"/>
```
devoncarew: It's ok for these project open processors to not have ids?

resources/META-INF/plugin.xml
stevemessick: None of the others have id's, so I think this is good. It is a very small name-space

resources/META-INF/studio-contribs.xml
stevemessick: Hardly any extensions have id's. I should probably add one, just in case.

PR #1909, opened 2018-03-12
Use hot reload for rendering when possible.

scheglov:
When we can, we render about `10x` faster.

Without hot reload - about `1000ms`.

With hot reload - about `60-80ms` for reload and `20-60ms` for rendering, depending on complexity of the widget. Edits, switching between widgets in the same file, and even in the same package should be hot realoadable.

### DIFF
--- a/src/io/flutter/preview/RenderHelper.java
+++ b/src/io/flutter/preview/RenderHelper.java
@@ -5,7 +5,6 @@
  */
 package io.flutter.preview;
 
-import com.google.common.base.Stopwatch;
 import com.google.common.io.Files;
 import com.google.common.io.Resources;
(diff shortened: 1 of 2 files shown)


src/io/flutter/preview/RenderHelper.java
```diff
@@ -153,7 +161,100 @@ private void scheduleRendering() {
...
+  final int width;
+  final int height;
+
+  final RenderHelper.Listener listener;
+
+  RenderRequest(String testerPath, VirtualFile packages, VirtualFile file,
+                String codeToRender,
+                String widgetClass,
+                String widgetConstructor,
+                int width, int height,
```
devoncarew: nit: `height` on its own line

PR #1917, opened 2018-03-15
experiment with different colors for the preview

devoncarew:
- experiment with different colors for the preview
- use the classname as a label in the widget preview
- rotate colors for different depths in the tree; sibling nodes will share the same color
- add a border to help separate adjacent widgets

Happy to iterate on this, or remove portions of the PR.

screen shot 2018-03-15 at 1 55 31 am

screen shot 2018-03-15 at 1 55 12 am

### DIFF
--- a/src/io/flutter/preview/PreviewArea.java
+++ b/src/io/flutter/preview/PreviewArea.java
@@ -39,13 +39,13 @@ import java.util.Map;
 public class PreviewArea {
   public static int BORDER_WITH = 5;
 
-  // https://material.io/guidelines/style/color.html#color-color-tool
   @SuppressWarnings("UseJBColor")
-  private static final Color[] widgetColors = new Color[]{
-    Color.decode("#0D47A1"),
-    Color.decode("#1976D2"),
-    Color.decode("#2196F3"),
-    Color.decode("#64B5F6")
+  private static final Color[] pastelColors = new Color[]{
+    new Color(255, 132, 203),
+    new Color(168, 231, 234),
+    new Color(251, 255, 147),
+    new Color(143, 255, 159),
+    new Color(193, 149, 255),
   };
 
   private final Listener myListener;
@@ -224,11 +224,9 @@ public class PreviewArea {
         final JLabel label = new JLabel(outline.getClassName());
         label.setBorder(JBUI.Borders.empty(2, 4, 0, 0));
         widget.add(label, BorderLayout.NORTH);
-        widget.setBackground(widgetColors[widgetDepth % widgetColors.length]);
+        widget.setBackground(pastelColors[widgetDepth % pastelColors.length]);
         widget.setBounds(rect);
-        //widget.setBorder(new EtchedBorder());
         widget.setBorder(IdeBorderFactory.createRoundedBorder(2));
-        //widget.validate();
 
         outlineToComponent.put(outline, widget);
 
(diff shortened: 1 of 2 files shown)


src/io/flutter/preview/RenderHelper.java
```diff
@@ -153,6 +153,10 @@ private void scheduleRendering() {
       return;
     }
 
+    if (myTesterPath == null) {
+      return;
```
scheglov: We probably need TODO here as a reminder to expose this error to the user.

src/io/flutter/preview/PreviewArea.java
```diff
@@ -218,9 +220,15 @@ private void renderWidgetOutline(@NotNull FlutterOutline outline) {
...
 
-        final JPanel widget = new JPanel();
-        widget.setBackground(pastelColors[(colorIndex++) % pastelColors.length]);
+        final JPanel widget = new JPanel(new BorderLayout());
+        final JLabel label = new JLabel(outline.getClassName());
+        label.setBorder(JBUI.Borders.empty(2, 4, 0, 0));
+        widget.add(label, BorderLayout.NORTH);
+        widget.setBackground(widgetColors[widgetDepth % widgetColors.length]);
         widget.setBounds(rect);
+        //widget.setBorder(new EtchedBorder());
```
scheglov: Remove?

src/io/flutter/preview/PreviewArea.java
```diff
@@ -218,9 +220,15 @@ private void renderWidgetOutline(@NotNull FlutterOutline outline) {
...
+        widget.setBorder(IdeBorderFactory.createRoundedBorder(2));
+        //widget.validate();
```
scheglov: Remove?

src/io/flutter/preview/PreviewArea.java
```diff
@@ -218,9 +220,15 @@ private void renderWidgetOutline(@NotNull FlutterOutline outline) {
         final int y = BORDER_WITH + rect.y - rootWidgetBounds.y;
         rect = new Rectangle(x, y, rect.width, rect.height);
 
-        final JPanel widget = new JPanel();
-        widget.setBackground(pastelColors[(colorIndex++) % pastelColors.length]);
+        final JPanel widget = new JPanel(new BorderLayout());
+        final JLabel label = new JLabel(outline.getClassName());
+        label.setBorder(JBUI.Borders.empty(2, 4, 0, 0));
```
scheglov: I'm not an expert in Swing, do you know by a chance - will border expand inside the widget or outside of it? Well, maybe we actually don't care about pixel-perfect quality at all, if we don't get actual screenshots.

src/io/flutter/preview/PreviewArea.java
devoncarew: will do

src/io/flutter/preview/PreviewArea.java
devoncarew: I'll check - my assumption was that this would not expand the containing widget's size.

devoncarew:
I'll keep the text and the border, and rotating colors through based on z-index, and revert the change to the actual colors.

devoncarew:
updated, ptal:

screen shot 2018-03-15 at 10 13 07 am

screen shot 2018-03-15 at 10 13 13 am

devoncarew:
(the Text padding does not impact the swing widgets size, at least for the layouts we're using)

devoncarew:
My idea for the in-line text labels was a slightly smaller font than the IntelliJ default, using something like upper title case (all caps, with the first letters of words slightly larger), and aligned in the top-left.

PR #1921, opened 2018-03-15
Heap use display initial implementation.

pq:
A work in progress but ready for iteration.

Currently enabled with a compile time flag (`PerfService. DISPLAY_HEAP_USE`) but will be promoted to a preference or menu enablement down the road.  Lots of TODOs captured in the source.

Disabled state:

image

Enabled:

image

(This is the sample app and so not a lot of heap use; nonetheless, scaling needs work and that dip is odd.)

Lots of next steps documented in the source but a few immediate ones include:

* tweaking scaling
* refactoring value caching
* adding a label for current/total heap
* handling GCs
* consolidating observatory actions into the panel (maybe into one combined over-flow style action)

@devoncarew

### DIFF
--- a/src/io/flutter/inspector/HeapDisplay.java
+++ b/src/io/flutter/inspector/HeapDisplay.java
@@ -118,7 +118,7 @@ public class HeapDisplay extends JPanel {
   }
 
   public static class HeapState implements HeapListener {
-    // Running count of the max help (in bytes).
+    // Running count of the max heap (in bytes).
     private int heapMaxInBytes;
 
     private final HeapSamples samples;
--- a/src/io/flutter/inspector/HeapDisplay.java
+++ b/src/io/flutter/inspector/HeapDisplay.java
@@ -1,0 +1,234 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.inspector;
+
+import com.intellij.openapi.Disposable;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.actionSystem.Presentation;
+import com.intellij.openapi.actionSystem.ex.CustomComponentAction;
+import com.intellij.openapi.util.Disposer;
+import com.intellij.ui.JBColor;
+import com.intellij.util.ui.JBUI;
+import io.flutter.perf.HeapMonitor.HeapListener;
+import io.flutter.perf.HeapMonitor.HeapSample;
+import io.flutter.perf.HeapMonitor.HeapSpace;
+import io.flutter.perf.HeapMonitor.IsolateObject;
+import io.flutter.perf.PerfService;
+import io.flutter.run.daemon.FlutterApp;
+import org.jetbrains.annotations.NotNull;
+
+import javax.swing.*;
+import java.awt.*;
+import java.util.*;
+import java.util.List;
+
+// TODO(pq): make capacity setting dynamic
+// TODO(pq): add label displaying curent/total heap use
+// TODO(pq): handle GCs
+// TODO(pq): tweak scaling
+// TODO(pq): fix duplicate value caching (remove int cache from graph)
+public class HeapDisplay extends JPanel {
+
+  public static class ToolbarComponentAction extends AnAction implements CustomComponentAction, HeapListener, Disposable {
+
+    private final List<JPanel> panels = new ArrayList<>();
+    private final List<HeapDisplay> graphs = new ArrayList<>();
+
+    // TODO(pq): be smart about setting capacity based on available width.
+    final HeapState heapState = new HeapState(30);
+
+    public ToolbarComponentAction(@NotNull Disposable parent, @NotNull FlutterApp app) {
+      final PerfService service = app.getPerfService();
+      assert service != null;
+      service.addListener(this);
+      service.start();
+      Disposer.register(parent, this);
+    }
+
+    @Override
+    public void update(AnActionEvent e) {
+      // No-op.
+    }
+
+    @Override
+    public void actionPerformed(AnActionEvent e) {
+      // No-op.
+    }
+
+    @Override
+    public JComponent createCustomComponent(Presentation presentation) {
+      final JPanel panel = new JPanel(new GridBagLayout());
+      final HeapDisplay graph = new HeapDisplay();
+
+      panel.add(graph,
+                new GridBagConstraints(0, 0, 1, 1, 1, 1,
+                                       GridBagConstraints.CENTER, GridBagConstraints.BOTH,
+                                       JBUI.insets(0, 3, 0, 3), 0, 0));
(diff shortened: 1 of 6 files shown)


devoncarew:
It would be nice if we had a slightly bugger space to display this. It would give us a bit more detail in the sparkline, and would let us provide a numerical value (`68MB of 90MB`) to provide a bit more info to the user. The inspector UI is under pretty active development; perhaps some likely spot will open up.

src/io/flutter/inspector/HeapDisplay.java
```diff
@@ -0,0 +1,234 @@
...
+  public static class ToolbarComponentAction extends AnAction implements CustomComponentAction, HeapListener, Disposable {
+
+    private final List<JPanel> panels = new ArrayList<>();
+    private final List<HeapDisplay> graphs = new ArrayList<>();
+
+    // TODO(pq): be smart about setting capacity based on available width.
+    final HeapState heapState = new HeapState(30);
+
+    public ToolbarComponentAction(@NotNull Disposable parent, @NotNull FlutterApp app) {
+      final PerfService service = app.getPerfService();
```
devoncarew: perhaps make getPerfService() nonnull?

src/io/flutter/inspector/HeapDisplay.java
```diff
@@ -0,0 +1,234 @@
...
+      return samples.size();
+    }
+
+    HeapSample get(int index) {
+      return samples.get(index);
+    }
+  }
+
+  public static class HeapState implements HeapListener {
+    // Running count of the max help (in bytes).
```
devoncarew: `help` ==> `heap`

src/io/flutter/inspector/HeapDisplay.java
```diff
@@ -0,0 +1,234 @@
...
+
+    int get(int index) {
+      return list.get(index);
+    }
+  }
+
+  private static final int PREFFERED_WIDTH = 60;
+  private static final int PREFFERED_HEIGHT = 16;
+
+  private static final Color GRAPH_COLOR = JBColor.LIGHT_GRAY;
```
devoncarew: We'll want to try and take this from the current theme.

I see a `UIUtil.isUnderDarcula()` util method, and some general color manager classes; not sure which is the best option for getting a good foreground color.

src/io/flutter/perf/HeapMonitor.java
```diff
@@ -0,0 +1,183 @@
...
+import org.dartlang.vm.service.element.RPCError;
+import org.dartlang.vm.service.element.Sentinel;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.*;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+// TODO(pq): handle GC events
```
devoncarew: We should move to a time based graph (time as the x axis instead of sample count). That'll let us do things like incorporate GC event info, without distorting the graph.

src/io/flutter/perf/PerfService.java
```diff
@@ -0,0 +1,93 @@
...
+      isRunning = true;
+    }
+  }
+
+  /**
+   * Stop the Perf service.
+   */
+  public void stop() {
+    if (isRunning) {
+      // The vmService does not have a way to remove listeners, so we can only stop paying attention.
```
devoncarew: Hmm, that seems like an oversight :( (re: removing listeners)

devoncarew: https://github.com/dart-lang/vm_service_drivers/pull/176

src/io/flutter/view/FlutterView.java
```diff
@@ -110,17 +114,22 @@ public void actionPerformed(AnActionEvent event) {
...
     toolbarGroup.add(registerAction(new DebugDrawAction(app)));
     toolbarGroup.add(registerAction(new TogglePlatformAction(app)));
     toolbarGroup.add(registerAction(new PerformanceOverlayAction(app)));
     toolbarGroup.addSeparator();
+    // TODO(pq): incorporate these into the heap display (maybe into a single combo selection).
     toolbarGroup.add(registerAction(new OpenTimelineViewAction(app)));
     toolbarGroup.add(registerAction(new OpenObservatoryAction(app)));
-    toolbarGroup.addSeparator();
+    // TODO(pq): push this into preferences or a menu option.
+    if (PerfService.DISPLAY_HEAP_USE) {
```
devoncarew: +1 to using the gear menu (and defaulting to on) in the future (and top-level pref would work as well)

src/io/flutter/inspector/HeapDisplay.java
devoncarew: nevermind

src/io/flutter/perf/PerfService.java
pq: Excellent.  Thanks!

src/io/flutter/inspector/HeapDisplay.java
```diff
@@ -0,0 +1,234 @@
...
+// TODO(pq): tweak scaling
+// TODO(pq): fix duplicate value caching (remove int cache from graph)
+public class HeapDisplay extends JPanel {
+
+  public static class ToolbarComponentAction extends AnAction implements CustomComponentAction, HeapListener, Disposable {
+
+    private final List<JPanel> panels = new ArrayList<>();
+    private final List<HeapDisplay> graphs = new ArrayList<>();
+
+    // TODO(pq): be smart about setting capacity based on available width.
```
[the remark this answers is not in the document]
pq: Since we're getting samples at the rate of 1 per second, is this the same thing as keeping the last 60 values?

src/io/flutter/inspector/HeapDisplay.java
pq: 😊

src/io/flutter/inspector/HeapDisplay.java
pq: If I'm understanding JBColor right, this is done for us.  For example, light gray is declared like this:

     new JBColor(Gray._192, Gray._64);

where `Gray._192` is for regular presentation and `Gray._64` for dark.

Let me know if I'm missing something though!

src/io/flutter/perf/HeapMonitor.java
pq: Ah.  I was probably oversimplifying, thinking that we could infer time from the count (since we're polling at a steady rate).  Let's discuss in person.  (I'm open to whatever!)

src/io/flutter/inspector/HeapDisplay.java
devoncarew: Ah, cool; I wasn't aware that `JBColor.LIGHT_GRAY` auto adjusted to the theme.

devoncarew: (from digging into the source, we may want to prefer something like `UIUtil.getLabelForeground()`)

pq: Hmmmm.  I like the idea but that feels a little dark:

image

I'll poke around and see if there's something else a little more semantic than "gray".

PR #1924, opened 2018-03-15
some tweaks to the memory usage view

devoncarew:
A follow up to https://github.com/flutter/flutter-intellij/pull/1921, some tweaks to the memory usage view:

- track and graph samples by sample time (setting us up to be able to record GC events)
- dock the view samples to the right, so the samples always grow out of the right hand side
- switch the y axis (to account for the swing coordinate system)
- add a tooltip for the memory used

@pq

### DIFF
--- a/src/io/flutter/inspector/HeapDisplay.java
+++ b/src/io/flutter/inspector/HeapDisplay.java
@@ -13,6 +13,7 @@ import com.intellij.openapi.actionSystem.ex.CustomComponentAction;
 import com.intellij.openapi.util.Disposer;
 import com.intellij.ui.JBColor;
 import com.intellij.util.ui.JBUI;
+import io.flutter.perf.HeapMonitor;
 import io.flutter.perf.HeapMonitor.HeapListener;
 import io.flutter.perf.HeapMonitor.HeapSample;
 import io.flutter.perf.HeapMonitor.HeapSpace;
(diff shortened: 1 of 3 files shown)


src/io/flutter/inspector/HeapDisplay.java
```diff
@@ -13,33 +13,33 @@
...
 import java.awt.*;
-import java.util.*;
+import java.text.DecimalFormat;
+import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
 
 // TODO(pq): make capacity setting dynamic
-// TODO(pq): add label displaying curent/total heap use
+// TODO(pq): add label displaying current/total heap use
```
pq: Good catch!

src/io/flutter/inspector/HeapDisplay.java
```diff
@@ -82,153 +82,147 @@ public void dispose() {
...
-      return samples.samples;
-    }
+  private void updateFrom(HeapState state) {
+    this.heapState = state;
 
-    public int getMaxHeapInBytes() {
-      return heapMaxInBytes;
+    if (!heapState.getSamples().isEmpty()) {
+      final HeapSample sample = heapState.getSamples().get(0);
+      setToolTipText(printMb(sample.getBytes()) + " of " + printMb(heapState.getMaxHeapInBytes()));
```
pq: Neat!

I was imagining we'd display this next to the sample graph but I guess that depends on how much horizontal space we have.  Great idea to put it in the tooltip. 👍

devoncarew: Sounds great! We can remove this this tooltip then.

Issue #1920, opened 2018-03-15
only auto-convert modules types when running in android studio


devoncarew:
/cc @jwren

PR #1925, opened 2018-03-16
dial back our module auto-conversion a bit

devoncarew:
- dial back our module auto-conversion a bit; only perform this for Android Studio
- fix https://github.com/flutter/flutter-intellij/issues/1920

@jwren


jwren:
LGTM

PR #1926, opened 2018-03-16
Run render server and instrumented code from a temporary directory.

scheglov:


scheglov:
Note, that you will need this change https://dart-review.googlesource.com/c/sdk/+/46920 in your Dart SDK for imports rewrite to work.

PR #1930, opened 2018-03-17
handle GC events

devoncarew:
Some misc work on the memory view:
- add samples for GC events
- fix an issue where the display values wouldn't update after a full restart
- ensure that the foreground color works in dark themes

@pq


src/io/flutter/inspector/HeapDisplay.java
```diff
@@ -147,7 +157,7 @@ protected void paintComponent(Graphics g) {
       graphPoints.add(new Point(x, y));
     }
 
-    graphics2D.setColor(JBColor.LIGHT_GRAY);
+    graphics2D.setColor(UIUtil.getLabelDisabledForeground());
```
pq: 👍

PR #1940, opened 2018-03-19
Clear preview area if no rendered widget.

scheglov:


scheglov:
I had to remove `Rendering...` message, because when we are in a widget and edit some text, it blinks from rendered state to `Rendering...`, and then back to rendered state.

PR #1941, opened 2018-03-19
Group observatory actions.

pq:
Adds an overflow style group for observation actions.

image

image

I'm imagining that we could add a separator and include heap display config too.  For example something like:

```
Open Observatory
Open the Timeline View
----------------------
[ ] Display Heap Usage
[ ] Show Heap Use Summary
...
```

@devoncarew


devoncarew:
lgtm!

Once we have small areas for live memory usage and frame rate graphs, we can add actions to those, to be able to open the observatory and break out into more detailed views.

PR #1942, opened 2018-03-19
More incremental work for commented out Bazel test run config.

jwren:
Override getSourcePosition() in BazelPositionMapper to open the correct file when a breakpoint is hit.


jwren:
Not sure what you mean under test.  Lets discuss in person.

pq:
I just mean tested (preferably unit tested).

jwren:
Not sure yet.  Do you know if we have tests for the Flutter test run config, io.flutter.run.test?

pq:
I don't think so but you can see how far Brian got in the `testSrc` directory.

PR #1944, opened 2018-03-19
Move Inspector View toolbar overflow to the East.

pq:
IMO feels more consistent with where similar menus are rendered for views and plays a bit nicer with resizing.

image

@devoncarew @jacob314


jacob314:
Looks good!

Issue #1936, opened 2018-03-19
Open in Android Studio does not work for plugins


stevemessick:
We need to detect when `Open in Android Studio` is used in either the plugin/android module or the plugin/example/android module. In both cases Android Studio needs to open the Android project in the plugin/example app.

PR #1955, opened 2018-03-20
Open plugin modules in AS

stevemessick:
`Open Android module in Android Studio` needs special handling when working with plugins. There are two Android modules, one for the plugin and another for the example app. However, opening the plugin's module directly is not useful since class paths are not set correctly. Therefore, always open the module defined by the example app when editing Android modules in plugins.

There are some side-issues here. We don't want to display the Gradle project-import dialog, which asks for the location of Gradle. That's solved. What isn't solved is the way the importer changes the default project creation location. After opening a module, new Flutter projects default to that modules's location for their default directory. Fixing that problem will come later.

Fixes https://github.com/flutter/flutter-intellij/issues/1936

@pq @devoncarew

### DIFF
--- a/src/io/flutter/actions/OpenInAndroidStudioAction.java
+++ b/src/io/flutter/actions/OpenInAndroidStudioAction.java
@@ -124,10 +124,13 @@ public class OpenInAndroidStudioAction extends AnAction {
     return null;
   }
 
+  // Return true if the directory is named android and contains either an app (for applications) or a src (for plugins) directory.
   private static boolean isAndroidWithApp(@NotNull VirtualFile file) {
     return file.getName().equals("android") && (file.findChild("app") != null || file.findChild("src") != null);
   }
 
+  // Return true if the directory has the structure of a plugin example application: a pubspec.yaml and an
+  // android directory with an app. The example app directory name is not specified in case it gets renamed.
   private static boolean isExampleWithAndroidWithApp(@NotNull VirtualFile file) {
     boolean hasPubspec = false;
     boolean hasAndroid = false;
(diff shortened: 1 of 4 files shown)


src/io/flutter/actions/OpenInAndroidStudioAction.java
```diff
@@ -125,19 +125,60 @@ private static VirtualFile getProjectForFile(@Nullable VirtualFile file) {
   }
 
   private static boolean isAndroidWithApp(@NotNull VirtualFile file) {
-    return file.getName().equals("android") && file.findChild("app") != null;
+    return file.getName().equals("android") && (file.findChild("app") != null || file.findChild("src") != null);
```
devoncarew: Can you add some text describing the project structure we're looking for? This is the android/ subdir, with both app/ and src/ subdirs (flutter apps but not flutter plugins)?

src/io/flutter/actions/OpenInAndroidStudioAction.java
```diff
@@ -125,19 +125,60 @@ private static VirtualFile getProjectForFile(@Nullable VirtualFile file) {
...
+        final Project project = e.getProject();
+        assert (project != null);
+        final VirtualFile projectDir = project.getBaseDir();
+        for (PubRoot root : PubRoots.forProject(project)) {
+          if (root.isFlutterPlugin()) {
+            VirtualFile rootFile = root.getRoot();
+            VirtualFile aFile = file;
+            while (aFile != null) {
+              if (aFile.equals(rootFile)) {
+                // We know a plugin resource is selected. Find the eample app for it.
```
devoncarew: sp: eample

stevemessick:
I made the comments more descriptive.

PR #1956, opened 2018-03-20
Use 2px blue line as selection feedback.

scheglov:


src/io/flutter/preview/MoveHandle.java
```diff
@@ -14,7 +15,10 @@
 public class MoveHandle extends Handle {
   @Override
   public void paint(Graphics g) {
-    g.drawRect(0, 0, getWidth(), getHeight());
+    final Graphics2D g2 = (Graphics2D)g;
+    g2.setStroke(new BasicStroke(2));
+    g2.setColor(JBColor.BLUE);
```
devoncarew: Is there a 'selection' color?

scheglov: Is this any better?
https://github.com/flutter/flutter-intellij/pull/1958

PR #1957, opened 2018-03-20
Improvements to the Flutter Test (Bazel) launch configuration UI.

jwren:
bazel-test-config


src/io/flutter/run/bazelTest/FlutterBazelTestConfigurationEditorForm.form
```diff
@@ -83,9 +61,36 @@
...
+          <text value="The absolute path to a Bazel launching script."/>
+        </properties>
+      </component>
+      <component id="eff55" class="javax.swing.JLabel">
+        <constraints>
+          <grid row="5" column="1" row-span="1" col-span="1" vsize-policy="0" hsize-policy="0" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+        </constraints>
+        <properties>
+          <enabled value="false"/>
+          <text value="The Bazel target, some //a/b/c:name."/>
```
devoncarew: Here, we should say `Generally, something like...` for similarity with above, or use an e.g.

jwren: Yes, I'll be following up to make the two bazel launch configs more consistent. I went back and forth on the "Generally".

PR #1964, opened 2018-03-21
Turn 'Show only widget' ON by default, send analytics.

scheglov:


src/io/flutter/preview/PreviewViewState.java
```diff
@@ -55,5 +57,19 @@ public void removeListener(ChangeListener listener) {
...
+
+  private void sendToAnalytics() {
+    final Analytics analytics = FlutterInitializer.getAnalytics();
+
+    // Send the number of times the state is initialized.
+    analytics.sendEvent("previewState", "ping");
+
+    // Send boolean properties turned on.
+    if (showOnlyWidgets) {
+      analytics.sendEvent("previewState", "showOnlyWidgets");
```
devoncarew: I actually think we should try and find a way to send this as part of our other settings, on order to simplify the job of interpreting the data on the back end,

devoncarew: We could move the `showOnlyWidgets` setting out of this class and into `FlutterSettings` (but still show the UI in the outline view).

scheglov: OK, done.
https://github.com/flutter/flutter-intellij/pull/1969

PR #1965, opened 2018-03-21
Remove stale dart-plugin jars.

pq:
We're building against `171.4694.29`.


pq:
@devoncarew: this is safe to remove?

devoncarew:
👍

PR #1966, opened 2018-03-21
Show header with the widget name above the preview area.

scheglov:
It seems that users are confused otherwise, and selection in the tree is not enough, so we decided to spend some vertical space for explanation. OTOH, this header might eventually host some additional action buttons (which don't fit / not appropriate for the main view toolbar?).

image

### DIFF
--- a/src/io/flutter/preview/PreviewArea.java
+++ b/src/io/flutter/preview/PreviewArea.java
@@ -7,6 +7,9 @@ package io.flutter.preview;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import com.intellij.openapi.actionSystem.*;
+import com.intellij.openapi.actionSystem.ex.CustomComponentAction;
+import com.intellij.openapi.ui.SimpleToolWindowPanel;
 import com.intellij.ui.IdeBorderFactory;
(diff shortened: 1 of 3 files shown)


devoncarew:
Should we use the regular toolbar class here to show the name? It would likely have similar padding and look as other areas in IntelliJ.

I think we should also just use the name as the label (`MyWidget`) instead of the longer `UI preview of MyWidget's build method:`.

src/io/flutter/preview/PreviewArea.java
```diff
@@ -152,10 +159,18 @@ public void show(@NotNull FlutterOutline unitOutline, @NotNull JsonObject render
       return;
     }
 
+    final Element widgetClassElement = widgetOutline.getDartElement();
+    if (widgetClassElement != null) {
+      header.setText("UI preview of " + widgetClassElement.getName() + "’s build() method:");
```
pq: Perhaps something a little more terse?  What about just:

    widgetClassElement.getName() + ".build():"

yielding something like:

 **MyWidget.build():**

scheglov: PTAL

scheglov: image

PR #1967, opened 2018-03-21
Bazel run config polish

jwren:


src/io/flutter/run/bazel/FlutterBazelConfigurationEditorForm.form
```diff
@@ -112,7 +112,7 @@
         </constraints>
         <properties>
           <enabled value="false"/>
-          <text value="The Bazel target, some //a/b/c:name."/>
+          <text value="The Bazel target. For example: //a/b/c:name."/>
```
devoncarew: Once this lands, I think I'll experiment with changing all of our `, some...` and `. For example:` text to using something like ` (e.g. ...)`.

devoncarew: (not clear it's going to be better, but worth trying it out)

jwren: SGTM.  Cleanup, iterating & polish is what it is all about right now before the release.

PR #1969, opened 2018-03-21
Move 'Show only widgets' to settings.

scheglov:


src/io/flutter/settings/FlutterSettings.java
```diff
@@ -136,11 +140,21 @@ public void setOrganizeImportsOnSaveKey(boolean value) {
     fireEvent();
   }
 
+  public boolean isShowOnlyWidgets() {
```
devoncarew: I'd move this down next to the `setShowOnlyWidgets` call.

PR #1973, opened 2018-03-21
update our build versions

devoncarew:
- update the various versions we build against
- build releases into a new `releases/` dir, instead of the `artifacts/` dir (which has downloaded artifacts as well)

@stevemessick

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 out/
 artifacts/
 build/
+releases/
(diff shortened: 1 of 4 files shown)


stevemessick:
Looks like a test needs to be updated.

devoncarew:
Updated! re-building now...

stevemessick:
Updated test LGTM

Issue #1963, opened 2018-03-21
update the meaning of the preview dart 2 flag


devoncarew:
It's still opt-in.

In a M23.1 release, we'll change the meaning to opt-out.

devoncarew:
Or possible, always send a flag in, to deal with the differences between master/dev/beta.

devoncarew:
We could do a 3-state pulldown here.

PR #1975, opened 2018-03-21
add a ternary setting for the preview dart 2 flag

devoncarew:
- add a ternary setting for the preview dart 2 flag; user's can take the sdk default, can explicitly enable it, or can explicitly disable it
- fixes https://github.com/flutter/flutter-intellij/issues/1963

screen shot 2018-03-21 at 3 41 54 pm

### DIFF
--- a/src/io/flutter/settings/FlutterSettings.java
+++ b/src/io/flutter/settings/FlutterSettings.java
@@ -18,12 +18,28 @@ public class FlutterSettings {
   private static final String reloadOnSaveKey = "io.flutter.reloadOnSave";
   private static final String openInspectorOnAppLaunchKey = "io.flutter.openInspectorOnAppLaunch";
   private static final String verboseLoggingKey = "io.flutter.verboseLogging";
-  private static final String disablePreviewDart2Key = "io.flutter.disablePreviewDart2";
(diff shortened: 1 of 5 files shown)


src/io/flutter/settings/FlutterSettings.java
```diff
@@ -18,13 +18,29 @@
...
 
+  public enum Dart2ModeSettings {
+    useSdkDefault(0),
+    enablePreviewDart2(1),
+    disablePreviewDart2(2);
+
+    private final int ordinal;
+
+    Dart2ModeSettings(int ordinal) {
+      this.ordinal = ordinal;
```
pq: ~Don't you get ordinals for free with Java enums?~

EDIT: OK, I think this is clearer.

devoncarew: I could use, but I just want to make the order very explicit, since the UI has the same order.

Alternatively, I could pull the text from the enum itself (a user displayName field), and populate the combobox with that.

pq: Whatever works.  To be honest I tend to abuse enums and lean towards overly depending on things like natural order and text.  It's cute but leads to possible bugs when someone else comes along and decides to re-sort or re-name...  Color me on the fence!

PR #1976, opened 2018-03-22
Fix bugs in StdoutJsonParser when a line spanned multiple calls to appendOutput.

jacob314:
The downside of this approach is we will not display a line until a \n occurs but that appears to be fine.

An alternate robust approach would be to use an actual streaming JSON parser to track when the JSON blocks are complete.

The previous approach failed if the JSON had an }] too early in the block and failed on json responses larger than a single call to appendOutput.

### DIFF
--- a/src/io/flutter/utils/StdoutJsonParser.java
+++ b/src/io/flutter/utils/StdoutJsonParser.java
@@ -10,59 +10,50 @@ import java.util.List;
 
 /**
  * A class to process regular text output intermixed with newline-delimited JSON.
+ *
+ * JSON lines starting with [{ are never be split into multiple lines even
+ * if they are emitted over the course of multiple calls to appendOutput.
+ * Regular lines on the other hand are emitted immediately so that debug output
+ * always shows up when expected..
  */
 public class StdoutJsonParser {
   private final StringBuilder buffer = new StringBuilder();
+  private boolean bufferIsJson = false;
   private final List<String> lines = new ArrayList<>();
 
   /**
    * Write new output to this [StdoutJsonParser].
    */
   public void appendOutput(String output) {
-    buffer.append(output);
+    for (int i = 0; i < output.length(); ++i) {
+      final char c = output.charAt(i);
+      buffer.append(c);
+      if (!bufferIsJson && buffer.length() == 2 && buffer.charAt(0) == '[' && c == '{') {
+        bufferIsJson = true;
+      }
+      if (c == '\n') {
+        flushLine();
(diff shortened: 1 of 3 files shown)


jacob314:
Added tests covering the failure.
Also removed flush() from the API as it is not needed now that we always flush as long as we aren't in a suspected `[{` JSON block.
Verified that this doesn't create extra new lines and does cause text to display even before the first line feed is issued.

src/io/flutter/utils/StdoutJsonParser.java
```diff
@@ -10,59 +10,50 @@
 
 /**
  * A class to process regular text output intermixed with newline-delimited JSON.
+ *
+ * JSON lines starting with [{ are never be split into multiple lines even
```
devoncarew: ==> `are never split`

src/io/flutter/utils/StdoutJsonParser.java
```diff
@@ -10,59 +10,50 @@
...
+ * if they are emitted over the course of multiple calls to appendOutput.
+ * Regular lines on the other hand are emitted immediately so that debug output
+ * always shows up when expected..
```
devoncarew: fewer periods :)

testSrc/unit/io/flutter/utils/StdoutJsonParserTest.java
```diff
@@ -58,12 +57,43 @@ public void split_json() throws Exception {
...
 
     assertArrayEquals(
       "validating parser results",
       new String[]{"hello\n", "there\n", "[{'foo':'bar'}]\n", "bye\n"},
       parser.getAvailableLines().toArray()
     );
   }
+
+  @Test
+  public void deep_nested_json() throws Exception {
```
devoncarew: nit: I believe our test names use lowerCamelCase

src/io/flutter/utils/StdoutJsonParser.java
jacob314: done

src/io/flutter/utils/StdoutJsonParser.java
jacob314: done..........

testSrc/unit/io/flutter/utils/StdoutJsonParserTest.java
jacob314: done. for some reason the existing tests in this file used_under_scores_so_i_copied_them :)

PR #1980, opened 2018-03-22
Make InspectorService supporting issuing most of its requests over the Flutter Daemon instead of observatory.

jacob314:
This avoids problematic cases where the observatory protocol allowed
requests to issue at surprising times such as right in the middle of
a flutter method call.
It is expected this fixes many hard or impossible to reproduce inspector
bugs.

To simplify the code I have also changed the semantics so that
the supportedServiceMethods set is computed before an InspectorService
is considered ready. This reduces the number of future.thenCompose
calls within the InspectorService code dealing with functionality only
exposed by the latest version.

### DIFF
--- a/src/io/flutter/inspector/InspectorService.java
+++ b/src/io/flutter/inspector/InspectorService.java
@@ -17,6 +17,7 @@ import io.flutter.run.daemon.FlutterApp;
 import io.flutter.utils.VmServiceListenerAdapter;
 import org.dartlang.vm.service.VmService;
 import org.dartlang.vm.service.element.*;
+import org.jetbrains.annotations.NotNull;
 
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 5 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -199,12 +207,23 @@ private void addInspectorPanel(String displayName,
    * Called when a debug connection starts.
    */
   public void debugActive(@NotNull FlutterViewMessages.FlutterDebugEvent event) {
+    final FlutterApp app = event.app;
+
+    whenCompleteUiThread(InspectorService.create(app.getFlutterDebugProcess(), app.getVmService()),
+                         (InspectorService inspectorService, Throwable throwable) -> {
+                           if (throwable != null) {
+                             LOG.error(throwable);
```
devoncarew: I think we want `warn` here, so we don't show a dialog to the user.

src/io/flutter/view/InspectorPanel.java
```diff
@@ -82,6 +82,7 @@
   private final InspectorService.FlutterTreeType treeType;
   @NotNull
   private final FlutterApp flutterApp;
+  private final InspectorService inspectorService;
```
devoncarew: `@NotNull`, to match the constructor param.

src/io/flutter/view/InspectorPanel.java
```diff
@@ -192,7 +195,7 @@ public void onAppChanged() {
 
   @Nullable
```
devoncarew: This should now be `@NotNull` I believe.

src/io/flutter/inspector/InspectorService.java
```diff
@@ -32,18 +33,60 @@
...
-  private final Set<InspectorServiceClient> clients;
+  @NotNull private final FlutterDebugProcess debugProcess;
+  @NotNull private final VmService vmService;
+  @NotNull private final Set<InspectorServiceClient> clients;
   private EvalOnDartLibrary inspectorLibrary;
-  private CompletableFuture<Set<String>> supportedServiceMethods;
+  @NotNull private final Set<String> supportedServiceMethods;
+
+  // TODO(jacobr): remove this field as soon as
+  // `ext.flutter.debugCallWidgetInspectorService` has been in two revs of the
```
devoncarew: This is a flutter framework service extension?

src/io/flutter/inspector/InspectorService.java
```diff
@@ -120,19 +162,69 @@ public void addClient(InspectorServiceClient client) {
...
     if (arg == null || arg.getId() == null) {
       return getInspectorLibrary().eval("WidgetInspectorService.instance." + methodName + "(null, \"" + groupName + "\")", null);
     }
     return getInspectorLibrary()
       .eval("WidgetInspectorService.instance." + methodName + "(\"" + arg.getId() + "\", \"" + groupName + "\")", null);
   }
 
-  CompletableFuture<InstanceRef> invokeServiceMethodOnRef(String methodName, InstanceRef arg) {
+  /**
+   * This API requires using the observatory service a the input parameter is an
```
devoncarew: some grammar nits here

src/io/flutter/view/FlutterView.java
jacob314: done

src/io/flutter/view/InspectorPanel.java

src/io/flutter/view/InspectorPanel.java

src/io/flutter/inspector/InspectorService.java
jacob314: discussed offline. Switched to
ext.flutter.inspector.methodName
and cleaned up how we were passing parameters.
Parameters are now passed as either
{'arg': arg, 'objectGroup': objectGroup}
{'objectGroup': objectGroup}
or
{'arg0': arg0, 'arg1': arg1, 'arg2': arg2, ...}
for the one vararg like case.

src/io/flutter/inspector/InspectorService.java
jacob314: rewrote this comment.

devoncarew:
Jacob, does this address the two crashers on https://github.com/flutter/flutter-intellij/milestone/34?

jacob314:
I've updated the two bugs. The first should be fixed and the second should be mitigated to the point where it can be pushed to M24.

PR #1983, opened 2018-03-22
Restore Project Structure editor

stevemessick:
Allow the Project Structure menu item to show up in Android Studio. Tested by building a 3.0 plugin and installing it into AS 3.0.1, and by running debuggable AS from sources.

It still uses the one that comes with IntelliJ, so needs some work, but at least it can be used to add (non-Dart) dependencies to Flutter modules. There is some code that is commented-out that will be useful in fixing up support for the new one (I hope).

@devoncarew @pq

### DIFF
--- a/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
+++ b/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
@@ -9,6 +9,7 @@ import com.intellij.openapi.actionSystem.ActionManager;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.startup.StartupActivity;
+import io.flutter.actions.FlutterShowStructureSettingsAction;
 import io.flutter.actions.OpenAndroidModule;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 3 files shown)


devoncarew:
I don't think I have enough context here why we're adding the menu item back into Android Studio - let's plan to discuss ~tomorrow am.

flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
```diff
@@ -30,5 +31,6 @@ public void runActivity(@NotNull Project project) {
     // The IntelliJ version of this action spawns a new process for Android Studio.
     // Since we're already running Android Studio we want to simply open the project in the current process.
     replaceAction("flutter.androidstudio.open", new OpenAndroidModule());
+    replaceAction("ShowProjectStructureSettings", new FlutterShowStructureSettingsAction());
```
devoncarew: Can we only do this replace if the given `project` is a Flutter project? We have some utils for detecting this, that look inside the pubspec.yaml file.

devoncarew:
I also think we should investigate ways we can get the new, JAVA_MODULE based flutter project type to be closer to something that Android Studio is happy with.

stevemessick:
I have implemented the guard clause and will add it with the next PR.

PR #1987, opened 2018-03-23
add shadows to the panels in the preview view

devoncarew:
@scheglov, some UI adjustments to the preview area:

- don't have a border for the area
- give the panels for widgets a left-bottom shadow
- change the label color based on the container background color
- use new colors for the panels

screen shot 2018-03-23 at 7 22 06 am

Let's talk in person re: the colors. I'm not committed to these ones, but want to iterate a bit on what's currently in master.

### DIFF
--- a/src/io/flutter/preview/PreviewArea.java
+++ b/src/io/flutter/preview/PreviewArea.java
@@ -10,10 +10,13 @@ import com.google.gson.JsonObject;
 import com.intellij.openapi.actionSystem.*;
 import com.intellij.openapi.actionSystem.ex.CustomComponentAction;
 import com.intellij.openapi.ui.SimpleToolWindowPanel;
-import com.intellij.ui.IdeBorderFactory;
+import com.intellij.ui.ColorUtil;
+import com.intellij.ui.components.JBLabel;
 import com.intellij.util.ui.JBUI;
+import com.intellij.util.ui.UIUtil;
 import org.dartlang.analysis.server.protocol.Element;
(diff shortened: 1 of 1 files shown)


src/io/flutter/preview/PreviewArea.java
```diff
@@ -41,17 +44,19 @@
 // TODO: we need to take the layer (or z-index?) of the widget into account (see Scaffold and its FAB)
 
 public class PreviewArea {
-  public static int BORDER_WITH = 5;
+  public static int BORDER_WIDTH = 0;
 
   @SuppressWarnings("UseJBColor")
   private static final Color[] pastelColors = new Color[]{
```
scheglov: Change the name?

devoncarew:
More tweaks to the colors:

screen shot 2018-03-23 at 9 37 38 am

scheglov:
`DropShadowBorder` prints these times. Sad!
image

devoncarew:
That seems crazy.

devoncarew:
But, good to know that it mostly just takes 0-1 ms to render :)

scheglov:
It has been like this for [at least 7 years](https://stackoverflow.com/questions/4627977/stop-dropshadowborder-console-logging) ;-)

devoncarew:
I'd like to keep a shadow like border, but we'll definitely need to find a different implementation.

devoncarew:
@scheglov, I'm going to use a different shadow

PR #1998, opened 2018-03-23
Report the kind of rendering failure to the listener.

scheglov:


src/io/flutter/preview/PreviewView.java
```diff
@@ -182,8 +173,23 @@ public void onResponse(FlutterOutline widget, JsonObject response) {
...
     @Override
-    public void onFailure(Throwable e) {
-      noWidget();
+    public void onFailure(RenderProblemKind kind) {
+      ApplicationManager.getApplication().invokeLater(() -> {
+        if (previewArea != null) {
+          switch (kind) {
+            case EXCEPTION:
+            case NO_TEMPORARY_DIRECTORY:
+              previewArea.clear("There was an exception during rendering");
```
pq: Should we encourage them to file a bug report here?  Maybe a link to submit feedback?

src/io/flutter/preview/PreviewView.java
```diff
@@ -182,8 +173,23 @@ public void onResponse(FlutterOutline widget, JsonObject response) {
...
+              break;
+            case NO_WIDGET:
+              previewArea.clear("The selection does not correspond to a widget");
```
pq: Will this also show when the widget is missing a `forDesignTime()` constructor?

pq: It might be good to distinguish between widgets in general and un-renderable widgets.  Maybe as simple as changing the text to:

    "The selection does not correspond to a renderable widget"

In a future release we could add a link to take users to docs explaining what this means...

src/io/flutter/preview/PreviewView.java
scheglov: In theory these should never happen.

src/io/flutter/preview/PreviewView.java
scheglov: Yes, right now we cannot distinguish "no widget" and "there is a widget, but it does not have `forDesignTime()` constructor". I will enhance the Flutter Outline API to fix this.

pq: Great.  Thanks!

src/io/flutter/preview/PreviewView.java
pq: 👍

PR #1999, opened 2018-03-23
Register the outer panel as the widget component.

scheglov:
I don't fully understand why, but otherwise the blue selection line disappears as you type in the widget.


devoncarew:
Can you file an issue to fix the off by one pixel selection? I'm happy to take a look in a bit.

Issue #1250, opened 2017-07-28
Extract Widget


brianegan:
Idea: Extract Widget

How it works: Highlight some widgets in a large widget tree you'd like to extract. Currently, it's east to extract these as a method in the current class using the "extract method" action. However, I often find myself creating a StatelessWidget out of this selection instead of a method. It would be very handy if you could "extract widget".

This would allow you to create a Stateless Widget, with a constructor and fields that match the necessary params. For example, if you extracted a piece of Code, and it required a `Todo` with the variable name `todo`, the StatelessWidget would extract a `final Todo todo` field and a Constructor to match. The build method would return the contents of your highlighted selection.

After creating the class, it would replace your current highlighted selection with the `new MyTodoWidget(todo)`;

Essentially, this would just be a slightly different extraction than a method.

bwilkerson:
That sounds perfectly reasonable, but I'll note that it would require work in both the analysis server and in each client. Refactorings are a little harder to add because they allow for / require user interaction, and that requires building a UI in the client. That doesn't make this impossible, but does make it a bigger task than just adding a quick fix or quick assist.

brianegan:
It's never as simple as it appears, is it? :) 

Totally understand if this is high effort / low reward, just an idea as it's a natural refactoring I'd noticed myself doing a number of times.

scheglov:
I agree, this would be a good feature. And we actually had a lot of experience with [editing UIs in the past](http://help.eclipse.org/kepler/index.jsp?topic=%2Forg.eclipse.wb.doc.user%2Fhtml%2Ffeatures%2Fvisual_inheritance.html&cp=64_4_4). We even had the "Extract Composite" refactoring in one of the previous versions of our visual UI builder :-)

Yes, this is pretty much the "Extract Method" factoring with wrapping the extract code into a class with moving variables into constructor arguments instead of method arguments.

`StatelessWidget` seems simple enough. Would `StatefulWidget` any use? I also don't know whether passing in (potentially mutable) widgets, like mentioned `Todo`, is considered a good pattern in Flutter.

Ideally it should be a Flutter specific refactoring, and special integration with IDE(s). In the interim it could be implemented as a Quick Assist, which will generate a unique class name and initiate linked rename, or just leave it as is and allow you to use Rename refactoring as the next step.

sethladd:
I'd love a "extract this function/method into a new widget". Originally requested here: https://github.com/flutter/flutter/issues/11919

[...]

PR #2005, opened 2018-03-26
Integrate 'Extract Widget' refactoring.

scheglov:
image

https://github.com/flutter/flutter-intellij/issues/1250

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -192,6 +192,11 @@
       <add-to-group group-id="RunMenu" anchor="after" relative-to-action="Stop"/>
     </group>
 
+    <!-- refactoring menu -->
+    <action class="io.flutter.actions.ExtractWidgetAction" id="Flutter.ExtractWidget" text="Flutter Widget...">
+      <add-to-group group-id="IntroduceActionsGroup" anchor="after" relative-to-action="ExtractMethod"/>
+    </action>
+
     <!-- help menu -->
     <action class="io.flutter.actions.FlutterGettingStartedAction" id="Flutter.FlutterHelp" text="Flutter Plugin Help">
       <add-to-group group-id="HelpMenu" anchor="after" relative-to-action="HelpTopics"/>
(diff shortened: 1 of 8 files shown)


resources/META-INF/plugin.xml
```diff
@@ -192,6 +192,11 @@
       <add-to-group group-id="RunMenu" anchor="after" relative-to-action="Stop"/>
     </group>
 
+    <!-- refactoring menu -->
+    <action class="io.flutter.actions.ExtractWidgetAction" id="Flutter.ExtractWidget" text="Flutter Widget...">
```
devoncarew: Will this show up in an 'extract' menu? I'm wondering if the user will have enough context to know that `Flutter Widget...` means extract flutter widget. (if so, then lets keep the shorter text)

src/io/flutter/actions/ExtractWidgetAction.java
```diff
@@ -0,0 +1,127 @@
...
+
+  public ExtractWidgetDialog(@NotNull Project project,
+                             @Nullable Editor editor,
+                             @NotNull ExtractWidgetRefactoring refactoring) {
+    super(project, editor, refactoring);
+    myRefactoring = refactoring;
+    setTitle("Extract Widget");
+    init();
+
+    myNameField.setText("NewWidget");
```
devoncarew: We don't know whether this new name will be legal `NewWidget` - if we did we could do something like `NewWidget7`.

Another option is to autogenerate interesting names that users would need to change. So, `BluePony` or `TinySunrise`; this is what DartPad does. I'm not advocating for this, but thought I would mention it.

resources/META-INF/plugin.xml
scheglov: The name and location is consistent with other `extract` refactorings.

image

slightfoot:
@scheglov @devoncarew Is it possible to get this option in the Quick Assist popup menu?
Re: https://twitter.com/devangelslondon/status/988779563990814720

scheglov:
This is not something that can be easily done for IntelliJ.

Quick Assists are contributed by the Dart plugin, and they show only the specific set of actions returned from Dart Analysis Server. There are no ways to include arbitrary actions, from the Flutter plugin.

But you can add a keyboard shortcut, so it will be easily accessible.
image
image

slightfoot:
I still think its worth looking into as an addition, the quick actions is the go to solution for Widget actions when doing Flutter development.

dahabit:
@slightfoot Thanks for your suggestion, @scheglov it will be great if we can implement that

PR #2006, opened 2018-03-26
add an fps view; implement a larger memory view

devoncarew:
- add a larger version of the memory view at the bottom of the inspector
- implement an FPS view
- both memory and fps views show in the toolbar if the inspector tool window is in a horizontal orientation, and at the bottom of the view if in a vertical orientation

Still left to do (likely in a follow-up) is to make the preference to show opt-out, and move it to the overflow menu of the inspector view (or similar location).

screen shot 2018-03-25 at 9 55 08 pm

screen shot 2018-03-25 at 9 56 56 pm

### DIFF
--- a/src/io/flutter/inspector/FPSDisplay.java
+++ b/src/io/flutter/inspector/FPSDisplay.java
@@ -93,8 +93,7 @@ public class FPSDisplay {
   }
 
   public static class ToolbarComponentAction extends AnAction implements CustomComponentAction {
-    // The width of the drawable fps area.
-    private static final int HEAP_GRAPH_WIDTH = 60;
+    private static final int DRAWABLE_HEAP_WIDTH = 60;
 
     final @NotNull Disposable parent;
     final @NotNull FlutterApp app;
@@ -127,7 +126,7 @@ public class FPSDisplay {
 
       // Graph component.
       final FPSPanel fpsPanel = new FPSPanel(flutterFramesMonitor);
-      fpsPanel.setPreferredSize(new Dimension(HEAP_GRAPH_WIDTH, -1));
+      fpsPanel.setPreferredSize(new Dimension(DRAWABLE_HEAP_WIDTH, -1));
 
       final FlutterFramesMonitor.Listener listener = event -> {
         fpsPanel.update();
--- a/src/io/flutter/inspector/FPSDisplay.java
+++ b/src/io/flutter/inspector/FPSDisplay.java
@@ -1,0 +1,370 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.inspector;
+
+import com.intellij.openapi.Disposable;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.actionSystem.Presentation;
+import com.intellij.openapi.actionSystem.ex.CustomComponentAction;
+import com.intellij.openapi.util.Disposer;
+import com.intellij.ui.IdeBorderFactory;
+import com.intellij.ui.SideBorder;
+import com.intellij.ui.components.JBLabel;
+import com.intellij.util.ui.JBUI;
+import com.intellij.util.ui.UIUtil;
+import io.flutter.perf.FlutterFramesMonitor;
+import io.flutter.run.daemon.FlutterApp;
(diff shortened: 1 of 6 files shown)


devoncarew:
We may also want some badging to indicate that the user is in `debug` or `profile` mode; the data for both views will be much closer to what users will see in release builds when run in profile mode.

src/io/flutter/inspector/FPSDisplay.java
```diff
@@ -0,0 +1,370 @@
...
+  private final Map<FlutterFramesMonitor.FlutterFrameEvent, JComponent> frameWidgets = new HashMap<>();
+
+  private Rectangle lastSavedBounds;
+
+  FPSPanel(FlutterFramesMonitor framesMonitor) {
+    this.framesMonitor = framesMonitor;
+
+    setLayout(null);
+    final Color color = UIUtil.getLabelDisabledForeground();
+    //noinspection UseJBColor
```
pq: How does this look w/ a light theme?

src/io/flutter/inspector/HeapDisplay.java
```diff
@@ -22,40 +24,96 @@
...
+    assert app.getPerfService() != null;
+    app.getPerfService().addListener(listener);
+    Disposer.register(parentDisposable, () -> app.getPerfService().removeListener(listener));
+
+    return panel;
+  }
 
   public static class ToolbarComponentAction extends AnAction implements CustomComponentAction, HeapListener, Disposable {
+    // The width of the drawable heap area.
+    private static final int HEAP_GRAPH_WIDTH = 60;
```
pq: Maybe just rename this so you can delete the comment?  That is: `DRAWABLE_HEAP_WIDTH`?

src/io/flutter/view/FlutterView.java
```diff
@@ -132,14 +133,18 @@ private DefaultActionGroup createToolbar(@NotNull ToolWindow toolWindow, @NotNul
...
+    if (FlutterSettings.getInstance().isShowHeapDisplay() && isHorizontal) {
+      toolbarGroup.add(HeapDisplay.createToolbarAction(parentDisposable, app));
+      toolbarGroup.addSeparator();
+      toolbarGroup.add(FPSDisplay.createToolbarAction(parentDisposable, app));
+      toolbarGroup.addSeparator();
+      //toolbarGroup.add(new ObservatoryActionGroup(this, app));
     }
+
+    toolbarGroup.add(registerAction(new OpenTimelineViewAction(app)));
+    toolbarGroup.add(registerAction(new OpenObservatoryAction(app)));
```
pq: Curious why we're losing the group here?  FWIW: I was thinking this would be a good place to hang display configuration options (assuming we have any)

src/io/flutter/view/FlutterView.java
```diff
@@ -180,6 +187,17 @@ private void addInspector(FlutterApp app, InspectorService inspectorService, Too
                       true);
     addInspectorPanel("Render Tree", tabs, state, InspectorService.FlutterTreeType.renderObject, app, inspectorService, toolWindow,
                       toolbarGroup, false);
+
+    final boolean isVertical = !toolWindow.getAnchor().isHorizontal();
```
pq: Smart!  Curious what happens if you change orientations when the view is rendered.  (I'm guessing nothing bad and that you just get the appropriate settings on next run.)

pq: Incidentally, another thing I was considering was making the displays themselves be (re)size-aware.  For example, they could widen given more space and/or show more details (like summaries, etc.) . Anyway, more to consider!

src/io/flutter/inspector/FPSDisplay.java
devoncarew: It works - the label color does the right thing here; I'll update w/ a screenshot.

src/io/flutter/inspector/HeapDisplay.java
devoncarew: sgtm

src/io/flutter/view/FlutterView.java
devoncarew: generally I was thinking that we'll want the memory related UI and settings together, and the timeline ui and settings together, but that's not really cohesive in this PR - I'll add the group back in here

src/io/flutter/view/FlutterView.java
devoncarew: Yeah, it updates on the next app run. We should definitely plan to make this more dynamic in the future.

devoncarew:
screen shot 2018-03-27 at 7 15 09 am

PR #2022, opened 2018-03-28
Disable user config

stevemessick:
Set the 'disable user config' flag at start up.

@devoncarew

### DIFF
--- a/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
+++ b/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
@@ -11,6 +11,7 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.startup.StartupActivity;
 import io.flutter.actions.FlutterShowStructureSettingsAction;
 import io.flutter.actions.OpenAndroidModule;
+import io.flutter.project.FlutterProjectCreator;
 import io.flutter.utils.FlutterModuleUtils;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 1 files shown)


flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
```diff
@@ -36,5 +37,7 @@ public void runActivity(@NotNull Project project) {
     // Since we're already running Android Studio we want to simply open the project in the current process.
     replaceAction("flutter.androidstudio.open", new OpenAndroidModule());
     replaceAction("ShowProjectStructureSettings", new FlutterShowStructureSettingsAction());
+    // Unset this flag for all projects, mainly to ease the upgrade from 3.0.1 to 3.1.
+    FlutterProjectCreator.disableUserConfig(project);
```
devoncarew: Can we add some code here, so that we only automatically disable the setting once per project? This would be similar to the 'one time set' logic from this PR:

https://github.com/flutter/flutter-intellij/commit/e312527a265d2f8ab8665470e9dd7e594147f266#diff-ea5ca6788a3794828791e5cb1ef0af2fR137

but per-project based.

stevemessick: I thought this was a short-term fix to deal with migration problems. Rather than future-proofing it, which lends it undeserved legitimacy, I think we should plan to remove it ASAP.

devoncarew: It's unclear to me whether we'll see this with new projects as well; it seems safer to auto-set once, and then not again on a project, but I'll defer to you.

Happy to publish whatever fix you think best once this is in.

PR #2044, opened 2018-04-02
Update build artifacts

stevemessick:
Drop 3.0.1; update 3.1; add 3.2.

@devoncarew

### DIFF
--- a/resources/META-INF/studio-contribs.xml
+++ b/resources/META-INF/studio-contribs.xml
@@ -9,7 +9,7 @@
   </extensions>
 
   <extensions defaultExtensionNs="com.android.project">
-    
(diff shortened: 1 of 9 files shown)


resources/META-INF/studio-contribs.xml
```diff
@@ -9,7 +9,7 @@
   </extensions>
 
   <extensions defaultExtensionNs="com.android.project">
-    
+    <projectsystem implementation="io.flutter.project.FlutterProjectSystemProvider"/>
```
devoncarew: Is this change intended?

stevemessick: Yes.

PR #2050, opened 2018-04-04
tweak the message for non-renderable widgets

devoncarew:
Make the text a bit shorter (so its less likely to be truncated in narrow layouts); @scheglov.


src/io/flutter/preview/PreviewView.java
```diff
@@ -185,7 +185,7 @@ public void onFailure(RenderProblemKind kind) {
               previewArea.clear(PreviewArea.NO_WIDGET_MESSAGE);
               break;
             case NOT_RENDERABLE_WIDGET:
-              previewArea.clear("Selection does not correspond to a renderable widget");
+              previewArea.clear("Selection not a renderable widget");
```
scheglov: Some other alternatives.
"not" => "is not"?
Or maybe just "Not a renderable widget"?

devoncarew: I'll change to `is not`

PR #2053, opened 2018-04-04
MultiIconSimpleColoredComponent

jacob314:
The commit has two commits. First is a baseline forking the SimpleColoredComponent class from
IntelliJ and the second commit customizes it to support icons interleaved with text.
Interleaving icons with text is needed to show flutter properties where the description contains icon(s) where the icons are interleaved with text as in the case of the color icons shown here.
image

### DIFF
--- a/src/io/flutter/view/MultiIconSimpleColoredComponent.java
+++ b/src/io/flutter/view/MultiIconSimpleColoredComponent.java
@@ -3,7 +3,6 @@
  * Use of this source code is governed by a BSD-style license that can be
  * found in the LICENSE file.
  */
-// TODO(jacobr): which should we keep.
 /*
  * Copyright 2000-2017 JetBrains s.r.o.
  *
@@ -21,8 +20,6 @@
  */
 package io.flutter.view;
 
-
-
 import com.intellij.ide.BrowserUtil;
 import com.intellij.ide.ui.UISettings;
 import com.intellij.openapi.application.Application;
@@ -31,10 +28,7 @@ import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.util.Comparing;
 import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.util.text.StringUtil;
-import com.intellij.ui.ColoredTextContainer;
-import com.intellij.ui.Gray;
-import com.intellij.ui.JBColor;
-import com.intellij.ui.SimpleTextAttributes;
+import com.intellij.ui.*;
 import com.intellij.ui.paint.EffectPainter;
 import com.intellij.util.ui.JBInsets;
 import com.intellij.util.ui.JBUI;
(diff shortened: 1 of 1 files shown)


src/io/flutter/view/MultiIconSimpleColoredComponent.java
```diff
@@ -19,10 +18,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+// TODO(jacobr): which license or both should we keep?
```
devoncarew: Lets keep both.

src/io/flutter/view/MultiIconSimpleColoredComponent.java
```diff
@@ -224,6 +217,20 @@ public void append(@NotNull final String fragment, @NotNull final SimpleTextAttr
     revalidateAndRepaint();
   }
 
+  /**
+   * Adds the icon at the beginning of the line unlike addIcon which adds the
```
devoncarew: `unlike addIcon, which`

src/io/flutter/view/MultiIconSimpleColoredComponent.java
```diff
@@ -224,6 +217,20 @@ public void append(@NotNull final String fragment, @NotNull final SimpleTextAttr
...
+   * icon after the already appended text.
+   */
+  @Override
+  public void setIcon(@Nullable Icon icon) {
+    assert(myIcons.size() == 0);
+    if (icon == null) {
+      return;
+    }
+
+    myIcons.add(new PositionedIcon(icon,0));
```
devoncarew: I'd do a format of this file before committing it.

src/io/flutter/view/MultiIconSimpleColoredComponent.java
```diff
@@ -582,19 +589,26 @@ private static boolean needFontFallback(Font font, String text) {
...
-      curX += iconRight;
-    }
 
     Font font = getBaseFont();
 
     int baseSize = font.getSize();
     boolean wasSmaller = false;
-    for (int i = 0; i < myAttributes.size(); i++) {
+    int i = 0;
+    int j = 0;
```
devoncarew: `j` ==> `iconIndex`?

devoncarew: I don't see where j is incremented.

src/io/flutter/view/MultiIconSimpleColoredComponent.java
jacob314: done

src/io/flutter/view/MultiIconSimpleColoredComponent.java
jacob314: done.

src/io/flutter/view/MultiIconSimpleColoredComponent.java
jacob314: done

src/io/flutter/view/MultiIconSimpleColoredComponent.java
jacob314: good catch. fixed this code block. The other similar code block is called.

PR #2063, opened 2018-04-06
address a few nits

devoncarew:
- some unused imports
- a few LOG statements changed to match our other uses


src/io/flutter/FlutterInitializer.java
```diff
@@ -14,6 +14,7 @@
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.diagnostic.Logger;
```
stevemessick: Why add this?

devoncarew: oops - that was cruft left over from an experiment

Issue #1806, opened 2018-02-22
Spawning an Isolate freezes the Widget and Render Tree Inspector


sroddy:
Spawning an Isolate freezes the Widget and Render Tree Inspector, until all the spawned isolates are killed.
This is sub-optimal because inside our apps we spawn a couple of isolates at the very beginning of the app lifecycle and we almost never kill them. 

## Steps to Reproduce

1) Run these commands:
```
cd flutter/examples/layers/
flutter run -t services/isolate.dart
```
2) Click the Start button
Notice that the Flutter Ispector is frozen inside its Widgets and Render Tree views until the Isolate is killed.

## Version info

Tested on a Nexus 5 with a very recent framework build

```
[✓] Flutter (on Mac OS X 10.12.6 16G1212, locale it-IT, channel bsp_alpha)
    • Flutter version 0.1.5-pre.11 at /Users/sroddy_bs/flutter/flutter
    • Framework revision 26356ea2cf (15 hours ago), 2018-02-21 22:29:22 +0100
    • Engine revision 13cf22c284
    • Dart version 2.0.0-dev.27.0-flutter-0d5cf900b0

[✓] Android toolchain - develop for Android devices (Android SDK 27.0.0)
    • Android SDK at /Users/sroddy_bs/Library/Android/sdk
    • Android NDK at /Users/sroddy_bs/Library/Android/sdk/ndk-bundle
    • Platform android-27, build-tools 27.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-915-b08)

[✓] iOS toolchain - develop for iOS devices (Xcode 9.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 9.2, Build version 9C40b
    • ios-deploy 1.9.2
    • CocoaPods version 1.4.0

[✓] Android Studio (version 3.0)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-915-b08)

[✓] IntelliJ IDEA Community Edition (version 2017.3.4)
    • Flutter plugin version 21.2.3
    • Dart plugin version 173.4548.30
```

devoncarew:
@jacob314

devoncarew:
[...]

Issue #2016, opened 2018-03-27
in the memory and fps views, clearly call out when we're in debug mode


devoncarew:
So that users know that the data they see isn't entirely representative of profile or release mode performance.

devoncarew:
And perhaps, have them be collapsed by default when run in debug mode (if we eventually do move to a accordion style container for the performance area).

PR #2066, opened 2018-04-09
handle apps with multiple isolates in the inspector

devoncarew:
- add logic in `PerfService` (as a singleton that already has knowledge of the service protocol and events) to track the current flutter isolate (at app startup, and during full retarts)
- have it fire events for when the isolate id for the flutter app's isolate changes
- have the inspector listen to those events; fix https://github.com/flutter/flutter-intellij/issues/1806 - the inspector would assert for apps that started isolates

@jacob314, I know you have a big refactor out. If this would conflict, I can roll back the inspector portions, land the rest, and re-apply once your work lands. We will want to get your changes in soon however, so we can resolve this issue as well as https://github.com/flutter/flutter-intellij/issues/2016 (which will require some UI work in the inspector view).

### DIFF
--- a/src/io/flutter/inspector/EvalOnDartLibrary.java
+++ b/src/io/flutter/inspector/EvalOnDartLibrary.java
@@ -5,7 +5,6 @@
  */
 package io.flutter.inspector;
 
-import com.google.common.collect.Iterables;
 import com.google.gson.JsonObject;
 import com.intellij.openapi.Disposable;
 import com.intellij.openapi.diagnostic.Logger;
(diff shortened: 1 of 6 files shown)


src/io/flutter/inspector/EvalOnDartLibrary.java
```diff
@@ -32,46 +33,35 @@
 public class EvalOnDartLibrary implements Disposable {
   private String isolateId;
   private final VmService vmService;
+  @SuppressWarnings("FieldCanBeLocal") private final PerfService perfService;
```
jacob314: why the suppress warnings annotation? The field appears to be final.

src/io/flutter/perf/PerfService.java
```diff
@@ -46,22 +56,33 @@ public void connectionClosed() {
...
             public void onError(RPCError error) {
             }
 
             @Override
             public void received(Isolate isolate) {
+              // Populate flutter isolate info.
+              if (flutterIsolateRef == null) {
+                for (String extensionName : isolate.getExtensionRPCs()) {
+                  if (extensionName.startsWith("ext.flutter.")) {
+                    flutterIsolateRef = isolateRef;
```
jacob314: maybe assert that there is only one flutter isolate 
and keep the first one?
It is theoretically possible someone could add ex.flutter extensions on a background isolate.

src/io/flutter/inspector/EvalOnDartLibrary.java
devoncarew: It wants me to remove this field and instead just use a local in the constructor. I'd prefer to keep it a field so it's available for other uses :)

src/io/flutter/perf/PerfService.java
devoncarew: Currently, this would happen if there was more than one flutter view running in the app. This is possible, but not a use case that we hit currently. I'd prefer to take the first flutter view and ignore the others instead of asserting (an assert will manifest as a crash to the user).

Perhaps just adding a TODO: to note that the best effort handling of multiple flutter view can be improved in the future?

PR #2070, opened 2018-04-09
Project Import Support (#1889)

pq:
Importing Flutter (and vanilla Dart) projects "just work". 🤘 

image

image

Also makes initialization a little more multi-module friendly.

Fixes: #1889.

@devoncarew @stevemessick

### DIFF
--- a/resources/META-INF/plugin_template.xml
+++ b/resources/META-INF/plugin_template.xml
@@ -237,6 +237,8 @@
     <iconProvider implementation="io.flutter.project.FlutterIconProvider" order="first"/>
 
     <library.type implementation="io.flutter.sdk.FlutterPluginLibraryType"/>
+
+    <projectStructureDetector implementation="io.flutter.project.FlutterProjectStructureDetector"/>
   </extensions>
 
   <!-- Dart Plugin extensions -->
--- a/src/io/flutter/FlutterInitializer.java
+++ b/src/io/flutter/FlutterInitializer.java
@@ -15,6 +15,8 @@ import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.fileEditor.FileEditorManager;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.ProjectRootManager;
 import com.intellij.openapi.startup.StartupActivity;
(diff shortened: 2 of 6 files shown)


resources/META-INF/plugin_template.xml
```diff
@@ -237,6 +237,8 @@
     <iconProvider implementation="io.flutter.project.FlutterIconProvider" order="first"/>
 
     <library.type implementation="io.flutter.sdk.FlutterPluginLibraryType"/>
+
+    <projectStructureDetector implementation="io.flutter.project.FlutterProjectStructureDetector"/>
```
stevemessick: Can you give this an id attr? If possible, it might make adding an Android-specific one easier. If a bunch of others are already defined and none of them have id's then it isn't worth the trouble.

src/io/flutter/FlutterInitializer.java
```diff
@@ -130,26 +132,33 @@ public void runActivity(@NotNull Project project) {
     FlutterViewFactory.init(project);
 
     // If the project declares a Flutter dependency, do some extra initialization.
-    final PubRoot root = PubRoot.singleForProjectWithRefresh(project);
-    if (root != null && root.declaresFlutter()) {
-      // Set Android SDK.
-      if (root.hasAndroidModule(project)) {
-        ensureAndroidSdk(project);
-      }
+    for (Module module : ModuleManager.getInstance(project).getModules()) {
```
devoncarew: re: better supporting multiple modules - this is fantastic! likely a cause of a long tail of issues for users

devoncarew: Should we deprecate `PubRoot.singleForProjectWithRefresh`, or are there still valid use cases for it?

src/io/flutter/project/FlutterProjectStructureDetector.java
```diff
@@ -0,0 +1,90 @@
...
+                                               @NotNull File base,
+                                               @NotNull List<DetectedProjectRoot> result) {
+    final VirtualFile dir = VfsUtil.findFileByIoFile(parent, false);
+    if (dir != null) {
+      final PubRoot pubRoot = PubRoot.forDirectory(dir);
+      if (pubRoot != null) {
+        if (pubRoot.declaresFlutter()) {
+          result.add(new FlutterProjectRoot(parent));
+        } else {
+          // TODO(pq): consider pushing pure dart project detection down into the Dart Plugin.
```
devoncarew: /cc @alexander-doroshko

src/io/flutter/project/FlutterProjectStructureDetector.java
```diff
@@ -0,0 +1,90 @@
...
+  public String getDetectorId() {
+    return "Flutter";
+  }
+
+  @Override
+  public void setupProjectStructure(@NotNull Collection<DetectedProjectRoot> roots,
+                                    @NotNull ProjectDescriptor projectDescriptor,
+                                    @NotNull ProjectFromSourcesBuilder builder) {
+    final List<ModuleDescriptor> modules = new ArrayList<>();
+    for (DetectedProjectRoot root : roots) {
```
devoncarew: These roots are the ones that we detected?

src/io/flutter/project/FlutterProjectStructureDetector.java
```diff
@@ -0,0 +1,90 @@
...
+      modules.add(new ModuleDescriptor(root.getDirectory(), JavaModuleType.getModuleType(), Collections.emptyList()));
```
devoncarew: We could also use:

```
ModuleTypeManager.getInstance().findByID(FlutterModuleUtils.getModuleTypeIDForFlutter())
```

src/io/flutter/FlutterInitializer.java
pq: Not sure!  Will think about it...

src/io/flutter/project/FlutterProjectStructureDetector.java
pq: Yep.

src/io/flutter/project/FlutterProjectStructureDetector.java
pq: Good catch!

resources/META-INF/plugin_template.xml
pq: It doesn't look like this is done in general but I'm happy to add one if it ends up coming in handy.

Issue #2071, opened 2018-04-09
apps don't consistently resume from the PauseStart state


devoncarew:
It looks like a regression due to https://github.com/flutter/flutter-intellij/commit/7bd4f0f65293c226d6acff7f53437388feebb901.

Note, the logic to resume at start is handled by the Dart plugin, in or around:
https://github.com/JetBrains/intellij-plugins/blob/2f85a778ce22c7f2592269de32fbf4aeb825672c/Dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/VmServiceWrapper.java#L164

PR #2073, opened 2018-04-10
Fix race condition in DartVmServiceDebugProcessZ

jacob314:
Also fix race condition determining what Flutter extensions are available resulting in 
inspector toolbar buttons typically being disabled.

Also add code to be a bit paranoid about ensuring all updates to PerfService occur on the UI thread.
We could optimize that code a bit and only switch to the UI thread when about to update fields.

fixes https://github.com/flutter/flutter-intellij/issues/2071

### DIFF
--- a/src/io/flutter/utils/EventStream.java
+++ b/src/io/flutter/utils/EventStream.java
@@ -18,14 +18,14 @@ import java.util.function.Consumer;
 /**
  * Simple class for listening for a value that can be repeatedly updated.
  * <p>
- * The  benefit of this class instead of listening for events is you don't have
- * to worry about missing values posted to the stream before your code started
- * listening.
+ * The  benefit of using this class instead of listening for events is you
+ * don't have to worry about missing values posted to the stream before your
+ * code started listening which can lead to hard to find bugs.
+ *
  * The benefit of using this class over awaiting a Future is that the value can
  * update multiple times.
  * <p>
- * New values to the EventStream can be posted on any thread to listen can be
- * made from any thread.
+ * The value associated with the EventStream can be set on any thread.
  * <p>
  * The class is inspired by listen method on the Stream class in Dart.
  */
@@ -34,6 +34,10 @@ public class EventStream<T> {
   protected final HashSet<StreamSubscription<T>> subscriptions = new LinkedHashSet<>();
 
   private volatile T currentValue;
+
+  public EventStream() {
+    this(null);
+  }
 
   public EventStream(T initialValue) {
     currentValue = initialValue;
--- a/src/io/flutter/utils/EventStream.java
+++ b/src/io/flutter/utils/EventStream.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.utils;
+
+
+import com.intellij.openapi.application.ApplicationManager;
+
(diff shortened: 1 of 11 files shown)


third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
```diff
@@ -267,6 +269,39 @@ private void onConnectSucceeded(VmService vmService,
     final FlutterLaunchMode launchMode = FlutterLaunchMode.getMode(executionEnvironment);
     if (launchMode.supportsDebugConnection()) {
       myVmServiceWrapper.handleDebuggerConnected();
+
+      // TODO(jacobr): the following code is a workaround for
```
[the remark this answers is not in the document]
jacob314: done

jacob314:
I wasn't satisfied by the fragile code and boilerplate in the initial pull request so I added an "EventStream" class borrowing ideas from the Dart stream class that makes the code a bit simpler.

src/io/flutter/utils/EventStream.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Simple class for listening for a value that can be repeatedly updated.
+ * <p>
+ * The  benefit of this class instead of listening for events is you don't have
+ * to worry about missing values posted to the stream before your code started
+ * listening.
+ * The benefit of using this class over awaiting a Future is that the value can
+ * update multiple times.
+ * <p>
+ * New values to the EventStream can be posted on any thread to listen can be
```
devoncarew: some grammar fixes needed here -

src/io/flutter/utils/EventStream.java
```diff
@@ -0,0 +1,106 @@
...
+ * <p>
+ * The class is inspired by listen method on the Stream class in Dart.
+ */
+public class EventStream<T> {
+
+  protected final HashSet<StreamSubscription<T>> subscriptions = new LinkedHashSet<>();
+
+  private volatile T currentValue;
+
+  public EventStream(T initialValue) {
```
devoncarew: Perhaps also have a 

```
public EventStream() {
  currentValue = null;
}
```

constructor (and use this in the call sites that currently pass in null)?

third_party/intellij-plugins-dart/src/com/jetbrains/lang/dart/ide/runner/server/vmService/DartVmServiceDebugProcessZ.java
```diff
@@ -267,6 +269,39 @@ private void onConnectSucceeded(VmService vmService,
...
+                  });
+                }
+              }
+            });
+          }
+        }
+
+        @Override
+        public void onError(RPCError error) {
+          System.out.println("XXX got an error" + error);
```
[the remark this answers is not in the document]
devoncarew: We should remove the System.out.println here -

src/io/flutter/utils/EventStream.java
jacob314: done.

src/io/flutter/utils/EventStream.java
jacob314: cleaned up a bit. feel free to send a follow up CL with to cleanup nits and polish the docs or api. this is the minimum for a useful Stream api. As a followup when time allows I'll adopt this API in more existing locations where it is simpler than what I am doing today.

PR #2078, opened 2018-04-11
remove uses of the singleForProject/singleForModule calls

devoncarew:
Some follow up to https://github.com/flutter/flutter-intellij/pull/2070 - remove uses of the singleForProject* calls. These would explicitly not handle the case of multiple flutter modules per project, or multiple pub roots per module. This removes the methods, and re-writes their previous call sites to handle multiple modules/pub roots.

Also, remove some older init code in `ProjectOpenActivity.java`.

@pq


pq:
(Looking at the build -- you may have missed a rename somewhere?)

devoncarew:
Yup, looks like it - re-trying the build now.

PR #2080, opened 2018-04-11
Update for Canary 10

stevemessick:
@pq @devoncarew 

Update the sources (and travis) to build all supported platforms (assuming 3.0 is no longer supported).

If you want to try this at home, be aware that the sed command on MacOS is different from linux. You need to give an empty value to the in-place option:
```bash
sed -i '' 's/npw\.model\.NewModuleModel/npw.module.NewModuleModel/' flutter-studio/src/io/flutter/module/FlutterDescriptionProvider.java
```
I chose to edit the file in-place because we do a fresh check-out for each build.

An exception is thrown during project creation in Android Studio 3.2 but since that also happens when creating plain Android projects I'm ignoring it.

### DIFF
--- a/flutter-studio/src/io/flutter/module/FlutterDescriptionProvider.java
+++ b/flutter-studio/src/io/flutter/module/FlutterDescriptionProvider.java
@@ -7,7 +7,7 @@ package io.flutter.module;
 
 import com.android.tools.idea.npw.module.ModuleDescriptionProvider;
 import com.android.tools.idea.npw.module.ModuleGalleryEntry;
-import com.android.tools.idea.npw.module.NewModuleModel;
+import com.android.tools.idea.npw.model.NewModuleModel;
 import com.android.tools.idea.observable.core.OptionalValueProperty;
 import com.android.tools.idea.wizard.model.SkippableWizardStep;
 import icons.FlutterIcons;
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -7,8 +7,7 @@
       "ideaVersion": "173.4670197",
       "dartPluginVersion": "181.4203.498",
       "sinceBuild": "173.0",
-      "untilBuild": "173.*",
-      "filesToSkip": ["src/io/flutter/editor/FlutterCompletionContributor.java"]
+      "untilBuild": "173.*"
     },
     {
       "name": "IntelliJ",
@@ -17,8 +16,7 @@
       "ideaVersion": "181.4705630",
       "dartPluginVersion": "181.4203.498",
       "sinceBuild": "181.0",
-      "untilBuild": "181.*",
-      "filesToSkip": ["src/io/flutter/editor/FlutterCompletionContributor.java"]
+      "untilBuild": "181.*"
     },
     {
       "name": "IntelliJ",
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -7,7 +7,8 @@
       "ideaVersion": "173.4670197",
       "dartPluginVersion": "181.4203.498",
       "sinceBuild": "173.0",
-      "untilBuild": "173.*"
+      "untilBuild": "173.*",
+      "filesToSkip": ["src/io/flutter/editor/FlutterCompletionContributor.java"]
     },
     {
       "name": "IntelliJ",
@@ -16,7 +17,8 @@
       "ideaVersion": "181.4705630",
       "dartPluginVersion": "181.4203.498",
       "sinceBuild": "181.0",
-      "untilBuild": "181.*"
+      "untilBuild": "181.*",
+      "filesToSkip": ["src/io/flutter/editor/FlutterCompletionContributor.java"]
     },
     {
       "name": "IntelliJ",
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -5,16 +5,17 @@
       "version": "3.1",
       "ideaProduct": "android-studio",
       "ideaVersion": "173.4670197",
-      "dartPluginVersion": "173.4700",
+      "dartPluginVersion": "181.4203.498",
       "sinceBuild": "173.0",
-      "untilBuild": "173.*"
+      "untilBuild": "173.*",
+      "filesToSkip": ["src/io/flutter/editor/FlutterCompletionContributor.java"]
     },
     {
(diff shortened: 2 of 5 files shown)


product-matrix.json
```diff
@@ -5,18 +5,20 @@
       "version": "3.1",
       "ideaProduct": "android-studio",
       "ideaVersion": "173.4670197",
-      "dartPluginVersion": "173.4700",
+      "dartPluginVersion": "181.4203.498",
```
devoncarew: 3.1 is still based on 2017.3, no? The Dart plugin version here should likely be `173.4700`.

product-matrix.json
```diff
@@ -5,18 +5,20 @@
...
     {
       "name": "IntelliJ",
       "version": "3.2",
       "ideaProduct": "android-studio",
-      "ideaVersion": "173.4688006",
-      "dartPluginVersion": "173.4700",
-      "sinceBuild": "173.0",
-      "untilBuild": "173.*"
+      "ideaVersion": "181.4705630",
+      "dartPluginVersion": "181.4203.498",
```
devoncarew: In my local copy of Android Studio 3.2, I see `181.2784.29` as the Dart plugin version -

product-matrix.json
```diff
@@ -5,18 +5,20 @@
...
+      "sinceBuild": "181.0",
+      "untilBuild": "181.*",
+      "filesToSkip": ["src/io/flutter/editor/FlutterCompletionContributor.java"]
```
devoncarew: Can we include the completion contributor here? I assume it requires >= 2018.1?

tool/travis.sh
```diff
@@ -38,6 +38,10 @@ else
   # Run some validations on the repo code.
   ./bin/plugin lint
 
+if [ "$IDEA_VERSION" = "3.1" ] ; then
```
devoncarew: Can you add a comment in the script about what this line is doing?

product-matrix.json
```diff
@@ -5,18 +5,20 @@
...
     },
     {
       "name": "IntelliJ",
```
devoncarew: We should probably remove this third entry, otherwise we'll be building two plugins that declare compatibility with 2018.1.

flutter-studio/src/io/flutter/module/FlutterDescriptionProvider.java
```diff
@@ -7,7 +7,7 @@
 
 import com.android.tools.idea.npw.module.ModuleDescriptionProvider;
 import com.android.tools.idea.npw.module.ModuleGalleryEntry;
-import com.android.tools.idea.npw.module.NewModuleModel;
+import com.android.tools.idea.npw.model.NewModuleModel;
```
devoncarew: This will work in both 3.1 and 3.2?

product-matrix.json
stevemessick: Yes, that's true for me as well. But it won't compile with that version.

flutter-studio/src/io/flutter/module/FlutterDescriptionProvider.java
stevemessick: Exactly. Without adding all that machinery to the plugin tool. Which might be a more robust solution, long term.

stevemessick:
PTAL

I forgot to mention that the change to InstallSdkAction fixes a problem reported by a couple users. They were seeing an exception thrown when we tried to do a write action (modify the project model) in a read-only context (wrong thread).

product-matrix.json
devoncarew: I suspect that we'll need to compile against the version that users will have at runtime. Are these compile issues w/ APIs from the Dart plugin? If so, we can probably resolve that (push changes to those versions).

devoncarew: It looks like that version of the Dart plugin does not have the `com.jetbrains.lang.dart.ide.completion.DartCompletionExtension` class.

@pq @alexander-doroshko - the latest AS (3.2 release 10) selects v `181.2784.29` of the Dart plugin, not `181.4203.498`, the version w/ the latest changes.

devoncarew: Perhaps we can just exclude `src/io/flutter/editor/FlutterCompletionContributor.java` from this build? We were doing something similar for the 2017.1 builds.

stevemessick: Silly me, I knew that. It was added after the version of Dart plugin that works with 3.2 was created.

FWIW the 173 branch of the Dart plugin source has all the right stuff. That's what I use during development, and everything compiles. But there is no plugin distro that matches that source that is selectable for Android Studio 3.2.

There are a number of changes to the analysis server in the Dart plugin that are present in 173 but not 181.2784.29, which means any Flutter plugin we build for 3.2 won't run until the Dart plugin is updated.

stevemessick:
Summary of current status (with latest version of this PR):

AS 3.1: Builds, runs
AS 3.2: Builds, does not run due to analysis server incompatibility
IJ 2018.1: Builds, runs

To get 3.2 working we need to either roll-back some changes to the Flutter plugin that depend on new additions to the analysis server in the Dart plugin, or get a new version of the Dart plugin based on branch 181.2784 that includes the changes cherry-picked into branch 173.

devoncarew:
lgtm

> To get 3.2 working we need to either roll-back some changes to the Flutter plugin that depend on new additions to the analysis server in the Dart plugin, or get a new version of the Dart plugin based on branch 181.2784 that includes the changes cherry-picked into branch 173.

Let's plan on the 2nd option, and have the 1st option as a fallback.

alexander-doroshko:
> the latest AS (3.2 release 10) selects v 181.2784.29 of the Dart plugin, not 181.4203.498, the version w/ the latest changes

That's unfortunate. 181.2784 branch is outdated long ago, 181.4203 is the original 2018.1 release and the latest current branch is 181.4445 that is used for already released 2018.1.1. 

It might be that that Android Studio *does* use one of the latest branches but still says that it is 181.2784-compatible. I didn't check this hypothesis.

Well, if we can't influence AS to say that it is at least 181.4203-compatible then I think we may create a special Dart plugin build with all the 181-branch changes and since=until=181.2784.

PR #2081, opened 2018-04-11
Fixes to make outline preview working on Windows.

scheglov:
image

### DIFF
--- a/src/io/flutter/preview/RenderHelper.java
+++ b/src/io/flutter/preview/RenderHelper.java
@@ -311,11 +311,15 @@ class RenderThread extends Thread {
       boolean newProcess = false;
       if (myProcess == null) {
         myProcessRequest = request;
-        myProcess = new ProcessBuilder(request.testerPath,
-                                       "--non-interactive",
-                                       "--use-test-fonts",
-                                       "--packages=" + request.packages.getPath(),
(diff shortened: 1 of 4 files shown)


src/io/flutter/preview/RenderHelper.java
```diff
@@ -311,11 +311,15 @@ private void render(@NotNull RenderRequest request) {
...
       if (myProcess == null) {
         myProcessRequest = request;
-        myProcess = new ProcessBuilder(request.testerPath,
-                                       "--non-interactive",
-                                       "--use-test-fonts",
-                                       "--packages=" + request.packages.getPath(),
-                                       renderServerPath).start();
+        final ProcessBuilder processBuilder =
+          new ProcessBuilder(request.testerPath,
+                             "--enable-checked-mode",
```
devoncarew: `--enable-checked-mode` is only necessary due to the code in the flutter framework, where some platform detection happens in an assert()? If so, we should open an issue for that to see if its the best way to do things.

scheglov: I opened an issue for Flutter to comment on this.
https://github.com/flutter/flutter/issues/16470

kdy1:
Is this patch published?

Issue #1923, opened 2018-03-15
Flutter hangs for new Flutter project if not internet


timekarma:
Thanks for the feedback! If your issue is related to the Flutter framework itself,
please open an issue at
[github.com/flutter/flutter](https://github.com/flutter/flutter/issues/new).

## Steps to Reproduce

_Please tell us what you were doing and what went wrong_
**Creating a new Flutter project, if there is no internet connection, Flutter/Android Studio hangs. Only way out is kill the process & restart**

## Version info
beta
Please paste the output of running `flutter doctor -v` here (available from the command
line or from `Tools > Flutter > Flutter Doctor`). It will provide the version of the
Flutter framework as well as of the IntelliJ plugin.

stevemessick:
The `flutter create` command never finishes.
```
$ flutter create foo2
Creating project foo2...
  foo2/ios/Runner.xcworkspace/contents.xcworkspacedata (created)
  foo2/ios/Runner/Info.plist (created)
...
  foo2/.idea/workspace.xml (created)
Wrote 66 files.

Running "flutter packages get" in foo2...
Could not resolve URL "https://pub.dartlang.org".
pub get failed (69) -- attempting retry 1 in 1 second...
Could not resolve URL "https://pub.dartlang.org".
pub get failed (69) -- attempting retry 2 in 2 seconds...
Could not resolve URL "https://pub.dartlang.org".
pub get failed (69) -- attempting retry 3 in 4 seconds...
Could not resolve URL "https://pub.dartlang.org".
pub get failed (69) -- attempting retry 4 in 8 seconds...
Could not resolve URL "https://pub.dartlang.org".
pub get failed (69) -- attempting retry 5 in 16 seconds...
Could not resolve URL "https://pub.dartlang.org".
pub get failed (69) -- attempting retry 6 in 32 seconds...
Could not resolve URL "https://pub.dartlang.org".
pub get failed (69) -- attempting retry 7 in 64 seconds...
```
I had just created a project named foo1 so pub should not have needed to download anything.

Since the OP didn't include it:
```
$ flutter doctor -v
[✓] Flutter (Channel master, v0.2.1-pre.63, on Mac OS X 10.13.3 17D102, locale en-US)
    • Flutter version 0.2.1-pre.63 at /Users/messick/src/flutter/flutter
    • Framework revision c8fd99fb33 (4 days ago), 2018-03-12 01:05:23 -0700
    • Engine revision 1d0da77995
[...]

PR #2095, opened 2018-04-17
Option to create projects in “offline” mode (#1923).

pq:
Adds an option to create projects "offline".

image

To make room for the options, the help is trimmed a bit too.  Previously:

image

To follow are some smarts to auto-select based on the reachability of pub.

See: #1923

@scheglov @devoncarew


devoncarew:
Looks great!

We'll need to update the Android Studio wizard as well - that has a separate UI (/cc @stevemessick). Can you create an issue for that?

pq:
@devoncarew : opened https://github.com/flutter/flutter-intellij/issues/2099 to track AS support for offline project creation.

PR #2096, opened 2018-04-17
Auto-toggle offline mode if we’re obviously offline (#1923).

pq:
Checks to see if pub (the website) is reachable as a proxy for "are we online and should we expect flutter create online to work".  If it's not, it auto-toggles the offline option and adds a label decoration to make it hopefully not too subtle.

image

with a tooltip:

image

Online defaults to:

image

See: #1923.

@scheglov @devoncarew

### DIFF
--- a/src/io/flutter/module/settings/FlutterCreateParams.java
+++ b/src/io/flutter/module/settings/FlutterCreateParams.java
@@ -8,10 +8,21 @@ package io.flutter.module.settings;
 import org.jetbrains.annotations.NotNull;
 
 import javax.swing.*;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.Socket;
 
 public class FlutterCreateParams {
   private JCheckBox createProjectOfflineCheckBox;
   private JPanel mainPanel;
+  private JLabel infoLabel;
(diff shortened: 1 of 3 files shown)


src/io/flutter/module/settings/FlutterCreateParams.java
```diff
@@ -22,4 +33,16 @@ public boolean isOfflineSelected() {
     return createProjectOfflineCheckBox.isSelected();
   }
 
+  private static boolean isPubAvailable() {
+    // Check to see if the pub site is accessible as an indicator for whether we're online
+    // and if we should expect pub commands to succeed.
+    try {
+      new Socket(InetAddress.getByName("pub.dartlang.org"), 80);
```
scheglov: Do we need to close this socket?
Probably not very important anyway, we don't create projects often.

pq: Ah.  But why not. Will fix.

pq:
We still want `flutter create` to be more robust but for now I think this fixes things on the IDE side in the meantime.

Fixes: #1923.

PR #2098, opened 2018-04-18
build against android studio 3.2 canary 11

devoncarew:
build against android studio 3.2 canary 11:

- move the 3.1 import re-writing logic to dart code (from the shell script) - this let's the local build progress
- just build two products (down from three); we were building two products compatible with 2018.1
- update to the latest version of AS and the dart plugin versions they resolve against

We don't compile against the version of the Dart plugin that resolves in AS 3.2; we'll need to backport Dat plugin changes to that version, or add back in some reflection code to the Flutter plugin.

@stevemessick @alexander-doroshko

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,5 @@ script: ./tool/travis.sh
 # Testing product matrix; the values are generated from product-matrix.json.
 env:
   - DART_BOT=true
-  - IDEA_VERSION=3.1
-  - IDEA_VERSION=3.2
+  - IDEA_VERSION=2017.3
   - IDEA_VERSION=2018.1
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -2,32 +2,23 @@
   "list": [
     {
       "name": "IntelliJ",
-      "version": "3.1",
+      "version": "2017.3",
       "ideaProduct": "android-studio",
-      "ideaVersion": "173.4670197",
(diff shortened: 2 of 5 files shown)


devoncarew:
The good news is we only see a few compile error:

https://travis-ci.org/flutter/flutter-intellij/jobs/368349752#L931

All against missing members of the `DartAnalysisServerService` class.

devoncarew:
Yeah, this isn't necessarily for landing, but to see where we are.

We'll likely want to:
- change to the version of the dart plugin that selects against IntelliJ 2018.1
- put in some runtime detection code to disable the flutter plugin (and warn) at runtime for AS 3.2

stevemessick:
Gotcha. I wonder what is the best way to warn about the Canary incompatibility? It would be easy enough to restore the warning in #1463 FlutterStudioInitializer but I wonder if that's best.

devoncarew:
@stevemessick, I moved this PR back to building against IntelliJ 2018.1 for our 2018.1 platform support. This fixes the compilation issues w/ the Dart plugin that gets selected for AS 3.2.

This PR now:
- move the 3.1 import re-writing logic to dart code (from the shell script) - this let's the local build progress
- just build two products (down from three); we were building two products compatible with 2018.1

and is ready for another look -

(we'll need to add some logic to message users at runtime that AS 3.2 is not yet supported)

PR #2109, opened 2018-04-20
use the normal intellij background color

devoncarew:
- use the normal intellij background color for the inspector trees (otherwise we get a strange looking artifact with smaller panes that the scrollable area)
- remove the tab for "Details"
- remove a small overscroll area - I think visually this didn't pay for itself

screen shot 2018-04-20 at 10 17 10 am

### DIFF
--- a/src/io/flutter/inspector/InspectorTree.java
+++ b/src/io/flutter/inspector/InspectorTree.java
@@ -93,9 +93,6 @@ public class InspectorTree extends Tree implements DataProvider, Disposable {
     super(treemodel);
     setUI(new InspectorTreeUI());
     final BasicTreeUI ui = (BasicTreeUI)getUI();
-    if (!legacyMode) {
-      setBackground(VERY_LIGHT_GREY);
-    }
     this.detailsSubtree = detailsSubtree;
 
     setRootVisible(rootVisible);
--- a/src/io/flutter/inspector/InspectorTree.java
+++ b/src/io/flutter/inspector/InspectorTree.java
@@ -93,12 +93,17 @@ public class InspectorTree extends Tree implements DataProvider, Disposable {
     super(treemodel);
     setUI(new InspectorTreeUI());
     final BasicTreeUI ui = (BasicTreeUI)getUI();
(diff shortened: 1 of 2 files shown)


src/io/flutter/inspector/InspectorTree.java
```diff
@@ -93,9 +93,6 @@ public InspectorTree(final DefaultMutableTreeNode treemodel,
     super(treemodel);
     setUI(new InspectorTreeUI());
     final BasicTreeUI ui = (BasicTreeUI)getUI();
-    if (!legacyMode) {
-      setBackground(VERY_LIGHT_GREY);
```
jacob314: Any objections if I instead move this as an option. I think we will at least want to try having it on in some user studies to see whether it helps users understand the context for the details tree. the challenge is sometimes when you change selection in the master tree we don't want to change the root of the details tree and seeing this rectangle helps explain that.

jacob314:
lgtm with added text describing what the blank details area is before user selects a widget.

src/io/flutter/inspector/InspectorTree.java
devoncarew: Just for context, the visual artifact I was trying to address here was that the details area often looked like it wasn't taking the full scrollable space in the scrollpane. This would happen when the details didn't have much content - the bk color of the details would be white, the bk color of the scrollpane would be light gray, and it looked like the details just weren't stretched to fill the available space.

devoncarew: No objection to an option, though I don't know where this would be displayed. I'll retain the code w/ a TODO:.

devoncarew:
@jacob314, ptal. This has been updated to:
- use the regular background color for the details area; we keep the background color override (`VERY_LIGHT_GREY`) for the master tree
- add a todo to investigate why the getEmptyText() for the details area is not displaying when there are no detail nodes (this should be saying `Widget subtree of the selected Widget`)

jacob314:
lgtm.
The light grey background color wasn't useful for the details tree.

PR #2110, opened 2018-04-20
Remove "directoryPatterns" logic from Flutter module reasoning for Bazel workspace project configurations.

jwren:


jwren:
I did some more testing, I am confident enough to land it, but will continue to test.

PR #2112, opened 2018-04-20
Fix issues turning --track-widget-creation on and off.

jacob314:
Now you can turn the flag on and off and click the debug or run button
and have the app relaunch correctly.

Work around https://github.com/flutter/flutter/issues/16766
with horrible hacks to delete the build directory if the config
change.

### DIFF
--- a/src/io/flutter/run/LaunchState.java
+++ b/src/io/flutter/run/LaunchState.java
@@ -342,7 +342,6 @@ public class LaunchState extends CommandLineState {
               }
               return launchState.launch(env);
             }
-            app.getCommand().getParametersList().getList();
             final FlutterLaunchMode launchMode = FlutterLaunchMode.getMode(env);
             if (launchMode.supportsReload() && app.isStarted()) {
               // Map a re-run action to a flutter full restart.
@@ -360,18 +359,7 @@ public class LaunchState extends CommandLineState {
     }
 
     private static boolean identicalCommands(GeneralCommandLine a, GeneralCommandLine b) {
-      final List<String> paramsA = a.getParametersList().getList();
-      final List<String> paramsB = b.getParametersList().getList();
-      // TODO(jacobr): there is probably an existing helper that compares lists of strings.
-      if (paramsA.size() != paramsB.size()) {
-        return false;
-      }
-      for (int i = 0; i < paramsA.size(); ++i) {
-        if (!Objects.equals(paramsA.get(i), paramsB.get(i))) {
-          return false;
-        }
-      }
-      return true;
+      return a.getParametersList().getList().equals(b.getParametersList().getList());
     }
 
     @Nullable
--- a/src/io/flutter/run/LaunchState.java
+++ b/src/io/flutter/run/LaunchState.java
@@ -7,6 +7,7 @@ package io.flutter.run;
 
 import com.intellij.execution.*;
 import com.intellij.execution.configurations.CommandLineState;
+import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.configurations.RunProfile;
 import com.intellij.execution.configurations.RunProfileState;
 import com.intellij.execution.executors.DefaultDebugExecutor;
@@ -41,9 +42,7 @@ import io.flutter.view.OpenFlutterViewAction;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
+import java.util.*;
 
 /**
  * Launches a flutter app, showing it in the console.
@@ -238,6 +237,9 @@ public class LaunchState extends CommandLineState {
 
     @NotNull
     LaunchState getState(@NotNull Executor executor, @NotNull ExecutionEnvironment environment) throws ExecutionException;
+
+    @NotNull
+    GeneralCommandLine getCommand(ExecutionEnvironment environment, FlutterDevice device) throws ExecutionException;
   }
 
   /**
@@ -326,6 +328,21 @@ public class LaunchState extends CommandLineState {
 
         if (app != null && StringUtil.equals(app.deviceId(), selectedDeviceId)) {
           if (executorId.equals(app.getMode().mode())) {
+
+            if (!identicalCommands(app.getCommand(), launchState.runConfig.getCommand(env, app.device()))) {
+              // To be safe, relaunch as the arguments to launch have changed.
+              try {
+                // TODO(jacobr): ideally we shouldn't be synchronously waiting
+                // for futures like this but I don't see a better option.
+                // In practice this seems fine.
+                app.shutdownAsync().get();
+              }
+              catch (InterruptedException | java.util.concurrent.ExecutionException e) {
+                LOG.error(e);
+              }
+              return launchState.launch(env);
+            }
+            app.getCommand().getParametersList().getList();
             final FlutterLaunchMode launchMode = FlutterLaunchMode.getMode(env);
             if (launchMode.supportsReload() && app.isStarted()) {
               // Map a re-run action to a flutter full restart.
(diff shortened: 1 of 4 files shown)


src/io/flutter/run/LaunchState.java
```diff
@@ -326,6 +328,21 @@ protected final RunContentDescriptor doExecute(@NotNull RunProfileState state, @
 
         if (app != null && StringUtil.equals(app.deviceId(), selectedDeviceId)) {
           if (executorId.equals(app.getMode().mode())) {
+
+            if (!identicalCommands(app.getCommand(), launchState.runConfig.getCommand(env, app.device()))) {
+              // To be safe, relaunch as the arguments to launch have changed.
+              try {
+                // TODO(jacobr): ideally we shouldn't be synchronously waiting
```
devoncarew: Do we expect people to change launch config options and then do a full restart? I would expect that the new options wouldn't be applied until the next full app launch.

src/io/flutter/run/SdkRunConfig.java
```diff
@@ -84,11 +98,56 @@ public LaunchState getState(@NotNull Executor executor, @NotNull ExecutionEnviro
...
+        if (Files.exists(cachedParametersPath)) {
+          try {
+            existingJson = new String(Files.readAllBytes(cachedParametersPath), StandardCharsets.UTF_8);
+          }
+          catch (IOException e) {
+            LOG.warn("Unable to get existing json from " + cachedParametersPath);
+          }
+        }
+        if (!json.equals(existingJson)) {
+          // We don't have proof the current run is consistent with the existing run.
```
devoncarew: Here's a simpler, less elegant solution:

If the launch args contain `--track-widget-creation`, then always delete `build/snapshot_blob.bin.d.fingerprint` before a run.

Thoughts?

jacob314: unfortunately that doesn't quite work as there is also a problem currently that if you run with the flag and then without the flag you will have a bad combination of widgets tracking and not tracking locations due to over eager caching

src/io/flutter/run/LaunchState.java
jacob314: this isn't the "full restart" case.
This is if the user clicks on "Debug main.dart"
after changing the config.
Currently even though the config changes, surprisingly all they get is a "full restart".
If the user triggers an actual "full restart" with this change, they still get a "full restart" even though the config has changed.

src/io/flutter/run/LaunchState.java
```diff
@@ -342,6 +359,21 @@ protected final RunContentDescriptor doExecute(@NotNull RunProfileState state, @
       return launchState.launch(env);
     }
 
+    private static boolean identicalCommands(GeneralCommandLine a, GeneralCommandLine b) {
+      final List<String> paramsA = a.getParametersList().getList();
+      final List<String> paramsB = b.getParametersList().getList();
+      // TODO(jacobr): there is probably an existing helper that compares lists of strings.
```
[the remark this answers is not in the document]
jacob314: doh! forgot Dart and Java diverged on this.

src/io/flutter/run/SdkRunConfig.java
```diff
@@ -84,11 +98,56 @@ public LaunchState getState(@NotNull Executor executor, @NotNull ExecutionEnviro
...
+        final JsonArray jsonArray = new JsonArray();
+        for (String parameter : command.getParametersList().getList()) {
+          for (String allowedParameter : parametersToTrack) {
+            if (parameter.startsWith(allowedParameter)) {
+              jsonArray.add(new JsonPrimitive(parameter));
+              break;
+            }
+          }
+        }
+        String json = new Gson().toJson(jsonArray);
```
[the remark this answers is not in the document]
jacob314: done

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -93,14 +93,16 @@
   private FlutterDebugProcess myFlutterDebugProcess;
   private @Nullable VmService myVmService;
   private PerfService myPerfService;
+  private GeneralCommandLine myCommand;
```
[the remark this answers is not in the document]
jacob314: done. Also moved to the right place in the command list

src/io/flutter/run/LaunchState.java
devoncarew: OK, it sounds like this is here to reduce surprise where the cli launch args change, the user hits the run or debug buttons, but we recycle an existing launch with different launch args.

👍

src/io/flutter/run/LaunchState.java
```diff
@@ -326,6 +328,21 @@ protected final RunContentDescriptor doExecute(@NotNull RunProfileState state, @
...
+                // for futures like this but I don't see a better option.
+                // In practice this seems fine.
+                app.shutdownAsync().get();
+              }
+              catch (InterruptedException | java.util.concurrent.ExecutionException e) {
+                LOG.error(e);
+              }
+              return launchState.launch(env);
+            }
+            app.getCommand().getParametersList().getList();
```
devoncarew: Are we using the result of this call?

src/io/flutter/run/SdkRunConfig.java
```diff
@@ -84,11 +98,56 @@ public LaunchState getState(@NotNull Executor executor, @NotNull ExecutionEnviro
...
-      final GeneralCommandLine command = fields.createFlutterSdkRunCommand(project, device, mode);
+      final GeneralCommandLine command = getCommand(env, device);
+      {
+        // Workaround for https://github.com/flutter/flutter/issues/16766
+        // TODO(jacobr): remove once flutter tool incremental building works
+        // properly with --track-widget-creation.
+        final Path buildPath = command.getWorkDirectory().toPath().resolve("build");
+        final Path cachedParametersPath = buildPath.resolve("last_build_run.json");
+        final String[] parametersToTrack = {"--preview-dart-2", "--track-widget-creation"};
+        final JsonArray jsonArray = new JsonArray();
```
devoncarew: If we wanted to simplify this, we could just put all the launch args into a text file, do a string compare on the current args, and delete the stamp file if there were differences.

devoncarew: (looks good either way)

src/io/flutter/run/SdkRunConfig.java
```diff
@@ -84,11 +98,56 @@ public LaunchState getState(@NotNull Executor executor, @NotNull ExecutionEnviro
...
+        String existingJson = null;
+        if (Files.exists(cachedParametersPath)) {
+          try {
+            existingJson = new String(Files.readAllBytes(cachedParametersPath), StandardCharsets.UTF_8);
+          }
+          catch (IOException e) {
+            LOG.warn("Unable to get existing json from " + cachedParametersPath);
+          }
+        }
+        if (!json.equals(existingJson)) {
```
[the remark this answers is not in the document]
jacob314: done

src/io/flutter/run/LaunchState.java
jacob314: nope. removed accidental line.

src/io/flutter/run/SdkRunConfig.java
jacob314: The problem with using all the args is then we get spurious deletes if you just change the device id related args this way we don't do invasive things to the build unless you change the --track-widget-creation or --preview-dart-2 flags.
Technically we only really need to use --track-widget-creation but I'm being a bit paranoid.

src/io/flutter/run/LaunchState.java
```diff
@@ -326,6 +328,21 @@ protected final RunContentDescriptor doExecute(@NotNull RunProfileState state, @
...
         if (app != null && StringUtil.equals(app.deviceId(), selectedDeviceId)) {
           if (executorId.equals(app.getMode().mode())) {
+
+            if (!identicalCommands(app.getCommand(), launchState.runConfig.getCommand(env, app.device()))) {
+              // To be safe, relaunch as the arguments to launch have changed.
+              try {
+                // TODO(jacobr): ideally we shouldn't be synchronously waiting
+                // for futures like this but I don't see a better option.
+                // In practice this seems fine.
+                app.shutdownAsync().get();
```
devoncarew: Thinking about this, you may not actually need to shut down the old launch; launching the new one will cause the old one to be terminated by flutter_tools.

jacob314: I'm surprised the shutdown call is required but it is currently.
unfortunately the shutdown is required as otherwise we surprisingly end up with some stale state. Even if it wasn't required, clarifying the behavior by shutting down the application immediately is probably better as otherwise you would get 10 confusing seconds with an existing old app running even though a rebuild is in progress.

PR #2115, opened 2018-04-22
add a 'performance' tab to the inspector

devoncarew:
add a 'performance' tab to the inspector:
- add a performance tab to the inspector
- due to the number of tabs, move the toolbar to the line above the inspector tabs (otherwise we run out of horizontal space)
- in debug mode runs, have a warning in the tab about preferring to profile in profile mode
- in profile mode, show the regular inspector tabs as disabled, and have the performance tab selected by default (but keep the widgets/render/perf tab order the same)
- only poll for memory usage when the tab is active
- remove the user preference in the experimental section of the preferences UI
- fix https://github.com/flutter/flutter-intellij/issues/2016

screen shot 2018-04-22 at 3 22 26 pm

@jacob314 @pq

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -42,8 +42,6 @@ import com.intellij.util.ui.UIUtil;
 import icons.FlutterIcons;
 import io.flutter.FlutterBundle;
 import io.flutter.FlutterInitializer;
-import io.flutter.inspector.FPSDisplay;
-import io.flutter.inspector.HeapDisplay;
 import io.flutter.inspector.InspectorService;
 import io.flutter.run.daemon.FlutterApp;
 import io.flutter.run.daemon.FlutterDevice;
@@ -83,8 +81,9 @@ public class FlutterView implements PersistentStateComponent<FlutterViewState>, 
 
   public static final String TOOL_WINDOW_ID = "Flutter Inspector";
 
-  public static final String WIDGET_TREE_LABEL = "Widgets";
-  public static final String RENDER_TREE_LABEL = "Render Tree";
+  public static final String WIDGET_TAB_LABEL = "Widgets";
+  public static final String RENDER_TAB_LABEL = "Render Tree";
+  public static final String PERFORMANCE_TAB_LABEL = "Performance";
 
   protected final EventStream<Boolean> shouldAutoHorizontalScroll = new EventStream<>(FlutterViewState.AUTO_SCROLL_DEFAULT);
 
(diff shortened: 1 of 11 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -213,49 +201,53 @@ private void addInspector(FlutterApp app, @Nullable InspectorService inspectorSe
...
-        dashboardsPanel.add(FPSDisplay.createJPanelView(runnerTabs, app), BorderLayout.NORTH);
-        dashboardsPanel.add(HeapDisplay.createJPanelView(runnerTabs, app), BorderLayout.SOUTH);
-      }
-    }
   }
 
   private ActionCallback onTabSelectionChange(TabInfo info, boolean requestFocus, @NotNull ActiveRunnable doChangeSelection) {
-    final InspectorPanel panel = (InspectorPanel)info.getComponent();
-    panel.setVisibleToUser(true);
+    if (info.getComponent() instanceof InspectorTabPanel) {
```
jacob314: Aren't all the tabs now InspectorTabPanel objects?

src/io/flutter/view/FlutterView.java
```diff
@@ -296,16 +287,46 @@ private void addInspectorPanel(String displayName,
...
+  private void addDisabledTab(String displayName,
+                              JBRunnerTabs runnerTabs,
+                              FlutterApp app,
+                              DefaultActionGroup toolbarGroup) {
+    final JPanel panel = new JPanel(new BorderLayout());
+    final JBLabel label = new JBLabel("Widget info not available in profile mode", SwingConstants.CENTER);
+    label.setForeground(UIUtil.getLabelDisabledForeground());
+    panel.add(label, BorderLayout.CENTER);
+
+    final TabInfo tabInfo = new TabInfo(panel)
```
jacob314: idea: might as well make this placeholder tab an InspectorTabPanel just to normalize the UI?

src/io/flutter/view/InspectorPerfTab.java
```diff
@@ -0,0 +1,62 @@
...
+
+    Box labelBox = Box.createHorizontalBox();
+    labelBox.add(new JLabel("Running in debug mode"));
+    labelBox.add(Box.createHorizontalGlue());
+    labelBox.setBorder(JBUI.Borders.empty(3, 10));
+    add(labelBox);
+
+    if (app.getLaunchMode() == FlutterLaunchMode.DEBUG) {
+      labelBox = Box.createHorizontalBox();
+      labelBox.add(new JLabel("Note: for best results, re-run in profile mode"));
```
jacob314: In a follow up CL tomorrow lets make the inspector indicate that for best results you should rerun with --track-widget-creation.

src/io/flutter/view/InspectorPerfTab.java
```diff
@@ -0,0 +1,62 @@
...
+  private @NotNull FlutterApp app;
+
+  InspectorPerfTab(Disposable parentDisposable, @NotNull FlutterApp app) {
+    this.app = app;
+
+    setLayout(new BoxLayout(this, BoxLayout.Y_AXIS));
+
+    add(Box.createVerticalStrut(6));
+
+    Box labelBox = Box.createHorizontalBox();
```
pq: Make labelBox `final`?

devoncarew:
> I'm somewhat worried about this getting lost in its own tab but I understand the rationale. A couple of thoughts:

Yes, I think this improves the situation for show performance info during debug mode vs profile mode, but don't love that we loose a bit of the casual awareness of perf information.

> do we have analytics tracking tab activations (sorry, forgot to look)

I don't believe that we do; would definitely be good to have.

> might we consider re-introducing a sparkline in the toolbar? (This could be opted-into.)

I think a small sparkline in the toolbar - or the bottom 20-30px of the tool window - would be worth investigating.

src/io/flutter/view/FlutterView.java
devoncarew: All but one; we could make that one an `InspectorTabPanel`, but w/o `TabInfo` being parametrizable, I like the instanceof checks; happy to change this however.

src/io/flutter/view/FlutterView.java
devoncarew: commented above

src/io/flutter/view/InspectorPerfTab.java
devoncarew: will do

devoncarew: actually, it's reassigned below

PR #2116, opened 2018-04-22
Simplify logic calling setPubRootDirectories using

jacob314:
hasServiceExstension to ensure setPubRootDirectories is called
each time an isolate is available that supports setPubRootDirectories.

There was a bug where this method wasn't called on full restart of the app.


jacob314:
TBR as the summary tree is broken on hot restart without this as pub root directories aren't always set.

devoncarew:
lgtm

PR #2132, opened 2018-04-24
Add "Track Widget Creation" checkbox to FlutterConfigurationEditorForm

jacob314:
and
"Default to Track Widget Creation (requires Dart 2.0)"
checkbox to FlutterSettingsConfigurable.
Logic is that if you create an new FlutterConfiguration you get the value
from FlutterSettings as the default.

By default both are off so users won't be forced into running the transform.
Next step is to warn users at the bottom of the inspector that it is more awesome if they check the option.

image

image

### DIFF
--- a/src/io/flutter/sdk/FlutterSdk.java
+++ b/src/io/flutter/sdk/FlutterSdk.java
@@ -193,6 +193,10 @@ public class FlutterSdk {
       args.add("--no-preview-dart-2");
     }
 
+    if (!FlutterSettings.getInstance().isDisablePreviewDart2() && FlutterSettings.getInstance().isTrackWidgetCreation()) {
+      args.add("--track-widget-creation");
+    }
+
     if (device != null) {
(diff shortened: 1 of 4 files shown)


devoncarew:
I have some suggestions for UI changes, but feel free to disagree:

- given that we expect this setting to be ~temporary, I think we can omit it from the launch configurations themselves (otherwise, we likely would want to have two toggles)
- I would move the `Track Widget Creation` setting to under the Dart 2.0 one, and indent it slightly (see the `Organize imports on save` setting); then we won't need to explicitly say that dart 2.0 is required
- perhaps rename to `Run applications with --track-widget-creation`, and have a tooltip that explains a bit more about what it does, and why it's not on by default (allows the inspector to know locations where widgets are creates in source, and, slightly experimental)

jacob314:
I agree with those suggestions for the UI. It is simpler and plans for the successful outcome where we can enable the transformer by default.
image

PR #2133, opened 2018-04-24
Make highlight nodes shown in both trees optional.

jacob314:
The highlighting of nodes shown in both trees doesn't currently provide much value.
It is now behind an option which is unchecked by default.

Maybe I should just give up and remove it completely but I think there is a good feature hiding in this highlighting. My next step is to only highlight nodes that are actually scrolled into view in the details tree. I suspect it will become much more valuable in that case but I may just be overly attached to the highlighting.

Highlighting off looks much less busy.
image


devoncarew:
Do you think this is something that is important to get in by M24?

jacob314:
I would like to get it in to M24 as the highlighting is a bit distracting
but it isn't critical. The impact of this change is very limited and easy
to verify.

PR #2138, opened 2018-04-24
Track inspector tab selection analytics.

pq:
With this in place we'll start seeing analytics view entries like these:

* `flutter inspector/widgets`
* `flutter inspector/render tree`
* `flutter inspector/performance`

corresponding with explicit Inspector tab selections.

This will give us some window into how well used the secondary tabs are and maybe indicate whether we have discovery problems.

Follow-up from: #2115.

@devoncarew @jacob314

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -38,6 +38,7 @@ import com.intellij.ui.content.ContentManager;
 import com.intellij.ui.content.ContentManagerAdapter;
 import com.intellij.ui.content.ContentManagerEvent;
 import com.intellij.ui.tabs.TabInfo;
+import com.intellij.ui.tabs.TabsListener;
 import com.intellij.util.ui.UIUtil;
 import icons.FlutterIcons;
 import io.flutter.FlutterBundle;
@@ -86,7 +87,8 @@ public class FlutterView implements PersistentStateComponent<FlutterViewState>, 
   public static final String PERFORMANCE_TAB_LABEL = "Performance";
 
   protected final EventStream<Boolean> shouldAutoHorizontalScroll = new EventStream<>(FlutterViewState.AUTO_SCROLL_DEFAULT);
-  protected final EventStream<Boolean> highlightNodesShownInBothTrees = new EventStream<>(FlutterViewState.HIGHLIGHT_NODES_SHOWN_IN_BOTH_TREES_DEFAULT);
+  protected final EventStream<Boolean> highlightNodesShownInBothTrees =
+    new EventStream<>(FlutterViewState.HIGHLIGHT_NODES_SHOWN_IN_BOTH_TREES_DEFAULT);
 
 
   @NotNull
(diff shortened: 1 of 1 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -234,6 +236,14 @@ private void addInspector(FlutterApp app, @Nullable InspectorService inspectorSe
       }
 
       addPerformanceTab(runnerTabs, app, !hasInspectorService);
+
+      // Track inspector tab selection analytics.
+      runnerTabs.addListener(new TabsListener.Adapter() {
```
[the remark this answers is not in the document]
pq: I'm not sure this is what we want because this fires gratuitously, for example, when the tab activates.  Really we just want explicit selections (which we get with a `TabsListener`).  (Actually, we may be able to infer that from previous selection --- if it's null, then that suggests the kind of selection we want.  I'll investigate and report back.)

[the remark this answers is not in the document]
jacob314: leaving the code as addListener makes sense in that case.  Maybe add a comment explaining why.
minor nit: use a lambda expression instead of an anonymous class.

jacob314:
lgtm

pq:
@devoncarew : do these analytics entries look reasonable?

src/io/flutter/view/FlutterView.java
```diff
@@ -234,6 +233,7 @@ private void addInspector(FlutterApp app, @Nullable InspectorService inspectorSe
       }
 
       addPerformanceTab(runnerTabs, app, !hasInspectorService);
+      runnerTabs.setSelectionChangeHandler(this::onTabSelectionChange);
```
devoncarew: You may have answered this from Jacob's review - so I apologize in advance - but why are we registering this here, when we're likely already `onTabSelectionChange()` is (I assume) already being called on tab changes?

src/io/flutter/view/FlutterView.java
```diff
@@ -248,7 +248,15 @@ private ActionCallback onTabSelectionChange(TabInfo info, boolean requestFocus,
       final InspectorTabPanel panel = (InspectorTabPanel)info.getComponent();
       panel.setVisibleToUser(true);
     }
+
     final TabInfo previous = info.getPreviousSelection();
+
+    // Track analytics for explicit inspector tab selections.
+    // (The initial selection will have no previous, so we filter that out.)
+    if (previous != null) {
+      FlutterInitializer.getAnalytics().sendScreenView("flutter inspector/" + info.getText().toLowerCase());
```
devoncarew: Prefer instead `FlutterView.TOOL_WINDOW_ID.toLowerCase()`?

pq: 👍

src/io/flutter/view/FlutterView.java
pq: This registration used to happen above (line 222) and used to only apply if `hasInspectorService`; Jacob pointed out that this was a bug, so I fixed it.  As for why we're explicitly registering at all, I think we need to set the handler explicitly for the runnerTabs or we won't get the changes.  @jacob314 can maybe speak to that though.

devoncarew: Gotcha; sounds good (when I originally read this, it seemed like we were doing the selection change listening twice)

PR #2140, opened 2018-04-24
Validate project location

stevemessick:
Ensure the location for project creation does not already exist. Flutter allows re-creating a project in an existing location but doing that with the Project Creation Wizard produces really confusing results.

@devoncarew @pq


flutter-studio/src/io/flutter/project/FlutterProjectStep.java
```diff
@@ -235,6 +236,13 @@ private void updateSdkField(JTextComponent sdkEditor) {
     if (moduleName.isEmpty()) {
       return errorResult("Please enter a name for the " + getContainerName() + ".");
     }
+    //noinspection ConstantConditions
+    if (getModel().projectType().get().get() == FlutterProjectType.APP) {
+      File loc = new File(FileUtil.toSystemDependentName(getModel().projectLocation().get()), getModel().projectName().get());
```
devoncarew: `final`?

PR #2141, opened 2018-04-24
Not ready for Canary 12

stevemessick:
I don't think we're ready to "officially" support Android Studio 3.2 Canary releases yet. We can discuss it, but please try creating a new project first.

@devoncarew @pq

### DIFF
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -16,7 +16,7 @@
       "ideaProduct": "android-studio",
       "ideaVersion": "181.4729833",
       "dartPluginVersion": "181.4203.498",
-      "sinceBuild": "181.0",
+      "sinceBuild": "181.4445",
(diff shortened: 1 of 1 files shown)


stevemessick:
No, I have not had problems with window management. I'm getting Gradle upgrade requests, which if I ignore them cause exceptions, but it isn't reliably reproducible.

stevemessick:
Do you think it will be easier to triage issues when the old plugin is used in C12? Or would it be less confusing to keep them all in sync?

devoncarew:
> I'm getting Gradle upgrade requests, which if I ignore them cause exceptions

Exceptions from flutter run later? Or, in IntelliJ?

stevemessick:
Deleting this in favor of displaying a warning

PR #2146, opened 2018-04-25
minor changes to the inspector label provider

devoncarew:
minor changes to the inspector label provider:
- don't bold nodes anymore - when `--track-widget-creation` is the default, all the nodes in the main inspector tree will be from user code
- in the render tree, don't show the `child: ` prefix - most nodes have this prefix, so it doesn't add much. We still show `child 1:`, `child 2:`, ...

Also, added a todo: - when using the inspector on an app, one thing I really wanted was to have a suffix on the nodes for the widgets that were defined in user code. So, `MaterialApp` - while created in my build() method - wouldn't have a suffix. But `MyCustomPage` would, and would give me an indication for where it was defined (`lib/pages/foo_page.dart`).

@jacob314

### DIFF
--- a/src/io/flutter/view/DiagnosticsTreeCellRenderer.java
+++ b/src/io/flutter/view/DiagnosticsTreeCellRenderer.java
@@ -30,17 +30,16 @@ import java.util.regex.Pattern;
 import static io.flutter.utils.JsonUtils.getIntMember;
 
 class DiagnosticsTreeCellRenderer extends InspectorColoredTreeCellRenderer {
-  // TODO(jacobr): enable this experiment once we make the link actually
-  // clickable.
+  // TODO(jacobr): enable this experiment once we make the link actually clickable.
   private static final boolean SHOW_RENDER_OBJECT_PROPERTIES_AS_LINKS = false;
 
   private final InspectorPanel panel;
+
   /**
-   * Split text into two groups, word characters at the start of a string
-   * and all other characters. Skip an <code>-</code> or <code>#</code> between the
-   * two groups.
+   * Split text into two groups, word characters at the start of a string and all other
+   * characters. Skip an <code>-</code> or <code>#</code> between the two groups.
    */
-  private final Pattern primaryDescriptionPattern = Pattern.compile("(\\w+)[-#]?(.*)");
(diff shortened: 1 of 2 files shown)


devoncarew:
My looking, it seems like this data requires a round trip to the VM to get, so not sure of the costs of doing it.

devoncarew:
(I don't think this is generally something we need to land for M24)

src/io/flutter/view/DiagnosticsTreeCellRenderer.java
```diff
@@ -217,6 +209,9 @@ else if (!node.getDescription().isEmpty()) {
         appendText(node.getDescription(), textAttributes);
       }
 
+      // TODO(devoncarew): For widgets that are definied in the current project, we could consider
```
jacob314: I like this idea. The relative path perhaps shown in grey italics could provide some good context.
This is the sort of feature we should put behind a checkbox in the settings menu to get user feedback and provide an escape hatch if users find it annoying.

src/io/flutter/view/DiagnosticsTreeCellRenderer.java
```diff
@@ -182,28 +180,22 @@ else if (isLinkedChild || panel.currentShowNode == value) {
...
         if (node.getShowSeparator()) {
-          // Is this good?
           appendText(node.getSeparator(), SimpleTextAttributes.GRAY_ATTRIBUTES);
         }
         else {
           appendText(" ", SimpleTextAttributes.GRAY_ATTRIBUTES);
         }
       }
 
-      if (panel.isCreatedByLocalProject(node)) {
```
jacob314: change this to 
if (panel.detailsSubtree && panel.isCreatedBylocalProject(node) && !panel.isHighlightNodesShownInBothTrees) {
  textAttributes = textAttributes.derive(SimpleTextAttributes.REGULAR_BOLD_ATTRIBUTES.getStyle(), null, null, null);
}|

that way it is still clear which nodes in the details tree are created by the local project but the entire summary tree is not bold.

devoncarew: gotcha, will do

devoncarew:
Finally made it back to this PR; updated, ptal.

screen shot 2018-04-26 at 3 33 58 pm

PR #2147, opened 2018-04-25
some improvements to how we pass --profile and --release info to launches

devoncarew:
some improvements to how we pass --profile and --release info to launches:
- no longer piggy-back on the  `additionalArgs` field
- instead, make sure the `FlutterLaunchMode` field is properly populated on the ExecutionEnvironment
- use that info when creating the params to `flutter run`
- and - now that that info is properly populated - don't pass in flags like `--track-widget-creation` when not run in `--debug` launches

### DIFF
--- a/src/io/flutter/actions/RunFlutterAction.java
+++ b/src/io/flutter/actions/RunFlutterAction.java
@@ -62,19 +62,18 @@ public abstract class RunFlutterAction extends AnAction {
     final SdkRunConfig sdkRunConfig = (SdkRunConfig)configuration.clone();
     final SdkFields fields = sdkRunConfig.getFields();
     final String additionalArgs = fields.getAdditionalArgs();
-    final String runArg = "--" + myLaunchMode.getCliCommand();
-    String flavorArg = "";
+
+    String flavorArg = null;
     if (fields.getBuildFlavor() != null) {
-      flavorArg = " --flavor=" + fields.getBuildFlavor();
+      flavorArg = "--flavor=" + fields.getBuildFlavor();
     }
 
-    if (additionalArgs == null) {
-      fields.setAdditionalArgs(runArg + flavorArg);
-    }
-    else {
-      if (!additionalArgs.contains(runArg)) {
-        fields.setAdditionalArgs(additionalArgs + flavorArg + " " + runArg);
-      }
+    if (additionalArgs != null && flavorArg != null) {
+      fields.setAdditionalArgs(additionalArgs + " " + flavorArg);
+    } else if (flavorArg != null) {
+      fields.setAdditionalArgs(flavorArg);
+    } else if (additionalArgs != null) {
+      fields.setAdditionalArgs(additionalArgs);
     }
 
     final Executor executor = getExecutor(myExecutorId);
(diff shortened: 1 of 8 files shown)


src/io/flutter/actions/RunFlutterAction.java
```diff
@@ -62,19 +62,18 @@ public void actionPerformed(AnActionEvent e) {
...
     }
 
-    if (additionalArgs == null) {
-      fields.setAdditionalArgs(runArg + flavorArg);
-    }
-    else {
-      if (!additionalArgs.contains(runArg)) {
-        fields.setAdditionalArgs(additionalArgs + flavorArg + " " + runArg);
-      }
+    if (additionalArgs != null && flavorArg != null) {
```
jacob314: nit:  I think the intent is a little clearer if written like:
List<String> args = new ArrayList<>();
if (additionalArgs != null) {
  args.add(additionalArgs);
}
if (flavorArg != null) {
  args.add(flavorArg);
}
if (args.isNotEmpty()) {
  fields.setAdditionalArgs(joiner.on(" ").join(args));
}

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -193,19 +194,26 @@ else if (FlutterSettings.getInstance().isDisablePreviewDart2()) {
       args.add("--no-preview-dart-2");
     }
 
-    if (!FlutterSettings.getInstance().isDisablePreviewDart2() && FlutterSettings.getInstance().isTrackWidgetCreation()) {
-      args.add("--track-widget-creation");
+    if (flutterLaunchMode == FlutterLaunchMode.DEBUG) {
+      if (!FlutterSettings.getInstance().isDisablePreviewDart2() && FlutterSettings.getInstance().isTrackWidgetCreation()) {
+        args.add("--track-widget-creation");
```
jacob314: we might actually want to put this fix into the release. Shipping release builds with track widget location was bad.

src/io/flutter/run/LaunchState.java
```diff
@@ -119,7 +124,7 @@ private RunContentDescriptor launch(@NotNull ExecutionEnvironment env) throws Ex
       }
     }
 
-    final FlutterLaunchMode launchMode = FlutterLaunchMode.getMode(env);
+    final FlutterLaunchMode launchMode = FlutterLaunchMode.getFromEnvironment(env);
```
pq: For parallelism, maybe this could be `fromEnv(..)` like `RunMode.fromEnv(..)`?

devoncarew: 👍

src/io/flutter/actions/RunFlutterAction.java
devoncarew: sounds good

src/io/flutter/run/LaunchState.java
devoncarew: will do

src/io/flutter/sdk/FlutterSdk.java
devoncarew: I ran across this when thinking about how this might interact with launching apps in `--profile` mode; it would be a good fix for the milestone.

PR #2158, opened 2018-04-26
use the verbose flag for flutter doctor

devoncarew:
- use the verbose flag for flutter doctor

This will require an associated fix to flutter_tools.


devoncarew:
I'm going to close this PR - I think we should remove the `-v` flag from `flutter doctor`.

PR #2170, opened 2018-04-28
Set FlutterError.onError and send back exception response.

scheglov:


src/io/flutter/preview/render_server_template.txt
```diff
@@ -12,6 +12,20 @@ import 'package:flutter/material.dart';
 /// import, which widget to instantiate and render, the size of the preview
 /// canvas, etc.
 void main() async {
+  // Report errors to the clients in JSON format.
```
devoncarew: not for this PR, but we may consider setting up a task on travis to copy this file to a dart file, set up a .packages file such that the file could be analyzed correctly, and analyze the file. That would let us know about any analysis issues in the template file that we'd otherwise miss.

PR #2172, opened 2018-04-29
fix an issue where reload on save would stop working

devoncarew:
- fix an issue where reload on save would stop working

If a reload-in-save was occurring when an app shut down, the bit that indicated that we were in a reload on save wouldn't get cleared. We now properly pass the exception back through the reload result Future (by terminating all outstanding daemon calls), allowing the reload bit to be cleared.

### DIFF
--- a/src/io/flutter/run/daemon/DaemonApi.java
+++ b/src/io/flutter/run/daemon/DaemonApi.java
@@ -22,10 +22,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
-import java.util.ArrayDeque;
-import java.util.Deque;
-import java.util.LinkedHashMap;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/daemon/DaemonApi.java
```diff
@@ -135,12 +145,12 @@ else if (outputType.equals(ProcessOutputTypes.STDOUT)) {
       }
 
       @Override
-      public void processWillTerminate(ProcessEvent event, boolean willBeDestroyed) {
+      public void processWillTerminate(@NotNull ProcessEvent event, boolean willBeDestroyed) {
```
pq: Ah.  This is good.  I've been meaning to look and see if this is annotated in `2017.3`.  (I think we want that to be true?)

pq: EDIT: I took a quick look and think we're good. 👍

devoncarew: awesome, thanks for checking. I remember a bunch of annotations being introduced in 2017.3, so somewhat assume that missing annotations are not safe to start adding in...

PR #2173, opened 2018-04-29
Live preview tweaks

devoncarew:
several tweaks to live preview:
- when showing an exception in the messages view, clear previous content
- remove the `depth` field from the rendering data returned from the preview server. I believe this field was just the tree depth of the render node, and not something like the z-order, or something that could be used to compare sibling render nodes. From reading the source (for classes like Column), and experimentation previewing various widgets, I think that taking the render siblings list, reversing the order, and translating that to swing components, gives us the right z order
- give the smaller version of the preview panel slightly more space
- don't fail (assert) if we encounter off-stage widgets
- allow slightly more time for previewing (esp. useful for the first render)
- don't fail for non-fatal errors (those w/o stack traces); we probably still do want to find ways to pass these back to the user (the one I was hitting was `"A RenderFlex overflowed by 12 pixels on the right."`)

@scheglov


scheglov:
Does it work with `FloatingActionButton` without depth? It used not to work, but maybe the order or iteration has been changed since then.

devoncarew:
It does; I don't think we were reversing the returned order before? (to account for the paint order difference between flutter and swing)

scheglov:
We were reversing, I think.
Good thing that it works now without depth.

devoncarew:
I tested w/ the hello world app, with the DartConf one, and with flutter_planets.

Issue #2165, opened 2018-04-27
Extract Widget Error


brianegan:
Hey there :) Tried out the new extract Widget refactoring and ran into a bug.

## Steps to Reproduce

  1. Highlight a Widget for extraction
  2. Perform "Extract Widget" with a user-defined Keyboard Shortcut
  3. Extraction seems to work sometimes, doesn't work other times
  4. I always see an IntelliJ error, reported below

IntelliJ IDEA `2018.1.2` • Flutter plugin `io.flutter 24.0.2` • Dart plugin `181.4668.60`

Flutter 0.3.1 • channel beta • https://github.com/flutter/flutter.git
Framework • revision 12bbaba9ae (7 days ago) • 2018-04-19 23:36:15 -0700
Engine • revision 09d05a3891
Tools • Dart 2.0.0-dev.48.0.flutter-fe606f890b

## Exception

Assertion failed: Refactorings should be invoked inside transaction

```
java.lang.Throwable: Assertion failed: Refactorings should be invoked inside transaction
	at com.intellij.openapi.diagnostic.Logger.assertTrue(Logger.java:163)
	at com.intellij.refactoring.ui.RefactoringDialog.show(RefactoringDialog.java:66)
	at io.flutter.actions.ExtractWidgetAction.actionPerformed(ExtractWidgetAction.java:38)
	at com.intellij.openapi.actionSystem.ex.ActionUtil$1.run(ActionUtil.java:220)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAware(ActionUtil.java:237)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher$1.performAction(IdeKeyEventDispatcher.java:576)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.lambda$processAction$2(IdeKeyEventDispatcher.java:625)
	at com.intellij.openapi.application.TransactionGuardImpl.performUserActivity(TransactionGuardImpl.java:195)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.processAction(IdeKeyEventDispatcher.java:624)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.inInitState(IdeKeyEventDispatcher.java:479)
	at com.intellij.openapi.keymap.impl.IdeKeyEventDispatcher.dispatchKeyEvent(IdeKeyEventDispatcher.java:213)
[...]

PR #2174, opened 2018-04-29
Run ExtractWidgetAction in transaction.

scheglov:
https://github.com/flutter/flutter-intellij/issues/2165

### DIFF
--- a/src/io/flutter/actions/ExtractWidgetAction.java
+++ b/src/io/flutter/actions/ExtractWidgetAction.java
@@ -38,6 +38,11 @@ public class ExtractWidgetAction extends DumbAwareAction {
       new ExtractWidgetDialog(project, editor, refactoring).show();
     }
   }
+
+  @Override
+  public boolean startInTransaction() {
+    return true;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/actions/ExtractWidgetAction.java
```diff
@@ -38,6 +38,11 @@ public void actionPerformed(AnActionEvent event) {
       new ExtractWidgetDialog(project, editor, refactoring).show();
     }
   }
+
+  @Override
+  public boolean startInTransaction() {
+    return true;
+  }
```
pq: Well that was easy! 🎉

scheglov: :-D

devoncarew:
@scheglov, is this fix worth back-porting to the current release?

scheglov:
If we are going to do any back-porting anyway, we might add it.
Otherwise it is not worth it.

I think this happens only when errors are logged.
Otherwise it is benign.

We compute refactoring in Analysis Server, so we probably don't actually need this transaction.

PR #2179, opened 2018-04-30
Add `flutter clean` top-level action.

pq:
Adds top-level action.

image

Shows results in console:

image

@devoncarew @stevemessick

### DIFF
--- a/resources/META-INF/plugin_template.xml
+++ b/resources/META-INF/plugin_template.xml
@@ -75,6 +75,10 @@
               text="Flutter Packages Upgrade"
               description="Run 'flutter packages upgrade'"/>
       <separator/>
+      <action id="flutter.clean" class="io.flutter.actions.FlutterCleanAction"
+              text="Flutter Clean"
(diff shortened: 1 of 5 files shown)


resources/META-INF/plugin_template.xml
```diff
@@ -75,6 +75,10 @@
               text="Flutter Packages Upgrade"
               description="Run 'flutter packages upgrade'"/>
       <separator/>
+      <action id="flutter.clean" class="io.flutter.actions.FlutterCleanAction"
+              text="Flutter Clean"
+              description="Run 'flutter clean'"/>
```
stevemessick: I'm not sure if the description shows up anywhere but if it does then something more descriptive would be nice.
"Remove build artifacts" perhaps.

pq: It doesn't show anywhere as far as I can see but you make a good point.  All of the action descriptions should be more descriptive.  I've made a note to follow up on that.

PR #2187, opened 2018-05-02
Flutter Log View first steps.

pq:
First steps.

Optionally swaps out the default console view for flutter runs w/ a super-charged logging one if `FlutterLog.LOGGING_ENABLED` is set to true.  (Obviously this will change after some iteration.)

Lots of `TODO`s in the source but this puts a stake in the ground.  From here we can experiment with UI/X refinements.

image

/cc @devoncarew


pq:
@devoncarew : mind taking a quick look?

src/io/flutter/logging/FlutterLog.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class FlutterLog {
+
+  // (Temporary) flag to enable logging support.
+  public static final boolean LOGGING_ENABLED = false;
+  public static final String LOGGING_STREAM_ID = "_Logging";
+
+  private static final FlutterLog INSTANCE = new FlutterLog();
```
devoncarew: I think instead of a global singleton, we should have a single instance per app session. The singelton could live off a `FlutterApp` or a `PerfService` instance.

src/io/flutter/logging/FlutterLog.java
```diff
@@ -0,0 +1,102 @@
...
+
+  private void onEntry(@Nullable FlutterLogEntry entry) {
+    if (entry != null) {
+      for (Listener listener : listeners) {
+        listener.onEvent(entry);
+      }
+    }
+  }
+
+  public void listenToProcess(@NotNull ProcessHandler processHandler, @NotNull Disposable parent) {
```
devoncarew: We probably want to invert things here, so that the FlutterLog doesn't know how to listen to n different things, but other places translate and log events to it (service protocol log messages, flutter frame events, ...).

src/io/flutter/logging/FlutterLog.java
```diff
@@ -0,0 +1,102 @@
...
+      }
+
+      @Override
+      public void connectionClosed() {
+        onVmConnectionClosed();
+      }
+    });
+
+    // Listen for logging events (note: no way to unregister).
+    vmService.streamListen(LOGGING_STREAM_ID, VmServiceConsumers.EMPTY_SUCCESS_CONSUMER);
```
devoncarew: If we refactor so that the singleton lives off the PerfService object, we can piggy back on the existing listener there.

src/io/flutter/logging/FlutterLogEntry.java
```diff
@@ -0,0 +1,41 @@
...
+package io.flutter.logging;
+
+import com.intellij.openapi.util.text.StringUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class FlutterLogEntry {
+
+  @NotNull
+  private final String timestamp;
```
devoncarew: Perhaps use a long instead? I assume that would be something like ms since the epoch. Elsewhere, would could then convert that to a user string for rendering.

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+
+import static io.flutter.logging.FlutterLog.LOGGING_STREAM_ID;
+
+public class FlutterLogEntryParser {
+
+  // Known entry categories.
+  public static final String LOG_CATEGORY = "LOG";
```
devoncarew: Perhaps use lowercaseCamelCase for log categories, with subcategories separated by periods? So, something like `flutter.frame`, `flutter.someOtherEvent`, `stdio.stdout`, ...

That pattern would also give framework and app developers easy places to plug in (`flutter.redux`, `fooApp.pageOpen`, ...)

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -0,0 +1,99 @@
...
+        case LOGGING_STREAM_ID:
+          return parseLoggingEvent(event);
+      }
+    }
+
+    return null;
+  }
+
+  @Nullable
+  static FlutterLogEntry parseDaemonEvent(@NotNull ProcessEvent event, @Nullable Key outputType) {
```
devoncarew: My suspicion is that the daemon events aren't that interesting to app developers - that we want to give users insight into their app at runtime, not the steps that flutter_tools goes through when it launches their app.

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -0,0 +1,367 @@
...
+
+public class FlutterLogView extends JPanel implements ConsoleView, DataProvider, FlutterLog.Listener {
+
+  private final SimpleToolWindowPanel toolWindowPanel;
+
+  // TODO(pq): migrate to defining columninfo objects and then add rendering to them.
+  // see: ListTreeTableModelOnColumns use PropertiesPanel setup
+  // TODO(pq): render tooltip with complete event details.
+  private enum LogTreeColumn {
+    TIME(100, "time", String.class) {
```
devoncarew: We could consider not displaying this data; it's lower value then category / message / severity.

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -0,0 +1,367 @@
...
+    public void setTree(JTree tree) {
+      this.tree = tree;
+      treeTable = ((TreeTableTree)tree).getTreeTable();
+    }
+
+    // A temporary child counter; to be replaced w/ root.getChildCount()
+    int childCount;
+
+    public void onEvent(FlutterLogEntry entry) {
+      final MutableTreeNode root = (MutableTreeNode)getRoot();
```
devoncarew: I think the event could come in on any thread; we should only modify the model here on the EDT thread. That may address the issues below.

src/io/flutter/run/LaunchState.java
```diff
@@ -81,6 +85,16 @@ public LaunchState(@NotNull ExecutionEnvironment env,
     DaemonConsoleView.install(this, env, workDir);
   }
 
+
+  @Nullable
+  protected ConsoleView createConsole(@NotNull final Executor executor) throws ExecutionException {
+    if (FlutterLog.LOGGING_ENABLED) {
+      return new FlutterLogView();
+    }
+
```
devoncarew: perhaps add an `else` to clarify its one or the other?

src/io/flutter/logging/FlutterLog.java
devoncarew: Some other easy events we could log are GC events and frame events.

devoncarew: (the events we currently get from flutter include `Flutter.FrameworkInitialization`, `Flutter.FirstFrame`, `Flutter.Frame`)

src/io/flutter/run/LaunchState.java
pq: Done.

src/io/flutter/logging/FlutterLog.java
pq: Another benefit is centralizing disposal logic.

src/io/flutter/logging/FlutterLog.java
pq: I'm not sure we want to piggy-back on the PerfService since I think whether it is running is tied to the the `HeapDisplay`.  But maybe it shouldn't be. Anyway, I've migrated the log to live in the app and have added a TODO to consider moving this bit too.

src/io/flutter/logging/FlutterLogEntry.java
pq: Adding a TODO to reconsider.  We may make this go away altogether as discussed above.

src/io/flutter/logging/FlutterLog.java
pq: (I think this isn't right actually. Will follow-up in another PR.)

devoncarew:
awesome, will take another look

> FWIW build issues trace back to

it should be green w/ the next merge to master

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -151,6 +153,11 @@ public void onDebuggerResumed() {
     };
   }
 
+  @NotNull
+  public FlutterLog getLog() {
```
devoncarew: `getFlutterLog()`?

src/io/flutter/logging/FlutterLog.java
devoncarew: Yeah, I have no real opinion of `FlutterApp` vs. `PerService`; I was mostly interested in there being one per flutter app session (not one per Intellij instance). looks good either was -

src/io/flutter/logging/FlutterLogEntry.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.intellij.openapi.util.text.StringUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class FlutterLogEntry {
+
+  // TODO(pq): assuming we keep a timestamp, consider making it a long.
+  @NotNull
+  private final String timestamp;
```
devoncarew: I think having timestamps for the log entries is potentially useful, even if we son't display them. Just for comparison purposes and other, this should likely not be already converted to user facing text, but be in some more semantic type, like a `long` / `Date` / ...

src/io/flutter/logging/FlutterLogEntryParser.java
devoncarew: > Anyway, this we can certainly iterate on.

👍

src/io/flutter/logging/FlutterLogView.java
devoncarew: Sure; something to consider as we iterate on the UI.

src/io/flutter/run/FlutterLaunchMode.java
```diff
@@ -20,11 +22,21 @@
   RELEASE("release");
 
   private static final Key<FlutterLaunchMode> LAUNCH_MODE_KEY = Key.create("FlutterLaunchMode");
+  private static final Key<FlutterApp> APP_KEY = Key.create("FlutterApp");
 
   public static void addToEnvironment(ExecutionEnvironment env, FlutterLaunchMode mode) {
     env.putUserData(FlutterLaunchMode.LAUNCH_MODE_KEY, mode);
   }
 
+  public static void addToEnvironment(@NotNull ExecutionEnvironment env, @NotNull FlutterApp app) {
```
devoncarew: These add / from env. methods should likely be moved to `FlutterApp` (to parallel `FlutterLaunchMode` and `RunMode`).

src/io/flutter/run/daemon/FlutterApp.java
pq: Done.

src/io/flutter/logging/FlutterLogEntry.java
pq: Sounds good.  Agreed!

src/io/flutter/run/FlutterLaunchMode.java
pq: Good idea; moved.

PR #2189, opened 2018-05-02
Use 'flutter run -d=flutter-tester' to render preview.

scheglov:


src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -226,6 +226,15 @@ else if (FlutterSettings.getInstance().isDisablePreviewDart2()) {
     return new FlutterCommand(this, root.getRoot(), FlutterCommand.Type.RUN, args.toArray(new String[]{}));
   }
 
+  public FlutterCommand flutterRunOnTester(@NotNull PubRoot root, @NotNull String mainPath) {
+    final List<String> args = new ArrayList<>();
+    args.add("--machine");
+    args.add("--track-widget-creation");
```
devoncarew: We're using the kernel transformer now?

scheglov: We don't use it yet, but might switch to it (and inspector) for widget / location mapping.
It will help with constant widgets.

Remove it for now maybe and reintroduce when we really going to use it?

src/io/flutter/preview/RenderHelper.java
```diff
@@ -222,27 +216,21 @@ private void scheduleRendering() {
     this.height = height;
     this.listener = listener;
   }
+
+  Module getModule() {
+    return pubRoot.getModule(project);
```
pq: Might pubRoot be null?

src/io/flutter/preview/RenderHelper.java
```diff
@@ -325,78 +317,62 @@ private void render(@NotNull RenderRequest request) {
         }
       }
 
-      // If the is no rendering server process, start a new one.
-      boolean newProcess = false;
-      if (myProcess == null) {
+      // If there is no the rendering server process, start a new one.
```
pq: => If there is no rendering server process

src/io/flutter/sdk/FlutterSdk.java
devoncarew: You're choice, I just wasn't sure it was in use here.

/cc @jacob314 for fyi

jacob314: Either option sounds fine to me.

src/io/flutter/preview/RenderHelper.java
scheglov: Fixed.

src/io/flutter/preview/RenderHelper.java
scheglov: No, but marking fields and parameters explicitly with `@NotNull` is a good idea.

PR #2202, opened 2018-05-04
 Logger View auto-scroll.

pq:
@devoncarew


pq:
> We'll want to move to using behavior from the console view 

Agreed.  That's a little trickier but I think better long-term.

PR #2207, opened 2018-05-04
log gc events

devoncarew:
- log gc events
- show the full name for categories in the log

@pq

### DIFF
--- a/src/io/flutter/logging/FlutterLogEntryParser.java
+++ b/src/io/flutter/logging/FlutterLogEntryParser.java
@@ -11,15 +11,26 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import com.intellij.execution.process.ProcessEvent;
 import com.intellij.openapi.util.Key;
+import io.flutter.perf.HeapMonitor;
+import org.dartlang.vm.service.VmService;
 import org.dartlang.vm.service.element.Event;
+import org.dartlang.vm.service.element.IsolateRef;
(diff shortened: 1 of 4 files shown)


src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -32,6 +43,35 @@ public static FlutterLogEntry parse(@Nullable String id, @Nullable Event event)
...
           return parseLoggingEvent(event);
+        case VmService.GC_STREAM_ID:
+          final IsolateRef isolateRef = event.getIsolate();
+          final HeapMonitor.HeapSpace newHeapSpace = new HeapMonitor.HeapSpace(event.getJson().getAsJsonObject("new"));
+          final HeapMonitor.HeapSpace oldHeapSpace = new HeapMonitor.HeapSpace(event.getJson().getAsJsonObject("old"));
+
+          // TODO(devoncarew): update the VM library - timestamp is a long
+          final long timestamp;
+
+          if (event.getJson().get("timestamp") == null) {
```
pq: Do you mean:

    if (event.getJson().get("timestamp") != null)

Also, different (though maybe no better):

    if (event.getJson().has("timestamp"))

devoncarew: yes, either of these are better then the current check :)

devoncarew: updated

PR #2216, opened 2018-05-07
Add a command bar to editors that can open in a native-code editor

stevemessick:
Add "Flutter commands" at the top of an editor when it can be opened in a native code editor like Xcode or Android Studio. This reuses the menu actions to do all the heavy lifting, which is nice since the Android Studio version does not try (and fail) to re-launch itself.

It would be nice to make the just-opened file open for editing in the new window but I haven't figured out how to do that with either Xcode or IntelliJ yet. I'm pretty sure I can do it in the case where Android Studio opens a new Android Studio project, but I have less confidence when IntelliJ launches Android Studio or when either fires up Xcode.

@pq @devoncarew

### DIFF
--- a/src/io/flutter/editor/NativeEditorNotificationProvider.java
+++ b/src/io/flutter/editor/NativeEditorNotificationProvider.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.editor;
+
+import com.intellij.openapi.actionSystem.*;
+import com.intellij.openapi.editor.colors.EditorColors;
+import com.intellij.openapi.fileEditor.FileEditor;
+import com.intellij.openapi.project.DumbAware;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.Key;
+import com.intellij.openapi.util.io.FileUtil;
+import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.ui.EditorNotificationPanel;
+import com.intellij.ui.EditorNotifications;
+import com.intellij.ui.HyperlinkLabel;
+import icons.FlutterIcons;
+import org.jetbrains.annotations.NonNls;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class NativeEditorNotificationProvider extends EditorNotifications.Provider<EditorNotificationPanel> implements DumbAware {
+  private static final Key<EditorNotificationPanel> KEY = Key.create("flutter.native.editor.notification");
+
+  private final Project myProject;
+
+  public NativeEditorNotificationProvider(@NotNull Project project) {
+    myProject = project;
+  }
+
+  @NotNull
+  @Override
+  public Key<EditorNotificationPanel> getKey() {
+    return KEY;
+  }
+
+  @Nullable
+  @Override
+  public EditorNotificationPanel createNotificationPanel(@NotNull VirtualFile file, @NotNull FileEditor fileEditor) {
+    if (!file.isInLocalFileSystem()) {
+      return null;
+    }
+    VirtualFile root = findRootDir(file);
+    if (root == null) {
+      return null;
+    }
+
+    String actionName;
+    if (root.getName().equals("android")) {
+      actionName = "flutter.androidstudio.open";
+    } else if (root.getName().equals("ios")) {
+      actionName = "flutter.xcode.open";
+    } else {
+      return null;
+    }
+
+    NativeEditorActionsPanel panel = new NativeEditorActionsPanel(file, root, actionName);
+    return panel.isValidForFile() ? panel : null;
+  }
+
+  private VirtualFile findRootDir(@NotNull VirtualFile file) {
+    // Return the top-most parent of file that is a child of the project directory.
+    VirtualFile projectDir = myProject.getBaseDir();
+    VirtualFile parent = file.getParent();
+    if (projectDir.equals(parent)) {
+      return null;
+    }
+    VirtualFile root = parent;
+    while (parent != null) {
+      parent = parent.getParent();
+      if (projectDir.equals(parent)) {
+        return root;
+      }
+      root = parent;
+    }
+    return null;
+  }
+
+  class NativeEditorActionsPanel extends EditorNotificationPanel {
+    final VirtualFile myFile;
+    final VirtualFile myRoot;
+    final AnAction myAction;
+
+    NativeEditorActionsPanel(VirtualFile file, VirtualFile root, String actionName) {
+      super(EditorColors.GUTTER_BACKGROUND);
+      myFile = file;
+      myRoot = root;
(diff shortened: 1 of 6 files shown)


stevemessick:
Travis is complaining about something irrelevant to this.

src/io/flutter/editor/NativeEditorNotificationProvider.java
```diff
@@ -0,0 +1,138 @@
...
+    final AnAction myAction;
+
+    NativeEditorActionsPanel(VirtualFile file, VirtualFile root, String actionName) {
+      super(EditorColors.GUTTER_BACKGROUND);
+      myFile = file;
+      myRoot = root;
+      myAction = ActionManager.getInstance().getAction(actionName);
+
+      icon(FlutterIcons.Flutter);
+      text("Flutter commands");
```
pq: I wonder if we might have something more hand-holdy here; like:

"For the best results, edit ${platform} files in ${platform_tool}; open ${platform_tool}"

/cc @devoncarew @mit-mit

src/io/flutter/editor/NativeEditorNotificationProvider.java
```diff
@@ -0,0 +1,138 @@
...
+    return KEY;
+  }
+
+  @Nullable
+  @Override
+  public EditorNotificationPanel createNotificationPanel(@NotNull VirtualFile file, @NotNull FileEditor fileEditor) {
+    if (!file.isInLocalFileSystem()) {
+      return null;
+    }
+    VirtualFile root = findRootDir(file);
```
pq: Make all these locals final?

devoncarew:
(@stevemessick, I think the travis issue will be go away if you merge master into this branch)

stevemessick:
Here's a screen shot showing the new command bar and most of the hover-text description for the action. Sorry I couldn't get the whole thing; I'm not that coordinated. This is exactly what comes from the menu actions.

screen shot 2018-05-08 at 5 26 34 am

@mit-mit Do you want this replaced with the more helpful phrase @pq suggested? Does it replace the title, the action, or go in between them?

@pq I added the wordy final modifiers. Can you explain what value that adds?

src/io/flutter/editor/NativeEditorNotificationProvider.java
devoncarew: We may not have room for the longer text, and the action text (`Open Android module in Android Studio`) is ~pretty descriptive.

Instead of `Flutter commands`, perhaps just `Flutter`? Between that and the flutter icon, they'll have good context for who's providing the editor action bar.

devoncarew:
> Travis is complaining about something irrelevant to this.

I'd generally prefer not to merge to master when red if we can help it. Can you:
- git checkout master
- git pull
- git checkout edit-cmd-bar
- git merge master

That should address the travis issue; thanks!

src/io/flutter/editor/NativeEditorNotificationProvider.java
mit-mit: How about 'Open for editing in Android Studio' ?

pq:
> @pq I added the wordy final modifiers. Can you explain what value that adds?

If you don't do it, we all get spammed with warnings when we inspect the project.  I don't care strongly about this convention but if we decide not to heed it, we should elide it from our project  inspections.

src/io/flutter/editor/NativeEditorNotificationProvider.java
```diff
@@ -0,0 +1,138 @@
...
+      }
+      return false;
+    }
+
+    private void performAction() {
+      final Presentation present = myAction.getTemplatePresentation();
+      final DataContext context = makeContext();
+      final AnActionEvent event = AnActionEvent.createFromDataContext(ActionPlaces.EDITOR_TOOLBAR, present, context);
+      // Open Xcode or Android Studio. If already running AS then just open a new window.
+      myAction.actionPerformed(event);
```
pq: Here we'll want to update the actions so that analytics can track how they were performed. That is, we really want to know how often the actions were triggered from the banner (vs. context menu).

src/io/flutter/editor/NativeEditorNotificationProvider.java
stevemessick: @mit-mit That sounds like we are opening the file, which we're not (at least not yet). But if I can get that to happen then the change sounds good.

@devoncarew Should we make that same change in other places as well? I picked "Flutter commands" to be consistent.

stevemessick:
@devoncarew I forgot to do the rebase. That's fixed now.

@pq Sounds like it isn't so much adding value as avoiding annoyance. I don't get that annoyance since the Android Studio sources don't have that setting enabled. I see no reason to make Java more wordy than it is so I'd be glad to see the project settings changed. However, it only affects me when I send code for review.

Also @pq, the analytics change is a good catch. Do you think we can implement that in the actions themselves? I'm thinking we can introduce an interface that defines an alternative entry point for the body of the action that either does not record analytics or takes analytic info as a parameter.

PTAL

pq:
> Sounds like it isn't so much adding value as avoiding annoyance. 

I think we should unify on inspection settings and just stick to whatever we agree on.  I'm less invested in what that is than that we do it (so we can avoid churn like this 😉)

> Do you think we can implement that in the actions themselves? 

Absolutely.  I can think of a couple of ways to do this.  At the moment, I don't think we're actually sending analytics for these actions at all (yikes) so we have a lot of room to maneuver.  Maybe easiest would be to add a new interface like `ActionWithAnalytics` or something that defines a `setAnalyticsId()` (or something).  Given a handle on the action, we can override it in your banner code.  In other words something like:

```java
ActionWithAnalytics action = (ActionWithAnalytics)ActionManager.getInstance().getAction(actionName);
action.setAnalyticsId("openxcodefromfilebanner");
```

where `"openxcodefromfilebanner"` is replaced with something consistent with the our existing keys.

The existing actions will also have to have default ids set.

Anyway, this is a lot to type.  I'm happy to throw a PR together with what I'm proposing when this lands if that works for you?

stevemessick:
@pq Does getAction() copy the Action object? (I have not checked.) If it doesn't then setAnalyticsId() would change the shared object.

pq:
Um, yeah.  You're totally right.  Adding another action (and id) might be the best approach after all.

stevemessick:
Very few Flutter actions implement analytics. I think we can punt for now, and do a general analytics-clean-up step later.

Another idea might be something like this.
```dart
public interface AnalyticAction {

  void actionPerformed(AnActionEvent e, String analyticId);
}
```
Then reimplement the existing actionPerformed() to call that, which would do what actionPerformed() used to do. We have multiple superclasses for actions (AnAction, DumbAware, others?) so we'd have to mix this in for each. Clients such as this one, that want to invoke the action directly with different analytics, can call the two-arg method.

src/io/flutter/editor/NativeEditorNotificationProvider.java
devoncarew: In the other location we are actually showing 3-4 flutter_tools commands - packages get, packages upgrade, ... So I think `Flutter commands` makes sense there.

stevemessick:
@pq Sure, but we're probably going to collide. I'm trying to make the "Open in" ones pull up in the new IDE the file that is selected in the action, if any.

Issue #2205, opened 2018-05-04
Flutter Inspector Widgets Tab is empty


kunli0:
## Steps to Reproduce

Turn on the "Run application with --track-widget-creation" which I assume will bold items in the inspector and allow double clicking to automatically highlight the code where the widget was created. Upon doing so however, the Widgets tree becomes completely empty, not showing anything. 

## Version info

Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel beta, v0.3.1, on Microsoft Windows [Version 6.1.7601], locale en-US)
[√] Android toolchain - develop for Android devices (Android SDK 27.0.3)
[√] Android Studio (version 3.1)
[√] Connected devices (1 available)

Android Studio 3.1.2
Build #AI-173.4720617, built on April 13, 2018
JRE: 1.8.0_152-release-1024-b02 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Windows 7 6.1

Android Studio Flutter Plugin: Version: 24.0.1

devoncarew:
I believe this is a dup of https://github.com/flutter/flutter-intellij/issues/2029; can you try updating to the latest plugin version (just released) and see if it address the issue?

devoncarew:
If no, please re-open (/cc @jacob314).

kunli0:
Using Plugin Version: 24.0
/cc @jacob314

jacob314:
Can you try manually deleting your build directory and re-running to test whether that works around the issue?

kunli0:
Manually deleted build directory. Build directory was recreated however, same issue. This is what the Inspector looks like.

![image](https://user-images.githubusercontent.com/11678847/39646289-e92d4a44-4fa8-11e8-9eba-b50402b8ea72.png)

devoncarew:
Hmm, the inspector is connected, just no widget info is showing. Do the render tree or performance tabs show anything? Are there any errors in the Android Studio logs?

kunli0:
The render tree and performance tabs both work fine.

kunli0:
This is the only thing that looks of interests

[...]

PR #2219, opened 2018-05-07
Fix track-widget-creation on Windows.

jacob314:
https://github.com/flutter/flutter-intellij/issues/2205
There were two separate bugs in how we handled windows file paths. This isn't the ideal fix but fixes the issue without requiring a fix on the flutter engine side.

### DIFF
--- a/src/io/flutter/inspector/InspectorSourceLocation.java
+++ b/src/io/flutter/inspector/InspectorSourceLocation.java
@@ -7,6 +7,7 @@ package io.flutter.inspector;
 
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
+import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.xdebugger.XSourcePosition;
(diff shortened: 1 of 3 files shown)


src/io/flutter/inspector/InspectorSourceLocation.java
```diff
@@ -30,8 +31,13 @@ public VirtualFile getFile() {
       return parent != null ? parent.getFile() : null;
     }
 
-    // We have to strip the file:// prefix from the Dart file path for compatibility.
-    final String filePrefix = "file://";
+    // We have to strip the file:// or file:/// prefix depending on the
+    // operating to convert from paths stored as URIs to local file system
```
devoncarew: `operating system`

src/io/flutter/inspector/InspectorSourceLocation.java
```diff
@@ -30,8 +31,13 @@ public VirtualFile getFile() {
...
+    // paths.
+    // TODO(jacobr): remove this workaround after the code in package:flutter
+    // is fixed to return local file system paths instead of URIs.
```
devoncarew: Do we want to change to file system paths (on both sides of this call) or `file:/` urls?

src/io/flutter/inspector/InspectorSourceLocation.java
jacob314: done

PR #2232, opened 2018-05-09
Partial unification of inspections between android and flutter

stevemessick:
Comparing the Flutter inspection settings to those used by Android Studio reveals a few changes that I think we ought to make. I did not attempt to make every inspection the same, just the ones that cause churn.

Notable changes include relaxing the requirement to declare local variables as final and adding a suggestion to make methods static where possible. (I don't particularly like that second one but it is what JetBrains requires for Dart, and Android Studio does, too.)

@devoncarew @pq @jacob314 @scheglov @jwren @stereotype441 (Hope I got everyone)

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,12 +1,7 @@
 <component name="InspectionProjectProfileManager">
   <profile version="1.0">
     <option name="myName" value="Project Default" />
-    <inspection_tool class="CanBeFinal" enabled="false" level="WARNING" enabled_by_default="false">
-      <option name="REPORT_CLASSES" value="false" />
-      <option name="REPORT_METHODS" value="false" />
-      <option name="REPORT_FIELDS" value="true" />
-    </inspection_tool>
-    <inspection_tool class="FieldMayBeFinal" enabled="true" level="WARNING" enabled_by_default="true" />
+    <inspection_tool class="FieldMayBeStatic" enabled="true" level="WARNING" enabled_by_default="true" />
     <inspection_tool class="HardCodedStringLiteral" enabled="false" level="WARNING" enabled_by_default="false">
       <option name="ignoreForAssertStatements" value="true" />
       <option name="ignoreForExceptionConstructors" value="true" />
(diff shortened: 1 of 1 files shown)


pq:
Adding @stereotype441.

@stevemessick: How does this reconcile with the inspections for the Dart Plugin?

stevemessick:
Oops. I searched for Paul's ID, got it, clicked it, and somehow the wrong one got added. Sorry about that @paulb!

pq:
Thanks @stevemessick!  

> if you want to do a detailed analysis I'd be glad to point you to the sources for the Android Studio settings.

I'd actually vote to flip this.  I think we should[1] derive our standards from the Dart Plugin since that's the codebase we're most frequently editing in concert.

[1] And I think maybe we already are?

pq:
> Or the Dart sources?

I sometimes do this; never AS.

devoncarew:
Good to have a discussion about our styles here!

I guess a hope would be that the IDEs will pick up whatever styles we have in `.idea/inspectionProfiles/Project_Default.xml` - we won't get differences based on individual editors, and we'll ass see the same lints in-line. If people are seeing differences, we should try and track down why, and if that's something we can address.

The lints I seem to get the most value out of are the `@Nullable` / `@Notnull` ones (which, aren't under discussion here :) )

If there are issues getting IDEs to recognize out setup, perhaps we can add some cli tooling around this if automation would help? `stash` / `update` / `pop`? `git checkout /foo/dir` / `update` / `<overwrite pref files>`?

stevemessick:
I think the problem is that each project has its own .idea directory. Certainly, I have different ones depending on whether I'm using Android Studio as the base or the JetBrains repo.

Just to make this explicit, here are the paths to my .idea directories.

  1. `.../intellij-community/.idea`
  (Can't remember how I got this one, but it is less than a month old.)
  2. `.../intellij-plugins/Dart/.idea`
  This comes from following the configuration instructions for the Dart plugin.
  3. `.../flutter-intellij/.idea`
  This is the one most people have, and is created by following the Flutter instructions.
  4. `.../studio-master-dev/tools/idea/.idea`
  This one is created when following the Android Studio set-up instructions.

The last two were modified today. #3 was created when I configured a project to answer a question earlier in this thread. #4 is the one I normally work (that project has the Android Studio sources).

stevemessick:
Inspection settings I got after configuring the Dart sources in IntelliJ according to its README.md:
dart-final

I think this is ready to merge. Any objections?

Just to be clear, this isn't a discussion about style. It is just making everything consistent. The only thing that isn't is the Flutter plugin settings, and this PR corrects that issue.

pq:
I'm probably an outlier but I think I'd prefer the existing settings but won't resist if folks want to land this.  Thanks for all the follow-up @stevemessick!

Hixie:
What is your plan with this PR? (No particular rush, I'm just cleaning out the review queue.)

stevemessick:
The intent is to get some people in a room and talk through some IntelliJ-Java-coding consistency issues. We have not thought it sufficiently important to actually schedule that meeting yet. Not sure that qualifies as a "plan" but it is where we are currently.

stevemessick:
When we get back to this we also need to consider recent changes, such as the one reported in https://github.com/flutter/flutter-intellij/issues/2444#issuecomment-400700181.

Hixie:
Should we close this PR in the meantime?

Hixie:
I'm going to close this PR to get it off our review queue. Please reopen when it's relevant again.

PR #2233, opened 2018-05-09
Action analytics

pq:
An approach to getting unique analytics for the 2 flavors of "open in..." native editor actions.

I'm not married to the analytics keys ("openInXcodeFromBanner", etc. are long-winded but not inconsistent with what we have already).  Needless to say, very open to suggestions.

/cc @devoncarew @stevemessick


pq:
Oh I see a build break...  I'll fix that in a sec.

pq:
Ok, weird, I pulled in a conflicting FlutterView during a rebase.  Should be fixed now.

pq:
Sorry for the noise.  This is green now.

@devoncarew @stevemessick

src/io/flutter/actions/ActionWithAnalytics.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * An action that sends analytics when performed.
+ */
+public abstract class ActionWithAnalytics extends AnAction {
+
+  /**
+   * Template method.  Implement @performAction.
+   */
+  @Override
+  public final void actionPerformed(AnActionEvent e) {
```
stevemessick: Final methods often end up causing unexpected problems. It would be better to omit that modifier.

pq: 👍 In general I hear you.  In this case though it's a template method and I don't want it accidentally overridden since the analytics call won't happen unless they think to call super.

pq:
Ok, I've rebased and this PR is clean again.

What do you guys think?

cc @stevemessick @devoncarew

src/io/flutter/actions/ActionWithAnalytics.java
stevemessick: `@OverridingMethodsMustInvokeSuper` will do what you want and won't do what I don't want.

screen shot 2018-05-10 at 12 43 02 pm

You'll need to add a library dependency on jsr305 to enable that annotation, but I think it's worth doing. (Remember to add it to all relevant module definitions. I think there are two.)

In its current form, this method appears to contradict the instructions from JetBrains:
[https://www.jetbrains.org/intellij/sdk/docs/tutorials/action_system/working_with_custom_actions.html](https://www.jetbrains.org/intellij/sdk/docs/tutorials/action_system/working_with_custom_actions.html)

flutter-studio/src/io/flutter/actions/OpenAndroidModule.java
```diff
@@ -52,6 +52,12 @@ public void actionPerformed(AnActionEvent e) {
     openOrImportProject(projectFile, e.getProject(), sourceFile, forceOpenInNewFrame);
   }
 
+  @NotNull
+  @Override
+  public String getAnalyticsId() {
+    return "openModuleInAndroidStudio";
```
devoncarew: `openModuleInAndroidStudio` ==> `OpenModuleInAndroidStudio` (TitleCamelCase)

src/io/flutter/actions/OpenInAndroidStudioAction.java
```diff
@@ -39,8 +38,14 @@ public void update(AnActionEvent event) {
     presentation.setVisible(enabled);
   }
 
+  @NotNull
   @Override
-  public void actionPerformed(AnActionEvent e) {
+  public String getAnalyticsId() {
+    return "openInAndroidStudio";
```
devoncarew: `OpenInAndroidStudio`

src/io/flutter/actions/OpenInXcodeAction.java
```diff
@@ -141,8 +140,14 @@ public void update(AnActionEvent event) {
     }
   }
 
+  @NotNull
   @Override
-  public void actionPerformed(AnActionEvent e) {
+  public String getAnalyticsId() {
+    return "openInXcode";
```
devoncarew: OpenInXcode

src/io/flutter/editor/NativeEditorNotificationProvider.java
```diff
@@ -17,13 +17,30 @@
...
 import org.jetbrains.annotations.Nullable;
 
 public class NativeEditorNotificationProvider extends EditorNotifications.Provider<EditorNotificationPanel> implements DumbAware {
   private static final Key<EditorNotificationPanel> KEY = Key.create("flutter.native.editor.notification");
 
+  private static final ActionWithAnalytics OPEN_IN_XCODE_ACTION = new OpenInXcodeAction() {
+    @NotNull
+    @Override
+    public String getAnalyticsId() {
+      return "openInXcodeFromBanner";
```
devoncarew: We don't have an existing pattern for this; perhaps `OpenInXcode.banner`?

src/io/flutter/editor/NativeEditorNotificationProvider.java
```diff
@@ -17,13 +17,30 @@
...
+    }
+  };
+
+  private static final ActionWithAnalytics OPEN_IN_ANDROID_STUDIO_ACTION = new OpenInXcodeAction() {
+    @NotNull
+    @Override
+    public String getAnalyticsId() {
+      return "openInAndroidStudioFromBanner";
```
devoncarew: `OpenInAndroidStudio.banner`?

src/io/flutter/actions/ActionWithAnalytics.java
devoncarew: I don't have an opinion here, but we may want to move discussion out of a PR; perhaps when you're in the office Monday Steve?

src/io/flutter/editor/NativeEditorNotificationProvider.java
pq: 👍

src/io/flutter/actions/ActionWithAnalytics.java
pq: Sure.  We can discuss and rejigger.

PR #2237, opened 2018-05-09
log view opt-in preference

pq:
image

/cc @scheglov @devoncarew


devoncarew:
Do we have the name 'Flutter Log' in the UI for the console? If no, is there a place we could expose it? Or, place with the text here?

`Replace the Run and Debug console view with a a logging view`?
`Show Run console output using a Flutter logging view`?

pq:
Good call.  We don't currently have any say about the view label (and I'm not sure we can).  I'll investigate if we can/should.  In the meantime, I'll switch to:

> Show Run console output using a Flutter logging view

Thanks!

👍

PR #2239, opened 2018-05-09
Improve work of preview on Windows.

scheglov:
1. Distinction between initial start timeout and render timeout.
2. Set `FLUTTER_TEST` to `true` (required on Windows).
3. Don't call render listener for requests that were made inactive be another request.

There are still two issues on Windows.

1. The current beta does not have required `--run-forever`.
2. Hot reload does not really work - `flutter_tester.exe` (I think) keeps the `build\app.dill` file locked, so you cannot write into it during first hot reload.

### DIFF
--- a/src/io/flutter/preview/RenderHelper.java
+++ b/src/io/flutter/preview/RenderHelper.java
@@ -352,6 +352,9 @@ class RenderThread extends Thread {
       if (myApp == null) {
         final FlutterCommand command = request.flutterSdk.flutterRunOnTester(request.pubRoot, renderServerPath);
         final GeneralCommandLine commandLine = command.createGeneralCommandLine(request.project);
+
+        // Windows is not a supported Flutter target platform.
+        // Set FLUTTER_TEST to force using Android (everywhere, not just on Windows)
         commandLine.getEnvironment().put("FLUTTER_TEST", "true");
 
         final FlutterApp app = FlutterApp.start(
--- a/src/io/flutter/preview/RenderHelper.java
+++ b/src/io/flutter/preview/RenderHelper.java
@@ -8,6 +8,7 @@ package io.flutter.preview;
 import com.google.common.io.Resources;
 import com.google.common.util.concurrent.Uninterruptibles;
 import com.google.gson.JsonObject;
+import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.runners.ExecutionEnvironment;
 import com.intellij.openapi.module.Module;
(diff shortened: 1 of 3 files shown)


src/io/flutter/preview/RenderHelper.java
```diff
@@ -331,13 +351,16 @@ private void render(@NotNull RenderRequest request) {
       // Wait for it to start.
       if (myApp == null) {
         final FlutterCommand command = request.flutterSdk.flutterRunOnTester(request.pubRoot, renderServerPath);
+        final GeneralCommandLine commandLine = command.createGeneralCommandLine(request.project);
+        commandLine.getEnvironment().put("FLUTTER_TEST", "true");
```
devoncarew: Could you add a comment about why we need this?

a-siva:
Yes, the new engine code keeps a lock on the file until the main isolate goes away, this was done to ensure that all child isolates get the same version of the dill file as the parent isolate.
/cc @chinmaygarde 

Can you clarify your use case and why you are trying to overwrite the app.dill while the isolate is running.

chinmaygarde:
The various snapshots used by an isolate are inherited by all child isolates launched by that parent isolate. Child isolates may be launched by `Isolate.spawn` and related functions (typically called by the `compute` method in Foundation). Previously, the snapshots were either stored in their entirety in memory or referenced as paths. Now, we have moved to a model where we hold on to a mapping to the various snapshots and also reference the assets directory. The assets directory reference is probably what is causing you issues on Windows. In the future, we will make the mappings private to the process as well. Having child isolates use snapshots incompatible with their parents in undefined behavior and we are trying to make it harder for embedder to get into this situation.

My suggestion would be to put the updated `.dill` file in a temporary directory and specify an asset in that directory to the engine for use as an isolate snapshot.

scheglov:
On Mac I can both overwrite and remove `app.dill`, so it seems to be a Windows feature that locked file is locked by named.

I don't understand what you mean under "specify an asset in that directory to the engine for use as an isolate snapshot". What asset?

There are also `app.dill.incremental.dill` files, are these also locked and cannot be overwritten?

chinmaygarde:
cc @aam & @rmacnak-google 

I guess we are unsure what "preview" does. Can you elaborate on your use case?

scheglov:
https://medium.com/flutter-io/flutter-outline-hot-reload-and-the-implementation-of-a-live-widget-preview-69abd39aa3bb

And also running integration tests without a device connected.

PR #2242, opened 2018-05-11
Update the action label depending on selection

stevemessick:
Update the label (and hover-text description) of the action for "Open in Android Studio" based on whether a file is selected (or open for editing) or not. This implements the suggestion made by @mit-mit in #2216.

Also added the dep for jsr305. It is distributed with IJ/AS so we might as well use it.

@devoncarew

### DIFF
--- a/flutter-intellij-community.iml
+++ b/flutter-intellij-community.iml
@@ -44,5 +44,6 @@
     <orderEntry type="library" name="snakeyaml" level="project" />
     <orderEntry type="library" name="Dart SDK" level="project" />
     <orderEntry type="library" name="Dart Packages" level="project" />
+    <orderEntry type="library" name="jsr305" level="project" />
   </component>
 </module>
--- a/flutter-intellij-community.iml
+++ b/flutter-intellij-community.iml
@@ -44,6 +44,5 @@
     <orderEntry type="library" name="snakeyaml" level="project" />
     <orderEntry type="library" name="Dart SDK" level="project" />
     <orderEntry type="library" name="Dart Packages" level="project" />
-    <orderEntry type="library" name="jsr305" level="project" />
   </component>
 </module>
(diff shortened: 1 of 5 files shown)


src/io/flutter/FlutterBundle.properties
```diff
@@ -162,3 +162,8 @@ old.android.studio.message=Your path contains Android Studio 2.x,\n\
   which does not support Flutter.\n\n\
   You can change it using the flutter command-line tool via:\n\
   flutter\u00a0config\u00a0--android-studio-dir\u00a0{0}path{0}to{0}Android\u00a0Studio\u00a03.x
+
+flutter.androidstudio.open.module.text=Open Android module in Android Studio
+flutter.androidstudio.open.module.description=Open Android Studio in a new window to edit the top-level Android project
+flutter.androidstudio.open.file.text=Open for Editing in Android Studio
+flutter.androidstudio.open.file.description=Open Android Studio in a new window to edit this file
```
scheglov: FWIW, `*this* file` is inconsistent with `*the* top-level Android Project`.

src/io/flutter/editor/NativeEditorNotificationProvider.java
```diff
@@ -42,29 +41,38 @@ public EditorNotificationPanel createNotificationPanel(@NotNull VirtualFile file
...
+  private EditorNotificationPanel createPanelForAction(VirtualFile file, VirtualFile root, String actionName) {
+    if (actionName == null) {
+      return null;
+    }
+    NativeEditorActionsPanel panel = new NativeEditorActionsPanel(file, root, actionName);
+    return panel.isValidForFile() ? panel : null;
+  }
 
-    final String actionName;
+  private static String getActionName(VirtualFile root) {
```
devoncarew: can `root` be null? If no, consider a `@NotNull` annotation

flutter-intellij-community.iml
```diff
@@ -44,5 +44,6 @@
     <orderEntry type="library" name="snakeyaml" level="project" />
     <orderEntry type="library" name="Dart SDK" level="project" />
     <orderEntry type="library" name="Dart Packages" level="project" />
+    <orderEntry type="library" name="jsr305" level="project" />
```
devoncarew: I don't see these annotations used below?

devoncarew: If not, consider reverting this dep (and the one below).

src/io/flutter/editor/NativeEditorNotificationProvider.java
stevemessick: Good catch. Thanks.

flutter-intellij-community.iml
stevemessick: Done.

PR #2247, opened 2018-05-14
AS canary 14 mods

stevemessick:
Here are the changes required to get our plugin to build with Android Studio canary 14.

I still have some debugging to do to see exactly why hot reload is broken.

@devoncarew @pq

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,5 +23,5 @@ script: ./tool/travis.sh
 # Testing product matrix; the values are generated from product-matrix.json.
 env:
   - DART_BOT=true
-  - IDEA_VERSION=2017.3
-  - IDEA_VERSION=2018.1
+  - CI=true IDEA_VERSION=2017.3
+  - CI=true IDEA_VERSION=2018.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,5 +23,5 @@ script: ./tool/travis.sh
 # Testing product matrix; the values are generated from product-matrix.json.
 env:
   - DART_BOT=true
-  - CI=true IDEA_VERSION=2017.3
-  - CI=true IDEA_VERSION=2018.1
+  - IDEA_VERSION=2017.3
+  - IDEA_VERSION=2018.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,5 +23,4 @@ script: ./tool/travis.sh
 # Testing product matrix; the values are generated from product-matrix.json.
 env:
   - DART_BOT=true
-  - IDEA_VERSION=2017.3
-  - IDEA_VERSION=2018.1
+  - IDEA_VERSION=2018.1.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,5 @@ script: ./tool/travis.sh
 # Testing product matrix; the values are generated from product-matrix.json.
 env:
   - DART_BOT=true
-  - IDEA_VERSION=2018.1.2
+  - IDEA_VERSION=2017.3
+  - IDEA_VERSION=2018.1
--- a/flutter-intellij-community.iml
+++ b/flutter-intellij-community.iml
@@ -22,16 +22,16 @@
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
     <orderEntry type="module" module-name="Dart-community" />
-    <orderEntry type="module" module-name="platform-api" />
-    <orderEntry type="module" module-name="lang-api" />
-    <orderEntry type="module" module-name="lang-impl" />
-    <orderEntry type="module" module-name="xdebugger-api" />
-    <orderEntry type="module" module-name="xdebugger-impl" />
-    <orderEntry type="module" module-name="testFramework" scope="TEST" />
+    <orderEntry type="module" module-name="intellij.platform.ide" />
+    <orderEntry type="module" module-name="intellij.platform.lang" />
+    <orderEntry type="module" module-name="intellij.platform.lang.impl" />
+    <orderEntry type="module" module-name="intellij.platform.debugger" />
+    <orderEntry type="module" module-name="intellij.platform.debugger.impl" />
+    <orderEntry type="module" module-name="intellij.platform.testFramework" scope="TEST" />
     <orderEntry type="library" name="Velocity" level="project" />
-    <orderEntry type="module" module-name="smRunner" />
-    <orderEntry type="module" module-name="xml" />
-    <orderEntry type="module" module-name="idea-ui" />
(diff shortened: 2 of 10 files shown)


stevemessick:
This is now ready for review. Got a clean build on travis and it builds locally, in both dev and release modes. It installs into Android Studio and IntelliJ although I have not tested it (and suspect the hot reload issue I observed earlier is still there). 

The changes to the .iml files do not affect travis but will cause problems for development using old versions of IntelliJ.

.travis.yml
```diff
@@ -23,5 +23,5 @@ script: ./tool/travis.sh
 # Testing product matrix; the values are generated from product-matrix.json.
 env:
   - DART_BOT=true
-  - IDEA_VERSION=2017.3
-  - IDEA_VERSION=2018.1
+  - CI=true IDEA_VERSION=2017.3
+  - CI=true IDEA_VERSION=2018.1
```
pq: Curious what `CI=true` does?

flutter-intellij-community.iml
```diff
@@ -42,7 +42,25 @@
...
     </orderEntry>
     <orderEntry type="library" name="snakeyaml" level="project" />
-    <orderEntry type="library" name="Dart SDK" level="project" />
-    <orderEntry type="library" name="Dart Packages" level="project" />
+    <orderEntry type="library" exported="" name="Dart SDK" level="project" />
+    <orderEntry type="library" exported="" name="Dart Packages" level="project" />
+    <orderEntry type="module-library">
+      <library>
+        <CLASSES>
+          <root url="jar://$APPLICATION_HOME_DIR$/lib/miglayout-swing-5.0.jar!/" />
```
pq: This looks a little brittle to me.  If we're going to depend on a specific jar version would it be safer to put it in a deps dir like was done for `weberknecht-0.1.5.jar`?

tool/plugin/lib/plugin.dart
```diff
@@ -530,24 +530,44 @@ class BuildCommand extends ProductCommand {
...
+        newSource = newSource.replaceAll(
+          'import java.awt.*',
+          'import javax.swing.*',
+        );
+        newSource = newSource.replaceAll('public Image', 'public Icon');
+        newSource =
+            newSource.replaceAll('IconUtil.toImage', ''); // Leaves parens
+        processedFile1.writeAsStringSync(newSource);
+        processedFile2 = new File(
+            'flutter-studio/src/io/flutter/project/ChoseProjectTypeStep.java');
```
pq: Aside: should this file get renamed to `ChooseProjectTypeStep`?

.travis.yml
stevemessick: CI identifies the run as a travis CI run to the plugin tool. It is used during the build to auto-edit some files on travis, so no one has to hack the plugin.dart code to make a local build work.

flutter-intellij-community.iml
stevemessick: That would mean distributing the jar with the plugin rather than using the jar distributed with the IDE. @scheglov How was this dependency defined prior to 2018.3?

tool/plugin/lib/plugin.dart
stevemessick: Yes, but fixing bad file names is outside the scope of this PR :)

.travis.yml
pq: Cool.  Thanks for the info!

tool/plugin/lib/plugin.dart
pq: 😃

flutter-intellij-community.iml
pq: @devoncarew may have ideas but it was just an observation. I'm fine with whatever.

devoncarew: > @devoncarew may have ideas but it was just an observation. I'm fine with whatever.

I'd want to make sure that this import works for AS 3.1 / 3.2  and IntelliJ 2017.3 and 2018.1, on the bot and for users in the wild.

stevemessick:
Got the weird problems sorted out. Fixed a compile issue that caused an exception in Canary 15. The build still uses Canary 14. Deleted the mig-layout dependency and the CI parameter. Also fixed a problem that caused the build to use the wrong source dependencies in some situations.

PTAL, I think this one is ready to go.

devoncarew:
> Also fixed a problem that caused the build to use the wrong source dependencies in some situations.

awesome!

devoncarew:
@stevemessick, looks like travis is failing on a unneeded `@override`: https://travis-ci.org/flutter/flutter-intellij/jobs/381881343#L943

flutter-intellij-community.iml
```diff
@@ -42,7 +42,7 @@
       </library>
     </orderEntry>
     <orderEntry type="library" name="snakeyaml" level="project" />
-    <orderEntry type="library" name="Dart SDK" level="project" />
-    <orderEntry type="library" name="Dart Packages" level="project" />
+    <orderEntry type="library" exported="" name="Dart SDK" level="project" />
```
devoncarew: Is the `exported=""` necessary here? likely harmless w/ or w/o it?

tool/plugin/lib/plugin.dart
```diff
@@ -490,7 +490,7 @@ class BuildCommand extends ProductCommand {
         separator('Release mode (--release) implies --unpack');
       }
       if (!await performReleaseChecks(this)) {
-        return 1;
+        //return 1;
```
devoncarew: This should still return `1` (to fail)?

stevemessick:
That's fixed now, @devoncarew . The override is only valid for Canary 15.

tool/plugin/lib/plugin.dart
```diff
@@ -530,24 +531,46 @@ class BuildCommand extends ProductCommand {
       }
 
       // TODO(devoncarew): Remove this when we no longer support AS 3.1.
-      var processedFile = new File(
-          'flutter-studio/src/io/flutter/module/FlutterDescriptionProvider.java');
-      var oldSource;
+      var processedFile1, processedFile2, oldSource1, oldSource2, newSource;
```
devoncarew: If we end up with more of these, we'll likely want to wrap in a class, and have something along the lines off a list of them that we can `appy()` and `revert()`.

tool/plugin/lib/plugin.dart
stevemessick: Good catch. Too much git commanding. Fixed.

flutter-intellij-community.iml
stevemessick: No idea. I didn't edit this file by hand.

PR #2257, opened 2018-05-18
Improve logger daemon event handling.

pq:
Still a work in progress but this improves logging output significantly by properly parsing daemon API events (no more observatory spam! 🎉 ).

image

Work on the TODOs to follow.

/cc @jacob314

### DIFF
--- a/src/io/flutter/logging/FlutterLogEntryParser.java
+++ b/src/io/flutter/logging/FlutterLogEntryParser.java
@@ -67,6 +67,8 @@ public class FlutterLogEntryParser {
   @VisibleForTesting
   @Nullable
   public static FlutterLogEntry parseDaemonEvent(@NotNull String eventText) {
+    // TODO(pq): restructure parsing to ensure DaemonEvent messages are only parsed once
+    // (with daemon JSON going into one stream and regular log messages going elsewhere)
     stdoutParser.appendOutput(eventText);
     for (String line : stdoutParser.getAvailableLines()) {
       if (DaemonApi.parseAndValidateDaemonEvent(line.trim()) != null) {
--- a/src/io/flutter/logging/FlutterLogEntryParser.java
+++ b/src/io/flutter/logging/FlutterLogEntryParser.java
@@ -6,12 +6,13 @@
 package io.flutter.logging;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
 import com.intellij.execution.process.ProcessEvent;
+import com.intellij.execution.ui.ConsoleViewContentType;
 import com.intellij.openapi.util.Key;
 import io.flutter.perf.HeapMonitor;
+import io.flutter.run.daemon.DaemonApi;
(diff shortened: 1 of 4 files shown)


src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -63,33 +67,15 @@ static FlutterLogEntry parseDaemonEvent(@NotNull ProcessEvent event, @Nullable K
...
-    try {
-      // TODO(pq): there's lots to parse; for now, just extract a message value.
-      final JsonObject json = new JsonParser().parse(text).getAsJsonObject();
-      final JsonObject params = json.get("params").getAsJsonObject();
-
-      if (params.has("message")) {
-        return new FlutterLogEntry(timestamp(), TOOLS_CATEGORY, params.get("message").getAsJsonPrimitive().getAsString());
+    stdoutParser.appendOutput(eventText);
+    for (String line : stdoutParser.getAvailableLines()) {
+      if (DaemonApi.parseAndValidateDaemonEvent(line.trim()) != null) {
```
jacob314: It seems like we are now parsing the DaemonEvent messages a couple times (here and separately when we actually do something with it). It would be cleaner to instead layer things so the DaemonEvent messages are only parsed once with daemon JSON going into one stream and regular log messages going somewhere else.

src/io/flutter/run/daemon/DaemonApi.java
```diff
@@ -217,7 +217,7 @@ private Command takePending(int id) {
    * Parse the given string; if it is valid JSON - and a valid Daemon message - then return
    * the parsed JsonObject.
    */
-  static JsonObject parseAndValidateDaemonEvent(String message) {
+  public static JsonObject parseAndValidateDaemonEvent(String message) {
```
jacob314: it would be nice to keep this private. see comment above about only having to call it once per message.

src/io/flutter/logging/FlutterLogEntryParser.java
pq: Good idea.  Since that will likely touch a few other places, I'll add a TODO and loop back.

src/io/flutter/run/daemon/DaemonApi.java
pq: 👍

jacob314:
lgtm

pq:
@devoncarew: @jacob314 and I had a good conversation about how this might all get restructured (assuming we stick with this approach); let's discuss next week.  In the meantime, this is at least visibly a step (if only provisionally) in the right direction. 👍

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -27,6 +28,9 @@
   private static final NumberFormat nf = new DecimalFormat();
   private static final NumberFormat df1 = new DecimalFormat();
 
+  // TODO(pq): w/ state, we should now make the parser not static.
+  private static final StdoutJsonParser stdoutParser = new StdoutJsonParser();
```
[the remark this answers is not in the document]
pq: cool.  thanks for confirming.  on my list for today. 👍

PR #2259, opened 2018-05-21
Hyperlink support for log messages.

pq:
Adds basic hyperlink support to log message cells.

  * adds mouse event handling to the log tree
  * adds flutter console hyperlink handling

image

/cc @devoncarew

### DIFF
--- a/src/io/flutter/logging/FlutterLogView.java
+++ b/src/io/flutter/logging/FlutterLogView.java
@@ -7,6 +7,7 @@ package io.flutter.logging;
 
 import com.intellij.execution.filters.Filter;
 import com.intellij.execution.filters.HyperlinkInfo;
+import com.intellij.execution.filters.OpenFileHyperlinkInfo;
 import com.intellij.execution.process.ProcessHandler;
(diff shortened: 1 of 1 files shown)


src/io/flutter/logging/FlutterLogView.java
```diff
@@ -428,8 +459,55 @@ public FlutterLogTreeTable(@NotNull FlutterLogTreeTableModel model) {
     @SuppressWarnings("EmptyMethod")
     @Override
     public TableCellRenderer getCellRenderer(int row, int column) {
-      // TODO(pq): add cell renderer
+      // TODO(pq): do this more robustly
+      if (column == 2) {
```
devoncarew: What's the state of the art here? Do other table models have hard-coded column ==> cell renders logic? If so, a named constant here may help this code when other people refacgtor.

pq: 👍 Looking into just this.  There are a few approaches in the wild but still need to understand the ramifications.  Needless to say, definitely something better than `column == 2` in short order for sure!

PR #2267, opened 2018-05-23
Log Tree popup (and clear entry action)

pq:
Adds a popup menu to the logging tree.

Implements a clear entry action.

image

(FWIW: on the fence about Clear vs. Cut vs. Delete.  "Clear all" is implemented in the default console view.)

A few more to come (`Copy` to clipboard probably up next).

/cc @devoncarew


devoncarew:
👍 for `Clear All` 

`Copy` also makes a lot of sense.

I'm on the fence about an individual `Clear` command. Do you think this is something we'd want overall?

pq:
In experimenting, when I had spammy logs it was useful to be able to just delete entries.  That was my thinking anyway.  Maybe when we have filtering this will be less useful.

Happy to remove or reassess before release.

pq:
Removed!

PR #2275, opened 2018-05-25
Gradle deps

stevemessick:
@devoncarew 

Not quite sure how this will be used yet. It takes a while to run, so we'll need to design around that. I'd say it is comparable to analyzing a Dart project.

### DIFF
--- a/flutter-studio/src/io/flutter/android/GradleDependencyFetcher.java
+++ b/flutter-studio/src/io/flutter/android/GradleDependencyFetcher.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.android;
+
+import com.android.annotations.VisibleForTesting;
+import com.android.prefs.AndroidLocation;
+import com.android.tools.idea.gradle.project.common.GradleInitScripts;
+import com.android.tools.idea.gradle.util.EmbeddedDistributionPaths;
+import com.android.utils.SdkUtils;
+import com.google.common.collect.Lists;
+import com.intellij.execution.ExecutionException;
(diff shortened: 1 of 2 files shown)


flutter-studio/testSrc/io/flutter/GradleDependencyFetcherTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+  @BeforeClass
+  public static void setUp() {
+    Disposable disposable = Disposer.newDisposable();
+    ApplicationManager.setApplication(new MockApplication(disposable), disposable);
+    Extensions.registerAreaClass("IDEA_PROJECT", null);
+    ourProject = new MockProject(ApplicationManager.getApplication().getPicoContainer(), disposable);
+  }
+
+  private static String getDependencyReport() {
```
devoncarew: `createMockDependencyReport()`?

flutter-studio/src/io/flutter/android/GradleDependencyFetcher.java
```diff
@@ -0,0 +1,148 @@
...
+    myProject = project;
+    myDependencies = new HashMap<>(15);
+  }
+
+  @NotNull
+  public Map<String, List<String>> getDependencies() {
+    return myDependencies;
+  }
+
+  public void run() {
```
devoncarew: When this is used, I assume the result will either be available synchronously, or asynchronously via a Future?

flutter-studio/testSrc/io/flutter/GradleDependencyFetcherTest.java
stevemessick: OK

flutter-studio/src/io/flutter/android/GradleDependencyFetcher.java
stevemessick: Well, run() is synchronous. But I expect we'll have something driving the process that is async and also takes care of the progress UI.

Issue #2282, opened 2018-05-29
Expansions should not add explicit new


mit-mit:
The expansions `stful`, `stless`, etc. still create new keywords.

PR #2285, opened 2018-05-29
Don't use 'new' for stless, stfull, stanim.

scheglov:
Fixes https://github.com/flutter/flutter-intellij/issues/2282


devoncarew:
It's unclear to be that we need `stanim` -  suspect it's used much less than `stless` or `stfull`, yet we give it equal billing (it's own template).

PR #2299, opened 2018-05-31
Log model refactoring.

pq:
work to support filtering.

* stores events in the log (not just the tree model)
* migrates the clear action to empty the log (and view)

/cc @scheglov

### DIFF
--- a/src/io/flutter/logging/FlutterLog.java
+++ b/src/io/flutter/logging/FlutterLog.java
@@ -57,9 +57,9 @@ public class FlutterLog {
     // TODO(pq): add locking.
     entries.clear();
   }
-
+  
   public List<FlutterLogEntry> getEntries() {
-    return Collections.unmodifiableList(entries);
+    return Collections.unmodifiableList(new ArrayList<>(entries));
   }
 
   public void removeListener(@NotNull Listener listener) {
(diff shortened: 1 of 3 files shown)


src/io/flutter/logging/FlutterLog.java
```diff
@@ -45,12 +53,22 @@ public void addListener(@NotNull Listener listener, @Nullable Disposable parent)
     }
   }
 
+  public void clear() {
+    // TODO(pq): add locking.
+    entries.clear();
+  }
+
+  public List<FlutterLogEntry> getEntries() {
+    return Collections.unmodifiableList(entries);
```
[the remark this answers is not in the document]
pq: Quite right!  I thought it made a copy too.  Thanks!

src/io/flutter/logging/FlutterLog.java
```diff
@@ -57,9 +57,9 @@ public void clear() {
     // TODO(pq): add locking.
     entries.clear();
   }
-
+  
   public List<FlutterLogEntry> getEntries() {
-    return Collections.unmodifiableList(entries);
+    return Collections.unmodifiableList(new ArrayList<>(entries));
```
scheglov: You could use also `ImmutableList.copyOf()` from Guava.

pq: Nice.  Good idea!

PR #2331, opened 2018-06-06
Filter count labels.

pq:
Subject to move as toolbar contributions get finalized but a good starting point (and handy in the meantime).

image

/cc @scheglov


src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -302,6 +309,17 @@ public FlutterLogTree(@NotNull FlutterApp app, @NotNull Disposable parent) {
     this.model = model;
   }
 
+  public void addListener(@NotNull EventCountListener listener, @Nullable Disposable parent) {
```
scheglov: Do we repeat `com.intellij.util.EventDispatcher` here?
It seems quite popular in IntelliJ code.

pq: Oh!  Handy.  Thanks!

PR #2339, opened 2018-06-08
use the log category name from the dart:developer event

devoncarew:
- use the log category name from the dart:developer event

@pq


src/io/flutter/logging/FlutterLogView.java
```diff
@@ -217,7 +217,7 @@ public void updated(int filtered, int total) {
 
         final StringBuilder sb = new StringBuilder();
         sb.append(total).append(" event");
-        if (total > 1) {
+        if (total != 1) {
```
pq: 🎉

PR #2345, opened 2018-06-08
Clean up the quick fixes in build-17

stevemessick:
@devoncarew Canary 18 introduces more API changes. This fixes all known issues for that, in a way that continues to work in 3.1.

Hopefully, the `@SuppressWarnings("override")` will let us easily find all those places that need annotations once 3.2 goes stable.


flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
```diff
@@ -75,16 +81,31 @@ public ProjectSystemSyncManager getSyncManager() {
...
   }
 
+  @SuppressWarnings("override")
   @NotNull
   public Collection<PsiElementFinder> getPsiElementFinders() {
-    return null;//gradleProjectSystem.getPsiElementFinders();
+    Method finders = ReflectionUtil.getMethod(gradleProjectSystem.getClass(), "getPsiElementFinders");
+    if (finders == null) {
+      // Not reached; method exists in c18 and not called in c<18.
+      throw new NullPointerException();
```
devoncarew: Should we instead log a warning, and return an empty collection?

Issue #2200, opened 2018-05-04
An improved console for logging.


pq:
The current console view is limited.  Among other things, it:

* lacks structure,
* is not easily searched, and
* provides no affordances for filtering

The plan is to re-think the console log and feedback is most welcome.

A few leading questions:

* what kinds of problems are folks solving with the current console log; how might it better solve them?
* how might it be improved to solve other use cases?
* are there exemplars we can draw inspiration from?

Thanks in advance for your input!

See also: #2199 (cc @zoechi)

zoechi:
I use the [logging](https://pub.dartlang.org/packages/logging) package for printing logs.

Being able to filter for 
- `loggerName` (or parts of the `.`-separated hierarchical name), 
- `time`  
- `message` 
- `level`
would be great.

Show/hide system log messages would also be great, they are often noisy.

I use logs to see if things are executed in the order I expect and if the values I expect are used.
Breakpoints are too cumbersome for code that is called often.

pq:
Thanks @zoechi! 

Popping up, are there other logging tools you use to aid flutter development?  For example does network traffic interest you?

zoechi:
@pq I didn't run into a situation yet where I needed to debug the network traffic, but I'd use tools like WireShark for that.

johnpryan:
Network traffic would be useful for us. For what it's worth, we are using a simple cURL logger right now. It would be much nicer to have some basic features that [flutter_stetho](https://github.com/brianegan/flutter_stetho) provides more at-hand.  Off the top of my head, here's what I think we could use in order of usefulness:

1. view the URL, HTTP headers, and body of requests and responses
2. filtering requests by type (XHR, image) and by filter text
3. "preserve log" feature across app restarts (maybe this is useful for logging more generally?)
4. copy the request as cURL
5. timing of all requests the app has made
6. redo a request

More generally, yes, we use the logging package mentioned above, and being able to filter, color-code, etc would come in handy. 

What other solutions have people found for IntelliJ? Would there be a way to make the logging package compatible with an off-the-shelf IntelliJ plugin... perhaps writing to a .log file and opening with [ideolog](https://plugins.jetbrains.com/plugin/9746-ideolog)?

pq:
Thanks for the ideas @johnpryan!

Incidentally, we should also kick off a conversation about how we'd support [redux.dart](https://github.com/johnpryan/redux.dart). 👍 

[...]

PR #2349, opened 2018-06-10
support regex filter in log view

quangson91:
Ref: https://github.com/flutter/flutter-intellij/issues/2200

Implement support filter by regex in the log view.

Notes:
If user enable `regex` => we will automatically added `.*` before/after the filter text.

Example:
User input: `hello`
=> The actual text filter in mode regex is `.*hello.*`


quangson91:
I closed this PR because we have new one: https://github.com/flutter/flutter-intellij/pull/2350

PR #2350, opened 2018-06-10
support match case/regex filter in log view

quangson91:
Ref: https://github.com/flutter/flutter-intellij/issues/2200

+ Implement support filter by regex & match case in the log view.
+ Refactoring by creating a form handle filter actions.

Notes:
If user enable `regex` => we will automatically added `.*` before/after the filter text.

Example:
User input: `hello`
=> The actual text filter in mode regex is `.*hello.*`

Screenshot:

@pq the UI will look like: 
flutter_log_filter


src/io/flutter/logging/FlutterLogFilterPanel.java
```diff
@@ -0,0 +1,89 @@
...
+
+import javax.swing.*;
+
+
+public class FlutterLogFilterPanel {
+  private JPanel root;
+  private JCheckBox chbMatchCase;
+  private JCheckBox chbRegex;
+  private JPanel pnlExpression;
+  private final LogFilterTextField txtExpression;
```
pq: nit: could we expand these field names to be less abbreviated?  maybe:

`matchCaseCheckBox`
`regexCheckBox`
`panelExpression` and
`textExpression`

?

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -266,12 +266,22 @@ protected void customizeCellRenderer(JTable table, @Nullable Object value, boole
 
 
   public static class ContainsTextFilter implements EntryFilter {
```
pq: Since this is more than just matching containing text, perhaps rename it to something more general.  Maybe `LogEntryFilter`?  Or perhaps do away with the interface altogether and just have an `EntryFilter` class?  (I was thinking there'd be a separate implementation for each matching strategy but what you've go works great too. 👍

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -281,27 +291,36 @@ public boolean accept(@NotNull FlutterLogEntry entry) {
       if (text == null) {
         return true;
       }
-
-      if (entry.getMessage().contains(text)) {
+      final String standartText = isMatchCase ? text : text.toLowerCase();
```
pq: standartText => standardText?

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -281,27 +291,36 @@ public boolean accept(@NotNull FlutterLogEntry entry) {
...
+      final String standartMessage = isMatchCase ? entry.getMessage() : entry.getMessage().toLowerCase();
```
pq: standart => standard

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -281,27 +291,36 @@ public boolean accept(@NotNull FlutterLogEntry entry) {
...
+      final String standartCategory = isMatchCase ? entry.getCategory() : entry.getCategory().toLowerCase();
```
pq: standart => standard

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -281,27 +291,36 @@ public boolean accept(@NotNull FlutterLogEntry entry) {
...
         return true;
       }
+      return acceptByCheckingRegexOption(standartCategory, standartText);
+    }
 
-      //noinspection RedundantIfStatement
-      if (entry.getCategory().contains(text)) {
-        return true;
+    private boolean acceptByCheckingRegexOption(@NotNull String message, @NotNull String text) {
+      final String standardString = message.replaceAll("\n", "");
```
pq: I'm curious, why is this necessary?

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -281,27 +291,36 @@ public boolean accept(@NotNull FlutterLogEntry entry) {
...
+      final ContainsTextFilter filter = (ContainsTextFilter)o;
+      return isRegex == filter.isRegex &&
+             isMatchCase == filter.isMatchCase &&
+             Objects.equals(text, filter.text);
     }
 
     @Override
     public int hashCode() {
-      return 13 ^ Objects.hashCode(text);
+      return Objects.hash(text, isRegex, isMatchCase);
```
pq: 👍

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -29,77 +31,19 @@
...
-import java.awt.event.KeyEvent;
 import java.awt.event.MouseEvent;
 
 public class FlutterLogView extends JPanel implements ConsoleView, DataProvider, FlutterLog.Listener {
-
   // based on com.intellij.openapi.vcs.ui.SearchFieldAction
   private class FilterToolbarAction extends AnAction implements CustomComponentAction /*, RightAlignedToolbarAction */ {
-    private final JPanel myComponent;
-    private final SearchTextField myField;
+    private final FlutterLogFilterPanel pnlFlutterLogFilter = new FlutterLogFilterPanel(param -> doFilter());
```
pq: pnlFlutterLogFilter => filterPanel?

src/io/flutter/logging/LogFilterTextField.java
```diff
@@ -0,0 +1,62 @@
...
+  private void doFilterIfNeeded() {
+    if (onFilterListener != null) {
+      onFilterListener.onFilter();
+    }
+  }
+
+  public interface OnFilterListener {
+    void onFilter();
+  }
+}
```
pq: Add a trailing newline?

testSrc/unit/io/flutter/logging/FlutterLogTreeTest.java
```diff
@@ -0,0 +1,61 @@
...
+    final FlutterLogTree.ContainsTextFilter filterMatchCaseNormal = new FlutterLogTree.ContainsTextFilter("hello", true, false);
+    final FlutterLogTree.ContainsTextFilter filterMatchCaseRegex = new FlutterLogTree.ContainsTextFilter("hello", true, true);
+    final FlutterLogTree.ContainsTextFilter filterInvalidMatchCaseNormal = new FlutterLogTree.ContainsTextFilter("Hello", true, false);
+    final FlutterLogTree.ContainsTextFilter filterInvalidMatchCaseRegex = new FlutterLogTree.ContainsTextFilter("Hello", true, true);
+    assertTrue(filterMatchCaseNormal.accept(entry));
+    assertTrue(filterMatchCaseRegex.accept(entry));
+    assertFalse(filterInvalidMatchCaseNormal.accept(entry));
+    assertFalse(filterInvalidMatchCaseRegex.accept(entry));
+  }
+}
```
pq: Add a trailing newline?

src/io/flutter/logging/FlutterLogTree.java
pq: (This is not a blocking ask.  Feel free to ignore and we can do something on another pass.)

pq:
Super.  Thanks for the screenshot @quangson91.  This is great!

I wonder if we can remove the "Expression" label to get a bit more horizontal space?  I think it's pretty clear what the search box is for.  Perhaps we could add a nice tooltip if you think it's unclear?

quangson91:
@pq I take a screenshot from android logcat.
We may discuss more about UI specs for flutter view before we work.
flutter_log_view_specs

quangson91:
>I wonder if we can remove the "Expression" label to get a bit more horizontal space? I think it's pretty >clear what the search box is for. Perhaps we could add a nice tooltip if you think it's unclear?

Yeap. @pq I will do that.

src/io/flutter/logging/FlutterLogTree.java
quangson91: I choose a solution to merge them all in class `EntryFilter`.
If in the future we need more tyle filter -> we will refactor later.

src/io/flutter/logging/FlutterLogTree.java
quangson91: Oops. My typos mistake. I updated.

src/io/flutter/logging/FlutterLogTree.java
quangson91: updated.

src/io/flutter/logging/FlutterLogTree.java
quangson91: Updated.

src/io/flutter/logging/FlutterLogTree.java
quangson91: Because java regex doesn't recognize as a character.
For example:

```java
public class Main {
    public static void main(String args[]) {
        String input = "\n hello world \n";
        String regex = ".*hello.*";
        System.out.println("Result: " + input.matches(regex));
    }
}
```
Result: `false`.

But if we change input to:
```
String input = "\n hello world \n".replaceAll("\n", "");
```
=> Result: `true`.

quangson91: The affect of this code is user can't filter with string context break line.

src/io/flutter/logging/FlutterLogView.java
quangson91: Updated. Thanks!

src/io/flutter/logging/LogFilterTextField.java
quangson91: Done.

testSrc/unit/io/flutter/logging/FlutterLogTreeTest.java
quangson91: Done.

src/io/flutter/logging/FlutterLogTree.java
pq: Aha.  Thanks!  If this is just for the regex case, should we move the `replaceAll` into the if block?

IOW, something like:
```java
if (isRegex) {
  // Extra replace to support matching on strings w/ line breaks.
  return message.replaceAll("\n", "").matches(".*" + text + ".*");
}
```

quangson91: Yeap. I updated. Thank you!

pq: Cool!  One last question...  I'm rusty w/ regexps but I wonder if we there isn't some combination of Pattern flags we couldn't use to get around need to modify the message string.  Maybe look at [`DOTALL`](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#DOTALL) and [`MULTILINE`](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#MULTILINE)?

Happy to land an iterate as well.

Thanks again!

quangson91: Yes, It looks better.
We just need to add `(?s)` to support DOTALL.
```
return message.matches("(?s).*" + text + ".*");
```
That's great!.
Thank you!

quangson91:
@pq , There is something coming up for logging feature.

Ref: https://i.stack.imgur.com/q3dEU.gif

+ Log Level
+ Color by log level (can custom color by regex too).
+ Filter (match case, regex)
+ Integrate with GrepConsole (grep console only support log view extends from com.intellij.execution.impl.ConsoleViewImpl).
+ Support multi tab for filter. 
+ Clickable in reference in console. (Eg. In console we have xyz.dart:15 -> click open xyz.dart file)
+ Auto scroll at the end / Keep scroll position.
+ Append line able. (Eg. we have log 1, 2, 3, 4. user click append line => we append "=========" then the comming log will bellow "======";)
+ Copyable.
+ Wrap line option.
+ Folding option.

===
Other
+ select log by app.
+ select log by device.

pq:
@quangson91 : great list.  We do have a bunch of these already and a few are already in progress, but others aren't planned yet.  Maybe call out things you'd like to see happen on #2200 so we can discuss and coordinate after this PR is merged? I'd also like to be careful not to turn the logging view into a feature kitchen sink.  In particular, I think the prior art of loggers isn't the best model for usability and would like to start from the use cases rather from the implementations.  Anyway, yeah.  Exciting stuff.  Thanks for your continued contributions!

quangson91:
@pq, yeap, I got it.

PR #2359, opened 2018-06-11
Merge DartVmServiceDebugProcess and DartVmServiceDebugProcessZ

jacob314:
This pays down a lot of technical debt from the DartVmServiceDebugProcessZ class that extended and then reimplemented everything in DartVmServiceDebugProcess resulting in a fragile mix of Flutter and Dart functionality. The merge is far from robotic. If anything I did doesn't make sense let me know.


src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -76,32 +83,42 @@
...
   private final int myTimeout;
   @Nullable private final VirtualFile myCurrentWorkingDirectory;
   @Nullable protected String myRemoteProjectRootUri;
 
+
+  private VmOpenSourceLocationListener myVmOpenSourceLocationListener;
+
+  @NotNull private final ObservatoryConnector myConnector;
+
+  private boolean remoteDebug = false;
```
stevemessick: I'm going to assume we don't know if remote debugging makes sense for Flutter and that's why all the references to this exist. If we do know (or someone does and hasn't said) then there is some more cleanup to do.

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -141,19 +156,52 @@ public void onTextAvailable(@NotNull final ProcessEvent event, @NotNull final Ke
...
     else {
-      LOG.assertTrue(myExecutionResult != null && myDASExecutionContextId != null, myDASExecutionContextId + myExecutionResult);
+      LOG.assertTrue(myExecutionResult != null, myExecutionResult);
     }
+
+    this.executionEnvironment = executionEnvironment;
+    this.mapper = mapper;
+    myConnector = connector;
+
+    setLogger();
```
stevemessick: There is another setLogger() on line 130.

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -609,4 +533,241 @@ private static String threeSlashize(@NotNull final String uri) {
...
+        @Override
+        public void onError(RPCError error) {
+          LOG.error(error.toString());
+        }
+      });
+    }
+
+    // We re-enable the remote debug flag so that the service wrapper will call our guessRemoteProjectRoot()
+    // method with the list of loaded libraries for the isolate.
+    remoteDebug = true;
```
stevemessick: I'm confused by myRemoteDebug vs remoteDebug. The accessor that (I think) is used by the wrapper tests myRemoteDebug, which is always false. If I'm right, guessRemoteProjectRoot() won't be called. OTOH things seem to work...

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
jacob314: good catch. done.

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
jacob314: The remoteDebug logic concerns me and my preference is to remove it.
Essentially currently we are starting with it off to avoid undesirable code paths and then faking that it is off to avoid undesirable code paths.
Now that VmServiceWrapper is in the same project we should be able to clean this up. 

Unless we wanted to support server apps running locally I don't think we need to ever really support anything other than remoteDebug mode.

jacob314:
Pushed a CL that removes the remote-debug cruft.

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
jacob314: It is worse than that. It starts off but then gets turned on (surprise!) to make sure we went through the right path in code that was formerly in a separate plugin so couldn't be rolled consistently. I've stripped the remote debug flag as I think it just causes confusion and in practice all debugging we want to do for flutter is essentially remote.

Issue #2353, opened 2018-06-11
Logging view: filter panel layout issues


pq:
After patching in #2350, a few things to fix:

* notice the behavior on re-size
* ~also notice that the event count is no longer updating~ (likely unrelated)

![filter_resize](https://user-images.githubusercontent.com/67586/41249297-9d0aa21e-6d68-11e8-8612-9388779d1da1.gif)

/cc @quangson91

pq:
@quangson91 : let me know if you're up for taking a look; otherwise I'm happy to.

quangson91:
@pq let me fix that problem.

quangson91:
@pq, Do you know how to inspect layout while development plugin ?
(I mean how we inspect layout of filter-actions?)

pq:
Thanks @quangson91!

> (I mean how we inspect layout of filter-actions?)

Sorry, I don't. ☹️

quangson91:
@pq I created a new PR here: https://github.com/flutter/flutter-intellij/pull/2361
Please take a look!

pq:
Merged.  Thanks!

pq:
Fixed w/ #2361

PR #2361, opened 2018-06-12
fix log layout problem.

quangson91:
Issue: https://github.com/flutter/flutter-intellij/issues/2353

+ Improve logging layout.

Here is the record of how it works.
fix_logging_layout

I'm not good for swing/layout much.
=> please give me a suggestion to make it better.

Notes:
Actually, I want to make log layout still keep visible while user resizes window like this one:
expect_log_view

### DIFF
--- a/src/io/flutter/logging/FlutterLogFilterPanel.form
+++ b/src/io/flutter/logging/FlutterLogFilterPanel.form
@@ -1,25 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <form xmlns="http://www.intellij.com/uidesigner/form/" version="1" bind-to-class="io.flutter.logging.FlutterLogFilterPanel">
-  <grid id="27dc6" binding="root" layout-manager="GridLayoutManager" row-count="1" column-count="4" same-size-horizontally="false" same-size-vertically="false" hgap="3" vgap="-1">
+  <grid id="27dc6" binding="root" layout-manager="GridLayoutManager" row-count="1" column-count="3" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
     <margin top="0" left="0" bottom="0" right="0"/>
     <constraints>
-      <xy x="20" y="20" width="755" height="44"/>
+      <xy x="20" y="34" width="450" height="35"/>
     </constraints>
     <properties>
+      <background color="-1250068"/>
       <enabled value="true"/>
-      <preferredSize width="986" height="35"/>
+      <minimumSize width="450" height="35"/>
+      <preferredSize width="450" height="35"/>
     </properties>
     <border type="none"/>
     <children>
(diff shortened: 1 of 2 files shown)


src/io/flutter/logging/FlutterLogFilterPanel.form
```diff
@@ -1,25 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <form xmlns="http://www.intellij.com/uidesigner/form/" version="1" bind-to-class="io.flutter.logging.FlutterLogFilterPanel">
-  <grid id="27dc6" binding="root" layout-manager="GridLayoutManager" row-count="1" column-count="4" same-size-horizontally="false" same-size-vertically="false" hgap="3" vgap="-1">
+  <grid id="27dc6" binding="root" layout-manager="GridLayoutManager" row-count="1" column-count="3" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
     <margin top="0" left="0" bottom="0" right="0"/>
     <constraints>
-      <xy x="20" y="20" width="755" height="44"/>
+      <xy x="20" y="34" width="450" height="35"/>
```
quangson91: I'm not sure is this good.
Actually, I just change the width from `755` to `450`.
So the width of filter action is smaller  -> we can resize without hide filter action.
But if the Intellij IDEA window too narrow -> filter action still hide.
Please give me some suggestion.

src/io/flutter/logging/FlutterLogFilterPanel.form
```diff
@@ -1,25 +1,18 @@
...
     <properties>
+      <background color="-1250068"/>
       <enabled value="true"/>
-      <preferredSize width="986" height="35"/>
+      <minimumSize width="450" height="35"/>
+      <preferredSize width="450" height="35"/>
     </properties>
     <border type="none"/>
     <children>
-      <grid id="9d27c" binding="panelExpression" layout-manager="GridLayoutManager" row-count="1" column-count="1" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
```
quangson91: I remove `panelExpression` which is the container of `expressionTextField`.
Instead, I put `expressionTextField` directly in the layout.

src/io/flutter/logging/FlutterLogFilterPanel.form
```diff
@@ -36,11 +29,15 @@
           <text value="Regex"/>
         </properties>
       </component>
-      <hspacer id="6d179">
+      <component id="b7426" class="com.intellij.ui.SearchTextField" binding="textExpression" custom-create="true">
```
quangson91: Here is the code insert textExpression directly.

src/io/flutter/logging/FlutterLogFilterPanel.form
pq: I'll have to play around.... but this is a vast improvement! 👍

src/io/flutter/logging/FlutterLogFilterPanel.form
pq: 👍

PR #2367, opened 2018-06-12
Log entry sequence numbering.

pq:
Playing w/ sequence numbers as per: #2200. 

image

/cc @stevemessick


pq:
Thanks @stevemessick!  Looking into that as well. 👍

PR #2368, opened 2018-06-12
chore: For BazelRunConfig launches, print the command-line to the console

jwren:


src/io/flutter/run/LaunchState.java
```diff
@@ -126,6 +130,19 @@ private RunContentDescriptor launch(@NotNull ExecutionEnvironment env) throws Ex
 
     final ExecutionResult result = setUpConsoleAndActions(app);
 
+    // For Bazel run configurations;
+    // where the console is not null;
```
pq: Nit: I'm getting confused by the semi-colons...

Maybe commas?

```
// For Bazel run configurations,
// where the console is not null,
// and we find the expected process handler type,
// print the command line command to the console.
```

Issue #2372, opened 2018-06-14
improvements to analytics for hot reload


devoncarew:
@InMatrix, I investigated our current analytics. Currently, we sent:
- analytics events for hot reload and hot restart actions
- analytics timings for when reloads or restarts finish
- analytics 'workflow' events at the end of an app run, to mark a run as either using the hot reload workflow, the hot restart workflow, or neither

We do not send events for:
- hot reloads when performed via reload-on-save
- hot restarts when invoked from the run or debug buttons (clicking on the run button for an already running app)
- (we do send post-reload / restart timings for the above however)

Going forward, we want to track the ratio of reloads to restarts per user per day. I think this means:
- after a run session, or cumulatively per day, send a `reloadRestarts` ratio metric
- have the end of a run session count as one restart
- fix the bug where we don't send a ping for hot restarts when done through the run/debug button
- continue to not send individual analytics pings for reload-on-save

InMatrix:
As we discussed in person, it might be easier to view the results if we can send the ratio metric to the Flutter Tool GA account.

devoncarew:
Will plan to move (and update the issue w/ the latest, in terms of the time buckets to use).

PR #2374, opened 2018-06-14
add more analytics events

devoncarew:
- fix one issue from https://github.com/flutter/flutter-intellij/issues/2372
- add a reload / restart ratio (we'll plan to have the real metrics in flutter_tools, but this will be useful for investigation)

### DIFF
--- a/src/io/flutter/run/daemon/FlutterApp.java
+++ b/src/io/flutter/run/daemon/FlutterApp.java
@@ -234,6 +234,10 @@ public class FlutterApp {
         // Send analytics about whether this session used the reload workflow, the restart workflow, or neither.
         final String workflowType = app.reloadCount > 0 ? "reload" : (app.restartCount > 0 ? "restart" : "none");
         FlutterInitializer.getAnalytics().sendEvent("workflow", workflowType);
+
+        // Send the ratio of reloads to restarts.
+        final double reloadRestartRatio = app.reloadCount / (app.restartCount + 1);
+        FlutterInitializer.getAnalytics().sendEventMetric("workflow", "reloadRestartRatio", (int)Math.round(reloadRestartRatio));
       }
     });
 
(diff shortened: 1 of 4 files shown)


src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -234,6 +234,10 @@ public void processTerminated(@NotNull ProcessEvent event) {
         // Send analytics about whether this session used the reload workflow, the restart workflow, or neither.
         final String workflowType = app.reloadCount > 0 ? "reload" : (app.restartCount > 0 ? "restart" : "none");
         FlutterInitializer.getAnalytics().sendEvent("workflow", workflowType);
+
+        // Send the ratio of reloads to restarts.
+        final double reloadRestartRatio = app.reloadCount / (app.restartCount + 1);
```
jacob314: wouldn't it be better to send 
app.reloadCount / (app.reloadCount + app.restartCount)
and just call it reloadFraction?
I am assuming the +1 in restartCount is there just to avoid dividing by zero.

devoncarew: Multiplying by 100 (and getting two additional decimal points) sounds good.

The 1 is there partly to avoid divide by zero, and partly counting the process termination as a restart.

I'm not sure that measuring as a reload fraction gains us anything over using a reload ratio; @InMatrix, thoughts?

InMatrix: Either is fine. Reload fraction might have a slight advantage of being easier to talk about than reload/restart ratio?

devoncarew: I changed this to reloadFraction (reload / (reload + restart)), and send in the value multiplied by 100 (since we can only send ints. So, 100 == all reload, and 0 == no reloads.

devoncarew:
Updated to use a `reloadFraction` metric - `100` means all reloads in a session, `0` means no reloads in a session, and `75` means a ratio of 3 reloads to each restart.

jacob314:
lgtm

PR #2376, opened 2018-06-14
improve log console

quangson91:
Implement flutter log console by using TextConsole (not TreeView).
Video demo:
demoflutterlog

The reason I choose to implement log console by using TextConsole (not TreeView).
+ When we have many logs -> TreeView will generate many rows (view) -> I think that's not good. (For performance, memory).
+ Tree view may not familiar with the developer. (in android we have Logcat, it is a TextConsole view, in iOS, the log looks like TextConsole too).
+ By using TextConsole we can easily keep integrating with GrepConsole plugin (support grep in multi-tab, color highlight - those features we can support, but I think we can't implement all of them immediately)

Notes:
The full features are not completed yet.

+ I also fix the problem UI of filter pannel while resizing the IDE.

@pq, Please take a look when you have time!

### DIFF
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -19,6 +19,7 @@ import com.intellij.util.Alarm;
 import com.intellij.util.EventDispatcher;
 import com.intellij.util.ui.ColumnInfo;
 import io.flutter.console.FlutterConsoleFilter;
+import io.flutter.logging.v2.EntryFilter;
 import io.flutter.run.daemon.FlutterApp;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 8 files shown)


src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -353,63 +354,6 @@ protected void customizeCellRenderer(JTable table, @Nullable Object value, boole
     }
   }
 
-  public static class EntryFilter {
```
quangson91: Make this class as a single class (not inside FlutterLogTree).

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -163,8 +164,8 @@ public void actionPerformed(AnActionEvent e) {
     private void doFilter() {
       final FlutterLogFilterPanel.FilterParam param = filterPanel.getCurrentFilterParam();
       final String text = getText();
-      final FlutterLogTree.EntryFilter filter =
-        StringUtils.isEmpty(text) ? null : new FlutterLogTree.EntryFilter(text, param.isMatchCase(), param.isRegex());
+      final EntryFilter filter =
```
quangson91: Just a update reference to the `EntryFilter`, nothing change here.

src/io/flutter/logging/v2/FlutterLogFormater.kt
```diff
@@ -0,0 +1,43 @@
...
+ * found in the LICENSE file.
+ */
+
+package io.flutter.logging.v2
+
+import io.flutter.logging.FlutterLog
+import io.flutter.logging.FlutterLogEntry
+import java.text.SimpleDateFormat
+
+class FlutterLogFormater(
```
quangson91: This class support format console like a table.

src/io/flutter/logging/v2/FlutterLogView.kt
```diff
@@ -0,0 +1,141 @@
...
+      val popupMenu = ActionManager.getInstance().createActionPopupMenu(
+        ActionPlaces.UNKNOWN,
+        createPopupActionGroup()
+      )
+      popupMenu.component.show(component, component.width, 0)
+    }
+
+    private fun createPopupActionGroup(): DefaultActionGroup = DefaultActionGroup().apply {
+      add(ShowTimeStampsAction())
+      add(ShowLevelAction())
```
quangson91: I just implement those two actions
ShowTimeStampsAction & ShowLevelAction

I'm not sure, do we need to implement log line number.
The line-number of the log is relative (especially when we filtering).

quangson91:
@pq @devoncarew, I wrote new code in kotlin, it works well on the computer.
But it fails on Travis ci.
Do you have any suggestion for that problem?

Thanks!

pq:
Thanks for all of this work!  The decision to use a Tree View was in anticipation of a number of planned features.    While it’s possible that a text console would support those ideas,  a structured view was decided on as providing the most flexibility.   (For what it’s worth, while it’s true that logcat is text-based, the Android Studio folks are trending towards structured views for similar new features in progress.). This is not to dismiss any of the benefits you mention.  They’re significant!  For now though, as we round out the design and continue to explore the space, I think we should keep on course with our current implementation.  If that’s something we want to change, this is a great proof-of-concept.  Thank you for doing all this leg-work.

Thinking ahead to other feature work, a few things are in progress, not all tracked here unfortunately.  We can do better at that…  Sorry!  To help though maybe you could open fresh issues to discuss ideas so we can keep each other informed and avoid overlap and collision?

quangson91:
@pq Thank you for the explanation.
So is there backlog OR list of issues to development on next milestone?
I want to work more on improving logging feature ^^.

quangson91:
@pq we can close this PR.
next, I want to fix resize problem for log panel & add filter log by log level
https://github.com/flutter/flutter-intellij/issues/2377

Is that ok ?

pq:
Thanks @quangson91 !

> So is there backlog OR list of issues to development on next milestone?

Not just yet since things are moving fast.

Are there things you're specifically interested in?  Maybe open targeted issues and we can discuss?

> next, I want to fix resize problem for log panel & add filter log by log level

Super.  Sorry I didn't get to a review.  Busy day!

quangson91:
@pq Thank you!

Issue #2377, opened 2018-06-15
Fix resize flutter logview & add filter log by log level


quangson91:
Fix resize flutter log view & add filter log by log level

Log level:
```java
    NONE(0),
    FINEST(300),
    FINER(400),
    FINE(500),
    CONFIG(700),
    INFO(800),
    WARNING(900),
    SEVERE(1000),
    SHOUT(1200);
```

quangson91:
Done in #2380

PR #2379, opened 2018-06-15
fix flutter log view while resizing

quangson91:
Relate: https://github.com/flutter/flutter-intellij/pull/2361

Fix the filter panel is hidden when the window is narrow.

Reason:
It seems all actions will automatically hide when the window is narrow & auto visible when the mouse hover.

Solution
Keep the filter panel is just a Panel. (we don't wrap it in the action).

Demo:
fix_reisze_flutter_log_view

/cc @pq

PR #2380, opened 2018-06-15
add feature filter by log level

quangson91:
Issue: https://github.com/flutter/flutter-intellij/issues/2377
+ add feature filter by log level

Demo
filter_by_log_level

While testing, It seems the generate log level is not correct.
So I think we should double check after the issue for log level fixed.

/cc @pq what do you think?

### DIFF
--- a/src/io/flutter/logging/FlutterLogFilterPanel.form
+++ b/src/io/flutter/logging/FlutterLogFilterPanel.form
@@ -1,14 +1,14 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <form xmlns="http://www.intellij.com/uidesigner/form/" version="1" bind-to-class="io.flutter.logging.FlutterLogFilterPanel">
-  <grid id="27dc6" binding="root" layout-manager="GridLayoutManager" row-count="1" column-count="3" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
+  <grid id="27dc6" binding="root" layout-manager="GridLayoutManager" row-count="1" column-count="4" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
     <margin top="0" left="0" bottom="0" right="0"/>
     <constraints>
-      <xy x="20" y="34" width="450" height="35"/>
+      <xy x="20" y="34" width="500" height="35"/>
     </constraints>
     <properties>
       <enabled value="true"/>
-      <minimumSize width="450" height="35"/>
-      <preferredSize width="450" height="35"/>
+      <minimumSize width="500" height="35"/>
+      <preferredSize width="500" height="35"/>
     </properties>
     <border type="none"/>
     <children>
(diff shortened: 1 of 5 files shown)


src/io/flutter/logging/FlutterLogFilterPanel.form
```diff
@@ -37,6 +37,28 @@
...
+      <component id="4ad5d" class="javax.swing.JComboBox" binding="logLevelComboBox">
+        <constraints>
+          <grid row="0" column="3" row-span="1" col-span="1" vsize-policy="0" hsize-policy="6" anchor="8" fill="0" indent="0" use-parent-layout="false">
+            <minimum-size width="100" height="-1"/>
+            <preferred-size width="100" height="-1"/>
+            <maximum-size width="100" height="-1"/>
+          </grid>
+        </constraints>
+        <properties>
+          <model>
```
quangson91: I added some model here for display purpose while design on swing form.
The real model for combo box is: https://github.com/flutter/flutter-intellij/pull/2380/files#diff-cc5515b1514ea5d6adfd3685f8d96050R38

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -354,27 +357,20 @@ protected void customizeCellRenderer(JTable table, @Nullable Object value, boole
   }
 
   public static class EntryFilter {
-    @Nullable
-    private final String text;
-    private final boolean isRegex;
-    private final boolean isMatchCase;
+    @NotNull
+    private final FlutterLogFilterPanel.FilterParam filterParam;
```
quangson91: Replace all separate fields with a wrapper class.

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -354,27 +357,20 @@ protected void customizeCellRenderer(JTable table, @Nullable Object value, boole
...
-
-    @Nullable
-    public String getText() {
-      return text;
+    public EntryFilter(@NotNull FlutterLogFilterPanel.FilterParam filterParam) {
+      this.filterParam = filterParam;
     }
 
     public boolean accept(@NotNull FlutterLogEntry entry) {
+      if (entry.getLevel() < filterParam.getLogLevel().value) {
```
quangson91: For filter log by level, I simply check `entry.getLevel() < filterParam.getLogLevel().value`.
Because we define:
```java
  public enum Level {
    NONE(0),
    FINEST(300),
    FINER(400),
    FINE(500),
    CONFIG(700),
    INFO(800),
    WARNING(900),
    SEVERE(1000),
    SHOUT(1200);
```
Example, user choose level `INFO(800)` -> only log with level >= INFO(800) can be displayed. Otherwise, we should not accept.

pq:
Sorry I didn't get to this today; I'll patch it in on Monday and take a look.

Needless to say, super promising.  Thanks!

pq:
One thing to consider.  The levels that are defined in `logging` are a convention but still arbitrary.  Moreover, how would we support folks who define their own?

Suppose, I add:

```
 TRACE(750)
```

for example.  It'd be great to be able to filter on it.  In that case, allowing levels to be user-definable / configurable would be useful.  Thoughts?

quangson91:
@pq in android logcat they force us using limit level (info/verbose/debug/warm/error).
-> I think we should standard logging.
If we take care about user custom log level -> user can create custom log level up to a thousand :|

pq:
@quangson91: it looks like this needs a rebase after #2379...

quangson91:
@pq, I resolved the conflict.
PTAL when you have time !

pq:
Looks great.  Thanks @quangson91!

PR #2381, opened 2018-06-15
Log tree rendering refactor

pq:
Largely a structural refactoring to set us up for controlling row theming (e.g., foreground and background colors) in a rational way.  It did turn out that the way that renderers were getting mapped to active columns after a hide was incorrect so this does in fact fix that.

image

(Previous to this fix the hyperlink support was missing because the message row was wrongly getting rendered as simple text.)

Apologies for the churn in review; this is largely mechanical.  My hope is to land this one and put the more interesting changes in a follow-up where they'll be more obvious.

/cc @scheglov @devoncarew

### DIFF
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -78,12 +78,12 @@ public class FlutterLogTree extends TreeTable {
       @Override
       public TableCellRenderer getCustomizedRenderer(DefaultMutableTreeNode o, TableCellRenderer renderer) {
         if (renderer == null) {
-          renderer = getRenderer();
+          renderer = createRenderer();
         }
         return renderer;
       }
 
-      abstract TableCellRenderer getRenderer();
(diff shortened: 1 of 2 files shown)


src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -35,47 +34,248 @@
...
+
+      @Override
+      public TableCellRenderer getCustomizedRenderer(DefaultMutableTreeNode o, TableCellRenderer renderer) {
+        if (renderer == null) {
+          renderer = getRenderer();
+        }
+        return renderer;
+      }
+
+      abstract TableCellRenderer getRenderer();
```
scheglov: Is it more appropriate to call `createRenderer()`?
It seems that we call it once and cache the value.

pq: Sure.  I like that.

pq: Fixed. 👍

PR #2382, opened 2018-06-15
Log entry coloring.

pq:
Basic entry coloring (WIP).

image

As discussed in #2200, this will ultimately benefit from user configurability; this plumbs in the back-end for that.

/cc @devoncarew @scheglov

### DIFF
--- a/src/io/flutter/logging/FlutterLogView.java
+++ b/src/io/flutter/logging/FlutterLogView.java
@@ -17,9 +17,7 @@ import com.intellij.openapi.actionSystem.ex.CustomComponentAction;
 import com.intellij.openapi.actionSystem.impl.ActionButton;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.ui.SimpleToolWindowPanel;
-import com.intellij.ui.ColoredTableCellRenderer;
-import com.intellij.ui.PopupHandler;
-import com.intellij.ui.ScrollPaneFactory;
+import com.intellij.ui.*;
 import com.intellij.ui.components.JBLabel;
 import com.intellij.ui.content.Content;
 import com.intellij.ui.content.ContentFactory;
@@ -39,7 +37,31 @@ import java.awt.datatransfer.Clipboard;
 import java.awt.datatransfer.StringSelection;
 import java.awt.event.MouseEvent;
 
+import static com.intellij.ui.SimpleTextAttributes.STYLE_PLAIN;
+
 public class FlutterLogView extends JPanel implements ConsoleView, DataProvider, FlutterLog.Listener {
+
+  private class EntryModel implements FlutterLogTree.EntryModel {
+    boolean color;
+
+    @Override
+    public SimpleTextAttributes style(@Nullable FlutterLogEntry entry, int attributes) {
+      if (color && entry != null) {
+        final FlutterLog.Level level = FlutterLog.Level.forValue(entry.getLevel());
+        if (level != null) {
+          switch (level) {
+            case FINER:
+              return SimpleTextAttributes.GRAY_ATTRIBUTES;
+            case WARNING:
+              return ORANGE_ATTRIBUTES;
+            case SEVERE:
+              return SimpleTextAttributes.ERROR_ATTRIBUTES;
+          }
+        }
+      }
+      return SimpleTextAttributes.REGULAR_ATTRIBUTES;
+    }
+  }
 
   class ConfigureAction extends AnAction implements RightAlignedToolbarAction {
     private final DefaultActionGroup actionGroup;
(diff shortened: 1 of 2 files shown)


quangson91:
@pq, how did you get level `WARNING` in the console ?
I've tried:
```dart
  Logger.root.level = Level.ALL;
  Logger.root.onRecord.listen((LogRecord rec) {
    print('${rec.level.name}: ${rec.time}: ${rec.message}');
  });

  log.finest("LOG FINEST");
  log.finer("LOG FINER");
  log.fine("LOG FINE");
  log.config("LOG CONFIG");
  log.info("LOG INFO");
  log.warning("LOG WARNING");
  log.severe("LOG SEVERE");
  log.shout("LOG SHOUT");
```
But the log level all show `INFO`

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -39,8 +37,32 @@
 import java.awt.datatransfer.StringSelection;
 import java.awt.event.MouseEvent;
 
+import static com.intellij.ui.SimpleTextAttributes.STYLE_PLAIN;
+
 public class FlutterLogView extends JPanel implements ConsoleView, DataProvider, FlutterLog.Listener {
 
+  private class EntryModel implements FlutterLogTree.EntryModel {
+    boolean color;
```
quangson91: we should rename that variable. Eg, `isEnableLogColor` or something like that.
(`color` feels like the actual color value - Colors.RED)

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -39,8 +37,32 @@
...
+
+    @Override
+    public SimpleTextAttributes style(@Nullable FlutterLogEntry entry, int attributes) {
+      if (color && entry != null) {
```
quangson91: I'm not sure about the java coding style.
But I think we can use earlier return to reduce the deep nested.
Eg
```java
if (entry == null || notEnableLogColor) return;
// Do other things.
```

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -39,8 +37,32 @@
...
+        final FlutterLog.Level level = FlutterLog.Level.forValue(entry.getLevel());
+        if (level != null) {
```
quangson91: earlier return too.

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -300,8 +342,10 @@ String countString(int count) {
       return Integer.toString(count);
     }
   }
-
   @NotNull final FlutterApp app;
+  private final SimpleTextAttributes ORANGE_ATTRIBUTES = new SimpleTextAttributes(STYLE_PLAIN, JBColor.orange);
```
quangson91: Maybe add annotation `@NotNull`

pq:
Hey @quangson91.  To get that to work, you'll want to plumb level through to the `developer` log.  Try replacing your `print` with something like this:

```dart
  Logger.root.onRecord.listen((LogRecord rec) {
    developer.log(rec.message, level: rec.level.value);
  });
```

with an added import:

```dart
import 'dart:developer' as developer;
```


Hope that helps!

quangson91:
> Logger.root.onRecord.listen((LogRecord rec) {
>    developer.log(rec.message, level: rec.level.value);
> });

@pq it worked.
Thank you!

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -39,8 +37,32 @@
 import java.awt.datatransfer.StringSelection;
 import java.awt.event.MouseEvent;
 
+import static com.intellij.ui.SimpleTextAttributes.STYLE_PLAIN;
```
scheglov: We later use other values of SimpleTextAttributes.
Should we statically import all of them, or none?

pq: Good catch.  Thanks!

src/io/flutter/logging/FlutterLogView.java
pq: In a private classes I prefer shorter names, especially when the meaning should be clear from the context (and the type).  Of course, the fact that it surprised you, means it wasn't clear! 😄  I'll make it better.

Thanks!

src/io/flutter/logging/FlutterLogView.java
pq: I'm on the fence.  Many argue that having multiple return points in a function is as confusing (if not more). 🤔 

That said, when a function is very short and can be read in a glance I tend to think either approach is fine.

Thanks!

src/io/flutter/logging/FlutterLogView.java
pq: Cool.  Thanks.  See above for my rationale.

src/io/flutter/logging/FlutterLogView.java
pq: 👍

PR #2390, opened 2018-06-19
Reorder level and text filters

pq:
Sort log level first.

image

I *think* it looks a bit better and makes semantic sense... it's also consistent w/ logcat (for what that's worth).

/cc @quangson91

### DIFF
--- a/src/io/flutter/logging/FlutterLogFilterPanel.form
+++ b/src/io/flutter/logging/FlutterLogFilterPanel.form
@@ -14,7 +14,7 @@
     <children>
       <component id="20384" class="javax.swing.JCheckBox" binding="matchCaseCheckBox">
         <constraints>
-          <grid row="0" column="1" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+          <grid row="0" column="2" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
         </constraints>
         <properties>
           <text value="Match case"/>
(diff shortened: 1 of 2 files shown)


quangson91:
@pq I think it looks better 👍

quangson91:
@pq, Just for you know, I'm working on custom log color ^^.

quangson91:
I'm making something look like this: https://github.com/flutter/flutter-intellij/issues/2200#issuecomment-395977852

log cat

But it not in setting dialog (I will make it as a single dialog first)
(Because I don't know how to move it to setting page yet)

pq:
Cool.  Thanks for taking a look @quangson91!

Regarding color config, that's awesome.  Contributing the settings to the general preferences alongside logcat sounds idea too.  To get a hint for how that could be done, you might look at the Dart Plugin.  Here's where it contributes its color settings:

https://github.com/JetBrains/intellij-plugins/blob/ae11efd91d3588eff98dd355930f7852a35b23f9/Dart/resources/META-INF/plugin.xml#L52

Implementation here:

https://github.com/JetBrains/intellij-plugins/blob/master/Dart/src/com/jetbrains/lang/dart/highlight/DartColorsAndFontsPage.java

Cheers!

pq:
@quangson91: I took a little time today looking at how the Dart Plugin does it and I'm pretty sure that's the right approach.  Let me know what you think; and if I can help!

pq:
@quangson91: one more to look at:

https://github.com/JetBrains/android/blob/master/android/src/com/android/tools/idea/logcat/AndroidLogcatColorPage.java

Maybe the best model for what you're after 👍

quangson91:
Yeap. 
It's my pleasure

PR #2391, opened 2018-06-19
Add level combo renderer

pq:
Improves default rendering by padding items.

Before:

image

After:

image

Also leaves the door open for styling.  (Possibly using the level colors?)

/cc @quangson91 @scheglov


pq:
> As an aside, I find the filter text a bit loud (here and in the view); perhaps choose to lower case them?

I've been noticing the same thing.  I also wonder if `NONE` isn't distractung (though accurate) when no level has been selected; maybe the default should be empty (but map to NONE?)  Another alternative is a muted foreground color.

I'll experiment.

Thanks!

Issue #2396, opened 2018-06-20
Finish wiring up the Flutter Log Color page


pq:
Follow-up from: #2394.

For `M26`, we'll want to:

-  [x]  update log color by getting setting from FlutterLogColorPage
-  [x] set default color for logging (by level)

Alternatively, we can comment-out the contribution in `plugin.xml` and resume w/ `M27`.

/cc @quangson91

quangson91:
Default color can be found in this document: https://www.jetbrains.org/intellij/sdk/docs/reference_guide/color_scheme_management.html
(I'm working on this)

quangson91:
I implement default color scheme for flutter color log.
Ref: https://github.com/flutter/flutter-intellij/pull/2394/commits/4642f0805912362c39b42c612c2dd4ff523d09e0

quangson91:
I also found the way to get config
```
    Arrays.stream(ColorSettingsPages.getInstance().getRegisteredPages())
      .filter(page -> page instanceof FlutterLogColorPage)
      .findFirst().ifPresent(page -> {
      final List<AttributesDescriptor> attributeDescriptors = ColorSettingsUtil.getAllAttributeDescriptors(page);        
    });
```

`attributeDescriptors` will contain setting color for foreground/background and so on.

BUT I don't know why those variables are `null` :|

quangson91:
> update log color by getting setting from FlutterLogColorPage

I did it in this PR: https://github.com/flutter/flutter-intellij/pull/2399

quangson91:
@pq, @devoncarew as the pr #2399 already merged.
So I think we should close this issue.

PR #2399, opened 2018-06-21
setting log color from flutter log setting page

quangson91:
Issue: https://github.com/flutter/flutter-intellij/issues/2396

setting log color by taking value from flutter log color setting page.

Demo
flutter_log_color

/ cc @pq

### DIFF
--- a/src/io/flutter/logging/FlutterLogView.java
+++ b/src/io/flutter/logging/FlutterLogView.java
@@ -45,7 +45,6 @@ import java.awt.datatransfer.StringSelection;
 import java.awt.event.MouseEvent;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.BlockingDeque;
 import java.util.concurrent.ConcurrentHashMap;
 
 public class FlutterLogView extends JPanel implements ConsoleView, DataProvider, FlutterLog.Listener {
@@ -73,7 +72,7 @@ public class FlutterLogView extends JPanel implements ConsoleView, DataProvider,
   @NotNull
   private volatile EditorColorsScheme globalEditorColorsScheme = EditorColorsManager.getInstance().getGlobalScheme();
   @NotNull
-  private final Map<Level, SimpleTextAttributes> textAttributesByLogLevelCached = new ConcurrentHashMap<>();
+  private final Map<Level, SimpleTextAttributes> textAttributesByLogLevelCache = new ConcurrentHashMap<>();
 
   private class EntryModel implements FlutterLogTree.EntryModel {
     boolean showColors;
@@ -91,15 +90,15 @@ public class FlutterLogView extends JPanel implements ConsoleView, DataProvider,
 
     @NotNull
     private SimpleTextAttributes getTextAttributesByLogLevel(@NotNull Level level) {
-      if (textAttributesByLogLevelCached.containsKey(level)) {
-        return textAttributesByLogLevelCached.get(level);
+      if (textAttributesByLogLevelCache.containsKey(level)) {
+        return textAttributesByLogLevelCache.get(level);
       }
       try {
         final TextAttributesKey key = LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.get(level);
         final Color color = globalEditorColorsScheme.getAttributes(key).getForegroundColor();
 
         final SimpleTextAttributes result = new SimpleTextAttributes(SimpleTextAttributes.STYLE_PLAIN, color);
-        textAttributesByLogLevelCached.put(level, result);
+        textAttributesByLogLevelCache.put(level, result);
         return result;
       }
       catch (Exception e) {
@@ -369,7 +368,7 @@ public class FlutterLogView extends JPanel implements ConsoleView, DataProvider,
   public FlutterLogView(@NotNull FlutterApp app) {
     this.app = app;
     ApplicationManager.getApplication().getMessageBus().connect(this).subscribe(EditorColorsManager.TOPIC, scheme -> {
-      textAttributesByLogLevelCached.clear();
+      textAttributesByLogLevelCache.clear();
       globalEditorColorsScheme = EditorColorsManager.getInstance().getGlobalScheme();
     });
     final FlutterLog flutterLog = app.getFlutterLog();
--- a/src/io/flutter/logging/FlutterLogView.java
+++ b/src/io/flutter/logging/FlutterLogView.java
@@ -69,8 +69,7 @@ public class FlutterLogView extends JPanel implements ConsoleView, DataProvider,
     LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.SHOUT, FlutterLogConstants.SHOUT_OUTPUT_KEY);
   }
 
-  @NotNull
-  private volatile EditorColorsScheme globalEditorColorsScheme = EditorColorsManager.getInstance().getGlobalScheme();
+
   @NotNull
   private final Map<Level, SimpleTextAttributes> textAttributesByLogLevelCache = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/logging/FlutterLogView.java
```diff
@@ -35,8 +42,28 @@
...
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.FINE, FlutterLogConstants.FINE_OUTPUT_KEY);
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.CONFIG, FlutterLogConstants.CONFIG_OUTPUT_KEY);
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.INFO, FlutterLogConstants.INFO_OUTPUT_KEY);
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.WARNING, FlutterLogConstants.WARNING_OUTPUT_KEY);
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.SEVERE, FlutterLogConstants.SEVERE_OUTPUT_KEY);
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.SHOUT, FlutterLogConstants.SHOUT_OUTPUT_KEY);
+  }
+
+  @NotNull
+  private final EditorColorsScheme globalEditorColorsScheme = EditorColorsManager.getInstance().getGlobalScheme();
```
pq: NIT: perhaps make this static and rename to `GLOBAL_EDITOR_COLORS_SCHEME` or maybe just inline?

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -45,20 +72,26 @@
...
+    @NotNull
+    private SimpleTextAttributes getTextAttributesByLogLevel(@NotNull Level level) {
+      try {
+        final TextAttributesKey key = LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.get(level);
+        final Color color = globalEditorColorsScheme.getAttributes(key).getForegroundColor();
+        return new SimpleTextAttributes(SimpleTextAttributes.STYLE_PLAIN, color);
+      }
+      catch (Exception e) {
+        // Should never go here.
+        e.printStackTrace();
```
pq: Let's add a logger and log this instead of printing a stack trace.

Something like:

    private static final Logger LOG = Logger.getInstance(FlutterLogView.class);

and then:

    LOG.error(e);

?

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -45,20 +72,26 @@
...
+      }
+      return SimpleTextAttributes.GRAY_ATTRIBUTES;
```
pq: Maybe default to `REGULAR_ATTRIBUTES`?

quangson91: Yeah, I updated.

src/io/flutter/logging/FlutterLogView.java
quangson91: I updated.
Thanks!

quangson91:
@pq I have addressed your comments.
Please take a look!.

Also, I'm working on keep user preference setting.
(Mean, keep user setting filter eg regex/match case/show timestamp/show color entry, etc.)
The code almost finished, now I'm testing.
So I will create a new PR soon.

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -352,7 +358,9 @@ String countString(int count) {
 
   public FlutterLogView(@NotNull FlutterApp app) {
     this.app = app;
-
+    ApplicationManager.getApplication().getMessageBus().connect(this).subscribe(EditorColorsManager.TOPIC, scheme -> {
+      globalEditorColorsScheme = EditorColorsManager.getInstance().getGlobalScheme();
```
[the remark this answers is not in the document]
pq: Got it.  Thanks!

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -45,20 +78,26 @@
...
       }
       return SimpleTextAttributes.REGULAR_ATTRIBUTES;
     }
+
+    @NotNull
+    private SimpleTextAttributes getTextAttributesByLogLevel(@NotNull Level level) {
+      try {
+        final TextAttributesKey key = LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.get(level);
+        final Color color = globalEditorColorsScheme.getAttributes(key).getForegroundColor();
+        return new SimpleTextAttributes(SimpleTextAttributes.STYLE_PLAIN, color);
```
pq: Does this create a new `SimpleTextAttributes` object for each entry?  If so, would it make sense to cache them?

quangson91: Ah, you right.
We should be caching it. So we don't have re-created new `SimpleTextAttributes` value.
I implemented in this commit: https://github.com/flutter/flutter-intellij/pull/2399/commits/72ab8ff2668a61c0ff07747e2bca5dde782078ba

Please take a look!

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -45,20 +82,32 @@
...
+      if (textAttributesByLogLevelCached.containsKey(level)) {
+        return textAttributesByLogLevelCached.get(level);
+      }
+      try {
+        final TextAttributesKey key = LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.get(level);
+        final Color color = globalEditorColorsScheme.getAttributes(key).getForegroundColor();
+
+        final SimpleTextAttributes result = new SimpleTextAttributes(SimpleTextAttributes.STYLE_PLAIN, color);
+        textAttributesByLogLevelCached.put(level, result);
+        return result;
```
pq: I may be misreading, but does this ever get things from the cache?

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -35,8 +43,37 @@
...
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.INFO, FlutterLogConstants.INFO_OUTPUT_KEY);
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.WARNING, FlutterLogConstants.WARNING_OUTPUT_KEY);
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.SEVERE, FlutterLogConstants.SEVERE_OUTPUT_KEY);
+    LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP.put(Level.SHOUT, FlutterLogConstants.SHOUT_OUTPUT_KEY);
+  }
+
+  @NotNull
+  private volatile EditorColorsScheme globalEditorColorsScheme = EditorColorsManager.getInstance().getGlobalScheme();
+  @NotNull
+  private final Map<Level, SimpleTextAttributes> textAttributesByLogLevelCached = new ConcurrentHashMap<>();
```
pq: Maybe consider renaming to `textAttributesByLogLevelCache` ?

src/io/flutter/logging/FlutterLogView.java
pq: Also, how would a map of levels to colors get updated when `globalEditorColorsScheme` changes?

Another approach might be to keep a cache of colors to attributes...

For example:

```java
Map<Color, SimpleTextAttributes> attributeCache = new ConcurrentHashMap<>(); 

...

SimpleTextAttributes result = attributeCache.get(color); // or containsKey(...)
if (result == null) {
  result = ...
  attributeCache.put(color, result);
}
return result;
```

Then we don't have to worry about level cache consistency.

src/io/flutter/logging/FlutterLogView.java
quangson91: Done.

src/io/flutter/logging/FlutterLogView.java
quangson91: I think we have to cache: `Map<Level, SimpleTextAttributes> textAttributesByLogLevelCache`.

Reason:
Cached values is used in method:
```java
public SimpleTextAttributes style(@Nullable FlutterLogEntry entry, int attributes) {
```

=> input
```java
final FlutterLog.Level level = FlutterLog.Level.forValue(entry.getLevel());
```
=> Ouput
```java
SimpleTextAttributes
```

=> `Map<Level, SimpleTextAttributes>` is right choise.

quangson91: When global color scheme changed -> we simply reset cache data.

https://github.com/flutter/flutter-intellij/pull/2399/commits/72ab8ff2668a61c0ff07747e2bca5dde782078ba#diff-2a35badbf1bca121cb7f36039ac67315R372

quangson91: > I may be misreading, but does this ever get things from the cache?

Yes, I tested. I'm sure it will get value from cache.

quangson91:
@pq, I already refactoring logic storing text-attributes.
Please see this commit: https://github.com/flutter/flutter-intellij/pull/2399/commits/91b3c7be6b49be6744ed4d23bfca38d9904d1d1d

Issue #2403, opened 2018-06-22
lots of cpu usage with many rows in the Logging view


devoncarew:
I see heavy CPU usage (and what looks like a frozen UI) with lots of rows in the logging view.

<img width="321" alt="screen shot 2018-06-21 at 8 29 55 pm" src="https://user-images.githubusercontent.com/1269969/41756620-4ebea6c8-7592-11e8-85eb-8bedc7ab38a1.png">

<img width="333" alt="screen shot 2018-06-21 at 8 30 43 pm" src="https://user-images.githubusercontent.com/1269969/41756621-4ed8047e-7592-11e8-8f5d-acec673ddbcf.png">

devoncarew:
It may actually be in a non-terminating loop in some way? I don't see the UI un-freezing.

```
hashCode:-1, Object (java.lang)
hashCode:277, TreePath (javax.swing.tree)
put:464, Hashtable (java.util)
addMapping:693, VariableHeightLayoutCache (javax.swing.tree)
access$000:55, VariableHeightLayoutCache (javax.swing.tree)
setParent:1067, VariableHeightLayoutCache$TreeStateNode (javax.swing.tree)
insert:187, DefaultMutableTreeNode (javax.swing.tree)
add:411, DefaultMutableTreeNode (javax.swing.tree)
expand:1486, VariableHeightLayoutCache$TreeStateNode (javax.swing.tree)
expand:1288, VariableHeightLayoutCache$TreeStateNode (javax.swing.tree)
rebuild:743, VariableHeightLayoutCache (javax.swing.tree)
setModel:109, VariableHeightLayoutCache (javax.swing.tree)
configureLayoutCache:1821, BasicTreeUI (javax.swing.plaf.basic)
completeUIInstall:689, BasicTreeUI (javax.swing.plaf.basic)
completeUIInstall:127, WideSelectionTreeUI (com.intellij.util.ui.tree)
installUI:649, BasicTreeUI (javax.swing.plaf.basic)
setUI:666, JComponent (javax.swing)
setUI:700, JTree (javax.swing)
setUI:124, Tree (com.intellij.ui.treeStructure)
updateUI:716, JTree (javax.swing)
updateUI:54, TreeTableTree (com.intellij.ui.treeStructure.treetable)
updateUI:139, TreeTable (com.intellij.ui.treeStructure.treetable)
update:313, FlutterLogTree$TreeModel (io.flutter.logging)
[...]

PR #2409, opened 2018-06-22
Remove gratuitous logger model updates.

pq:
Greatly improves performance w/ busy event streams and long entry lists.

May address: #2403.

/cc @scheglov @devoncarew @quangson91

### DIFF
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -340,20 +340,24 @@ public class FlutterLogTree extends TreeTable {
     }
 
     public void appendNodes(List<FlutterLogEntry> entries) {
-      ApplicationManager.getApplication().invokeLater(() -> {
+      if (treeTable == null || uiThreadAlarm.isDisposed()) {
+        return;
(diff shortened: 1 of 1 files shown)


src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -340,20 +340,24 @@ public void clearEntries() {
     }
 
     public void appendNodes(List<FlutterLogEntry> entries) {
-      ApplicationManager.getApplication().invokeLater(() -> {
+      if (treeTable == null || uiThreadAlarm.isDisposed()) {
+        return;
+      }
+
+      uiThreadAlarm.cancelAllRequests();
+      uiThreadAlarm.addRequest(() -> {
```
devoncarew: I don't know enough about how this work, but if we cancel older requests, will the work they do to append `entries` not get scheduled?

pq: Right.  Luckily it won't matter.  The simple implementation is non-incremental; each reload fetches the entries fresh.

PR #2412, opened 2018-06-23
create flutter log preferences

quangson91:
Create a class to store flutter log setting.

// cc @pq, please take a look in next week!

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -592,6 +592,8 @@
                     serviceImplementation="io.flutter.dart.FlutterDartAnalysisServer"/>
     <projectService serviceInterface="io.flutter.bazel.WorkspaceCache"
                     serviceImplementation="io.flutter.bazel.WorkspaceCache"/>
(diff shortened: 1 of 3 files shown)


pq:
Super!

Looking at your other PR, it  seems like you were imaging that settings would "stick"?  Alternatively I wonder if we shouldn't consider a preference page to set defaults.

Thoughts?

What does logcat do?  (Does it remember settings between runs?)

quangson91:
> What does logcat do? (Does it remember settings between runs?)

=> the locat keeps preference too.

I think we should keep user preference. For example, someone prefer show `sequence` column -> if the default setting is hide `sequence` => each time user open => they have to enable show `sequence` column.

quangson91:
Also, the default value in `FlutterLogPreferences` is also default setting which will update when user change setting (eg show sequence column).

(https://github.com/flutter/flutter-intellij/pull/2412/files#diff-97e0b7151569d1ac6c45c4283a3eaa45R14)

PR #2415, opened 2018-06-24
[WIP] improve log filtering

quangson91:
The basic implement to improve log view.

There are some notes:

1. I use `JBTable` instead `TreeView` 
(I'm not sure our planning for logging view. If we that log event -> JBTable is enough).

2. That basic implement to show how filtering work without clear all row views & added it again.
  + Rref: https://docs.oracle.com/javase/tutorial/uiswing/components/table.html#sorting
  + I used `JBTable` instead `TreeView` because I don't know how to impelment `sorting/filtering` with `TreeView`. But if we have to force use `TreeView` -> I will double check.

I think the performance is better, I tested & tried with many logs at the same time.

Demo: https://www.youtube.com/watch?v=ZVp2eKEPdU4
(Please check youtube link, I tried using gif, but the file too big).

### DIFF
--- a/src/io/flutter/logging/FlutterLogConstants.java
+++ b/src/io/flutter/logging/FlutterLogConstants.java
@@ -22,6 +22,14 @@ public final class FlutterLogConstants {
     public static final String LEVEL = "Level";
     public static final String CATEGORY = "Category";
     public static final String MESSAGE = "Message";
+
+    public static final String[] ALL_COLUMNS = {
+      TIME,
+      SEQUENCE,
+      LEVEL,
+      CATEGORY,
+      MESSAGE
+    };
   }
 
   @NotNull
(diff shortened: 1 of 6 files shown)


pq:
Thanks for all of this work!

> I used JBTable instead TreeView because I don't know how to impelment sorting/filtering with TreeView. But if we have to force use TreeView -> I will double check.

If you have the time and energy, it would be great if you could double-check.  It wasn't obvious to me how to get it to work which is how I ended up with the current simple implementation.

Looking ahead, migrating to a table is definitely something to consider, especially if making a tree version performant continues to be a challenge.  Since a tree view affords us some flexibility for representing things like stack traces and complex structures (like JSON), which I hope we can implement soon, I'd like to push a little harder before we give up though.  Does that makes sense?

Again, thanks!

quangson91:
@pq , I'm closing this PR as we can able to implement it on `TreeTable`.

PR #2418, opened 2018-06-25
remove the Inspector's empty content message

devoncarew:
- remove the Inspector's empty content message

I hit another issue where an app was running, but the inspector was not displaying info about the current app (rather, no tab was selected - you had to manually click on the app's tab). I'm removing the empty content message so user's don't run into this issue.


devoncarew:
@jacob314

PR #2419, opened 2018-06-25
some tweaks to the open in android studio functionality

devoncarew:
Some tweaks to the open in android studio functionality:
- `Open for Editing in Android Studio` changed to `Open module in Android Studio` (which, I think adds some clarity, even if the action will later then open the individual file)
- provide instructions about how to use `flutter config --android-studio-dir ...`
- no longer cache the results of running `flutter config --machine`; this is an infrequent user operation, and this way we'll use the latest value if it was changed after AS started up
- added a todo: about using ourself as the AS location, if running from AS

@stevemessick


src/io/flutter/FlutterBundle.properties
```diff
@@ -165,7 +165,7 @@ old.android.studio.message=Your path contains Android Studio 2.x,\n\
 
 flutter.androidstudio.open.module.text=Open Android module in Android Studio
 flutter.androidstudio.open.module.description=Open Android Studio in a new window to edit the top-level Android project
-flutter.androidstudio.open.file.text=Open for Editing in Android Studio
+flutter.androidstudio.open.file.text=Open module in Android Studio
```
stevemessick: Since @mit-mit suggested the current label in https://github.com/flutter/flutter-intellij/pull/2216 I think he should sign off for this change.

mit-mit: I like that the old string suggested why one would want to open in AS (that the edit part). But perhaps an even longer strong would be better?

Simply 'open module' doesn't really say why one would want to click it.

devoncarew: I think it gives a better sense of what will be opened, and generally, think we should aim for brevity here...

mit-mit: Sure, but that is only useful if you actually understand what the reason for the dialog is in the first place

stevemessick:
My two cents: this isn't actually a dialog, it is the banner at the top of a text editor. As such, it has plenty of room, so shortening the string does not seem like a strong motivation. I'd cast a vote for the label that makes it most clear what's going on but I don't see a strong winner here. 

This whole area of opening (again) to edit the Android module is confusing and perhaps not well understood by users. Hopefully this will just go away in Q3.

Hixie:
@devoncarew What is your plan with this PR? (No particular rush, I'm just cleaning out the review queue.)

devoncarew:
@Hixie, it's less than a day old! My plan is to nurture it for awhile.

Hixie:
Oh, my bad. Sorry. :-) It was one of the least-recently-modified PRs, so I assumed it was old... I guess we've modified a lot of our PRs recently!

devoncarew:
Hmm, I forgot about this one; I'll merge in the current master, remove the label change, and see if what's left is still viable.

devoncarew:
@stevemessick, this is slightly simpler now, and freshly merged with the latest from master - ready for another look.

src/io/flutter/actions/OpenInAndroidStudioAction.java
```diff
@@ -189,20 +190,24 @@ private static String findAndroidStudio(@Nullable Project project) {
       return null;
     }
 
+    // TODO: If we're running in Android Studio (FlutterUtils.isAndroidStudio()), we should return ourself as the location.
```
stevemessick: If running in Android Studio we never get here. It uses a subclass of this class to open a new project window rather than trying to launch itself.

devoncarew: gotcha - will delete the comment

PR #2422, opened 2018-06-25
Update build for AS 3.2 beta

stevemessick:
@devoncarew @pq 

That was easy.

### DIFF
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -15,7 +15,7 @@
       "name": "IntelliJ",
       "version": "2018.1",
       "ideaProduct": "android-studio",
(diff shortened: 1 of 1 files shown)


product-matrix.json
```diff
@@ -15,7 +15,7 @@
       "name": "IntelliJ",
       "version": "2018.1",
       "ideaProduct": "android-studio",
-      "ideaVersion": "181.4830125",
+      "ideaVersion": "181.4847800",
```
devoncarew: We should also update the comment (`AS 3.2 Canary...`) above.

PR #2413, opened 2018-06-23
WIP update setting pref for flutter log

quangson91:
Update setting pref for flutter log including:
+ show column level/timestamp/category/seq, show color 
+ checkbox regex/match case/ log level.

PR #2430, opened 2018-06-26
extract renderers

pq:
Follow-up from: https://github.com/flutter/flutter-intellij/pull/2413#issuecomment-400318578

/cc @quangson91 @scheglov

### DIFF
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -93,7 +93,7 @@ public class FlutterLogTree extends TreeTable {
 
       @Override
       public TableCellRenderer getCustomizedRenderer(DefaultMutableTreeNode o, TableCellRenderer renderer) {
-        return renderer;
+        return this.renderer;
       }
     }
 
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -68,15 +68,17 @@ public class FlutterLogTree extends TreeTable {
       abstract void render(FlutterLogEntry entry);
     }
 
-    abstract class Column extends ColumnInfo<DefaultMutableTreeNode, FlutterLogEntry> {
+    class Column extends ColumnInfo<DefaultMutableTreeNode, FlutterLogEntry> {
       boolean show = true;
-      private TableCellRenderer renderer;
+      @NotNull
+      private final TableCellRenderer renderer;
 
-      Column(String name) {
+      Column(@NotNull String name, @NotNull TableCellRenderer renderer) {
         super(name);
+        this.renderer = renderer;
       }
 
-      boolean isVisible() {
+      boolean isShowing() {
         return show;
       }
 
@@ -91,13 +93,8 @@ public class FlutterLogTree extends TreeTable {
 
       @Override
       public TableCellRenderer getCustomizedRenderer(DefaultMutableTreeNode o, TableCellRenderer renderer) {
-        if (renderer == null) {
-          renderer = createRenderer();
-        }
         return renderer;
       }
-
-      abstract TableCellRenderer createRenderer();
     }
 
     private class TimeCellRenderer extends EntryCellRenderer {
@@ -184,36 +181,13 @@ public class FlutterLogTree extends TreeTable {
     ColumnModel(@NotNull FlutterApp app, @NotNull EntryModel entryModel) {
       this.app = app;
       this.entryModel = entryModel;
-      columns.add(new Column("Time") {
-        @Override
-        TableCellRenderer createRenderer() {
-          return new TimeCellRenderer();
-        }
-      });
-      columns.add(new Column("Sequence") {
-        @Override
-        TableCellRenderer createRenderer() {
-          return new SequenceCellRenderer();
-        }
-      });
-      columns.add(new Column("Level") {
-        @Override
-        TableCellRenderer createRenderer() {
-          return new LevelCellRenderer();
-        }
-      });
-      columns.add(new Column("Category") {
-        @Override
-        TableCellRenderer createRenderer() {
-          return new CategoryCellRenderer();
-        }
-      });
-      columns.add(new Column("Message") {
-        @Override
-        TableCellRenderer createRenderer() {
-          return new MessageCellRenderer(app.getModule());
-        }
-      });
+      columns.addAll(Arrays.asList(
+        new Column("Time", new TimeCellRenderer()),
+        new Column("Sequence", new SequenceCellRenderer()),
+        new Column("Level", new LevelCellRenderer()),
+        new Column("Category", new CategoryCellRenderer()),
+        new Column("Message", new MessageCellRenderer(app.getModule()))
+      ));
       // Cache for quick access.
       visible = columns.stream().filter(c -> c.show).collect(Collectors.toList());
     }
@@ -261,7 +235,7 @@ public class FlutterLogTree extends TreeTable {
     }
 
     public TableCellRenderer getRenderer(int column) {
-      return visible.get(column).createRenderer();
+      return visible.get(column).renderer;
     }
 
     public boolean isShowing(String column) {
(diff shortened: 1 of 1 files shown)


pq:
@quangson91 : are you able to "review" this?  If so, could you add one?

https://help.github.com/articles/about-pull-request-reviews/

Thanks for the quick feedback! 👍

quangson91:
@pq, Yeap. I'm checking this PR. ^^

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -68,6 +68,68 @@ void appendStyled(FlutterLogEntry entry, String text) {
...
+    abstract class Column extends ColumnInfo<DefaultMutableTreeNode, FlutterLogEntry> {
+      boolean show = true;
+      private TableCellRenderer renderer;
+
+      Column(String name) {
+        super(name);
+      }
+
+      boolean isVisible() {
+        return show;
```
quangson91: `show` vs `isVisible`
-> I think it should be same.
I mean, both should be `boolean show` && `boolean isShow()`
OR
`boolean isVisible` && `boolean isVisible()`

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -68,6 +68,68 @@ void appendStyled(FlutterLogEntry entry, String text) {
       abstract void render(FlutterLogEntry entry);
     }
 
+    abstract class Column extends ColumnInfo<DefaultMutableTreeNode, FlutterLogEntry> {
+      boolean show = true;
+      private TableCellRenderer renderer;
```
quangson91: add `@Nullable` maybe better for IDE lining.

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -68,6 +68,68 @@ void appendStyled(FlutterLogEntry entry, String text) {
...
+
+      Column(String name) {
```
quangson91: `@NotNull`

pq:
Cool.  Let's see if we can make it more formal.  👍 

@quangson91, when you go to:

https://github.com/flutter/flutter-intellij/pull/2430/files

do you see a review changes button in the top right?

image

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -68,6 +68,68 @@ void appendStyled(FlutterLogEntry entry, String text) {
...
+
+      @Override
+      public TableCellRenderer getCustomizedRenderer(DefaultMutableTreeNode o, TableCellRenderer renderer) {
+        if (renderer == null) {
+          renderer = createRenderer();
+        }
+        return renderer;
+      }
+
+      abstract TableCellRenderer createRenderer();
```
quangson91: Nullable OR NotNull
(So sub-class will know this method can be return null or not)

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -68,6 +68,68 @@ void appendStyled(FlutterLogEntry entry, String text) {
...
+        }
+        return null;
+      }
+
+      @Override
+      public TableCellRenderer getCustomizedRenderer(DefaultMutableTreeNode o, TableCellRenderer renderer) {
+        if (renderer == null) {
+          renderer = createRenderer();
+        }
+        return renderer;
```
quangson91: For me, I think we can make `render` as a `final` variable then we will init it in the constructor.
```java
      @NotNull
      private final TableCellRenderer renderer;

      Column(String name) {
        super(name);
        renderer = createRenderer();
      }
```

So in this method we can simply
```java
      @Override
      public TableCellRenderer getCustomizedRenderer(DefaultMutableTreeNode o, TableCellRenderer renderer) {
        return renderer;
      }
```

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -157,47 +187,25 @@ public TableCellRenderer getCustomizedRenderer(DefaultMutableTreeNode o, TableCe
       columns.add(new Column("Time") {
         @Override
         TableCellRenderer createRenderer() {
-          return new EntryCellRenderer() {
-            @Override
-            void render(FlutterLogEntry entry) {
-              appendStyled(entry, TIMESTAMP_FORMAT.format(entry.getTimestamp()));
-            }
-          };
+          return new TimeCellRenderer();
```
quangson91: Hmn, It looks verbose.
If we custom `XXXCellRender` => I think we could do it more simply by passing `cellRender` as a parameter in constructor.
```java
    class Column extends ColumnInfo<DefaultMutableTreeNode, FlutterLogEntry> {
      boolean show = true;
      @NotNull
      private final TableCellRenderer renderer;

      Column(String name, @NotNull TableCellRenderer renderer) {
        super(name);
        this.renderer = renderer;
      }
```

So the code add columns will become:

```java
      columns.add(new Column("Time", new TimeCellRenderer()));
      columns.add(new Column("Sequence", new SequenceCellRenderer()));
      columns.add(new Column("Level", new LevelCellRenderer()));
      columns.add(new Column("Category", new CategoryCellRenderer()));
      columns.add(new Column("Message", new MessageCellRenderer(app.getModule()));
```

OR
```java
      columns.addAll(Arrays.asList(
        new Column("Time", new TimeCellRenderer()),
        new Column("Sequence", new SequenceCellRenderer()),
        new Column("Level", new LevelCellRenderer()),
        new Column("Category", new CategoryCellRenderer()),
        new Column("Message", new MessageCellRenderer(app.getModule()))
      ));
```

quangson91:
@pq, Yeap, I can review the PR ^^.

Thank you!. I just added some comments.

quangson91:
Just notes:
After working in this class many times, I think this class becomes bigger.
So we should consider extracting `Column/CellRender` to a single class. (then make it as `package access`).

src/io/flutter/logging/FlutterLogTree.java
pq: Agreed!

src/io/flutter/logging/FlutterLogTree.java
pq: 👍

src/io/flutter/logging/FlutterLogTree.java
pq: 👍

src/io/flutter/logging/FlutterLogTree.java
pq: There's something clumsy here I agree.  I have some thoughts about this but let's follow-up on a bigger round of cleanup.

src/io/flutter/logging/FlutterLogTree.java
pq: Nice!  Agreed.

pq:
> After working in this class many times, I think this class becomes bigger.

Yeah.   This is a common problem.

> So we should consider extracting Column/CellRender to a single class.

I'll take a crack at this.  I agree it could be easier to work on down the road.

Thanks!

src/io/flutter/logging/FlutterLogTree.java
pq: 👍

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -68,6 +68,65 @@ void appendStyled(FlutterLogEntry entry, String text) {
...
+      public FlutterLogEntry valueOf(DefaultMutableTreeNode node) {
+        if (node instanceof FlutterEventNode) {
+          return ((FlutterEventNode)node).entry;
+        }
+        return null;
+      }
+
+      @Override
+      public TableCellRenderer getCustomizedRenderer(DefaultMutableTreeNode o, TableCellRenderer renderer) {
+        return renderer;
```
quangson91: We should return `this. renderer`
(Because method's parameters also named renderer)

pq: 👍

PR #2438, opened 2018-06-26
Update cursor on log entry link hover.

pq:
Update cursor on link hover.

image

/cc @devoncarew @scheglov


Hixie:
cc @devoncarew for review

Issue #2340, opened 2018-06-08
Logger view punchlist


devoncarew:
Some feedback from using the Logger view; feel free to break into separate issues:

- [x] when appending, the last item doesn't scroll into view (the last - 1 does)
- [x] it would be great if the text filter was live (filtering as you type, not just on return)
- [x] long text is truncated - we need to query the VM for the full value (which will be async, so we'll need to be careful to keep the original ordering of the log items) (#3052)
- [ ] we should display the 'error' and 'stackTrace' fields from dart:developer log() calls
- [x] we should have a way to display the severity (or at least, call out error level and warning level)
- [x] decide whether we do need to display the timestamp column (#2457)
- [x] we should devote more of the horizontal space to the message portion of the log entry (#2457)
- [x] stack traces don't look clickable, and don't jump to a location when clicked (#2402)
- [x] we should provide a visual separator after reloads and restarts (in the text console, we clear the console) (#2458)

/cc @pq

devoncarew:
One more item; when run in Debug mode, the view has less horizontal space available to it:

<img width="1348" alt="screen shot 2018-06-19 at 7 33 01 pm" src="https://user-images.githubusercontent.com/1269969/41664514-c2e87184-745a-11e8-854c-b9613d58e85d.png">

Much of the text that I want to see (the log message text) is now off the screen. We should consider making the non-message portion of the log entry more compact. Some ideas are to remove the entry timestamp, to move the entry timestamp column to the right side of the view, to use icons instead of text labels for log severity, ...

pq:
Awesome feedback.  Along these lines a few things I've been mulling over:

* how "sticky" should modifications be?  (IOW: if I hide a column once, should it stick to the next session?)
* would auto-hiding columns in narrow contexts be surprising?
* would it be odd to have different default layouts in run vs. debug views?

Input welcome!

quangson91:
@pq @devoncarew for me, I think we should make two things:
1. Bring console as a TAB beside `Debugger` tab.
2. Make log message in FlutterLogTree can `wrap text` (I mean display full text in multiple lines).

devoncarew:
> would auto-hiding columns in narrow contexts be surprising?

It would be great if the user didn't need to hide columns - if the default layout was one we thought they could run with.

quangson91:
I would be great if we can make default layout like this one:

[...]

PR #2439, opened 2018-06-27
Save & Load flutter log filter preferences.

quangson91:
https://github.com/flutter/flutter-intellij/issues/2340
+ Save & Load flutter log filter preferences.
So the user can keep filtering option between session.

// cc @pq for review

### DIFF
--- a/src/io/flutter/logging/FlutterLogFilterPanel.java
+++ b/src/io/flutter/logging/FlutterLogFilterPanel.java
@@ -98,12 +98,12 @@ public class FlutterLogFilterPanel {
   private JComboBox<FlutterLog.Level> logLevelComboBox;
 
   public FlutterLogFilterPanel() {
-    matchCaseCheckBox.addItemListener(e -> filterListener().onFilter(getCurrentFilterParam()));
-    regexCheckBox.addItemListener(e -> filterListener().onFilter(getCurrentFilterParam()));
+    matchCaseCheckBox.addItemListener(e -> getFilterListener().onFilter(getCurrentFilterParam()));
+    regexCheckBox.addItemListener(e -> getFilterListener().onFilter(getCurrentFilterParam()));
     final List<FlutterLog.Level> logLevels = Arrays.stream(FlutterLog.Level.values())
       .collect(Collectors.toList());
     logLevelComboBox.setModel(new CollectionComboBoxModel<>(logLevels));
-    logLevelComboBox.addActionListener(event -> filterListener().onFilter(getCurrentFilterParam()));
+    logLevelComboBox.addActionListener(event -> getFilterListener().onFilter(getCurrentFilterParam()));
     logLevelComboBox.setRenderer(new ColoredListCellRenderer<FlutterLog.Level>() {
       @Override
       protected void customizeCellRenderer(@NotNull JList<? extends FlutterLog.Level> list,
@@ -130,9 +130,9 @@ public class FlutterLogFilterPanel {
   }
 
   public void updateFromPreferences(@NotNull FlutterLogPreferences flutterLogPreferences) {
-    regexCheckBox.setSelected(flutterLogPreferences.TOOL_WINDOW_REGEX);
-    matchCaseCheckBox.setSelected(flutterLogPreferences.TOOL_WINDOW_MATCH_CASE);
-    logLevelComboBox.setSelectedItem(FlutterLog.Level.forValue(flutterLogPreferences.TOOL_WINDOW_LOG_LEVEL));
+    regexCheckBox.setSelected(flutterLogPreferences.isToolWindowRegex());
+    matchCaseCheckBox.setSelected(flutterLogPreferences.isToolWindowMatchCase());
+    logLevelComboBox.setSelectedItem(FlutterLog.Level.forValue(flutterLogPreferences.getToolWindowLogLevel()));
   }
 
   public void setOnFilterListener(@Nullable OnFilterListener onFilterListener) {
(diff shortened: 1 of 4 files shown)


src/io/flutter/logging/FlutterLogFilterPanel.java
```diff
@@ -66,30 +71,39 @@ public boolean equals(Object o) {
...
   public interface OnFilterListener {
     void onFilter(@NotNull FilterParam param);
   }
 
-  @NotNull
-  private final OnFilterListener onFilterListener;
+  private static final OnFilterListener DEFAULT_ON_FILTER_LISTENER = param -> LOG.info("Requests filter: " + param);
+
+  @Nullable
+  private volatile OnFilterListener onFilterListener;
```
quangson91: I make `onFilterListener` as nullable because we have update preferences.
So while update preferences (user setting before) => all listener should not call.

src/io/flutter/logging/FlutterLogFilterPanel.java
```diff
@@ -139,7 +163,13 @@ private void createUIComponents() {
   @NotNull
   private SearchTextField createSearchTextField() {
     final LogFilterTextField logFilterTextField = new LogFilterTextField();
-    logFilterTextField.setOnFilterListener(() -> onFilterListener.onFilter(getCurrentFilterParam()));
+    logFilterTextField.setOnFilterListener(() -> filterListener().onFilter(getCurrentFilterParam()));
     return logFilterTextField;
   }
+
+  @NotNull
+  private OnFilterListener filterListener() {
```
pq: Perhaps `getFilterListener()`?

src/io/flutter/logging/FlutterLogFilterPanel.java
```diff
@@ -139,7 +163,13 @@ private void createUIComponents() {
...
+    final OnFilterListener listener = onFilterListener;
+    return listener == null ? DEFAULT_ON_FILTER_LISTENER : listener;
```
pq: Curious why not just:

    return onFilterListener == null ? DEFAULT_ON_FILTER_LISTENER : onFilterListener;

src/io/flutter/logging/FlutterLogPreferences.java
```diff
@@ -18,7 +18,7 @@
   public boolean SHOW_LOG_CATEGORY = true;
   public boolean SHOW_COLOR = true;
 
-  public int TOOL_WINDOW_LOG_LEVEL = 800; // Level INFO
+  public int TOOL_WINDOW_LOG_LEVEL = FlutterLog.Level.NONE.value;
```
pq: Looking at this again I realize that the naming confuses me.  In general, I think ALL_CAPS are reserved for static finals.

Maybe consider changing these?

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -156,11 +156,12 @@ private DefaultActionGroup createPopupActionGroup() {
 
     @Override
     public boolean isSelected(AnActionEvent e) {
-      return logModel.shouldShowTimestamps();
+      return flutterLogPreferences.SHOW_TIMESTAMP;
```
pq: Again, this looks odd.  It reads like you're accessing a static final field...

I suppose I'd expect access (since it's in another class / file) to look like:

    flutterLogPreferences.shouldShowTimestamps();

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -287,7 +292,7 @@ public void actionPerformed(AnActionEvent e) {
         if (paths != null) {
           final Clipboard clipboard = Toolkit.getDefaultToolkit().getSystemClipboard();
           final StringBuilder sb = new StringBuilder();
-          for (final TreePath path: paths) {
+          for (final TreePath path : paths) {
```
pq: It looks like our formatting settings don't agree...  When I format it removes the space before `:`.  Does it not for you?

src/io/flutter/logging/FlutterLogFilterPanel.java
quangson91: Ah, because of this one
screen shot 2018-06-27 at 10 11 06 am

src/io/flutter/logging/FlutterLogFilterPanel.java
quangson91: Yeap. I rename that method.

src/io/flutter/logging/FlutterLogPreferences.java
quangson91: 👍 
I updated.

src/io/flutter/logging/FlutterLogView.java
quangson91: Oh, I use IDE generate get/set.
So it generated to `isShowTimestamps()`.
Should we change to `shouldShowTimestamps()` ?

src/io/flutter/logging/FlutterLogView.java
quangson91: Look likes the coding style is different.
How did you setting code style ?

pq: Interesting.  I'm using our project-level settings.  (And I don't believe I modified them?)

@devoncarew or @stevemessick : what happens when you format

```java
        for (String s : new ArrayList<String>()) {

        }
```

in the flutter-intellij project?

quangson91: My Intellij IDE will keep space before `:`.

pq: Thanks!  I think this is an issue w/ defaults in EAP 2018.2.

quangson91: Oh, I got it.

alexander-doroshko: Looks like the final decision is that IntelliJ 2018.2 will be adding spaces before colons in for-each loops by default. It's configurable of course. See today's [blog post](https://blog.jetbrains.com/idea/2018/07/intellij-idea-2018-2-beta-2-is-available/):
`Starting with this build IntelliJ IDEA will, by default, add spaces before colons in a for-each loop.`

pq: Thanks for following up @alexander-doroshko !

PR #2441, opened 2018-06-27
lower case the log level names

devoncarew:
- display the log level names in lower case
- default the filter level to `config`
- do display the log filter level all the time (previously, we didn't display for the `none` level)

@pq @quangson91

### DIFF
--- a/src/io/flutter/logging/FlutterLog.java
+++ b/src/io/flutter/logging/FlutterLog.java
@@ -50,16 +50,24 @@ public class FlutterLog {
       this.value = value;
     }
 
+    @NotNull
     public static Level forValue(int value) {
       final Level[] levels = Level.values();
-      for (int i = 0; i < levels.length; ++i) {
-        if (value >= levels[i].value && ((i >= levels.length - 1) || value < levels[i + 1].value)) {
+
+      for (int i = levels.length - 1; i >= 0; i--) {
+        if (value >= levels[i].value) {
           return levels[i];
         }
       }
-      return null;
+
+      return NONE;
+    }
+
+    public String toDisplayString() {
+      return name().toLowerCase();
     }
   }
+
   private final EventDispatcher<Listener>
     dispatcher = EventDispatcher.create(Listener.class);
 
--- a/src/io/flutter/logging/FlutterLogFilterPanel.java
+++ b/src/io/flutter/logging/FlutterLogFilterPanel.java
@@ -103,6 +103,7 @@ public class FlutterLogFilterPanel {
     final List<FlutterLog.Level> logLevels = Arrays.stream(FlutterLog.Level.values())
       .collect(Collectors.toList());
     logLevelComboBox.setModel(new CollectionComboBoxModel<>(logLevels));
+    logLevelComboBox.setSelectedItem(FlutterLog.Level.CONFIG);
     logLevelComboBox.addActionListener(event -> getFilterListener().onFilter(getCurrentFilterParam()));
     logLevelComboBox.setRenderer(new ColoredListCellRenderer<FlutterLog.Level>() {
       @Override
@@ -111,8 +112,7 @@ public class FlutterLogFilterPanel {
                                            int index,
                                            boolean selected,
                                            boolean hasFocus) {
-        final String label = index == -1 && value == FlutterLog.Level.NONE ? "" : value.name().toLowerCase();
-        append(label);
+        append(value.toDisplayString());
       }
     });
   }
--- a/src/io/flutter/logging/FlutterLogFilterPanel.java
+++ b/src/io/flutter/logging/FlutterLogFilterPanel.java
@@ -5,7 +5,9 @@
  */
 package io.flutter.logging;
 
-import com.intellij.ui.*;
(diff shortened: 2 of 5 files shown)


src/io/flutter/logging/FlutterLog.java
```diff
@@ -50,16 +50,24 @@
...
+        if (value >= levels[i].value) {
           return levels[i];
         }
       }
-      return null;
+
+      return NONE;
+    }
+
+    public String toDisplayString() {
```
quangson91: Maybe add `@NotNull`

src/io/flutter/logging/FlutterLogFilterPanel.java
```diff
@@ -80,15 +82,15 @@ public int hashCode() {
...
-  ) {
+  public FlutterLogFilterPanel(@NotNull OnFilterListener onFilterListener) {
     this.onFilterListener = onFilterListener;
+
     matchCaseCheckBox.addItemListener(e -> onFilterListener.onFilter(getCurrentFilterParam()));
     regexCheckBox.addItemListener(e -> onFilterListener.onFilter(getCurrentFilterParam()));
     final List<FlutterLog.Level> logLevels = Arrays.stream(FlutterLog.Level.values())
       .collect(Collectors.toList());
     logLevelComboBox.setModel(new CollectionComboBoxModel<>(logLevels));
+    logLevelComboBox.setSelectedItem(FlutterLog.Level.CONFIG);
```
quangson91: I just confused.
In https://github.com/flutter/flutter-intellij/pull/2441/files#diff-c8d60b1f6627b5f4700899a7d7f97b88R36
we defined:
```java
public static int GC_EVENT_LEVEL = FlutterLog.Level.FINER.value;
```
But here you make default selected item is `Level.CONFIG`

=> what's the different ?

src/io/flutter/logging/FlutterLogFilterPanel.java
```diff
@@ -97,8 +99,7 @@ protected void customizeCellRenderer(@NotNull JList<? extends FlutterLog.Level>
                                            int index,
                                            boolean selected,
                                            boolean hasFocus) {
-        final String label = index == -1 && value == FlutterLog.Level.NONE ? "" : value.name().toLowerCase();
-        append(label);
+        append(value.toDisplayString());
```
quangson91: In previous code -> `Level.NONE` will display nothing (empty string).
But for new code -> I think it will display `none`.

src/io/flutter/logging/FlutterLog.java
```diff
@@ -50,16 +50,24 @@
...
+      return name().toLowerCase();
```
quangson91: should we update this look like:
```
return this == NONE ? "" else name().toLowerCase();
```

src/io/flutter/logging/FlutterLog.java
```diff
@@ -50,16 +50,24 @@
       this.value = value;
     }
 
+    @NotNull
```
quangson91: Great! 👍 
I love NotNull ^^.

src/io/flutter/logging/FlutterLog.java
devoncarew: Generally useful I agree, but I tend not to use it on methods where there's no logic in the method, or on people who depend on it.

src/io/flutter/logging/FlutterLogFilterPanel.java
devoncarew: GC events are printed at a slightly more verbose setting than the default. While using the logging view, I found GC events noisy and not adding a ton of value.

We should either filter them out by severity, or add the ability to show and hide categories (gc event, network events, ...).

src/io/flutter/logging/FlutterLogFilterPanel.java
devoncarew: Yup, this is an explicit change. I thought eliding none would be useful, but after using the view for a while I'm not sure. Definitely odd to not show it if the default setting is `config`.

src/io/flutter/logging/FlutterLog.java
devoncarew: I'd like to show the value for none now.

src/io/flutter/logging/FlutterLog.java
devoncarew: 👍

src/io/flutter/logging/FlutterLogFilterPanel.java
pq: >  or add the ability to show and hide categories 

👍 I'd like to see this happen.  https://github.com/flutter/flutter-intellij/issues/2445 to track.

src/io/flutter/logging/FlutterLog.java
quangson91: Yeap. Got it.

src/io/flutter/logging/FlutterLog.java
quangson91: Yeap. Got it.

PR #2453, opened 2018-06-30
Add InheritedWidget and Stateful Widget with Animation live templates

brianegan:
Howdy, y'all! I've been doing more and more animations, and thought the `stanim` snippet from VSCode was pretty cool. Same kinda thing with InheritedWidgets...

I've added two new live templates to quickly stub out the boilerplate for these classes.

They will produce the following code.

### stanim

```dart
class Anim extends StatefulWidget {
  @override
  _AnimState createState() => _AnimState();
}

class _AnimState extends State<Anim> with SingleTickerProviderStateMixin {
  AnimationController _controller;

  @override
  void initState() {
    _controller = AnimationController(vsync: this);
    super.initState();
  }

  @override
  void dispose() {
    _controller.dispose();
    super.dispose();
  }
  
  @override
  Widget build(BuildContext context) {
    return Container();
  }
}
```

### inh

```dart
class Inher extends InheritedWidget {
  const Inher({
    Key key,
    @required Widget child,
  })
      : assert(child != null),
        super(key: key, child: child);

  static Inher of(BuildContext context) {
    return context.inheritFromWidgetOfExactType(Inher) as Inher;
  }

  @override
  bool updateShouldNotify(Inher old) {
    return /* CURSOR WILL BE PLACED HERE AFTER NAMING THE WIDGET SO YOU CAN WRITE THE FUNCTIONALITY */;
  }
}
```

### Example in action

new_templates

### DIFF
--- a/resources/liveTemplates/flutter_miscellaneous.xml
+++ b/resources/liveTemplates/flutter_miscellaneous.xml
@@ -11,4 +11,17 @@
       <option name="DART_TOPLEVEL" value="true" />
     </context>
   </template>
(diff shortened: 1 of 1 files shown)


pq:
Awesome!

@stevemessick did the other assists so adding him to take a peek.  Otherwise, this looks good to me.

Thanks!

Issue #2402, opened 2018-06-21
Hyperlinks not working in log view.


pq:
A regression since column hiding was added.

devoncarew:
This would be nice to have for M26 (but not critical).

pq:
The regression that broke file hyperlinks (`lib/main.dart`) has been fixed; updating to reflect missing support for stack traces.

quangson91:
@pq, I'm working on this issue.
The code almost done, currently, I'm testing.
I will create a PR soon.

quangson91:
I created a new PR for support hyperlink: https://github.com/flutter/flutter-intellij/pull/2454
Please take a look!
Thanks!

devoncarew:
Wow, @quangson91, great stuff!

PR #2454, opened 2018-06-30
support hyperlinks in flutter console log

quangson91:
Issue: https://github.com/flutter/flutter-intellij/issues/2402
+ support hyperlinks in flutter console log

Use `DartConsoleFilter` & `UrlFilter` to support hyper link stacktrace & url in log console.

Notes: stack trace can clickable, but `url` can't click to open in the browser.

Screenshot:
hyperlink_flutter_log

// cc @pq please take a look in next week :)

### DIFF
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -133,18 +133,24 @@ public class FlutterLogTree extends TreeTable {
       @NotNull
       private final FlutterApp app;
       @NotNull
-      private final List<Filter> linkFilters;
+      private final Filter[] filters;
 
       MessageCellRenderer(@NotNull FlutterApp app) {
         this.app = app;
-        linkFilters = new ArrayList<>();
+        filters = createMessageFilters().toArray(new Filter[0]);
+      }
+
+      @NotNull
+      private List<Filter> createMessageFilters() {
+        final List<Filter> filters = new ArrayList<>();
         if (app.getModule() != null) {
-          linkFilters.add(new FlutterConsoleFilter(app.getModule()));
+          filters.add(new FlutterConsoleFilter(app.getModule()));
         }
-        linkFilters.addAll(Arrays.asList(
+        filters.addAll(Arrays.asList(
           new DartConsoleFilter(app.getProject(), app.getProject().getBaseDir()),
           new UrlFilter()
         ));
+        return filters;
       }
 
       @Override
@@ -156,7 +162,7 @@ public class FlutterLogTree extends TreeTable {
           return;
         }
         final List<Filter.ResultItem> resultItems = new ArrayList<>();
-        for (Filter filter : linkFilters) {
+        for (Filter filter : filters) {
           final Filter.Result result = filter.applyFilter(message, message.length());
           if (result == null) {
             continue;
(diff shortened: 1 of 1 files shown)


src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -128,37 +129,48 @@ void render(FlutterLogEntry entry) {
...
         // TODO(pq): SpeedSearchUtil.applySpeedSearchHighlighting
         // TODO(pq): setTooltipText
         final String message = entry.getMessage();
-        int cursor = 0;
+        final List<Filter.ResultItem> resultItems = linkFilters.stream()
+          .map(filter -> filter.applyFilter(message, message.length()))
+          .filter(Objects::nonNull)
+          .flatMap(result -> result.getResultItems().stream())
+          .sorted(Comparator.comparingInt(Filter.ResultItem::getHighlightStartOffset))
+          .collect(Collectors.toList());
```
pq: Oh cool!

One concern: might we consider performance here?  There's a bit of conflicting information about the impact of streams on performance and though I'm generally skeptical of micro-benchmarks, this is a case where I think we should proceed cautiously.  As we all know, `render(..)` gets called very frequently and we should be sure it's very performant with huge lists of entries.

My gut says that it would be worth looking into a more straight-forward array-based implementation here.

Does that feel like premature optimization to you?

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -239,7 +251,7 @@ public TableCellRenderer getRenderer(int column) {
     }
 
     public boolean isShowing(String column) {
-      for (Column c: columns) {
+      for (Column c : columns) {
```
pq: That space again! 😄

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -128,37 +129,48 @@ void render(FlutterLogEntry entry) {
     }
 
     private class MessageCellRenderer extends EntryCellRenderer {
-      @Nullable
-      private final FlutterConsoleFilter consoleFilter;
+      @NotNull
+      private final FlutterApp app;
+      @NotNull
+      private final List<Filter> linkFilters;
```
pq: (See below but I wonder if we might not migrate to a speedier native array here.)

pq:
> Notes: stack trace can clickable, but url can't click to open in the browser.

I wonder if `UrlFilter` in `com.intellij.execution.filters` wouldn't "just work" here?

src/io/flutter/logging/FlutterLogTree.java
quangson91: You right,
I updated to:
```java
        final List<Filter.ResultItem> resultItems = new ArrayList<>();
        for (Filter filter : linkFilters) {
          final Filter.Result result = filter.applyFilter(message, message.length());
          if (result == null) {
            continue;
          }
          resultItems.addAll(result.getResultItems());
        }
        resultItems.sort(Comparator.comparingInt(Filter.ResultItem::getHighlightStartOffset));
```

src/io/flutter/logging/FlutterLogTree.java
quangson91: Yeap, I changed to: `private final Filter[] filters;`

pq:
> But don't know why we click to the link the browser doesn't open :|

Hmmm.  Without a bit of debugging, I'm not sure but happy to help.  This is a great start!

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -128,37 +130,60 @@ void render(FlutterLogEntry entry) {
     }
 
     private class MessageCellRenderer extends EntryCellRenderer {
-      @Nullable
-      private final FlutterConsoleFilter consoleFilter;
+      @NotNull
+      private final FlutterApp app;
+      @NotNull
+      private final Filter[] filters;
```
pq: 💎

Issue #2452, opened 2018-06-30
Improve Copy / Paste on new Logging View


brianegan:
Hey hey :) I've been trying out the new logging view and wanted to provide a bit of feedback wrt copying and pasting logged messages. It's a pretty common task when running into bugs in a larger app where another team might be responsible for the feature you're navigating through and you encounter an error. I might want to create a new ticket with the StackTrace, or paste it into slack for discussion.

### Issue 1: Use cmd+c to copy a log Message

  1. Enable new logging view (applies to all of these)
  2. Run the app
  3. See a log message or many messages you're interested in copying 
  4. Highlight it and press `cmd+c` to copy
  5. Get output like this when pasting: 

```
io.flutter.logging.FlutterLogEntry@76b5b7da | io.flutter.logging.FlutterLogEntry@76b5b7da | io.flutter.logging.FlutterLogEntry@76b5b7da | io.flutter.logging.FlutterLogEntry@76b5b7da
-- | -- | -- | --
```

Workaround: Secondary-click on the message, click "Copy asText". This now produces a result I was expecting:

```
13:30:48.572 2508 800 flutter.tools Restarted app in 1,230ms.
```

or with multiple lines:

```
13:30:48.572 2508 800 flutter.tools Restarted app in 1,230ms.
13:30:47.834 2509 400 runtime.gc collection time 1ms • 26.2MB used of 27.2MB • isolates/21837764
13:30:48.007 2510 400 runtime.gc collection time 2ms • 30.0MB used of 31.2MB • isolates/21837764
```

### Issue 2: Copy a portion of the message

Say I want to filer down to a specific message. Rather than typing it in (error-prone, sometimes long), it'd be great to copy only a portion of the message rather than "Copy as Text" and have to remove the bits I don't want.

  1. Run app with new logging view
  2. Attempt to select only a portion of the logged message, e.g. I wanted to copy "Restarted app" from the message "Restarted app in 1,230ms."
  3. Can't quite see a way to do this

Thanks for all the hard work and community effort on this feature!

quangson91:
I already create a new PR for fixing copy by shortcut: https://github.com/flutter/flutter-intellij/pull/2455
@pq, please check it next week :)

For the issue 2 ( Copy a portion of the message) I think we should have more discussion :)

brianegan:
Sounds good, thanks for helping out!

pq:
[...]

PR #2455, opened 2018-06-30
Improve Copy / Paste on new Logging View

quangson91:
Issue: https://github.com/flutter/flutter-intellij/issues/2452

This PR fix problem when user copy log by shortcut (eg: `Ctrl + C` / `CMD + C`)
I fix by register copy shortcut & handle it.
(By default, the system will automatically copy the value of each cell - com.intellij.ui.treeStructure.treetable.ListTreeTableModelOnColumns#getValueAt then passing to the clipboard).

**NOTE**
Currently, when user copy we still copy `invisible` columns.
We should improve that.

See my comments for more detail.
// cc @pq please review on next week!

### DIFF
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -445,7 +445,7 @@ public class FlutterLogTree extends TreeTable {
       buffer
         .append(TIMESTAMP_FORMAT.format(entry.getTimestamp()))
         .append(" ").append(entry.getSequenceNumber())
-        .append(" ").append(FlutterLog.Level.forValue(entry.getLevel()).toDisplayString())
+        .append(" ").append(entry.getLevelName())
         .append(" ").append(entry.getCategory())
         .append(" ").append(entry.getMessage());
       if (!entry.getMessage().endsWith("\n")) {
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -445,7 +445,7 @@ public class FlutterLogTree extends TreeTable {
       buffer
         .append(TIMESTAMP_FORMAT.format(entry.getTimestamp()))
         .append(" ").append(entry.getSequenceNumber())
-        .append(" ").append(entry.getLevel())
+        .append(" ").append(FlutterLog.Level.forValue(entry.getLevel()).toDisplayString())
         .append(" ").append(entry.getCategory())
         .append(" ").append(entry.getMessage());
       if (!entry.getMessage().endsWith("\n")) {
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -525,7 +525,7 @@ public class FlutterLogTree extends TreeTable {
     super(model);
     model.setTree(this.getTree());
     this.model = model;
-    registerShortcut();
+    registerCopyHandler();
     flutterLogPopup = new FlutterLogEntryPopup();
     addMouseListener(new SimpleMouseListener() {
       @Override
@@ -567,7 +567,7 @@ public class FlutterLogTree extends TreeTable {
     });
   }
 
-  private void registerShortcut() {
+  private void registerCopyHandler() {
     final AnAction actionCopy = ActionManager.getInstance().getAction(IdeActions.ACTION_COPY);
     final ShortcutSet copyShortcutSet = actionCopy.getShortcutSet();
     final String copyCommand = "flutter.log.copyCommand";
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -9,7 +9,9 @@ import com.intellij.execution.filters.Filter;
 import com.intellij.execution.filters.HyperlinkInfo;
 import com.intellij.execution.filters.UrlFilter;
 import com.intellij.openapi.Disposable;
+import com.intellij.openapi.actionSystem.*;
 import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.ui.ColoredTableCellRenderer;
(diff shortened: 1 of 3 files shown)


src/io/flutter/logging/FlutterLogView.java
```diff
@@ -285,24 +282,7 @@ public void setSelected(AnActionEvent e, boolean state) {
 
     @Override
     public void actionPerformed(AnActionEvent e) {
-      ApplicationManager.getApplication().invokeLater(() -> {
```
quangson91: I refactoring by move this code to the `FlutterLogTree`.
So we can call `logTree.sendToClipboardSelectedItems()`

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -482,6 +493,48 @@ public FlutterLogTree(@NotNull FlutterApp app,
...
+
+        final StringSelection selection = new StringSelection(sb.toString());
+        clipboard.setContents(selection, selection);
+      }
+    });
+  }
+
+  private void registerShortcut() {
+    final AnAction actionCopy = ActionManager.getInstance().getAction(IdeActions.ACTION_COPY);
+    final ShortcutSet copyShortcutSet = actionCopy.getShortcutSet();
```
quangson91: Query shortcut for copy action then register to FlutterLogTree.
When user trigger that action -> we simply call `sendToClipboardSelectedItems`

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -404,7 +415,7 @@ public void describeTo(@NotNull StringBuilder buffer) {
       buffer
         .append(TIMESTAMP_FORMAT.format(entry.getTimestamp()))
         .append(" ").append(entry.getSequenceNumber())
-        .append(" ").append(entry.getLevel())
+        .append(" ").append(FlutterLog.Level.forValue(entry.getLevel()).toDisplayString())
```
quangson91: I think we should display log level as text.
(Because user see in LogView is text eg `info/fine` instead of 800/400).

pq: 👍

src/io/flutter/logging/FlutterLogView.java
pq: Good idea!

src/io/flutter/logging/FlutterLogTree.java
pq: I need to think about this some more but a quick question: how does this handler get **de**-registered?

Curious how this is done in other views... Or maybe you looked?

src/io/flutter/logging/FlutterLogTree.java
pq: One concern here is for custom user levels but we'll want to support those elsewhere too.  (Just food for thought!)

src/io/flutter/logging/FlutterLogTree.java
quangson91: @pq, `de-registered` mean?

> Curious how this is done in other views... Or maybe you looked?

Currently, the default action copy/paste it copies string values of the selected row in the tree view. But the string take by method
```java
com.intellij.ui.treeStructure.treetable.ListTreeTableModelOnColumns#getValueAt
  public Object getValueAt(Object value, int column) {
    return myColumns[column].valueOf(value);
  }
```

In our code
```java
      @Nullable
      @Override
      public FlutterLogEntry valueOf(DefaultMutableTreeNode node) {
        if (node instanceof FlutterEventNode) {
          return ((FlutterEventNode)node).entry;
        }
        return null;
      }
```

=> we return `FlutterLogEntry` for each cell.
=> when user copy => the result looks like  this:
```
io.flutter.logging.FlutterLogEntry@76b5b7da | io.flutter.logging.FlutterLogEntry@76b5b7da | io.flutter.logging.FlutterLogEntry@76b5b7da | io.flutter.logging.FlutterLogEntry@76b5b7da
```

quangson91: Another solution is
```
      @Nullable
      @Override
      public FlutterLogEntry valueOf(DefaultMutableTreeNode node) {
        if (node instanceof FlutterEventNode) {
          // TODO: Instead return FlutterLogEntry we will return STRING value for each column.
          return ((FlutterEventNode)node).entry;

        }
        return null;
      }
```

src/io/flutter/logging/FlutterLogTree.java
quangson91: I created a new method `getLevelName`.
So we can use it in the future.
https://github.com/flutter/flutter-intellij/pull/2455/commits/8ae8b519887e207fb4992c032095f71c337a9e9b

quangson91:
@pq, I have addressed your comment, please take a look!

pq:
> @pq, de-registered mean?

Sorry, I wasn't clear.  I just mean I can see how we're registering this key-binding handler but am wondering how we remove that registration.  (Or if we don't have to, why not?)

quangson91:
@pq, I think we don't have too.
By register this key-binding we already override `copy by shortcut action`.

Another way that doesn't have to register key-binding is: https://github.com/flutter/flutter-intellij/pull/2455#discussion_r199664579

pq:
Thanks for following up @quangson91! Looks like we need another rebase. ⚒

quangson91:
@pq, I updated this PR. please take a look!

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -550,6 +556,37 @@ private String getSelectedLog() {
...
+      final String log = getSelectedLog();
+      if (log != null) {
+        final Clipboard clipboard = Toolkit.getDefaultToolkit().getSystemClipboard();
+        final StringSelection selection = new StringSelection(log);
+        clipboard.setContents(selection, selection);
+      }
+    });
+  }
+
+  private void registerShortcut() {
```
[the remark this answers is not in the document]
quangson91: Yes, I updated.

pq:
Thanks!

PR #2459, opened 2018-07-02
Refactor FlutterPluginsLibraryManager

stevemessick:
Moves the pure library management code out of FlutterPluginsLibraryManager into an abstract superclass. The intent is to define a new subclass, specific to Android Studio, to manage discovered Android Libraries.

Lots of copy-paste-delete going on. Lightly tested.

@devoncarew @pq


stevemessick:
Changes made in new PR.

Issue #2463, opened 2018-07-03
Fix or disable URL hyperlinks in the Log View.


pq:
Follow-up from: https://github.com/flutter/flutter-intellij/pull/2454.

URL hyperlinks render but do not navigate.  For `M27`, we should fix or disable.

PR #2464, opened 2018-07-03
Fix log entry browser links (#2463)

pq:
Fixes: #2463.

/cc @devoncarew @quangson91


quangson91:
@pq, Great. Now I know why it doesn't work :D

Issue #2457, opened 2018-07-02
Settle on Log View entry presentation defaults


pq:
Breakout from #2340,

> decide whether we do need to display the timestamp column

I'm inclined to hide it by default; showing only level, category and message and coloring off.

@devoncarew : thoughts?

pq:
Sensible defaults will also address:

> we should devote more of the horizontal space to the message portion of the log entry 

👍

pq:
Proposed defaults would look like this:

![image](https://user-images.githubusercontent.com/67586/42240093-008c7f2a-7ebb-11e8-934e-0a17c45bbc5b.png)

With the option to enable hidden views from the gear pull-down.

![image](https://user-images.githubusercontent.com/67586/42240103-0dfa363e-7ebb-11e8-829d-450445fb42cb.png)

@devoncarew: you suggested dropping the timestamp; curious how this sits with you?

Alternatively, we might disable them all:

![image](https://user-images.githubusercontent.com/67586/42240217-6dad83ce-7ebb-11e8-82d7-3a9e7911bb83.png)

and nudge them to configure what they want?

Either way, maybe a toast on first start that points to configuration options?

PR #2465, opened 2018-07-03
Update logger defaults (#2457)

pq:
Update logger defaults 

image

Fixes: #2457

/cc @devoncarew @quangson91

### DIFF
--- a/src/io/flutter/logging/FlutterLogFilterPanel.java
+++ b/src/io/flutter/logging/FlutterLogFilterPanel.java
@@ -103,7 +103,6 @@ public class FlutterLogFilterPanel {
     final List<FlutterLog.Level> logLevels = Arrays.stream(FlutterLog.Level.values())
       .collect(Collectors.toList());
     logLevelComboBox.setModel(new CollectionComboBoxModel<>(logLevels));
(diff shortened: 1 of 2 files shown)


src/io/flutter/logging/FlutterLogFilterPanel.java
```diff
@@ -113,6 +112,7 @@ protected void customizeCellRenderer(@NotNull JList<? extends FlutterLog.Level>
                                            boolean selected,
                                            boolean hasFocus) {
         append(value.toDisplayString());
+    logLevelComboBox.setSelectedItem(FlutterLog.Level.CONFIG);
```
devoncarew: Needs formatting? Or, is this meant to live a few lines up?

pq: Oh, yeah.  That was dumb.

And actually, this should really be properly set in the prefs -- I didn't see the default there until just now.

Will fix.

Thanks!

pq:
Fixed now.  Thanks!

quangson91:
@pq, I have a question. Why don't we enable show color by default?
(For me, if the log levels like `warning/shout` => we should be coloring them. It is better for eyes to recognize).

PR #2474, opened 2018-07-09
Add Android module libraries to Flutter project

stevemessick:
@devoncarew @pq 

Analyzes the Android module and adds all libraries required to compile its code to the top-level Flutter project. This should fix problems related to not having indexed all the code in the Android module (i.e. completion, navigation, etc.)

It probably still needs work but I'm not sure what the issues are. It takes noticeably more time to open a project since a Gradle sync is required, but that is no worse than what Android Studio does for an Android project.


devoncarew:
Awesome! Looking now...

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
```diff
@@ -0,0 +1,198 @@
...
+  private static final Logger LOG = Logger.getInstance(AndroidModuleLibraryManager.class);
+  private static final String BUILD_FILE_NAME = "build.gradle";
+  private final AtomicBoolean isUpdating = new AtomicBoolean(false);
+
+  public AndroidModuleLibraryManager(@NotNull Project project) {
+    super(project);
+  }
+
+  public void update() {
+    doGradleSync(getProject(), (Project x) -> updateAndroidLibraryContent(x));
```
devoncarew: I assume you've thought about this, but do we want to drive the gradle sync, or notice when one has happened and then query gradle somewhow?

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  private Void updateAndroidLibraryContent(Project androidProject) {
```
devoncarew: `void`?

devoncarew: Have `androidProject` be `@Nonnullable`?

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
```diff
@@ -0,0 +1,198 @@
...
+    LibraryTable androidProjectLibraryTable = LibraryTablesRegistrar.getInstance().getLibraryTable(androidProject);
+    Library[] androidProjectLibraries = androidProjectLibraryTable.getLibraries();
+    if (androidProjectLibraries.length == 0) {
+      LOG.error("Gradle sync was incomplete -- no Android libraries found");
```
devoncarew: If we do want this to be a user-facing error, we should use the toast notifications (from the `FlutterMessages` utility class). If not, we should log at a warn level.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
```diff
@@ -0,0 +1,198 @@
...
+          public void rootsChanged(ModuleRootEvent event) {
+            manager.scheduleUpdate();
+          }
+        });
+        manager.scheduleUpdate();
+      }
+    }
+  }
+
+  private static void fileChanged(@NotNull final Project project, @NotNull final VirtualFile file) {
```
devoncarew: Just for my own edification, it looks like we watch the `build.gradle` (in any subdirectory?), and schedule a gradle sync when it changes.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
```diff
@@ -0,0 +1,198 @@
...
+    GradleSyncInvoker gradleSyncInvoker = ServiceManager.getService(GradleSyncInvoker.class);
+    gradleSyncInvoker.requestProjectSync(androidProject, request, listener);
+  }
+
+  @NotNull
+  public static AndroidModuleLibraryManager getInstance(@NotNull final Project project) {
+    return ServiceManager.getService(project, AndroidModuleLibraryManager.class);
+  }
+
+  public static void startWatching() {
```
devoncarew: It looks like this will only watch projects that are open at IntelliJ launch?

flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
```diff
@@ -85,7 +85,20 @@ public ProjectSystemSyncManager getSyncManager() {
   @SuppressWarnings("override")
   @Nullable
   public GradleCoordinate getAvailableDependency(@NotNull GradleCoordinate coordinate, boolean includePreview) {
-    return null;
+    //return gradleProjectSystem.getAvailableDependency(coordinate, includePreview);
+    Method finders = ReflectionUtil.getMethod(gradleProjectSystem.getClass(), "getAvailableDependency");
+    if (finders == null) {
+      LOG.error("No method found: GradleProjectSystem.getAvailableDependency()");
```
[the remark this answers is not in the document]
stevemessick: Oops! I forgot to change that. Done.

flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
```diff
@@ -85,7 +85,20 @@ public ProjectSystemSyncManager getSyncManager() {
...
+      return null;
+    }
+    try {
+      //noinspection unchecked
+      return (GradleCoordinate) finders.invoke(gradleProjectSystem, coordinate, includePreview);
+    }
+    catch (IllegalAccessException | InvocationTargetException e) {
+      LOG.error(e);
```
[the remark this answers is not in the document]
stevemessick: None of the exceptions will be thrown. The checking is all there to satisfy compilers and style guides. :) The method isn't used in 3.1 and exists in 3.2.

flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
```diff
@@ -85,7 +85,20 @@ public ProjectSystemSyncManager getSyncManager() {
   @SuppressWarnings("override")
   @Nullable
   public GradleCoordinate getAvailableDependency(@NotNull GradleCoordinate coordinate, boolean includePreview) {
-    return null;
+    //return gradleProjectSystem.getAvailableDependency(coordinate, includePreview);
```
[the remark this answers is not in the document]
stevemessick: We need that reflection as long as we build for 3.1. I added a TODO.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
stevemessick: Getting rid of type-checking warnings was a hassle. If there's a better way to do it I'd like to know. Using "void" will cause a compilation error now, though.

stevemessick: Yep, missed that one. Done.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
stevemessick: Done.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
stevemessick: Yes, there are two of them (at least) in an Android module. We might want to watch `settings.gradle` too.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
stevemessick: This runs each time a project is opened.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryType.java
```diff
@@ -0,0 +1,62 @@
...
+import com.intellij.openapi.roots.libraries.ui.LibraryPropertiesEditor;
+import com.intellij.openapi.vfs.VirtualFile;
+import icons.FlutterIcons;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.*;
+
+public class AndroidModuleLibraryType extends LibraryType<AndroidModuleLibraryProperties> {
+  public static final String LIBRARY_NAME = "Android Libraries";
```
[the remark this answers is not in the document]
stevemessick: screen shot 2018-07-10 at 11 54 51 am

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
stevemessick: That's a good question, so here's probably more of an answer than you expected. :) Currently, the only way to initiate a Gradle sync is to open the Android project in a new window, which we're trying to get away from. You could do it from the command line, I suppose, but I'm not really interested in that. Eventually we want to have some ability (such as Project Structure) to add dependents to Android modules without launching a new Android Studio for that purpose, and then we'd want to make sure a sync happened once but not twice, but that's in the future. Today, you can edit the Gradle build files (via AS open on a Flutter project) but that won't trigger a sync without this code.

flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
```diff
@@ -40,5 +41,7 @@ public void runActivity(@NotNull Project project) {
     // Unset this flag for all projects, mainly to ease the upgrade from 3.0.1 to 3.1.
     // TODO(messick) Delete once 3.0.x has 0 7DA's.
     FlutterProjectCreator.disableUserConfig(project);
+    // Monitor Android dependencies.
+    AndroidModuleLibraryManager.startWatching();
```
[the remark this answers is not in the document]
stevemessick: I added a flag.

stevemessick:
I made a few changes, PTAL.

stevemessick:
The latest change simplifies the library name. PTAL

screen shot 2018-07-11 at 7 25 16 am

There's no need to append the project name. The convention is:
 - `Dart *` contains Dart code
 - `Flutter *` contains Flutter code
 - `Android *` contains Android code

flutter-studio/src/io/flutter/android/AndroidModuleLibraryType.java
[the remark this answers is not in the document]
devoncarew: I see the latest is `Android Libraries`. It could still be useful to give a hint where these libraries come from (the Flutter plugin dependencies), but YMMV.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
stevemessick: This is one of those "What was I thinking?" moments. I removed the loop from `startWatching()` and added the Project as a parameter. Thanks for catching that!

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
```diff
@@ -0,0 +1,198 @@
...
+
+      @Override
+      public void syncStarted(@NotNull Project project, boolean skipped, boolean sourceGenerationRequested) {}
+
+      @Override
+      public void setupStarted(@NotNull Project project) {}
+
+      @Override
+      public void syncSucceeded(@NotNull Project project) {
+        callback.apply(androidProject);
```
pq: Does it make sense to put the flag update in a finally?  Something like:

```java
try {
  callback.apply(androidProject);
} finally {
  isUpdating.set(false);
}
```

EDIT: I'm probably over-thinking.

flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
```diff
@@ -9,12 +9,16 @@
...
 import io.flutter.actions.FlutterShowStructureSettingsAction;
 import io.flutter.actions.OpenAndroidModule;
+import io.flutter.android.AndroidModuleLibraryManager;
 import io.flutter.project.FlutterProjectCreator;
 import io.flutter.utils.FlutterModuleUtils;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.MissingResourceException;
+
 public class FlutterStudioStartupActivity implements StartupActivity {
```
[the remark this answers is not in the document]
stevemessick: Yes, done.

flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
```diff
@@ -40,5 +44,10 @@ public void runActivity(@NotNull Project project) {
     // Unset this flag for all projects, mainly to ease the upgrade from 3.0.1 to 3.1.
     // TODO(messick) Delete once 3.0.x has 0 7DA's.
     FlutterProjectCreator.disableUserConfig(project);
+    // Monitor Android dependencies.
+    if (System.getProperty("flutter.android.library.sync", null) != null) {
```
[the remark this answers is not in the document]
stevemessick: Could do; but later, I think.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
stevemessick: No, that doesn't work too well because the callback runs async. Effectively removes the flag. Speaking from experience here :)

stevemessick:
I made a slight modification to the code, in response to @devoncarew's comment about looping over projects. I checked it out and the smoke tests still work as expected so I'll submit without further review. Thanks for the comments!

Issue #2468, opened 2018-07-05
Validate android package names with multiple dots


g123k:
Hello,

Thanks for your amazing work, but I think I have found a tiny bug.

With the new project -> Flutter -> plugin dialog, you have to choose an organization.
If by mistake, there is a dot at the end of it, such as `com.example.`, it will generate a broken Android project : https://imgur.com/a/BzoAVbh

In my example, the Android package name will then be `com.example..test_plugin`, which is incorrect and won't compile.

A warning or an error message would be helpful here to prevent the generation of an incorrect Android project.

pq:
Great suggestion @g123k.  Thanks for the report!

Ruban2205:
Hello there, I've created a flutter project with the package name in.rubangino.newproject , But It throws an error called "Android resource linking failed" while compiling. But when I try with a new package name in new project like com.rubangino.newproject1 , It compiles and run without errors. As I am learning flutter, I can't able to rectify this issue. Can You please suggest me a solution to this error?

PR #2475, opened 2018-07-11
Port ADT package name validation logic.

pq:
Further validation of android package names.

Bad package names spell trouble and are winnowed out in standard Android project creation.

Follow-up from #2468. 

/cc @stevemessick @devoncarew


stevemessick:
LGTM. In case you didn't know, Android Studio takes it a bit farther, offering some auot-corrections as seen in these screenshots:

screen shot 2018-07-11 at 1 48 26 pm

screen shot 2018-07-11 at 1 48 47 pm

pq:
Those are awesome.  I so wish we could do something similar.  Unfortunately, I don't believe the wizard framework for vanilla IDEA supports that feedback.  I'll revisit though.

Thanks for the nudge!

PR #2477, opened 2018-07-12
Stop complaining about Canary versions while debugging

stevemessick:
When running from sources, like in a debugger, there's no reason to warn that Canary versions are not supported.

@devoncarew @pq @terrylucas


stevemessick:
We're getting random build failures with the message
`The command "flutter doctor" failed and exited with 50 during .`

The failing jobs succeeded in other builds so I'm committing anyway.

devoncarew:
You should be able to restart individual failed builds.

PR #2479, opened 2018-07-13
Update module definition

stevemessick:
Minor change to add a new module dependency that resolves compilation errors in AndroidUtils.

@pq


stevemessick:
TBR

This is just another bookkeeping update, like my two previous PRs. I'm pretty sure the reason the dependency added here wasn't needed by anyone else is because it is provided by the Dart plugin, which the "other" module definition uses. (I feel bad about so many TBRs but these things are holding me up and people seem to be busy with other work, and I think they fit in the TBR guidelines.)

PR #2480, opened 2018-07-13
Fix types

stevemessick:
@devoncarew The build is failing. I think this will fix it.


stevemessick:
TBR

This fixes three of four jobs and the fourth is unrelated to this.

PR #2481, opened 2018-07-13
Fix analysis options

stevemessick:
The build is failing. I think this will fix it, but not sure why this causes failure.


stevemessick:
TBR

This is blocking another PR I'm trying to get in.

PR #2482, opened 2018-07-13
Add "module" template to new-module and -project wizards in Android Studio

stevemessick:
@pq @devoncarew 

This is behind an experimental flag so it won't affect anyone yet. If anyone is using the feature they might notice that Flutter-specific entries no longer show up in the New Module wizard when it is run from an Android project. It didn't make sense for them to be included, anyway.

WIP: generated project does not open correctly.

Reformatting moved some code around.


stevemessick:
There's a build failure. That's not the fault of this code, but looks like a problem with the plugin tool. For some unknown reason the edits it performs for the 2017.3 build are not working after editing one of the affected classes. I'll look into it, but you can review this code as-is.

flutter-studio/src/io/flutter/module/FlutterDescriptionProvider.java
```diff
@@ -25,20 +30,39 @@
...
+        }
+      }
+    }
+    if (projectHasFlutter) {
+      // Makes no sense to add Flutter templates to Android projects...
+      res.add(new FlutterApplicationGalleryEntry(sharedModel));
+      res.add(new FlutterPluginGalleryEntry(sharedModel));
+      res.add(new FlutterPackageGalleryEntry(sharedModel));
+    }
+    if (System.getProperty("flutter.experimental.modules", null) != null) {
```
devoncarew: Let's definitely plan on moving this option into user preferences (in future PRs).

stevemessick:
Sounds good. I'm going to hold this for a bit while I look into the build failure. No point merging this if it breaks everything.

stevemessick:
#2489 makes the build green. Merging now.

Issue #2487, opened 2018-07-17
Don't scroll to bottom when scrolled away from bottom manually


zoechi:
The logging output should stay where it is when manually scrolled away from bottom.
Currently a log output scrolls to the bottom every time a new log output happens.

This makes it impossible to check previous log output while the app is running.

Scrolling to bottom (manually or using a scroll-to-end button) should restore the current behavior.

quangson91:
Ah, I understood @zoechi  problem.
It seems we have to detect if the log scroll-up => we will disable `auto scroll to end` action.

quangson91:
I created a new PR to fix this problem: https://github.com/flutter/flutter-intellij/pull/2488

zoechi:
@quangson91 
and you should detect when scrolled back to the end to re-enable `auto scroll to end`.
Does your fix include that?

quangson91:
My fix don't include that.
But I think for that action -> user just clicks action `auto-scroll to end`.

quangson91:
>  you should detect when scrolled back to the end to re-enable auto scroll to end

=> To fix that, we just need check new scroll value = 100 => we re-enable `auto scroll to end`.

Ah, I will fix that. I tested on Intellij, after user scroll to end -> `auto scroll to end` must re-enable

quangson91:
@zoechi I updated this PR: https://github.com/flutter/flutter-intellij/pull/2488

now when user scroll update -> disable auto scroll to end.
when user scroll to the end (bottom) -> re-enable `auto scroll to end`.

zoechi:
@quangson91 that sounds great. Thanks a lot!

pq:
Fixed w/ #2488 🎉

PR #2488, opened 2018-07-17
auto-disable scroll to end when user manual scroll log up

quangson91:
Issue: https://github.com/flutter/flutter-intellij/issues/2487
+ auto-disable scroll to end when user manual scroll log up

// cc @pq, @zoechi 

Screenshot

auto-disable-scroll-to-end-when-scroll-up

### DIFF
--- a/src/io/flutter/logging/FlutterLogView.java
+++ b/src/io/flutter/logging/FlutterLogView.java
@@ -36,6 +36,8 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
+import javax.swing.event.ChangeEvent;
+import javax.swing.event.ChangeListener;
 import javax.swing.table.TableCellRenderer;
 import javax.swing.table.TableColumn;
 import java.awt.*;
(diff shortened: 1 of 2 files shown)


src/io/flutter/logging/FlutterLogView.java
```diff
@@ -344,6 +346,31 @@ private String countString(int count) {
     }
   }
 
+  private static class LogVerticalScrollChangeListener implements ChangeListener {
+    private volatile int oldScrollValue = 0;
+
+    @Override
+    public void stateChanged(ChangeEvent e) {
+      final BoundedRangeModel model = (BoundedRangeModel)e.getSource();
```
pq: I'm not too familiar w/ `ChangeEvent`s but should we do an instance check to be safe or is this cast a sure thing?

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -344,6 +346,31 @@ private String countString(int count) {
...
+      if (isScrollUp) {
+        onScrollUp();
+      }
+      else if (isScrollToEnd) {
+        onScrollToEnd();
+      }
+      oldScrollValue = newScrollValue;
+    }
+
+    protected void onScrollUp() {
```
pq: Maybe make this an abstract class and these methods abstract to make it clear that implementers should implement?  Alternatively, maybe a comment saying something like:

    // implemented in subclass

?

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -441,11 +468,30 @@ private Object getTagForPosition(MouseEvent e) {
...
+    setupLogTreeScrollPane();
+  }
+
+  private void setupLogTreeScrollPane() {
     final JScrollPane pane = ScrollPaneFactory.createScrollPane(
       logTree,
       ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED,
       ScrollPaneConstants.HORIZONTAL_SCROLLBAR_AS_NEEDED);
+    pane.getVerticalScrollBar().getModel().addChangeListener(new LogVerticalScrollChangeListener() {
+      @Override
```
pq: It would be great if the ScrollToEnd toggle action could have its state updated as well.  I believe it will be out of sync?

src/io/flutter/logging/FlutterLogView.java
quangson91: Done, I make this class as an abstract class.

src/io/flutter/logging/FlutterLogView.java
quangson91: e.getSource() should be `BoundedRangeModel`.
But for sure, I just add check `instanceof BoundedRangeModel`

src/io/flutter/logging/FlutterLogView.java
quangson91: Ah, I moved that logic to ScrollToEndAction. so it is more clear.
Please check my record in the description of this PR.

pq: Fantastic.  Thanks!

PR #2494, opened 2018-07-19
remove unnecessary volatile keyword

quangson91:
as for suggestions of @jacob314, I removed `volatile` as it unnecessary.
Ref: https://github.com/flutter/flutter-intellij/commit/6c968f95ac54511d9deffdb244733d9946065941#r29766875

// @jacob314 please take a look!


jacob314:
Thanks for the quick fix!

pq:
Thanks!

PR #2496, opened 2018-07-19
Revert "Disable evaluating expressions when not on a call frame due t…

jacob314:
…o DartVM crashes. (#2436)"

This reverts commit ddd96076068fee3edca4132ce0e8c94b4164f977.


jacob314:
The Dart VM issues that lead to the initial CL have now been fixed.

PR #2506, opened 2018-07-27
Add a setting to control syncing Android libraries

stevemessick:
@devoncarew @pq 

The ability to automatically define an external library containing Android dependencies has been available behind a dev-only flag for a couple weeks. This PR adds a line to the Settings/Preferences panel to enable or disable it, but only for Android Studio.

I'm not sure whether this should go into the next release or not, so we should discuss that before merging.

No idea why BorderFactoryClass got removed from the form...

### DIFF
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.form
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.form
@@ -14,9 +14,6 @@
           <grid row="0" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="0" fill="3" indent="0" use-parent-layout="false"/>
         </constraints>
         <properties/>
-        <clientProperties>
-          <BorderFactoryClass class="java.lang.String" value="com.intellij.ui.IdeBorderFactory$PlainSmallWithIndent"/>
-        </clientProperties>
         <border type="etched" title="SDK"/>
         <children>
           <component id="a59e7" class="javax.swing.JLabel">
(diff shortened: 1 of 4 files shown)


devoncarew:
> No idea why BorderFactoryClass got removed from the form...

I tend to hand-edit this file to add it back in.

src/io/flutter/sdk/FlutterSettingsConfigurable.form
```diff
@@ -149,6 +140,15 @@
               <text value="Replace the Run and Debug console output with an experimental Flutter Logging view"/>
             </properties>
           </component>
+          <component id="33088" class="javax.swing.JCheckBox" binding="mySyncAndroidLibrariesCheckBox">
+            <constraints>
+              <grid row="3" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+            </constraints>
+            <properties>
+              <text value="Show Android libraries; for working with Java/Kotlin (requires reopening projects)"/>
```
devoncarew: 'Show Android libraries; for working with Java/Kotlin (requires reopening projects)'

Perhaps, 'When opening a Flutter project, run gradle to discover plugin native libraries'.?

The tooltip could then indicate why you'd want to  (for code completion and such).

stevemessick:
I updated the form to preserve formatting and change the text to focus on solving user problems. Info about how that happens is in the hover text.

PR #2516, opened 2018-07-31
Handle the device being disconnected more gracefully.

jacob314:
Fix race condition that sometimes occurs when the app is restarted resulting an easily avoidable error in the IntelliJ console.


pq:
This'd be good to cherry-pick into `M27`; just hit it in testing. 👍

PR #2525, opened 2018-08-01
remove “module” project type

pq:
To add back when add-top-app is more fully baked.

Note this was showing up in the project creation UI.

Before:

image

After:

image

/cc @stevemessick


pq:
Ah.  Question answered: removing the value isn't sufficient as it introduces build breaks.

@stevemessick : the other alternative is for me to filter the specific value out of the UI model.

Should I take a crack at that?

pq:
@stevemessick: took another crack.  PTAL.

Issue #2578, opened 2018-09-01
Android studio notification appeared with below mentioned exception. Debugging session also ended.


vivek2089:
## What happened

A bubble notification appeared with below mentioned exception. Debugging session also ended .I was debugging an async function. The function was reading value from sqlite database. The project uses sqflite plugin. (sqflite: ^0.11.2+2)

## Version information

Android Studio `3.1.4` • Flutter plugin `io.flutter 27.1.1` • Dart plugin `173.4700`

Error getting Flutter sdk information.

## Exception

null

```
java.lang.NullPointerException
	at org.dartlang.vm.service.element.InstanceRef.getKind(InstanceRef.java:42)
	at io.flutter.server.vmService.frame.DartVmServiceValue.computeVarHavingStringValuePresentation(DartVmServiceValue.java:189)
	at io.flutter.server.vmService.frame.DartVmServiceValue.computePresentation(DartVmServiceValue.java:148)
	at com.intellij.xdebugger.impl.ui.tree.nodes.XValueNodeImpl.<init>(XValueNodeImpl.java:70)
	at com.intellij.xdebugger.impl.ui.tree.nodes.XValueContainerNode.lambda$addChildren$0(XValueContainerNode.java:98)
	at com.intellij.util.containers.TransferToEDTQueue$2.process(TransferToEDTQueue.java:104)
	at com.intellij.util.containers.TransferToEDTQueue$2.process(TransferToEDTQueue.java:101)
	at com.intellij.util.containers.TransferToEDTQueue.processNext(TransferToEDTQueue.java:120)
	at com.intellij.util.containers.TransferToEDTQueue.access$300(TransferToEDTQueue.java:37)
	at com.intellij.util.containers.TransferToEDTQueue$1.run(TransferToEDTQueue.java:67)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:762)
	at java.awt.EventQueue.access$500(EventQueue.java:98)
	at java.awt.EventQueue$3.run(EventQueue.java:715)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:732)
[...]

PR #2586, opened 2018-09-05
fix an issue with an NPE on InstanceRef.getKind()

devoncarew:
- fix an issue with an NPE on InstanceRef.getKind()
- fix https://github.com/flutter/flutter-intellij/issues/2578

@jwren


jwren:
LGTM

PR #2589, opened 2018-09-06
Attach

stevemessick:
Add UI and basic framework for using 'flutter attach' to do hot reload of Flutter modules.

Debugging is not yet supported.

This code can be used to give demos of hot reload of modules.

@pq @devoncarew @jacob314


flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
```diff
@@ -169,7 +169,6 @@ private static void showHowToEditDialog() {
                          "edit the files as described in the <a href=\"" +
                          EDIT_INSTR_LINK +
                          "\">add2app documentation.</a></body></html>";
-      Messages.installHyperlinkSupport(myTextPane);
```
devoncarew: We don't need this anymore?

flutter-studio/src/io/flutter/run/attach/SdkAttachConfig.java
```diff
@@ -0,0 +1,128 @@
...
+  @Override
+  public LaunchState getState(@NotNull Executor executor, @NotNull ExecutionEnvironment env) throws ExecutionException {
+    try {
+      checkRunnable(env.getProject());
+    }
+    catch (RuntimeConfigurationError e) {
+      throw new ExecutionException(e);
+    }
+
+    SdkFields launchFields = getFields(); // TODO Check that this does not change from the constructor setting.
```
devoncarew: nit: `TODO(messick):`

src/io/flutter/run/LaunchState.java
```diff
@@ -90,6 +90,12 @@ public LaunchState(@NotNull ExecutionEnvironment env,
     DaemonConsoleView.install(this, env, workDir);
   }
 
+  @NotNull
+  protected Callback getCallback() {
```
devoncarew: I know this isn't part of your PR, but given that we're increasing the access level here, consider renameing the `Callback` interface to something more qualified - `LaunchStateCallback`, `ConnectionCompleteCallback`, ...

flutter-studio/src/io/flutter/actions/ConnectAndroidDebuggerAction.java
```diff
@@ -0,0 +1,134 @@
...
+import javax.swing.*;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ConnectAndroidDebuggerAction extends AndroidConnectDebuggerAction {
+  @SuppressWarnings("FieldCanBeLocal")
+  private static String RUN_DEBUG_LINK = "https://flutter.io/using-ide/#running-and-debugging";
+
+  @Override
+  public void actionPerformed(AnActionEvent e) {
```
devoncarew: If this method is copied from another (which I'm assuming it is given the `--flavor` and other code below), can you add comments in this method and the source method that the two should be kept in sync? Otherwise we risk fixes and changes in one not being applied to the other.

devoncarew:
I took one pass through -

flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
stevemessick: Correct, I forgot to remove this line after copying it further up the method.

flutter-studio/src/io/flutter/run/attach/SdkAttachConfig.java
stevemessick: Oops. I forgot to delete that TODO.

src/io/flutter/run/LaunchState.java
stevemessick: That ended up changing a bunch of identifiers but I think it is more readable now.

flutter-studio/src/io/flutter/actions/ConnectAndroidDebuggerAction.java
stevemessick: Will do.

stevemessick:
PTAL

That rename was more aggressive than I expected. I had to revert a bunch of invalid changes. I think I got them all but a second opinion would be welcome.

PR #2593, opened 2018-09-09
correct get selected log for row filter

quangson91:
###  correct get selected log for row filter
Reason: after apply row filter, the selected path will not correct. (because, some row is hidden by row-filter).

Screenshot:
**Before fix:**
screen shot 2018-09-09 at 10 06 05 am

**After fixed:**
screen shot 2018-09-09 at 10 08 33 am

// cc @pq, @devoncarew  for review.
Thanks!

### DIFF
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -554,27 +554,25 @@ public class FlutterLogTree extends TreeTable {
     });
   }
 
-  @Nullable
+  @NotNull
   private String getSelectedLog() {
     final int[] rows = getSelectedRows();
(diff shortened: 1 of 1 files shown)


src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -554,27 +554,25 @@ public void mouseDoublePressed(MouseEvent e) {
...
     }
-    return sb.toString();
+    return logBuilder.toString();
   }
 
   public void sendSelectedLogsToClipboard() {
     ApplicationManager.getApplication().invokeLater(() -> {
       final String log = getSelectedLog();
-      if (log != null) {
+      if (StringUtils.isNotEmpty(log)) {
```
pq: Nice. 👍

Issue #2485, opened 2018-07-16
Consider nesting stacktrace entries


pq:
It would be great to nest stack trace frames under log entries so they can be easily collapsed and hidden.

![image](https://user-images.githubusercontent.com/67586/42777364-b2c28b98-88ee-11e8-8751-0e1eb1b0ca1d.png)

/cc @quangson91

quangson91:
@pq, I'm trying to working on this issue.
But for now, do you have any suggestion to determine the log entry is Stacktrace ?

For me, I think we can determine the log is entry by check some first characters contain `#NUMBER`.

// cc @devoncarew

quangson91:
I just create an WIP pr for this one: https://github.com/flutter/flutter-intellij/pull/2594/files
(Still thinking about: how do we enable collapse for the FlutterLogTree)

pq:
Looking at the PR now.  I think in the medium term we'll want to introduce a structured format for entries coming from the Flutter framework itself so that nesting is something explicit.  FWIW: a bit of relevant conversation is going on in https://github.com/flutter/flutter/issues/21504 and https://github.com/flutter/flutter/pull/21505.

Thanks for pitching in!

DanTup:
We do have a structured way of getting stack traces, it's in the VM service :-)

It would be great if we could support breaking and let the user use all the debugging tools we have (eg. https://github.com/flutter/flutter/issues/17007). I don't know what the scale of work for it is though (or who could advise). The problem is that Flutter wants to catch the exceptions, but as a user I want them to break.

pq:
Another angle to consider is to use the new details pane:

![image](https://user-images.githubusercontent.com/67586/48084014-7e55a780-e1ab-11e8-8c18-c78619a20338.png)

So far this is just showing json "data" but could easily be used to show stack traces and that may be a nice alternative to nesting.

devoncarew:
Closing as this is related to the experimental logging view.

PR #2594, opened 2018-09-09
WIP - handle collapse for stacktrace

quangson91:
Issue: https://github.com/flutter/flutter-intellij/issues/2485

### DIFF
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -388,14 +388,31 @@ public class FlutterLogTree extends TreeTable {
       update();
     }
 
+    @Nullable
+    private MutableTreeNode parentStackTraceNode = null;
+
     public void appendNodes(List<FlutterLogEntry> entries) {
       if (treeTable == null || uiThreadAlarm.isDisposed()) {
         return;
       }
-
       uiThreadAlarm.addRequest(() -> {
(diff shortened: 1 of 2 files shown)


src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -388,14 +388,31 @@ public void clearEntries() {
...
         return;
       }
-
       uiThreadAlarm.addRequest(() -> {
         final MutableTreeNode root = getRoot();
-        entries.forEach(entry -> insertNodeInto(new FlutterEventNode(entry), root, root.getChildCount()));
+
+        entries.forEach(entry -> {
+          final FlutterEventNode newNode = new FlutterEventNode(entry);
+          // TODO: How do we enable collapse ????
```
quangson91: @pq Do you have any idea to make FlutterLogTree can collapse ?

pq: Hmmm.  I don't really know the right way to do it though I notice that `JTree`s have a `collapsePath` method.  So maybe something like:

      treeTable.getTree().collapsePath(...);

?

Of course it would be nicer if this was done in the model.  Maybe there's something I'm not seeing?  I'll do a little more looking.

Hixie:
@quangson91 What is your plan with this PR? (No particular rush, I'm just cleaning out the review queue.)

Issue #2597, opened 2018-09-11
"Navigate to testdata" does not work


maksimr:
## Steps to Reproduce

- Create a new project `flutter create foo`
- Open file `foo/test/widget_test.dart`
- Change line 18 to `expect(find.text('1'), findsOneWidget)`
- Run tests using "Flutter test run configuration"
- Try to navigate on the failed test from 'Run' panel 

## Version info

[✓] Flutter (Channel beta, v0.7.3, on Mac OS X 10.13.6 17G65, locale en-RU)
[✓] Android toolchain - develop for Android devices (Android SDK 27.0.3)
[✓] iOS toolchain - develop for iOS devices (Xcode 9.4.1)
[✗] Android Studio (not installed)
[✓] IntelliJ IDEA Ultimate Edition (version 2018.2.3)
[✓] VS Code (version 1.26.1)
[✓] Connected devices (1 available)

pq:
I've got this reproducing on the latest source and will take a look.

Thanks for the report @maksimr!

PR #2599, opened 2018-09-11
Fix flutter test navigation (#2597).

pq:
Fixes: #2597.

image

🤘 

See related PR: https://github.com/dart-lang/test/pull/853/

/cc @jacob314 @DanTup

### DIFF
--- a/src/io/flutter/test/DartTestEventsConverterZ.java
+++ b/src/io/flutter/test/DartTestEventsConverterZ.java
@@ -234,7 +234,7 @@ public class DartTestEventsConverterZ extends OutputToGeneralTestEventsConverter
   }
 
   /**
-   *  Hook to preprocess tests before adding location info and generating a service message.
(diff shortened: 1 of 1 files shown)


src/io/flutter/test/DartTestEventsConverterZ.java
```diff
@@ -730,6 +730,13 @@ public String toString() {
   }
 
   protected static class Test extends Item {
+
+    private static class LocationInfo {
```
jacob314: nit: it would be slightly cleaner to make all fields of this class final with a constructor that specifies all fields.

pq: In general I totally agree.  In this case since it's private and essentially a struct I didn't though.  Plus it would necessitate a bunch of locals to hold on to intermediate values.  But yeah, I'll let it simmer. 👍

Issue #2602, opened 2018-09-12
Flutter plugin is not compatible with IntelliJ 2018.3 EAP


kamilzych:
## Steps to Reproduce

1. Update IntelliJ IDEA to 2018.3 EAP
2. Open a Flutter project
3. There's no support for Flutter in the IDE. If you go to "Plugins" settings you'll see it says that Flutter plugin is not compatible with this version of IntelliJ.

## Version info

```
flutter doctor -v
[✓] Flutter (Channel dev, v0.8.1, on Mac OS X 10.13.5 17F77, locale en-PL)
    • Flutter version 0.8.1 at /Users/kamilzych/flutter
    • Framework revision 0235ec59d1 (6 days ago), 2018-09-06 20:23:13 -0700
    • Engine revision 9b297b3b51
    • Dart version 2.1.0-dev.3.1.flutter-760a9690c2

[✓] Android toolchain - develop for Android devices (Android SDK 28.0.2)
    • Android SDK at /Users/kamilzych/Library/Android/sdk
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-28, build-tools 28.0.2
    • ANDROID_HOME = /Users/kamilzych/Library/Android/sdk
    • Java binary at: /Applications/Android Studio Stable.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1024-b01)
    • All Android licenses accepted.

[✓] iOS toolchain - develop for iOS devices (Xcode 9.4.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 9.4.1, Build version 9F2000
    • ios-deploy 1.9.2
    • CocoaPods version 1.5.3

[!] IntelliJ IDEA Ultimate Edition (version 2018.3 EAP)
    • IntelliJ at /Users/kamilzych/Applications/JetBrains Toolbox/IntelliJ IDEA Ultimate.app
    ✗ Flutter plugin not installed; this adds Flutter specific functionality.
    ✗ Dart plugin not installed; this adds Dart specific functionality.
    • For information about installing plugins, see
      https://flutter.io/intellij-setup/#installing-the-plugins

[!] Connected devices
    ! No devices available

! Doctor found issues in 2 categories.
```

BTW. Although `flutter doctor` says that the Dart plugin is not installed, it's not true - it is.
[...]

PR #2603, opened 2018-09-12
2018.3 EAP support

pq:
Fixes: #2602.

(Build depends on 183.2407.10 getting uploaded to gstore; I don't have the perms for that but Steve said he'd take a look 👍 )

/cc @stevemessick


stevemessick:
The network is misbehaving. I'm trying to upload that file but it is so slow I may not have time today.

stevemessick:
LGTM

pq:
Thanks for the help!

Issue #2458, opened 2018-07-02
Visually mark reload and restarts in the Log View.


pq:
Exploring custom rendering of restart and reload entries to make it easier to see them at a glance.

---

More blue sky: 

* we might also consider showing reload/restart landmarks in the scrollbar.  As done in the editor:

    ![image](https://user-images.githubusercontent.com/67586/42177378-6c4249fa-7de1-11e8-94ae-f71972a9f284.png)

* also related: potential actions to quickly navigate to previous/next restart/reload.

zoechi:
Perhaps a dropdown that allows to select a specific run?
I find continuous logs from multiple runs cumbersome if not properly separated.
There doesn't seem to way with the new log view to get logs purged automatically on reload/rerun.

pq:
> There doesn't seem to way with the new log view to get logs purged automatically on reload/rerun.

👍 I was thinking we'd want this on restart for sure; do you think we want to support configuration for both?

zoechi:
Yes, restart, hot restart, hot reload, should purge the log with an option to not purge at all. 
Others might see this differently though.

I'm usually not interested in logs before any of these events and find it always cumbersome to find the start of the relevant log output.

pq:
Fodder for discussion a PR is out that adds some simple delineation.

![image](https://user-images.githubusercontent.com/67586/45509770-4b63da00-b74d-11e8-831e-34bc83a1e9ae.png)

![image](https://user-images.githubusercontent.com/67586/45509764-49018000-b74d-11e8-8310-e774a85300d0.png)

(See: #2605.)

pq:
Implemented w/ https://github.com/flutter/flutter-intellij/pull/2605/commits/1ceca4c8b3b7210b5c4cbcd27f437bf4244e8c06

PR #2604, opened 2018-09-13
Support for log clear on restart/reload

pq:
Clears log entries on restart / reload.

Configurable via the gear menu:

image

Default is to clear on restart.

/cc @devoncarew @stevemessick 

FYI @quangson91

### DIFF
--- a/src/io/flutter/logging/FlutterLogEntry.java
+++ b/src/io/flutter/logging/FlutterLogEntry.java
@@ -9,7 +9,15 @@ import com.intellij.openapi.util.text.StringUtil;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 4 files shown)


devoncarew:
FYI, I believe that the text console currently clears on both reloads and restarts. It's pretty useful for easily determining what happened after a restart.

pq:
> FYI, I believe that the text console currently clears on both reloads and restarts. It's pretty useful for easily determining what happened after a restart.

Ah!  I guess I've gotten overly used to the current functionality. 😄 

I'll update defaults to agree.

I've got a WIP for drawing a line to distinguish reloads (#2458) that should make the case where folks opt into not clearing more useful too.

pq:
> ... auto-rename was too aggressive and I had to undo a bunch of changes

This is interesting.  I feel like I've experienced this before too.  That said making `Kind` package protected was a mistake; fixed now.  Thanks!

PR #2605, opened 2018-09-13
visuals for reload/restart entries

pq:
Assuming clear entries is not set for reloads and/or restarts, add a marker in the log view to delineate reloads and restarts.

image

image

This is maybe a bit subtle but could serve as a jumping off point for future iteration.

See: #2458


devoncarew:
👍, that looks great!

PR #2614, opened 2018-09-17
Debug attach

stevemessick:
WIP for using breakpoints in modules. There's at least one line of code that is commented-out that I'm not sure won't be needed, so I added a TODO after it.

DartVmServiceDebugProcess.java got reformatted, making the changes look more substantial than they are. The main change is the addition of lines 555-558.

@pq @devoncarew @jacob314

### DIFF
--- a/src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
+++ b/src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
@@ -24,7 +24,11 @@ import com.intellij.openapi.wm.WindowManager;
 import com.intellij.testFramework.LightVirtualFile;
 import com.intellij.util.BitUtil;
 import com.intellij.util.TimeoutUtil;
-import com.intellij.xdebugger.*;
+import com.intellij.xdebugger.XDebugProcess;
+import com.intellij.xdebugger.XDebugSession;
+import com.intellij.xdebugger.XDebugSessionListener;
+import com.intellij.xdebugger.XDebuggerBundle;
+import com.intellij.xdebugger.XSourcePosition;
 import com.intellij.xdebugger.breakpoints.XBreakpointHandler;
 import com.intellij.xdebugger.evaluation.XDebuggerEditorsProvider;
 import com.intellij.xdebugger.evaluation.XDebuggerEvaluator;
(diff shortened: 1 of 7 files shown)


pq:
Sorry for the slowness...  Not sure I'm best qualified but happy to take a look if this is still something you want to land?

stevemessick:
Sure, I already have some changes, but it will be easier to review those after this lands.

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -35,78 +39,84 @@
...
   private final ExecutionEnvironment executionEnvironment;
-
   @NotNull
   private final PositionMapper mapper;
-
+  @Nullable protected String myRemoteProjectRootUri;
+  private boolean myVmConnected = false;
   @NotNull private final OpenDartObservatoryUrlAction myOpenObservatoryAction =
     new OpenDartObservatoryUrlAction(null, () -> myVmConnected && !getSession().isStopped());
+  private VmServiceWrapper myVmServiceWrapper;
```
jacob314: can you layer the formatting only changes as a separate patch so it is easier to tell what is new what what has moved? I'm having trouble quickly telling what is the diff to review.

src/io/flutter/server/vmService/VmServiceWrapper.java
```diff
@@ -204,6 +208,25 @@ public void received(Success response) {
     }
   }
 
+  public void attachIsolate(@NotNull IsolateRef isolateRef) {
+    //myDebugProcess.getSession().initBreakpoints(); // TODO(messick) Remove if not needed.
```
jacob314: you are calling initBreakpoints right before calling this method so I expect this isn't needed unless there is a separate code path.

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
stevemessick: I initially wrote "The main change is the addition of lines 555-558." I double-checked; that is the only change I intended to make. If you want, I can make a PR that just does the formatting. Once that's merged the diff viewer here will only show lines 555-558.

src/io/flutter/server/vmService/VmServiceWrapper.java
stevemessick: The are two paths to this code, and I think that's true for both.

stevemessick:
Thanks for the review!

jacob314:
Lgtm. I've reviewed that range of the file carefully. You are good to
submit.

PR #2621, opened 2018-09-20
Plugin dart2

stevemessick:
Needed to fix a few places where dynamic was causing problems. Also changed the "filesToSkip" list to accept directories, and to allow its entities to be nonexistent. That means it (hopefully) works both before and after Terry's profiler code is in. For good measure, added the "filesToSkip" for his PR.

@pq @devoncarew @jacob314 @terrylucas (gee, did I omit anyone?)

### DIFF
--- a/tool/plugin/lib/plugin.dart
+++ b/tool/plugin/lib/plugin.dart
@@ -525,7 +525,8 @@ class BuildCommand extends ProductCommand {
 
       // Handle skipped files.
       for (var file in spec.filesToSkip) {
-        var entity = FileSystemEntity.isFileSync(file) ? File(file) : Directory(file);
+        var entity =
+            FileSystemEntity.isFileSync(file) ? File(file) : Directory(file);
         if (entity.existsSync()) {
           await entity.rename('$file~');
         }
@@ -602,7 +603,8 @@ class BuildCommand extends ProductCommand {
         // Restore skipped files.
         for (var file in spec.filesToSkip) {
           var name = '$file~';
-          var entity = FileSystemEntity.isFileSync(name) ? File(name) : Directory(name);
+          var entity =
+              FileSystemEntity.isFileSync(name) ? File(name) : Directory(name);
           if (entity.existsSync()) {
             await entity.rename(file);
           }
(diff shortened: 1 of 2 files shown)


tool/plugin/lib/plugin.dart
```diff
@@ -728,7 +737,7 @@ class BuildSpec {
   final String untilBuild;
   final String pluginId = 'io.flutter';
   final String release;
-  final List<String> filesToSkip;
```
jacob314: how are these not strings?

jacob314: nm. I guess they are coming from json.

tool/plugin/lib/plugin.dart
```diff
@@ -525,7 +525,11 @@ class BuildCommand extends ProductCommand {
 
       // Handle skipped files.
       for (var file in spec.filesToSkip) {
```
jacob314: make this line
var (String file in spec.filesToSkip) {
as the type isn't being inferred.

tool/plugin/lib/plugin.dart
stevemessick: yup :)

tool/plugin/lib/plugin.dart
stevemessick: Will do!

stevemessick:
Thanks, Jacob!

tool/plugin/lib/plugin.dart
stevemessick: Wait, if the type isn't being inferred then how does this even work?

tool/plugin/lib/plugin.dart
```diff
@@ -524,7 +524,7 @@ class BuildCommand extends ProductCommand {
       await removeAll('build');
 
       // Handle skipped files.
-      for (var file in spec.filesToSkip) {
+      for (String file in spec.filesToSkip) {
```
jacob314:
```dart
Object object;
String file = object;
```
is the equivalent of
```java
Object object;
String file = (String) object;
```
in java

PR #2625, opened 2018-09-24
Release 28

stevemessick:
@pq Not sure this work. Does it have the state you want?


pq:
Hmmm.  What I really want is the state of the tree when we release so I can do some debugging.

I think the way to do that is maybe to create a new branch / tag from the commit you released from...

be51f02eeb9c0cf71b1a0a6e63b71f505f4dcba4 I think?

pq:
If that's a hassle, I can also just checkout that commit and debug from there.

pq:
Works great.  Thanks!

PR #2626, opened 2018-09-24
Attach debugger

stevemessick:
@jacob314 @devoncarew @pq 

Make breakpoints work after 'flutter attach'.

### DIFF
--- a/src/io/flutter/perf/PerfService.java
+++ b/src/io/flutter/perf/PerfService.java
@@ -37,16 +37,6 @@ public class PerfService {
 
   private boolean isRunning;
 
-  public static void updateRegisteredExtensions(Isolate isolate) {
-    List<FlutterApp> apps = FlutterApp.collectAllApps();
-    for (FlutterApp app : apps) {
-      PerfService service = app.getPerfService();
-      if (service != null) {
-        service.addRegisteredExtensionRPCs(isolate);
-      }
-    }
-  }
-
   public PerfService(@NotNull FlutterDebugProcess debugProcess, @NotNull VmService vmService) {
     this.heapMonitor = new HeapMonitor(vmService, debugProcess);
     this.flutterFramesMonitor = new FlutterFramesMonitor(vmService);
@@ -63,10 +53,6 @@ public class PerfService {
         onVmConnectionClosed();
       }
     });
-
-    //vmService.streamListen(VmService.ISOLATE_STREAM_ID, VmServiceConsumers.EMPTY_SUCCESS_CONSUMER);
-    //vmService.streamListen(VmService.EXTENSION_STREAM_ID, VmServiceConsumers.EMPTY_SUCCESS_CONSUMER);
-    //vmService.streamListen(VmService.GC_STREAM_ID, VmServiceConsumers.EMPTY_SUCCESS_CONSUMER);
 
     // Populate the service extensions info and look for any Flutter views.
     // TODO(devoncarew): This currently returns the first Flutter view found as the
--- a/src/io/flutter/perf/PerfService.java
+++ b/src/io/flutter/perf/PerfService.java
@@ -7,6 +7,7 @@ package io.flutter.perf;
 
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.util.text.StringUtil;
+import io.flutter.run.daemon.FlutterApp;
 import io.flutter.server.vmService.VmServiceConsumers;
 import gnu.trove.THashMap;
 import io.flutter.perf.HeapMonitor.HeapListener;
@@ -14,6 +15,7 @@ import io.flutter.run.FlutterDebugProcess;
 import io.flutter.utils.EventStream;
 import io.flutter.utils.StreamSubscription;
 import io.flutter.utils.VmServiceListenerAdapter;
+import java.util.List;
 import org.dartlang.vm.service.VmService;
 import org.dartlang.vm.service.consumer.GetIsolateConsumer;
 import org.dartlang.vm.service.consumer.VMConsumer;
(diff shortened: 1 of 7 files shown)


src/io/flutter/perf/PerfService.java
```diff
@@ -35,6 +37,16 @@
 
   private boolean isRunning;
 
+  public static void updateRegisteredExtensions(Isolate isolate) {
+    List<FlutterApp> apps = FlutterApp.collectAllApps();
+    for (FlutterApp app : apps) {
+      PerfService service = app.getPerfService();
+      if (service != null) {
+        service.addRegisteredExtensionRPCs(isolate);
```
jacob314: this doesn't seem right. The isolate likely is not a valid isolate for most of these apps.

src/io/flutter/perf/PerfService.java
```diff
@@ -52,9 +64,9 @@ public void connectionClosed() {
       }
     });
 
-    vmService.streamListen(VmService.ISOLATE_STREAM_ID, VmServiceConsumers.EMPTY_SUCCESS_CONSUMER);
-    vmService.streamListen(VmService.EXTENSION_STREAM_ID, VmServiceConsumers.EMPTY_SUCCESS_CONSUMER);
-    vmService.streamListen(VmService.GC_STREAM_ID, VmServiceConsumers.EMPTY_SUCCESS_CONSUMER);
+    //vmService.streamListen(VmService.ISOLATE_STREAM_ID, VmServiceConsumers.EMPTY_SUCCESS_CONSUMER);
```
jacob314: remove this code

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -552,8 +553,8 @@ public void received(Isolate isolate) {
                 else if (eventKind == EventKind.Resume) {
                   // Currently true if we got here via 'flutter attach'
                   ApplicationManager.getApplication().invokeLater(() -> {
-                    getSession().initBreakpoints();
-                    myVmServiceWrapper.attachIsolate(isolateRef);
+                    myVmServiceWrapper.attachIsolate(isolateRef, isolate);
+                    // TODO(messick) onVmConnected(vmService) here, to connect inspector? Seems to work without, not sure why.
```
jacob314: The inspector is smart enough to generally be fine even if the isolate has already started because the inspector window might be opened after the app is already started. Until the inspector window is actually opened we avoid doing work to keep the inspector from degrading app performance.

src/io/flutter/server/vmService/VmServiceWrapper.java
```diff
@@ -249,10 +277,11 @@ public void received(final Isolate isolate) {
     }
   }
 
-  private void setInitialBreakpoints(@NotNull IsolateRef isolateRef) {
+  private void setInitialBreakpointsAndCheckExtensions(@NotNull IsolateRef isolateRef, @NotNull Isolate isolate) {
     doSetBreakpointsForIsolate(myBreakpointHandler.getXBreakpoints(), isolateRef.getId(), () -> {
       myIsolatesInfo.setBreakpointsSet(isolateRef);
     });
+    PerfService.updateRegisteredExtensions(isolate);
```
jacob314: you should be able to just get just the perf service associated with this app instead of making this static.

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -194,6 +196,19 @@ public static FlutterApp fromProjectProcess(@NotNull Project project) {
     return null;
   }
 
+  @NotNull
+  public static List<FlutterApp> collectAllApps() {
+    List<FlutterApp> apps = Lists.newArrayList();
+    Project[] projects = ProjectManager.getInstance().getOpenProjects();
+    for (Project project : projects) {
+      FlutterApp app = fromProjectProcess(project);
```
jacob314: what is the project has multiple apps running? e.g. Android and iOS

src/io/flutter/perf/PerfService.java
stevemessick: Good point. I need to re-think this bit.

src/io/flutter/perf/PerfService.java
stevemessick: Done.

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
stevemessick: I deleted that TODO.

src/io/flutter/server/vmService/VmServiceWrapper.java
stevemessick: Right, this is part of that bit I need to rethink.

src/io/flutter/run/daemon/FlutterApp.java
stevemessick: Each app has its own OS process, doesn't it?

jacob314: so my concern is I don't quite understand how this can get all FlutterApp objects as a Project could have multiple running apps at the same time

jacob314: probably the solution is to modify 
 public static FlutterApp fromProjectProcess(@NotNull Project project)

to
 public static List<FlutterApp> fromProjectProcess(@NotNull Project project) {

src/io/flutter/server/vmService/VmServiceWrapper.java
```diff
@@ -249,10 +278,17 @@ public void received(final Isolate isolate) {
     }
   }
 
-  private void setInitialBreakpoints(@NotNull IsolateRef isolateRef) {
+  private void setInitialBreakpointsAndCheckExtensions(@NotNull IsolateRef isolateRef, @NotNull Isolate isolate) {
     doSetBreakpointsForIsolate(myBreakpointHandler.getXBreakpoints(), isolateRef.getId(), () -> {
       myIsolatesInfo.setBreakpointsSet(isolateRef);
     });
+    FlutterApp app = FlutterApp.fromEnv(myDebugProcess.getExecutionEnvironment());
+    if (app != null) {
```
jacob314: nit: probably add
assert app != null

jacob314:
LGTM!
Thanks for the cleanup.

src/io/flutter/server/vmService/VmServiceWrapper.java
stevemessick: Next PR :)

PR #2632, opened 2018-09-26
Release 29

stevemessick:
@devoncarew @pq


resources/META-INF/plugin.xml
```diff
@@ -10,7 +10,7 @@
   <vendor url="https://github.com/flutter/flutter-intellij">flutter.io</vendor>
 
   <category>Custom Languages</category>
-  
+  <version>29.0.1</version>
```
pq: I didn't think the version # was checked into `plugin.xml`?  (But I'm likely mis-remembering...)

stevemessick: It looks like the plugin tool added it when it shouldn't have. I manually deleted it for now.

PR #2636, opened 2018-09-27
Pop new project window to top

### LINKED ISSUE #2528: New window not visible
While a project window is open, create a new project. Make sure to open new projects in a new window (not replacing current window); change the preference if necessary. Note whether the new window is on top of the window stack or not. Sometimes it is, sometimes it isn't. It should always be on top.

stevemessick:
@pq @jacob314 @devoncarew 

Fixes #2528 

This works for Macs. If it does not work for Linux then we'll need more drastic measures, as in `DartVmServiceDebugProcess.focusProject()`.


stevemessick:
The only change here is the addition of this line:
```dart
ProjectUtil.focusProjectWindow(project, true);
```
I would like to add this to the build we're currently testing. This ensures the newly-created project window is on top when you have multiple project windows open (which is often true if mixing Flutter and Android).

PR #2640, opened 2018-09-28
[experimental] logging channel UI

pq:
Conditionally (if `ENABLE_LOGGING_CHANNELS` is toggled) contributes a simple UI for querying, enabling and disabling log channels as contributed by the [logs package](https://github.com/pq/logs).

image

image

/cc @devoncarew @jacob314

### DIFF
--- a/src/io/flutter/logging/FlutterLogView.java
+++ b/src/io/flutter/logging/FlutterLogView.java
@@ -330,7 +330,8 @@ public class FlutterLogView extends JPanel implements ConsoleView, DataProvider,
       app.getFlutterLog().getLoggingChannels().thenAccept(channels -> ApplicationManager.getApplication().invokeAndWait(() -> {
         final ChannelPanel panel = new ChannelPanel(channels);
         final Rectangle visibleRect = logTree.getVisibleRect();
-        final Point topRight = new Point(logTree.getLocationOnScreen().x + visibleRect.width - 150 /* TODO: make dynamic */,
+        // TODO(pq): make width dynamic based on channel name length
+        final Point topRight = new Point(logTree.getLocationOnScreen().x + visibleRect.width - 150,
                                          logTree.getLocationOnScreen().y + visibleRect.y);
         JBPopupFactory.getInstance()
           .createComponentPopupBuilder(panel, panel)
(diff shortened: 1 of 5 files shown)


src/io/flutter/logging/FlutterLogView.java
```diff
@@ -283,6 +295,56 @@ public void setSelected(AnActionEvent e, boolean state) {
...
+    ConfigureChannelsAction() {
+      super("Configure channels...");
+    }
+
+    @Override
+    public void actionPerformed(AnActionEvent e) {
+      app.getFlutterLog().getLoggingChannels().thenAccept(channels -> ApplicationManager.getApplication().invokeAndWait(() -> {
+        final ChannelPanel panel = new ChannelPanel(channels);
+        final Rectangle visibleRect = logTree.getVisibleRect();
+        final Point topRight = new Point(logTree.getLocationOnScreen().x + visibleRect.width - 150 /* TODO: make dynamic */,
```
jacob314: nit: make this TODO a // and maybe indicate how dynamic it should be.

pq: Done!

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -522,6 +520,7 @@ public String deviceId() {
 
   public void setFlutterDebugProcess(FlutterDebugProcess flutterDebugProcess) {
     myFlutterDebugProcess = flutterDebugProcess;
+    myFlutterLog.setFlutterDebugProcess(flutterDebugProcess);
```
jacob314: I wonder if it makes sense to have the FlutterLog get the debug process off the FlutterApp? it isn't great that the FlutterApp is the grab bag of all Flutter state but at least that means we only have to keep track of the state in one place.

pq: Maybe.  This definitely could benefit from some general cleaning up.  (On my list! ☑️ )

devoncarew:
👍 great to see this the UI control for this land! I think we've generally found the gear menu hard to discover (at least for the Inspector), so we may want to look at ways to expose many or most of the logging channels checkboxs along the toolbar?

PR #2645, opened 2018-10-01
Perf inspector review

jacob314:
This code is loosely based on the coverage tracking work and only exposed if a checkbox in flutter settings is set. The first patch in the CL is just renames followed by a main commit to change the functionality. Over time I tweaked the functionality a lot so the diff for some files isn't useful and I recommend just reviewing the entire files.

Key differences relative to the coverage tracking code:
1. Uses animated icons and highlights ranges of text instead of rendering line markers
2. Displays data for all visible files instead of just the single selected file to avoid confusion switching back and forth between open files.
3. I did a little refactoring to make the code easier to write unit tests for.

If the experimental flag is set then the perf tools are enabled by selecting  the "Track Widget Rebuilds" and "TrackWidgetRepaint" checkboxes from the performance tab.
screen shot 2018-10-01 at 8 39 27 am

Here is a screenshot of what the app looks like running. All visible tabs show the animations. CPU usage showing the running animations is around 30%.
screen shot 2018-10-01 at 8 39 48 am

If you click on a performance icon you get this view.
screen shot 2018-10-01 at 8 42 50 am


src/io/flutter/utils/AnimatedIcon.java
```diff
@@ -0,0 +1,194 @@
+/*
```
jacob314: This file is a pure copy of the matching file from IntelliJ which wasn't available for all versions we support. In the future we may want to tweak this class further to make the animations a bit more efficient.

src/io/flutter/view/InspectorPerfTab.java
```diff
@@ -27,14 +33,14 @@
...
-    labelBox.add(new JLabel("Running in " + app.getLaunchMode() + " mode"));
+    labelBox.add(new JBLabel("Running in " + app.getLaunchMode() + " mode"));
     labelBox.add(Box.createHorizontalGlue());
     labelBox.setBorder(JBUI.Borders.empty(3, 10));
     add(labelBox);
 
     if (app.getLaunchMode() == FlutterLaunchMode.DEBUG) {
       labelBox = Box.createHorizontalBox();
-      labelBox.add(new JLabel("Note: for best results, re-run in profile mode"));
+      labelBox.add(new JBLabel("<html><body><p style='color:red'>WARNING: for best results, re-run in profile mode</p></body></html>"));
```
jacob314: Unrelated one line change: show the warning in red so users realize they should use debug mode. I can revert this if desired.

src/io/flutter/sdk/FlutterSettingsConfigurable.form
```diff
@@ -167,6 +167,15 @@
               <toolTipText value="Memory Profiling tab located inside of the Inspector view displays memory statistics."/>
             </properties>
           </component>
+          <component id="6d6f1" class="javax.swing.JCheckBox" binding="myTrackWidgetPerfCheckBox">
+            <constraints>
+              <grid row="5" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+            </constraints>
+            <properties>
+              <text value="Enable Widget Rebuild and Repaint Profiler"/>
```
jacob314: We can alternately turn this on without an experimental flag. The UI is inactive unless the checkboxes in the perf tab are selected.

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -83,6 +84,8 @@
   private int userReloadCount;
   private int restartCount;
 
+  private long maxFileTimestamp;
```
jacob314: What I needed here was a way to tell that files are out of date from the previous run/hot reload. If there is a way to detect that more simply let me know.

src/io/flutter/perf/VmServiceWidgetPerfProvider.java
```diff
@@ -0,0 +1,187 @@
...
+import org.dartlang.vm.service.element.Event;
+import org.dartlang.vm.service.element.IsolateRef;
+import org.dartlang.vm.service.element.RPCError;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+
+
+public class VmServiceWidgetPerfProvider implements WidgetPerfProvider {
```
jacob314: @devoncarew  this code should generally look familiar although I did a bit of a quick and dirty job to refactor it out for testability.

jacob314:
CL isn't as big as it looks. There is an AnimatedIcon class copied from IntelliJ and a bunch of coverage related classes renamed and then tweaked to refactored to work for the perf icons.

devoncarew:
Will take a look now -

devoncarew:
For the UI in the Inspector performance tab, you might consider an update to separate the UI into `widget rebuild`, `fps`, and `memory usage` named sections (likely in that order). I'd borrow the look from the preference pages, ala:

screen shot 2018-10-01 at 2 49 18 pm

If you reserve some vertical space for the `widget rebuild` area, you should have enough space to show the rebuild details for a specific widget.

src/io/flutter/view/InspectorPerfTab.java
devoncarew: 👍 looks good

src/io/flutter/sdk/FlutterSettingsConfigurable.form
devoncarew: removing this pref sounds good; we can adjust the verbiage in the Inspector UI itself to indicate that this is early access (`[ ] Track widget rebuilds (experimental)` or similar).

src/io/flutter/perf/FlutterWidgetPerf.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    final Multimap<String, TextEditor> editorForPath = LinkedListMultimap.create();
+    final List<String> paths = new ArrayList<>();
+    for (TextEditor editor : fileEditors) {
+      final VirtualFile file = editor.getFile();
+      if (file == null) {
+        continue;
+      }
+      final String path = toSourceLocationUri(file.getPath());
```
devoncarew: rename the `path` var to `uri`?

src/io/flutter/inspector/WidgetPerfPanel.java
```diff
@@ -0,0 +1,88 @@
...
+   * Currently active file editor.
+   */
+  private FileEditor currentEditor;
+
+  /**
+   * Range within the current active file editor to show stats for.
+   */
+  private TextRange currentRange = null;
+
+  public WidgetPerfPanel(Disposable parentDisposable, FlutterApp app) {
```
devoncarew: Add non-nullable for `app`?

src/io/flutter/perf/EditorPerfDecorations.java
```diff
@@ -0,0 +1,502 @@
...
+      }
+    };
+  }
+
+  private void showPerfViewMessage() {
+    final FlutterView flutterView = ServiceManager.getService(getApp().getProject(), FlutterView.class);
+    final InspectorPerfTab inspectorPerfTab = flutterView.showPerfTab(getApp());
+    final StringBuilder sb = new StringBuilder("<html><body>");
+    final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("HH:mm:ss");
+    sb.append("<h2>Widget rebuild stats for ");
```
devoncarew: Use bold instead of h2 for the title? I think we generally want to avoid mixing too many fonts in any one UI.

src/io/flutter/perf/EditorPerfDecorations.java
```diff
@@ -0,0 +1,502 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.*;
+import java.awt.*;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.*;
+import java.util.List;
+
+class EditorPerfDecorations implements EditorMouseListener, EditorPerfModel {
```
devoncarew: I'd add general dartdoc for the (3-4?) larger classes in this PR to help people understand the roles of the different classes.

src/io/flutter/perf/EditorPerfDecorations.java
```diff
@@ -0,0 +1,502 @@
...
+  Color getErrorStripeMarkColor() {
+    // TODO(jacobr): tween from green or blue to red depending on the count.
+    final int count = getCountPastSecond();
+    if (count == 0) {
+      return null;
+    }
+    if (count > HIGH_LOAD_THRESHOLD) {
+      return JBColor.RED;
+    }
+    return JBColor.YELLOW; // TODO(jacobr): should we use green here instead?
```
devoncarew: Did you settle on a color here? I assume below the threshold, the color is there to indicate that we're repaiting the widget, but likely not at a rate which is an issue?

src/io/flutter/perf/EditorPerfDecorations.java
```diff
@@ -0,0 +1,502 @@
...
+
+  public void updateUI(boolean repaint) {
+    final int count = getCountPastSecond();
+    final TextAttributes textAttributes = highlighter.getTextAttributes();
+    assert textAttributes != null;
+    boolean changed = false;
+    if (count > 0) {
+      final Color targetColor = getErrorStripeMarkColor();
+      final double animateTime = (double)(System.currentTimeMillis()) * 0.001;
+      // TODO(jacobr): consider tracking a start time for the individual
```
devoncarew: Just an idea, but you could also tick an animation when a widget repaints. So the speed or rotation would be proportional to the repaint rate.

src/io/flutter/perf/FlutterWidgetPerf.java
```diff
@@ -0,0 +1,273 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static io.flutter.inspector.InspectorService.toSourceLocationUri;
+
+public class FlutterWidgetPerf implements Disposable, Repaintable {
+
+  // Retry requests if we do not receive a response within this interval.
+  private static final long REQUEST_TIMEOUT_INTERVAL = 2000;
+
+  // Intentionally use a low FPS as the animations in EditorPerfDecorations
```
devoncarew: Are we animating all windows or just visible ones?

src/io/flutter/perf/FlutterWidgetPerfManager.java
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  private boolean couldContainWidgets(@Nullable VirtualFile file) {
+    // TODO(jacobr): we might also want to filter for files not under the
+    // current project root.
+    return file != null && FlutterUtils.isDartFile(file);
+  }
+
+  private void updateCurrentAppChanged(@Nullable FlutterApp app) {
+    // TODO(jacobr): we currently only support showing stats for the last app
```
devoncarew: I believe we do have a way to get what we consider to be the front-most app - based on the last active Debug view? `FlutterAppManager.getActiveApp()`, if that helps for this use case

src/io/flutter/sdk/FlutterSettingsConfigurable.form
jacob314: done

src/io/flutter/perf/FlutterWidgetPerf.java
jacob314: done

src/io/flutter/inspector/WidgetPerfPanel.java
jacob314: done

src/io/flutter/perf/EditorPerfDecorations.java
jacob314: Agreed. Done.

src/io/flutter/perf/EditorPerfDecorations.java
jacob314: That could be fun. The one catch is we only get updated repaint count rates once per second currently which is good for keeping the traffic on the vm service protocol low but means we can't quite show the ticks at just the right time but we could fake it base on the data we have for the previous second.

src/io/flutter/perf/FlutterWidgetPerf.java
jacob314: just ones that are visible.

src/io/flutter/perf/FlutterWidgetPerfManager.java
jacob314: Expanded the comment. I think what we really want is to collect and aggregate thats from all running apps.

src/io/flutter/perf/EditorPerfDecorations.java
jacob314: Added a bunch more top level comments.

src/io/flutter/perf/EditorPerfDecorations.java
jacob314: I'm leaning towards yellow as yellow as a background color just seems like a good low intensity highlight but I could be convinced otherwise. It would be nice to settle on a good grey->green->yellow->red scheme for all performance related stats.

jacob314:
screen shot 2018-10-01 at 8 02 59 pm

Screenshot of the updated UI matching the style of the settings panel.

jacob314:
PTAL

src/io/flutter/perf/EditorPerfDecorations.java
devoncarew: Yeah, the position of some dial hand in a icon could give you a sense of both which specific widgets repainted, and the rate or repaint.

devoncarew:
And, consider moving the widget stats area to the first panel - I'm not sure if this will look better or not, and would depend on having a fixed height for the widget stats area.

jacob314:
Applied all the cleanup suggested except kept the widget stats area where it was for now. 
Also fixed a bug I'd missed on hot restart where we were sometimes calling a service extension before it was present causing problems.
Here is what the perf tab looks like. I've kept the new Widget Stats feature at the bottom as I think the right flow is users look at the FPS to decide if they might have an issue first before using the widget stats feature to diagnose the problem. I can be convinced otherwise though. The new UI looks good although there is a somewhat excessive number of horizontal lines due to lines in the graphs themselves.
screen shot 2018-10-02 at 4 56 57 pm

jacob314:
If there is anything you would like to see fixed in the patch I added after the lgtm let me know.

devoncarew:
Thanks for the changes! The UI looks great.

> there is a somewhat excessive number of horizontal lines due to lines in the graphs themselves

Yup, agreed; I may play with it locally as well.

jacob314:
Please feel free to send any and all polish for this feature. The current implementation is just a fully functional baseline implementation. The message displayed on clicking a gutter icon is barely acceptable, the icons used are a bit ugly, and it isn't clear whether the flashing borders for widgets being rebuilt are a good thing.

jacob314:
As far as the graphs go, it would be nice to reuse the Android Studio graphs instead of polishing the existing graphs much more ourselves. In particular highest priority is to highlight slow frames in red and provide tooltips overing over frames. @terrylucas

devoncarew:
I have some slight tweaks to the UI in a branch (and am currently fighting with some Swing layout issues):

screen shot 2018-10-02 at 7 26 32 pm

PR #2653, opened 2018-10-02
Attach debugger button for IntelliJ

stevemessick:
@pq @devoncarew 

- Add the attach button to the toolbar in IntelliJ (and AS). (We promised to do this.)
- Change the AS-specific action class to hide the Android attach button in Flutter projects.
- Change the name of the button to `Attach debugger to embedded Flutter process`
- API change: add DataContext parameter

/cc @matthew-carroll

### DIFF
--- a/flutter-studio/src/io/flutter/actions/ConnectAndroidDebuggerAction.java
+++ b/flutter-studio/src/io/flutter/actions/ConnectAndroidDebuggerAction.java
@@ -5,158 +5,25 @@
  */
 package io.flutter.actions;
 
-import com.google.common.base.Joiner;
-import com.intellij.execution.Executor;
-import com.intellij.execution.ProgramRunnerUtil;
-import com.intellij.execution.RunnerAndConfigurationSettings;
-import com.intellij.execution.configurations.RunConfiguration;
-import com.intellij.execution.runners.ExecutionEnvironment;
-import com.intellij.execution.runners.ExecutionEnvironmentBuilder;
 import com.intellij.openapi.actionSystem.AnActionEvent;
-import com.intellij.openapi.application.ApplicationManager;
-import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.project.Project;
-import com.intellij.openapi.ui.DialogWrapper;
-import com.intellij.openapi.ui.Messages;
-import com.intellij.openapi.wm.ToolWindowId;
-import io.flutter.FlutterInitializer;
-import io.flutter.FlutterUtils;
-import io.flutter.run.FlutterLaunchMode;
-import io.flutter.run.SdkFields;
-import io.flutter.run.SdkRunConfig;
-import io.flutter.run.SdkAttachConfig;
 import io.flutter.sdk.FlutterSdkUtil;
 import org.jetbrains.android.actions.AndroidConnectDebuggerAction;
-import org.jetbrains.annotations.Nullable;
-
-import javax.swing.*;
-import java.util.ArrayList;
-import java.util.List;
 
 public class ConnectAndroidDebuggerAction extends AndroidConnectDebuggerAction {
-  @SuppressWarnings("FieldCanBeLocal")
-  private static String RUN_DEBUG_LINK = "https://flutter.io/using-ide/#running-and-debugging";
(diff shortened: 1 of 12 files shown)


devoncarew:
Awesome! Reviewing now...

> Change the AS-specific action class to hide the Android attach button in Flutter projects.

I'm curious if we need to do this, as long as our attach button is easily differentiated from the pure Android one.

flutter-studio/src/io/flutter/actions/ConnectAndroidDebuggerAction.java
```diff
@@ -5,158 +5,25 @@
...
-      return;
-    }
     Project project = e.getProject();
     if (project == null || project.isDefault()) {
       super.update(e);
       return;
     }
-    if (!FlutterSdkUtil.hasFlutterModules(project)) {
-      super.update(e);
+    if (FlutterSdkUtil.hasFlutterModules(project)) {
```
devoncarew: I'd consider not changing the default behavior here - not extending `ConnectAndroidDebuggerAction` - and just making sure that our icon + text are distinct from this action.

resources/META-INF/plugin.xml
```diff
@@ -604,6 +604,13 @@
       <add-to-group group-id="ToolbarRunGroup" anchor="after" relative-to-action="RunnerActions"/>
       <keyboard-shortcut keymap="$default" first-keystroke="ctrl BACK_SLASH"/>
     </action>
+    <action id="AttachDebuggerAction"
+            class="io.flutter.actions.AttachDebuggerAction"
+            text="Attach debugger to embedded Flutter process"
+            description="Attach debugger to a Flutter process embedded in an Android app"
+            icon="AllIcons.Debugger.AttachToProcess">
```
devoncarew: Do we need to mock up a Flutter specific version of this?

src/io/flutter/actions/FlutterCleanAction.java
```diff
@@ -14,7 +15,7 @@
 
 public class FlutterCleanAction extends FlutterSdkAction {
   @Override
-  public void startCommand(@NotNull Project project, @NotNull FlutterSdk sdk, @Nullable PubRoot root) {
+  public void startCommand(@NotNull Project project, @NotNull FlutterSdk sdk, @Nullable PubRoot root, @NotNull DataContext context) {
```
devoncarew: What is the DataContext used for?

flutter-studio/src/io/flutter/actions/ConnectAndroidDebuggerAction.java
stevemessick: That alternative isn't really great, I think. It would leave the toolbar cluttered with an unused but active button. When clicked it presents this confusing pane:

screen shot 2018-10-03 at 12 57 14 pm

I believe that's a poor UX. It's inviting issue creation, don't you think?

src/io/flutter/actions/FlutterCleanAction.java
stevemessick: I traced it down to ExecutionEnvironment. I stopped tracing when I got to a platform class, so I don't have a good answer for that.

flutter-studio/src/io/flutter/actions/ConnectAndroidDebuggerAction.java
devoncarew: It's definitely a place users could get caught up. We'd be balancing that vs potential issues - in the current AS/IntelliJ release and in future ones - that subclassing the action and (slightly) modifying its behavior causes issues.

resources/META-INF/plugin.xml
devoncarew: Sure, I'll see about creating one.

src/io/flutter/actions/FlutterCleanAction.java
devoncarew: But why are we adding it as a parameter here?

devoncarew: OK, it looks like it's used in the startCommand() method of the `AttachDebuggerAction` class.

PR #2658, opened 2018-10-03
Perf ui tweaks

devoncarew:
- follow up to #2645, some tweaks to the UI for the performance tab in the inspector

screen shot 2018-10-03 at 8 40 15 am

screen shot 2018-10-03 at 8 40 21 am

@jacob314


jacob314:
I've re-landed. you should be good to go.

Issue #2617, opened 2018-09-19
NPE during debugging


mdudek:
## What happened

(please describe what you were doing when this exception occurred)

## Version information

Android Studio `3.1.4` • Flutter plugin `io.flutter 28.0.1` • Dart plugin `173.4700`

Error getting Flutter sdk information.

## Exception

null

```
java.lang.NullPointerException
	at org.dartlang.vm.service.element.Script.getSource(Script.java:43)
	at io.flutter.run.ObservatoryFile.createSnapshot(ObservatoryFile.java:109)
	at io.flutter.run.ObservatoryFile.<init>(ObservatoryFile.java:54)
	at io.flutter.run.ObservatoryFile$Cache.downloadOrGet(ObservatoryFile.java:152)
	at io.flutter.run.PositionMapper.getSourcePosition(PositionMapper.java:259)
	at io.flutter.run.PositionMapper.getSourcePosition(PositionMapper.java:232)
	at io.flutter.server.vmService.DartVmServiceDebugProcess.getSourcePosition(DartVmServiceDebugProcess.java:461)
	at io.flutter.server.vmService.frame.DartVmServiceStackFrame.<init>(DartVmServiceStackFrame.java:46)
	at io.flutter.server.vmService.VmServiceWrapper$9.lambda$received$0(VmServiceWrapper.java:414)
	at com.intellij.openapi.application.impl.ApplicationImpl$1.run(ApplicationImpl.java:315)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

jacob314:
That is a surprising exception. Do you have a Dart source project that can reproduce the problem?

mdudek:
Can you please send me your email address so I can send you my flutter project sources.

kevmoo:
I just hit this, too.

```
IntelliJ IDEA `2018.2.4` • Flutter plugin `io.flutter 29.0.4` • Dart plugin `182.4323.44`

Flutter 0.9.5 • channel dev • ssh://git@github.com/flutter/flutter
[...]

PR #2663, opened 2018-10-04
handle nulls from the service protocol

devoncarew:
- commit the latest version of the service protocol library (this version returns nulls for some fields, and documents which ones)
- update to some signature changes from the new library version
- be tolerant of nulls from Isolate.extensionRPCs
(fix https://github.com/dart-lang/vm_service_drivers/issues/185)
- handle cases where `Script.source` is null (this is new behavior in the VM, and happens for some `dart:` sources); fix https://github.com/flutter/flutter-intellij/issues/2617

The bulk of these changes are from the auto-generated protocol library.

@jacob314 @stevemessick


devoncarew:
We may want to patch this into the existing release - it's a major impediment to debugging against flutter master, and will be the same for flutter beta.

PR #2665, opened 2018-10-04
Remove fest dependency

stevemessick:
@terrylucas @jacob314


stevemessick:
TBR

@devoncarew We must have a problem with the travis config. This PR should have broken the build; it does break my local build (can't launch debugger).

PR #2669, opened 2018-10-05
compress pngs

kevmoo:
Drop ~400k in total


devoncarew:
Did you use command-line tooling for this? I ask because the images in `//resources/flutter/icons/` are generated, by the `grind icons` script. We don't run that script often, and I believe it depends on running on a mac for some of the cli tools it drives, but it would be nice to know how to augment the script for the next time we gen the icons.

kevmoo:
I use a tool (on homebrew) called `pngquant`

Here's my script

```
#!/usr/bin/env sh
ls -lR | grep -v '^d' | grep ".png" | awk '{total += $5} END {print "Before:", total}'
find . -name '*.png' -exec pngquant --speed 1 --ext .png --force 256 --skip-if-larger {} \;
ls -lR | grep -v '^d' | grep ".png" | awk '{total += $5} END {print " After:", total}'
```

PR #2681, opened 2018-10-09
drop support for 2017.3 (and AS 3.1)

devoncarew:
- drop support for 2017.3 (and AS 3.1)

I believe it's ok to drop support for these two platforms now.

@jwren @stevemessick @DaveShuckerow


jwren:
LGTM

PR #2687, opened 2018-10-11
color log categories

pq:
image

Colors up for discussion and more presets to come.

FWIW, I tried right-align and it looked more odd to my eye:

image

Some spacing could maybe help?  Dunno.

/cc @devoncarew


devoncarew:
Yeah, I don't have any better opinions here, but it does seem like we could tweak it a bit more?

I don't know if it helps, but I did right-align the web based logging UI; that - with some padding between the categories and the message - seemed to work.

screen shot 2018-10-11 at 2 50 58 pm

pq:
Agreed.  I'll follow-up w/ some tweaking.

PR #2716, opened 2018-10-22
process and present log data

pq:
image

/cc @devoncarew @jacob314

### DIFF
--- a/src/io/flutter/logging/FlutterLogEntry.java
+++ b/src/io/flutter/logging/FlutterLogEntry.java
@@ -27,6 +27,8 @@ public class FlutterLogEntry {
   private final int level;
   @NotNull
   private final String message;
+  @Nullable
+  private String data;
   private int sequenceNumber = -1;
 
   @NotNull
@@ -55,6 +57,15 @@ public class FlutterLogEntry {
 
   public FlutterLogEntry(long timestamp, @NotNull String category, @Nullable String message) {
     this(timestamp, category, UNDEFINED_LEVEL.value, message);
+  }
+
+  @Nullable
+  public String getData() {
+    return data;
+  }
+
+  public void setData(@Nullable String data) {
+    this.data = data;
   }
 
   @NotNull
--- a/src/io/flutter/logging/FlutterLogEntry.java
+++ b/src/io/flutter/logging/FlutterLogEntry.java
@@ -27,6 +27,7 @@ public class FlutterLogEntry {
   private final int level;
   @NotNull
   private final String message;
+  // TODO(pq): consider making data an Instance or JsonElement
   @Nullable
   private String data;
   private int sequenceNumber = -1;
--- a/src/io/flutter/logging/FlutterLogEntryParser.java
+++ b/src/io/flutter/logging/FlutterLogEntryParser.java
@@ -17,9 +17,8 @@ import io.flutter.run.FlutterDebugProcess;
 import io.flutter.run.daemon.DaemonApi;
 import io.flutter.utils.StdoutJsonParser;
 import org.dartlang.vm.service.VmService;
-import org.dartlang.vm.service.element.Event;
-import org.dartlang.vm.service.element.Instance;
-import org.dartlang.vm.service.element.IsolateRef;
+import org.dartlang.vm.service.consumer.GetObjectConsumer;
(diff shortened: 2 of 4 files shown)


pq:
(FWIW: if we don't feel 100% about this for the release we can put it behind a compile-time flag.)

src/io/flutter/logging/FlutterLogEntry.java
```diff
@@ -27,6 +27,8 @@
   private final int level;
   @NotNull
   private final String message;
+  @Nullable
+  private String data;
```
jacob314: maybe make data a JsonElement?

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -81,13 +80,40 @@ private static FlutterLogEntry parseLoggingEvent(@NotNull Event event) {
...
+      final Isolate isolate = new Isolate(json.get("isolate").getAsJsonObject());
+      debugProcess.getVmServiceWrapper().getObject(isolate.getId(), data.getId(), new GetObjectConsumer() {
+        @Override
+        public void received(Obj response) {
+          entry.setData(((Instance)response).getValueAsString());
+        }
+
+        @Override
+        public void received(Sentinel response) {
+          // ignored
```
jacob314: probably setData 
as null here

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -81,13 +80,40 @@ private static FlutterLogEntry parseLoggingEvent(@NotNull Event event) {
...
     // value and update this entry after creation.
     String messageStr = message.getValueAsString();
     if (message.getValueAsStringIsTruncated()) {
       messageStr += "...";
     }
-    return new FlutterLogEntry(timestamp(event), category, level, messageStr);
+
+    final FlutterLogEntry entry = new FlutterLogEntry(timestamp(event), category, level, messageStr);
+
+    final Instance data = new Instance(logRecord.getAsJsonObject().get("error").getAsJsonObject());
```
jacob314: consider using a CompletableFuture so you can await the future when showing the summary view so that you don't show null if there is a race condition.

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -603,8 +639,18 @@ protected void onScrollToEnd() {
...
-    toolWindowPanel.setContent(pane);
+    logModel.setScrollPane(treePane);
+
+    final JScrollPane dataPane = ScrollPaneFactory.createScrollPane(
+      dataPanel,
+      ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED,
+      ScrollPaneConstants.HORIZONTAL_SCROLLBAR_AS_NEEDED);
+
+    // TODO(pq): consider an affordance to hide/show data pane.
+    final Splitter treeSplitter = new Splitter(false, .75f);
```
jacob314: you can copy logic from the inspector to cache the splitter position.

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -405,19 +410,43 @@ public void disableIfNeeded() {
...
+  class DataPanel extends JEditorPane {
+    DataPanel() {
+      setEditable(false);
+    }
+
+    public void update() {
+      final StringBuilder sb = new StringBuilder();
+      for (FlutterLogTree.FlutterEventNode node : logTree.getSelectedNodes()) {
+        final String data = node.entry.getData();
+        if (data != null && !Objects.equals(data, "null")) {
```
jacob314: showing multiple selected nodes is a bit surprising to me but maybe it is helpful. I don't think any of other other master-detail views allow multiple selection like that... at least not yet.

src/io/flutter/logging/FlutterLogEntryParser.java
pq: Good idea. done.

src/io/flutter/logging/FlutterLogEntry.java
pq: I started to do this actually (`Instance`) and then realized there may be other related changes.  I've added a TODO to revisit.

Thanks!

PR #2720, opened 2018-10-23
preserve log selections; filter enhancements

pq:
preserve log selections / filter enhancements

* shows log counts for “empty” filters
* preserves selections on tree changes
* removes defunct entry filter class
* optimizes empty filter case
* mitigates occasional range check failure on tree initialization

/cc @jacob314 @devoncarew


src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -411,6 +413,9 @@ void update() {
 
       reload(getRoot());
       treeTable.updateUI();
+      // Preserve / restore selection state.
+      treeTable.getTree().setSelectionRows(treeTable.getSelectedRows());
```
jacob314: I'm a little surprised this is line is required.

pq: Me too.   There's something about the UI update that does not preserve the selection.  I tried a bunch of different things and this is the only thing that worked.

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -706,31 +675,33 @@ public void setFilter(@NotNull FlutterLogFilterPanel.FilterParam filter) {
...
+      try {
+        return RowFilter.regexFilter(
+          standardRegex,
+          FlutterTreeTableModel.ColumnIndex.MESSAGE.index,
+          FlutterTreeTableModel.ColumnIndex.CATEGORY.index,
+          FlutterTreeTableModel.ColumnIndex.LOG_LEVEL.index
+        );
+      }
+      catch (PatternSyntaxException e) {
+        // Fall through.
```
jacob314: not a new issue but we should notify the user if the regex is invalid.

src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -747,6 +718,8 @@ private void updateCounter() {
   }
 
   public void clearEntries() {
+    // Ensure selections are cleared on empty.
+    getTree().setSelectionRows(new int[]{});
```
jacob314: use 
    this.getTree().clearSelection();

instead

pq: Good idea.

src/io/flutter/logging/FlutterLogTree.java
pq: Yeah.  I'm not sure what the best UI is but ignoring isn't great.  I'll add a TODO.

PR #2728, opened 2018-10-26
Build for AS 3.3 beta

stevemessick:
@pq @devoncarew @terrylucas @jacob314 @keertip 

 - Update the plugin tool to build for AS 3.3 beta.
 - Patch sources to match API changes in FlutterProjectSystem and FlutterStudioMonitorStageView
 - Generate new build files
 - Delete dead code for building AS 3.1

After merging this PR you'll need to update your Android Studio dependency to use 3.3; the 3.2 version won't compile anymore unless using the plugin tool.

You won't be able to build using the plugin tool until the new Android Studio artifact is uploaded to the cloud. I'll do that next week, but if someone wants to do it sooner, the file name is android-studio-ide-182.5073496-linux.zip and can be downloaded from the Android Studio preview site.

@terrylucas I don't know if the change to FlutterStudioMonitorStageView is correct. If it isn't you'll need to update plugin.dart to match whatever it should be. See line 553.


terrylucas:
Thanks Steve the change to FlutterStudioMonitorStageView looks right.  I'll verify when I update.

devoncarew:
@stevemessick - I'm uploading the new archive (`android-studio-ide-182.5073496-linux.zip`) now :)

PR #2735, opened 2018-10-29
Read a few more fields to DiagnosticsNode json and wire up parent nodes

jacob314:
to child nodes by default.


jacob314:
These tweaks are useful for a follow on CL that calculates lints based on the shape of the widget tree combined with widget rebuild counts.

PR #2738, opened 2018-10-30
Build 3.3beta for M30 release

stevemessick:
Had to re-create this four day old PR because it was too out of date to manage.

@pq @devoncarew @terrylucas @jacob314 @keertip 

 - Update the plugin tool to build for AS 3.3 beta.
 - Patch sources to match API changes in FlutterProjectSystem and FlutterStudioMonitorStageView
 - Generate new build files
 - Delete dead code for building AS 3.1

After merging this PR you'll need to update your Android Studio dependency to use 3.3; the 3.2 version won't compile anymore unless using the plugin tool.

You won't be able to build using the plugin tool until the new Android Studio artifact is uploaded to the cloud. I'll do that next week, but if someone wants to do it sooner, the file name is android-studio-ide-182.5073496-linux.zip and can be downloaded from the Android Studio preview site.

@terrylucas I don't know if the change to FlutterStudioMonitorStageView is correct. If it isn't you'll need to update plugin.dart to match whatever it should be. See line 553.

### DIFF
--- a/flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
+++ b/flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
@@ -12,11 +12,15 @@ import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleManager;
 import com.intellij.openapi.project.DumbService;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.project.impl.ProjectImpl;
+import com.intellij.openapi.projectRoots.Sdk;
 import com.intellij.openapi.roots.ModuleRootEvent;
 import com.intellij.openapi.roots.ModuleRootListener;
+import com.intellij.openapi.roots.ModuleRootManager;
 import com.intellij.openapi.roots.OrderRootType;
 import com.intellij.openapi.roots.libraries.Library;
 import com.intellij.openapi.roots.libraries.LibraryTable;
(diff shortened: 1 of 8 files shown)


flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
```diff
@@ -58,13 +63,28 @@ public void update() {
   }
 
   private Void updateAndroidLibraryContent(@NotNull Project androidProject) {
+    ModuleManager mgr = ModuleManager.getInstance(androidProject);
+    Module module = null;
+    for (Module mod : mgr.getModules()) {
+      if (mod.getName().equals("android") || mod.getName().equals(".android")) {
```
jacob314: wow surprising that both these module names exist.

tool/plugin/lib/plugin.dart
```diff
@@ -532,73 +532,40 @@ class BuildCommand extends ProductCommand {
...
-        oldSource3 = processedFile3.readAsStringSync();
-        newSource = oldSource3;
-        newSource = newSource.replaceAll('.LightResourceClassService', '.*');
-        newSource =
-            newSource.replaceAll(' LightResourceClassService', ' Object');
-        processedFile3.writeAsStringSync(newSource);
-      } else if (spec.version == '2018.1') {
+      // TODO: Remove this when we no longer support AS 3.2.
+      File processedFile3, processedFile4;
+      String oldSource3, oldSource4, newSource;
```
jacob314: I know you didn't add it but it could be more robust to have
Map<String, File> oldSources;
so you can ensure you don't forget to restore an old source file or have a similar typo.

flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
stevemessick: I agree; this is a side-effect of the Add2App project which had a reason for using ".android" vs "android".

tool/plugin/lib/plugin.dart
stevemessick: Good point, and it was really hard to debug when that happened. Was hoping the problem would go away but I no longer think it will. We might want to see if this can be made data-driven from entries in the product matrix.

stevemessick:
Thanks for the review!

PR #2741, opened 2018-10-30
Keep edited files in a Map

stevemessick:
Following up a review comment from @jacob314, this changes the way edited files are managed during a build. It is less likely that a file will be left in its edited state.

/cc @devoncarew

### DIFF
--- a/tool/plugin/lib/plugin.dart
+++ b/tool/plugin/lib/plugin.dart
@@ -533,7 +533,7 @@ class BuildCommand extends ProductCommand {
       }
 
       // TODO: Remove this when we no longer support AS 3.2.
-      var files = Map<File, String>();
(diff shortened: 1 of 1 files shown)


tool/plugin/lib/plugin.dart
```diff
@@ -533,39 +533,34 @@ class BuildCommand extends ProductCommand {
       }
 
       // TODO: Remove this when we no longer support AS 3.2.
-      File processedFile3, processedFile4;
-      String oldSource3, oldSource4, newSource;
+      var files = Map<File, String>();
```
jacob314: nit: write as `<File, String>{}`   instead of `Map<File, String>()`

stevemessick: Done!

PR #2745, opened 2018-10-31
Build for EAP and Canary

stevemessick:
Update build to produce a single zip that works in IJ EAP and AS Canary IDEs.

After this is merged source builds need to use Dart plugin sources on branch 183.2153 of intellij-plugins. The Android Studio sources are sync'd to build 5081642.

@terry @jacob314 @devoncarew

### DIFF
--- a/flutter-intellij.iml
+++ b/flutter-intellij.iml
@@ -58,5 +58,6 @@
     </orderEntry>
     <orderEntry type="library" name="Dart SDK" level="project" />
     <orderEntry type="library" name="Dart Packages" level="project" />
+    <orderEntry type="library" name="netty-codec-http" level="project" />
   </component>
 </module>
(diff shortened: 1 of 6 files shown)


flutter-intellij.iml
```diff
@@ -58,5 +58,6 @@
     </orderEntry>
     <orderEntry type="library" name="Dart SDK" level="project" />
     <orderEntry type="library" name="Dart Packages" level="project" />
+    <orderEntry type="library" name="netty-codec-http" level="project" />
```
devoncarew: Is this change needed? I don't see any references in Java files.

devoncarew:
> After this is merged source builds need to use Dart plugin sources on branch 183.2153 of intellij-plugins.

Let me know if we need any new uploads to google storage -

stevemessick:
If we merge this PR then anyone working with AS sources would need to
switch to the latest Canary version. I'm not so sure that I want to go that
route. I'm thinking about rewriting the edits to be applied to the Beta
version, when compiling for EAP/Canary. That would keep us on a more stable
source base for the next couple months.

The change to the netty library is required when compiling from an EAP
base. The Dart plugin already does it that way.

And thanks, but I already uploaded the new AS binary.

stevemessick:
I'm going to optimistically merge this. If it causes too many problems then it will be reverted.

PR #2748, opened 2018-11-01
Update link for Opacity animation optimization.

jacob314:


devoncarew:
> For long-term production use, is there a way to detect this kind of URL change

It should be possible for us to locale our use of urls into one file, and have a build step in travis that checks those urls for 404s.

jacob314:
Verifying that the URLs are valid sounds good. We'll need to verify that the referenced locations in the document still exist as well as verifying the links don't 404 as in this case the html file exists but the link got renamed.

Another long term alternative if it turns out all the perf tips are links to dart docs is to encode the conditions of the perf tips directly in the dart docs with annotations. That way the tip rules and the suggested content wouldn't get out of date. We haven't had time to add widget category annotations to package:flutter instead of keeping them in a json file on the side so I expect this won't happen soon as the amount of work compared to the immediate user benefit is small.

PR #2749, opened 2018-11-01
Turn on TrackWidgetCreation when the Flutter version is new enough.

jacob314:
@DanTup I think you are doing something similar for VSCode.

### DIFF
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.form
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.form
@@ -169,7 +169,7 @@
           </component>
         </children>
       </grid>
-      <grid id="919ec" layout-manager="GridLayoutManager" row-count="3" column-count="2" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
+      <grid id="919ec" layout-manager="GridLayoutManager" row-count="4" column-count="2" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
         <margin top="0" left="0" bottom="0" right="0"/>
         <constraints>
           <grid row="1" column="0" row-span="1" col-span="1" vsize-policy="3" hsize-policy="3" anchor="0" fill="3" indent="0" use-parent-layout="false"/>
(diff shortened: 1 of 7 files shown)


DanTup:
Yup, I'm reading a default in the case where the user hasn't set it explicitly:

https://github.com/Dart-Code/Dart-Code/blob/a62545dcd5da2c884b5ccce12157f36da3eca9bf/src/providers/debug_config_provider.ts#L294-L301

And the default is based on the version:

https://github.com/Dart-Code/Dart-Code/blob/a62545dcd5da2c884b5ccce12157f36da3eca9bf/src/flutter/capabilities.ts#L13

I had to use `0.10.2-pre` as my version for comparing, since apparently semver considers `0.10.2-pre.something` to be less than `0.10.2` ("Pre-release versions have a lower precedence than the associated normal version"). I don't know whether that affects you here (if your version is only the x.y.z part, then probably not).

jacob314:
Tip of trunk flutter shows as 0.10.2 so this works when I've tested locally but 0.10.2-pre does seem more idiomatic.

DanTup:
> Tip of trunk flutter shows as 0.10.2

It probably depends how you get the version. I'm just reading a string directly out of the `version` file which has `0.10.2-pre.119` for my checkout of master. On the dev/beta channels I guess it won't have a suffix at all. If it already works like this, I guess your version has probably already been trimmed (or the comparisons for your version class behave differently! 🤔)

jacob314:
I'm iterating on this CL slightly. I don't like the current behavior if you switch back and forth between a post 0.10.2 version and a 0.10.1 version or if you first run the build with a 0.10.1 version and switch to 0.10.2.
The problem is we are saving the setting even when the user didn't actually change it.

jacob314:
Avoid the bugs mentioned above. If desired I can make the settings panel more polished hiding the option to disable track widget creation on old builds and hiding the option to enable on new builds.
Let me know.
Here is what the UI looks like now.
screen shot 2018-11-01 at 11 44 34 am

devoncarew:
For `Flutter builds before 0.10.2 only` - the setting no longer disables it after that point?

src/io/flutter/sdk/FlutterSettingsConfigurable.form
```diff
@@ -197,13 +197,22 @@
...
+              <text value="Disable tracking widget creation locations"/>
+              <toolTipText value="Disabling this setting breaks advanced Flutter Inspector functionality and performance tools. Only disable if you are noticing unexpected behavior differences between apps run in debug and release build."/>
+            </properties>
+          </component>
+          <component id="6d6f0" class="javax.swing.JCheckBox" binding="myLegacyTrackWidgetCreationCheckBox">
+            <constraints>
+              <grid row="3" column="0" row-span="1" col-span="2" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+            </constraints>
+            <properties>
+              <text value="[Flutter builds before 0.10.2 only] Run applications with --track-widget-creation"/>
```
devoncarew: Instead of `[Flutter builds before 0.10.2 only] Run applications with --track-widget-creation`, perhaps: `Run applications with --track-widget-creation (always enabled for Flutter v0.10.2 and later)`?

jacob314: Sounds good. Done.

jacob314:
Switched the UI so we only show the option appropriate for your Flutter SDK version.
Tweaked the logic so disabling doesn't apply when using an old sdk so we don't have a confusing case if you disabled and then switched to an old sdk.
smaller

PR #2751, opened 2018-11-01
move the memory tab into its own tool window

devoncarew:
- move the memory tab into its own tool window

This splits the memory tab out of the Inspector and moves it into its own tool window. It also makes it only available contributed in Android Studio; we can adjust that once the charting lib is available in IntelliJ. The functionality of the memory view should be unaffected.

The tool window is named `Flutter Performance`, but we can noodle on alternative names (once we ship it however we should stay consistent). Note that this work is not for the current (M30) release.

screen shot 2018-10-31 at 7 28 42 pm

### DIFF
--- a/src/io/flutter/view/FlutterPerfView.java
+++ b/src/io/flutter/view/FlutterPerfView.java
@@ -1,0 +1,197 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.view;
+
+import com.intellij.openapi.Disposable;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.SimpleToolWindowPanel;
+import com.intellij.openapi.util.Disposer;
+import com.intellij.openapi.wm.ToolWindow;
+import com.intellij.openapi.wm.ToolWindowManager;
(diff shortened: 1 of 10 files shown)


src/io/flutter/view/FlutterPerfView.java
```diff
@@ -0,0 +1,197 @@
...
+
+public class FlutterPerfView {
+  public static final String TOOL_WINDOW_ID = "Flutter Performance";
+
+  private static final Logger LOG = Logger.getInstance(FlutterPerfView.class);
+
+  @NotNull
+  private final Project myProject;
+
+  private final Map<FlutterApp, PerfViewAppState> perAppViewState = new HashMap<>();
```
jacob314: There does seem to be a lot of duplicate logic managing perAppViewState with FlutterView.
Perhaps add a TabPerAppView<T> to avoid the duplication unless it gets messy?

src/io/flutter/view/FlutterPerfView.java
```diff
@@ -0,0 +1,197 @@
...
+    addPerformanceViewContent(app, toolWindow);
+
+    app.getVmService().addVmServiceListener(new VmServiceListenerAdapter() {
+      @Override
+      public void connectionOpened() {
+        onAppChanged(app);
+      }
+
+      @Override
+      public void connectionClosed() {
```
jacob314: would be nice to share this with FlutterView for example.

devoncarew:
> There does seem to be a lot of duplicate logic managing perAppViewState with FlutterView.

This PR isn't time critical - I'll iterate a bit on it to look to share common code, and put it out for re-review.

devoncarew:
From an off-line discussion, I'll land this PR as it; we can look for opportunities to share code going forward.

PR #2768, opened 2018-11-05
Fix flutter setup links

ChangJoo-Park:
Flutter setup page url is changed to "get-started".

Please review this below. Is it right?
https://github.com/flutter/flutter-intellij/compare/master...ChangJoo-Park:master#diff-5a61e9f9f0b3221e1b12f6e24cf94ed4R153


pq:
Great catch.  Thank you @ChangJoo-Park !

Issue #2770, opened 2018-11-05
Add ChangJoo-Park to AUTHORS


pq:
Thanks again for #2768 @ChangJoo-Park!  We'd like to add you to [AUTHORS](https://github.com/flutter/flutter-intellij/blob/master/AUTHORS).  What's a good name / email we can use for attribution?

ChangJoo-Park:
Thanks @pq !
I'll PR to README.md with my email

pq:
Fixed w/ https://github.com/flutter/flutter-intellij/commit/8bdb005e922cc78ba8bb2d30c2caa831600d3e13.

PR #2773, opened 2018-11-06
Add author. ChangJoo Park

ChangJoo-Park:
reference here. https://github.com/flutter/flutter-intellij/issues/2770


ChangJoo-Park:
awesome 🎉 

Thanks pq.

PR #2777, opened 2018-11-06
overflow action button makeover

### LINKED ISSUE #2771: Ineffective icon in the inspector for additional tools
Feedback from a study participant today: the second gear icon (circled in red below) in the inspector was not a useful signifier for additional tools. He thought it’s usually for finding settings. Can we use a different icon such as the vertical three dots icon in Chrome, ">>", or a label such as "More tools"? 
 
image

pq:
Fixes #2771.

image

/cc @jacob314


jacob314:
Next step (in a separate cl) it would be good to move the observatory actions to the "More actions" list and promote possibly useful features like "Repaint Rainbows" to the main list so we don't lead users on the path to frustration.
We should similarly create a More Actions" buttom in the Run window and move the observatory actions to the "More actions" menu.

PR #2787, opened 2018-11-07
Disallow importing app templates into Android apps

stevemessick:
The module project template is not going away anytime soon. The app template does not work for importing into an Android app so don't allow it to be selected.

@pq


stevemessick:
Thanks for the quick review!

PR #2790, opened 2018-11-08
flutter make-host-app-editable

### LINKED ISSUE #2661: Don't make people type make-host-app-editable
Once https://github.com/flutter/flutter/pull/22635 lands we need to add support to AS and probably IJ as well to give people an IDE alternative to the CLI app conversion command. No one will want to type that command. ;)

/cc @devoncarew @pq

stevemessick:
Adds new command: flutter make-host-app-editable

Fixes #2661 

@devoncarew

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -559,6 +559,9 @@
       <action id="flutter.doctor" class="io.flutter.actions.FlutterDoctorAction"
               text="Flutter Doctor"
               description="Run 'flutter doctor'"/>
+      <action id="flutter.makeHostAppEditable" class="io.flutter.actions.FlutterMakeHostAppEditableAction"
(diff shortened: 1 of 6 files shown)


resources/META-INF/plugin.xml
```diff
@@ -570,6 +570,9 @@
       <action id="flutter.clean" class="io.flutter.actions.FlutterCleanAction"
               text="Flutter Clean"
               description="Run 'flutter clean'"/>
+      <action id="flutter.makeHostAppEditable" class="io.flutter.actions.FlutterMakeHostAppEditableAction"
+              text="Flutter Make Host App Editable"
```
devoncarew: I'm not sure whether we want this as `Flutter Make Host App Editable` - which makes other commands - or as `Make Flutter Host App Editable`, which is a bit more readable.

Issue #2763, opened 2018-11-05
IllegalStateException parsing json from FlutterSdk.queryFlutterConfig


cclink:
## What happened

Each time I opened my flutter project, I got this error.

## Version information

Android Studio `3.2.1` • Flutter plugin `io.flutter 30.0.1` • Dart plugin `181.5656`

Error getting Flutter sdk information.

## Exception

update failed for AnAction(io.flutter.actions.DeviceSelectorAction) with ID=Flutter.DeviceSelector: Not a JSON Object: null

```
java.lang.IllegalStateException: Not a JSON Object: null
	at com.google.gson.JsonElement.getAsJsonObject(JsonElement.java:91)
	at io.flutter.sdk.FlutterSdk.queryFlutterConfigImpl(FlutterSdk.java:483)
	at io.flutter.sdk.FlutterSdk.queryFlutterConfig(FlutterSdk.java:438)
	at io.flutter.android.IntelliJAndroidSdk.chooseAndroidHome(IntelliJAndroidSdk.java:118)
	at io.flutter.android.AndroidSdk.createFromProject(AndroidSdk.java:35)
	at io.flutter.actions.OpenEmulatorAction.getEmulatorActions(OpenEmulatorAction.java:22)
	at io.flutter.actions.DeviceSelectorAction.updateActions(DeviceSelectorAction.java:130)
	at io.flutter.actions.DeviceSelectorAction.lambda$update$2(DeviceSelectorAction.java:72)
	at com.intellij.openapi.application.impl.ApplicationImpl.invokeAndWait(ApplicationImpl.java:643)
	at io.flutter.FlutterUtils.invokeAndWait(FlutterUtils.java:61)
	at io.flutter.actions.DeviceSelectorAction.update(DeviceSelectorAction.java:71)
	at io.flutter.actions.DeviceSelectorAction.update(DeviceSelectorAction.java:66)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performDumbAwareUpdate(ActionUtil.java:136)
	at com.intellij.openapi.actionSystem.impl.Utils.doUpdate(Utils.java:259)
	at com.intellij.openapi.actionSystem.impl.Utils.expandActionGroup(Utils.java:202)
	at com.intellij.openapi.actionSystem.impl.Utils.expandActionGroup(Utils.java:153)
	at com.intellij.openapi.actionSystem.impl.Utils.expandActionGroup(Utils.java:228)
[...]

PR #2791, opened 2018-11-08
guard against bad config json (#2763)

pq:
Fixes: #2763.

I can't track down why it would happen but in at least this reported case the JSON coming back from the call to `flutter config --machine` is invalid.  In the absence of a fix on the flutter tools side, this guards against it.

/cc @devoncarew


pq:
@devoncarew : thoughts on this one?

devoncarew:
👍 

sorry for the delay here!

PR #2792, opened 2018-11-09
badge log entries w/ associated data

pq:
Adds a badge to log entries that have associated data.

image

image

**Note:** this is not currently user facing since only logging events produced by `package:logs` will associate JSON data with entries.

(Also: this is certainly not pressing to land; just playing around over ☕️.)

/cc @devoncarew


pq:
> Possibly unrelated, but one of the issues I hit with the logging view was being unable to see text for long lines. 

> I assume that this is now addressed with the new detail area (that it would help show text for long lines, or multi-line text)?

Not yet but this is a good idea.  #2808 opened to track.

PR #2796, opened 2018-11-09
shuffle complex inspection tools into menu (and simple out) (#2772)

### LINKED ISSUE #2772: Show simple tools before advanced tools in the inspector toolbar
A study participant commented that exploring difficult to use tools such as the Observatory from the inspector's toolbar discouraged him from exploring other tools, under the assumption that other, less accessible tools would just be as challenging if not more so. 

We should order the list of tools by ease of use and frequency of use. Advanced tools should probably go into the `more` menu in the end. 

image

pq:
As per @InMatrix' suggestion:

> I think we can move Timeline and Observatory to the menu, and move "Enable Slow Animations" and "Show Paint Baselines" to the toolbar.

image

Fixes #2772.

/cc @jacob314

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -60,8 +60,8 @@ import org.jetbrains.annotations.Nullable;
 
 import javax.swing.*;
 import java.awt.*;
-import java.util.*;
 import java.util.List;
+import java.util.*;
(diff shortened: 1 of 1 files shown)


pq:
I left the timeline and observatory icons in but to be honest, feel mixed...  (Looks like we may have a bikeshed to paint!)

PR #2799, opened 2018-11-09
Report dart errors

### LINKED ISSUE #2657: IntelliJ should handle StateExceptions thrown by Dart code
There are a number of plugin issues created because the Flutter tool returned exceptionally from a request by the IDE. These usually result in a CompletionException surfacing to our users. Many are from StateExceptions ("Bad state: No element") thrown by SDK library code; those indicate errors in the Flutter code but we don't have enough information to act on them. To do anything we'd need the full stack trace from Dart; I don't know if the Flutter tool includes a stack trace in its response currently or not.

The plugin should handle these conditions more gracefully. Rather than asking the user to file a bug it should explain there is an error and suggest updating the Flutter tooling. Alternatively, if the stack trace is available then ask them to file an issue on the Flutter project, including the stack trace.

/cc @devoncarew I'm thinking of closing a bunch of issues with a comment referring to this one.

stevemessick:
Recognize Dart stack traces. Send error reports to the Flutter repo when they are recognized as such. If we can't be sure what we have, but we think it is a Dart stack trace, send it to this repo. Unfortunately, DaemonApi was not properly arranged so the methods got moved around when it was reformatted. The main change in it was fetching the Dart stack trace at line 188.

Fixes #2657

Sample report (with simulated error):
[bug-report.txt](https://github.com/flutter/flutter-intellij/files/2567831/bug-report.txt)

@devoncarew

### DIFF
--- a/src/io/flutter/FlutterConstants.java
+++ b/src/io/flutter/FlutterConstants.java
@@ -98,6 +98,9 @@ public class FlutterConstants {
     "watcher",
     "yaml");
 
+  // Aligned w/ VSCode (https://github.com/flutter/flutter-intellij/issues/2682)
+  public static String RELOAD_REASON_SAVE = "save";
+  public static String RELOAD_REASON_MANUAL = "manual";
 
   public static final String FLUTTER_SETTINGS_PAGE_ID = "flutter.settings";
   public static final String INDEPENDENT_PATH_SEPARATOR = "/";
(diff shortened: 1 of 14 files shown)


devoncarew:
> Send error reports to the Flutter repo when they are recognized as such. If we can't be sure what we have, but we think it is a Dart stack trace, send it to this repo. 

Steve, do we think that these should be reported to the flutter/flutter repo? I wonder how many of the issues are due to us issuing requests when the isolate has restarted, or the app has exited. As in, instead of reporting these issues to the user or to flutter/flutter, which issues should we be tolerant of and silently ignore.

devoncarew:
> How about I redirect everything to our issue list so we can study the problem a bit more?

👍, sounds good to me.

stevemessick:
I didn't change 14 files! The only ones that belong here are FlutterErrorReportSubmitter and DaemonApi. Sorry for the confusion.

PR #2809, opened 2018-11-12
Move perf tab into Flutter Performance tool window

kenzieschmoll:


kenzieschmoll:
screen shot 2018-11-12 at 11 23 01 am
Note: memory tab will not show up in Intellij until memory profiling is supported.

src/io/flutter/view/FlutterView.java
```diff
@@ -252,8 +251,6 @@ private void addInspectorViewContent(FlutterApp app, @Nullable InspectorService
         addDisabledTab(WIDGET_TAB_LABEL, runnerTabs, app, toolbarGroup);
         addDisabledTab(RENDER_TAB_LABEL, runnerTabs, app, toolbarGroup);
       }
-
-      addPerformanceTab(runnerTabs, app, !hasInspectorService);
```
jacob314: Maybe we should temporarily add a placeholder tab that just has a link to open the new perf tab?

kenzieschmoll: Sure, I can add that. Probably would be good to have a lead for existing users.

kenzieschmoll:
@jacob314 Should I remove the toolbar items from the Flutter Inspector toolbar (with the exception of the first two inspector related items)?

devoncarew:
> Should I remove the toolbar items from the Flutter Inspector toolbar

Which items? Depending on what they are, some duplication (available in multiple places) could be useful; that would make them more likely to be discovered by users.

devoncarew:
Re: the layout of the `Performance` tab, we may want to adjust that a bit so that things line up a bit more horizontally. Perhaps just having the `Running in debug more` `Note: ...` on one line across the top of the view, and then the FPS section and the widget rebuild checkboxes will line up horizontally.

jacob314:
I think all the exisiting items should still be shown on the inspector toolbar.
In fact, we should add an additional item that is a link to open the performance tool window.

kenzieschmoll:
screen shot 2018-11-13 at 11 59 07 am
screen shot 2018-11-13 at 8 48 00 am
Will address toolbar in a follow up PR.

devoncarew:
👍 lgtm for my part

src/io/flutter/view/InspectorPerfTab.java
```diff
@@ -66,19 +66,22 @@
...
+    final JPanel headerPanel = new JPanel();
+    final BoxLayout boxLayout = new BoxLayout(headerPanel, BoxLayout.X_AXIS);
+    headerPanel.setLayout(boxLayout);
+    headerPanel.add(new JBLabel("Running in " + app.getLaunchMode() + " mode. "));
     if (app.getLaunchMode() == FlutterLaunchMode.DEBUG) {
       headerPanel.add(
-        new JBLabel("<html><body><p style='color:red'>Note: for best results, re-run in profile mode</p></body></html>"),
-        BorderLayout.SOUTH
-      );
+        new JLabel("<html><body><p style='color:red'>Note: for best results, re-run in profile mode</p></body></html>"));
```
jacob314: why did this go from JBLabel to JBLabel?
Generally we should use the JBContainerName classes instead of JContainerName classes to get a look and feel more consistent with intellij not that it matters in this case.

src/io/flutter/view/FlutterView.java
```diff
@@ -101,6 +102,9 @@
 
   private final Map<FlutterApp, PerAppState> perAppViewState = new HashMap<>();
 
+
+  private FlutterApp app;
```
jacob314: This needs to be removed. The FlutterView needs to support multiple apps

src/io/flutter/view/FlutterView.java
```diff
@@ -341,14 +336,38 @@ private void addDisabledTab(String displayName,
     runnerTabs.addTab(tabInfo);
   }
 
+  private void addPerformancePlaceholderTab(JBRunnerTabs runnerTabs,
+                                            boolean selectedTab) {
```
jacob314: You can get the app from the  `PerAppState state` object.

src/io/flutter/view/FlutterView.java
```diff
@@ -282,34 +290,20 @@ private ActionCallback onTabSelectionChange(TabInfo info, boolean requestFocus,
     return doChangeSelection.run();
   }
 
-  public void switchToRenderTree(FlutterApp app) {
```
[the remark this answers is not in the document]
kenzieschmoll: The method isn't being used anywhere as far as I can tell. Thought I'd go ahead and remove some dead code while I was in here.

src/io/flutter/view/FlutterView.java
```diff
@@ -341,14 +334,52 @@ private void addDisabledTab(String displayName,
...
+      @Override
+      public void linkSelected(LinkLabel aSource, String aLinkData) {
+        showFlutterPerformanceWindow();
+      }
+    }, null);
+    labelRow.add(linkLabel);
+
+    labelRow.add(new JLabel(" for performance information"));
+
+    final JPanel panel = new JPanel(new GridBagLayout());
```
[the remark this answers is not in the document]
kenzieschmoll: I just need to find a way to center a row of three labels within the panel. Vertical centering was working fine, but I was struggling to get them to center horizontally as well. I can try to go back to the drawing board if need be, but this was a solution I found on a forum. Is GridBagLayout expensive?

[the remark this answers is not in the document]
jacob314: no its fine

src/io/flutter/view/FlutterView.java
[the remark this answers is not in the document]
jacob314: go ahead and remove it.

src/io/flutter/view/InspectorPerfTab.java
kenzieschmoll: Just an oversight. Reverted to JBLabel.

src/io/flutter/view/FlutterView.java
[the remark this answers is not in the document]
kenzieschmoll: Implemented a simpler solution.

kenzieschmoll:
Updated UI.
screen shot 2018-11-13 at 2 37 22 pm
screen shot 2018-11-13 at 2 37 45 pm

src/io/flutter/view/FlutterView.java
```diff
@@ -335,32 +334,31 @@ private void addDisabledTab(String displayName,
...
+        showFlutterPerformanceWindow(app);
       }
     }, null);
-    labelRow.add(linkLabel);
-
-    labelRow.add(new JLabel(" for performance information"));
+    linkLabel.setBorder(JBUI.Borders.empty(3, 10));
+    linkLabel.setHorizontalAlignment(SwingConstants.CENTER);
+    if (FlutterUtils.isAndroidStudio()) {
+      // Remove underline for Android Studio to avoid LinkLabel bug where underline is left aligned even though text is center aligned.
```
jacob314: nit: add TODO to remove this check once AndroidStudio is based on IntelliJ 182 instead of 181

kenzieschmoll: Done.

Issue #2682, opened 2018-10-10
Pass reload reasons to Flutter tools


DanTup:
The reload/restart messages in Flutter Daemon can now be passed a `reason` (to provide better stats on user-initiated hot reloads/restarts versus things triggered automatically):

https://github.com/flutter/flutter/pull/21983

In VS Code, we're sending [the strings `save` and `manual`](https://github.com/Dart-Code/Dart-Code/blob/55687223bd2d9085b8fd37043d22d6f9e36799d9/src/constants.ts#L1-L2) depending on whether it was a reload triggered by the user saving, or an explicit action (toolbar button, executing command, explicit keybind - we can't tell the difference between any of those in VS Code):

https://github.com/Dart-Code/Dart-Code/pull/1243

```
[10:31:23 AM]: ==> [{"id":"8","method":"app.restart","params":{"appId":"711d65de-0764-49da-8103-cd8d756d1876","fullRestart":false,"pause":true,"reason":"manual"}}]

[10:31:37 AM]: ==> [{"id":"9","method":"app.restart","params":{"appId":"711d65de-0764-49da-8103-cd8d756d1876","fullRestart":false,"pause":true,"reason":"save"}}]
```

jwren:
Assigning to @devoncarew to triage

pq:
@devoncarew : are you looking at this one?

devoncarew:
hey - no, not currently.

PR #2812, opened 2018-11-13
Add reload reason analytics

pq:
Fixes: #2682.

/cc @DanTup @devoncarew

### DIFF
--- a/src/io/flutter/FlutterConstants.java
+++ b/src/io/flutter/FlutterConstants.java
@@ -98,6 +98,9 @@ public class FlutterConstants {
     "watcher",
     "yaml");
 
+  // Aligned w/ VSCode (https://github.com/flutter/flutter-intellij/issues/2682)
+  public static String RELOAD_REASON_SAVE = "save";
+  public static String RELOAD_REASON_MANUAL = "manual";
 
   public static final String FLUTTER_SETTINGS_PAGE_ID = "flutter.settings";
   public static final String INDEPENDENT_PATH_SEPARATOR = "/";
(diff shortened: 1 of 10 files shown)


DanTup:
@pq Sorry, I didn't see this till this morning. It all looks good to me too - though I didn't understand what the code in `RenderHelper` was for (and why it would be recorded as `save`).

pq:
No worries.  Thanks @DanTup !  The `RenderHelper` bit is part of the preview pane.  I opted not to distinguish it from an auto-save since I think the really interesting # is manual vs. non-manual.  If folks are interested in more precise analytics, I'm happy to make the distinction.

/cc @scheglov

pq:
Cool.  Alternatively, we could add a third category (like `tool` or something) that is used for things like the render helper if we want to keep that analytic distinct.  That has the benefit of leaving the existing data streams alone.

/cc @devoncarew

devoncarew:
Definitely agree that we shouldn't record is as a manual save. Since it's just plumbing for an automatic feature, perhaps we could decide not to record that event at all?

DanTup:
Do you mean not to record the reason? Not recording the event is tricky since it's recorded in flutter tools (it's the reload event). I think if we want to separate it from normal save-triggered-reloads, setting it to `tool` as @pq suggests may make most sense?

devoncarew:
Ah, gotcha. Then a separate category makes sense (`tool`), and flutter_tools could then choose to record it as that, or ignore it completely.

DanTup:
The tools just pass through the string directly to analytics in a custom dimension so I think it makes sense to just let it go through. If someone wants to filter them out, they can it at reporting time, so the raw data isn't lost.

PR #2815, opened 2018-11-13
Stop complaining about daemon process

stevemessick:
I think this is a recent change in IntelliJ that it complains about processes that do nothing, like a daemon, when launched normally. You just need to add this option to eliminate that complaint.

I haven't checked if this causes build problems yet.

@devoncarew @pq


stevemessick:
OK, it causes build problems that I don't feel like dealing with now. I'm closing it.

Issue #2762, opened 2018-11-03
Let users know the inspector needs a running project


InMatrix:
When the user opens the inspector without a running Flutter project, there is no explanation about why there is nothing displayed. We can add some information in this state such as: "Run your project to use the inspector." We could consider adding more information there, and effectively use that space as an on-boarding UI.

![image](https://user-images.githubusercontent.com/348942/47956260-f0857c80-df5e-11e8-9ccb-a18b06595bf9.png)

This was seen in user studies early on but reported again by a user recently on Twitter.

devoncarew:
We'd had this in earlier versions, but ran into a platform bug (since addressed?) where the content for a newly launched app wasn't auto-selected - even with an app running, the user would see the blank, 'no running apps' message.

PR #2816, opened 2018-11-13
Restore ‘no app running’ inspector message (#2762)

pq:
Fixes: #2762.

Confirmed this works in 2018.3 and AS 3.2.1.

Happy to add a check for earlier versions if you all think it's prudent?

/cc @jacob314 @devoncarew

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -82,6 +82,8 @@ public class FlutterView implements PersistentStateComponent<FlutterViewState>, 
     boolean sendRestartNotificationOnNextFrame = false;
   }
 
+  private Content emptyContent;
+
   public static final String TOOL_WINDOW_ID = "Flutter Inspector";
 
   public static final String WIDGET_TAB_LABEL = "Widgets";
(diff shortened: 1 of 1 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -473,6 +481,15 @@ private void updateForEmptyContent(ToolWindow toolWindow) {
     }
 
     toolWindow.setIcon(FlutterIcons.Flutter_13);
+
+    // Display a 'No running applications' message.
+    final ContentManager contentManager = toolWindow.getContentManager();
+    final JPanel panel = new JPanel(new BorderLayout());
+    final JLabel label = new JLabel("No running applications", SwingConstants.CENTER);
```
jacob314: nit: JLabel -> JBLabel

jacob314:
lgtm!

devoncarew:
> Happy to add a check for earlier versions if you all think it's prudent?

IntelliJ 2018.1 / AS 3.2 may still not have the fix; we should just keep an eye on this after it lands (the issue was manifested as the Inspector view not switching over to the live content tab when at app ran).

src/io/flutter/view/FlutterView.java
pq: 👍

PR #2818, opened 2018-11-14
Remove size specification from WidgetPerfPanel. 

kenzieschmoll:
Prevents perf tips from being clipped:
screen shot 2018-11-14 at 9 10 58 am
Now the layout can extend as much as it needs to (see exaggerated example):
screen shot 2018-11-14 at 9 58 43 am


kenzieschmoll:
@jacob314 Sorry, I should have made that more clear in the description. This isn't a bug - I did that manually so that the perf tips window would be especially long.

jacob314:
non-issues are my favorite kind of bugs.

PR #2823, opened 2018-11-14
Add toolbar to Flutter Performance window.

kenzieschmoll:
- Add toolbar to perf window containing relevant actions.
- Move common toolbar actions to their own classes.
- Call boolean service extensions from VMServiceManager instead of ViewAction classes on app start/restart.


kenzieschmoll:
screen shot 2018-11-14 at 9 05 21 am

src/io/flutter/server/vmService/VMServiceManager.java
```diff
@@ -245,6 +248,16 @@ private void addServiceExtension(String name) {
       else if (!stream.getValue()) {
         stream.setValue(true);
       }
+
+      if (serviceExtensionState.get(name) != null && serviceExtensionState.get(name).getValue()) {
+        // We should not call the service extension for the follwing extension.
+        if (StringUtil.equals(name, "ext.flutter.inspector.show")) {
```
devoncarew: Instead of an early exit from the method, perhaps invert the if condition?

src/io/flutter/server/vmService/VMServiceManager.java
```diff
@@ -245,6 +248,16 @@ private void addServiceExtension(String name) {
       else if (!stream.getValue()) {
         stream.setValue(true);
       }
+
+      if (serviceExtensionState.get(name) != null && serviceExtensionState.get(name).getValue()) {
```
jacob314: add a 
getServiceExtensionCurrentValue method that returns the value.
In that method you should synchronize on 
serviceExtensionsState as that is used to keep all additions to that map safe.

src/io/flutter/server/vmService/VMServiceManager.java
```diff
@@ -245,6 +248,16 @@ private void addServiceExtension(String name) {
...
+        // We should not call the service extension for the follwing extension.
```
jacob314: why doesn't show work?

I would have thought you could go through the same code path just with this code added as a bonus if the actual button was clicked. For a hot restart, bringing the emulator to the front is actually a bug more than a feature as I can't think of how it actually makes sense. 

      // If toggling inspect mode on, bring all devices to the foreground.
      // TODO(jacobr): consider only bringing the device for the currently open inspector TAB.
      if (isSelected()) {
        final FlutterDevice device = app.device();
        if (device != null) {
          device.bringToFront();
        }
      }

src/io/flutter/view/DebugDrawAction.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package io.flutter.view;
+
+import com.intellij.icons.AllIcons;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import io.flutter.FlutterBundle;
+import io.flutter.run.daemon.FlutterApp;
+import org.jetbrains.annotations.NotNull;
+
+class DebugDrawAction extends FlutterViewToggleableAction {
```
jacob314: Maybe call this DebugPaintAction to match the extension name.

src/io/flutter/view/DebugDrawAction.java
```diff
@@ -0,0 +1,28 @@
...
+  DebugDrawAction(@NotNull FlutterApp app) {
+    super(app, FlutterBundle.message("flutter.view.debugPaint.text"), FlutterBundle.message("flutter.view.debugPaint.description"),
+          AllIcons.General.TbShown);
+
+    setExtensionCommand("ext.flutter.debugPaint");
+  }
+
+  @Override
+  protected void perform(AnActionEvent event) {
```
jacob314: in a follow up cl it would be nice to get rid of all these redundant perform methods that are doing the same thing.

src/io/flutter/server/vmService/VMServiceManager.java
kenzieschmoll: Used the pre-existing getServiceExtensionState() method and called .getValue() on it. This should be fine as getServiceExtensionState() is synchronized on serviceExtensionState.

src/io/flutter/server/vmService/VMServiceManager.java
kenzieschmoll: I didn't actually write this code, just copied it over from where it used to be in FlutterView.

src/io/flutter/view/PerformanceOverlayAction.java
```diff
@@ -0,0 +1,30 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+class PerformanceOverlayAction extends FlutterViewToggleableAction {
+
+  public static final String SHOW_PERFORMANCE_OVERLAY = "ext.flutter.showPerformanceOverlay";
+
+  PerformanceOverlayAction(@NotNull FlutterApp app) {
+    super(app, "Toggle Performance Overlay", "Toggle Performance Overlay", AllIcons.Modules.Library);
+    setExtensionCommand(SHOW_PERFORMANCE_OVERLAY);
```
jacob314: make sure you have run the code format here and elsewhere

src/io/flutter/server/vmService/VMServiceManager.java
kenzieschmoll: Same for all the actions in the new files.

src/io/flutter/view/TimeDilationAction.java
```diff
@@ -0,0 +1,39 @@
...
+    setExtensionCommand("ext.flutter.timeDilation");
+  }
+
+  @Override
+  protected void perform(@Nullable AnActionEvent event) {
+    final Map<String, Object> params = new HashMap<>();
+    params.put("timeDilation", isSelected() ? 5.0 : 1.0);
+    if (app.isSessionActive()) {
+      app.callServiceExtension("ext.flutter.timeDilation", params);
+    }
```
jacob314: how are we avoiding calling the ext.flutter.timeDilation  with an invalid boolean value?
I would think that a bool for the state of ext.flutter.timeDilation is stored in VMServiceManager

src/io/flutter/view/TogglePlatformAction.java
```diff
@@ -0,0 +1,75 @@
...
+import com.intellij.execution.ui.ConsoleViewContentType;
+import com.intellij.icons.AllIcons;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import io.flutter.FlutterBundle;
+import io.flutter.run.daemon.FlutterApp;
+import io.flutter.utils.StreamSubscription;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.concurrent.CompletableFuture;
+
```
jacob314: add a todo here to normalize this with other code so the actions are in sync.

src/io/flutter/view/TimeDilationAction.java
```diff
@@ -0,0 +1,39 @@
...
+import com.intellij.icons.AllIcons;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import io.flutter.run.daemon.FlutterApp;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+
+class TimeDilationAction extends FlutterViewToggleableAction {
```
jacob314: I would think you need to extend FlutterViewAction instead of FlutterViewToggleableAction for this to be safe.

kenzieschmoll: I believe this extends FlutterViewToggleableAction because it needs access to isSelected(), which is not part of FlutterViewAction.

src/io/flutter/view/TimeDilationAction.java
kenzieschmoll: Addressed this case in VMServiceManager

src/io/flutter/server/vmService/VMServiceManager.java
```diff
@@ -249,12 +251,24 @@ else if (!stream.getValue()) {
...
+      if (getServiceExtensionState(name).getValue()) {
+        restoreServiceExtensionState(name);
+      }
+    }
+  }
+
+  private void restoreServiceExtensionState(String name) {
+    if (app.isSessionActive()) {
+      // We should not call the service extension for the follwing extensions.
+      if (!StringUtil.equals(name, "ext.flutter.inspector.show")
```
jacob314: negated logic and nested ifs are harder to read than repeated if else cases.
Rewrite as

if (StringUtil.equals(name, "ext.flutter.inspector.show") || 
 StringUtil.equals(name, "ext.flutter.platformOverride") {
  // Do not call service extension for these extensions because...
} else if (StringUtil.equals(name, "ext.flutter.timeDilation")) {
   final Map<String, Object> params = new HashMap<>();
          params.put("timeDilation", 5.0);
          app.callServiceExtension("ext.flutter.timeDilation", params);
} else {
  app.callBooleanExtension(name, true);
}

kenzieschmoll: Fixed.

src/io/flutter/view/DebugDrawAction.java
kenzieschmoll: Doing this in a follow up CL.

Issue #2784, opened 2018-11-07
update new project wizard images


devoncarew:
In Android Studio, we have ~3 images for the new project wizards:

<img width="727" alt="screen shot 2018-11-07 at 11 04 17 am" src="https://user-images.githubusercontent.com/1269969/48154224-e9bb7a00-e27c-11e8-9cf1-278522ba45ae.png">

we should update the 'plugin' and 'package' one w/ new images; this may just be as simple as using icon badges that are slightly larger. Several possible material design svgs are available here: https://material.io/tools/icons/?style=baseline.

stevemessick:
@devoncarew The module template is still in use.

devoncarew:
Gotcha - thanks; not easy to keep up with the latest here :) We should have an updated image for that as well then.

PR #2826, opened 2018-11-15
update the android studio wizard images

devoncarew:
- update the android studio wizard images
- fix https://github.com/flutter/flutter-intellij/issues/2784

@stevemessick


stevemessick:
I'll patch this in and take a look soon. It might be worth getting @pq
impression too.

devoncarew:
You should be able to see the images in-line in the 'files changes' tab: https://github.com/flutter/flutter-intellij/blob/19108ae30468d0047a750f5a0a9ee41494657c84/resources/icons/template_new_package.png

stevemessick:
Here is the package icon panel from the old wizard:
screen shot 2018-11-15 at 9 57 06 am

And here is the new version with this PR patched in:
screen shot 2018-11-15 at 9 54 37 am

My thoughts:
 - Module and App should be different.
 - The new Plugin is fine.
 - The old Package was better.

I think someone with a graphic design background should take a look.

stevemessick:
In general, I like the size changes. They do fit better with the Android Studio icons.

The original Plugin badge was from Material icons: the devices icon. It seemed reasonable at the time but I think the extensions icon is better; it seems more similar to other plugin icons elsewhere. The work badge on the Package icon doesn't do anything for me, though. It doesn't say this is a pure-Dart component, nor does it say it is device-independent. (My first thought was that it was a suitcase.) IIRC, when I was looking through the Material icons I couldn't find anything that really fit for Package, so I used the Dart icon. Is there a monochrome version of the Dart icon? (The Android Studio module selector does not use monochrome badges, BTW. I think they use charcoal grey instead of black in several icons.)

The reason I think App and Module should be different is we have no timeline for the change. Maybe App should have a badge suggesting it is short-lived? Because the change, as I understand, is to rename Module to Application then remove App. So, when it is all over, the Application icon would be the same as the App icon today.

This is why I lost interest in icon design. It ends up with a bunch of valid but conflicting opinions. Whoever is doing the editing gets final say. Therefore: LGTM.

Issue #2758, opened 2018-11-02
Track usage of rebuild indicators and perf tips


InMatrix:
It would be useful to track the following events in the analytics from the next release:
* open the perf tab in the inspector
* enable rebuild indicators
* disable rebuild indicators
* click on a perf tip

jacob314:
@pq you know the analytics system better than I do. Want to add these events?

pq:
Some implementation thoughts.

As for #2827, we've got perf tips covered.  What's left is  rebuild indicators.

@jacob314 and I have been chatting about ways to do this in a centralized place (e.g.,  `VMServiceManager`) and I've got a POC.  Having said that, I'm not convinced it's the right approach.  In instrumenting our instrumentation (heh) I notice that we're already tracking these service extension calls at the action level.  For example:

```
{an=Flutter IntelliJ Plugin, aiid=IntelliJ IDEA, tid=UA-67589403-7, ds=app, cd1=2018.3, t=event, av=183.4284.118, v=1, ul=en, ea=ext.flutter.showPerformanceOverlay, ec=vmservice, cid=b67a0e50-1506-4942-8940-9688e1c2265f, sr=2560x1600}
{an=Flutter IntelliJ Plugin, aiid=IntelliJ IDEA, tid=UA-67589403-7, ds=app, cd1=2018.3, t=event, av=183.4284.118, v=1, ul=en, ea=PerformanceOverlayAction, ec=intellij, cid=b67a0e50-1506-4942-8940-9688e1c2265f, sr=2560x1600}
{an=Flutter IntelliJ Plugin, aiid=IntelliJ IDEA, tid=UA-67589403-7, ds=app, cd1=2018.3, t=event, av=183.4284.118, v=1, ul=en, ea=ext.flutter.repaintRainbow, ec=vmservice, cid=b67a0e50-1506-4942-8940-9688e1c2265f, sr=2560x1600}
{an=Flutter IntelliJ Plugin, aiid=IntelliJ IDEA, tid=UA-67589403-7, ds=app, cd1=2018.3, t=event, av=183.4284.118, v=1, ul=en, ea=RepaintRainbowAction, ec=intellij, cid=b67a0e50-1506-4942-8940-9688e1c2265f, sr=2560x1600}
[...]

PR #2827, opened 2018-11-15
analytics for perf tip selections

pq:
Part of: #2758

Note that:

> open the perf tab in the inspector

is done already in `io.flutter.view.FlutterPerfView#onTabSelectionChange`.

/cc @devoncarew @jacob314


src/io/flutter/view/InspectorPerfTab.java
```diff
@@ -131,13 +138,17 @@ private void buildUI() {
   private void setTrackRebuildWidgets(boolean selected) {
     final FlutterWidgetPerfManager widgetPerfManager = FlutterWidgetPerfManager.getInstance(app.getProject());
     widgetPerfManager.setTrackRebuildWidgets(selected);
+    // Send analytics.
+    sendAnalyticEvent(selected, "TrackRebuildWidgets");
```
devoncarew: Actually, I think we should move the system of record for the rebuild/repaint prefs to the FlutterSettings class, and then track analytics from there. That will mean that we can better know how well used the setting is (esp. given that its currently opt-in).

(https://github.com/flutter/flutter-intellij/pull/2832 is related here)

pq: Great idea.  Now that you've done more consolidation in #2832, especially.

I'll update this once you land it.

Thanks!

jacob314: I think TrackRebuildWidgets is more like other boolean ext.flutter.* debugging extensions like 
RepaintRainbows, RepaintBaselines, and ShowPerfOverlays
so the better place to track it is in VMServiceManager which is being used as the single source of truth for those flags.
Another distinction is setting in FlutterSettings should persist across sessions but TrackRebuildWidgets shoudn't any more than DebugPaintRainbows should.

pq: @jacob314: are we tracking analytics for those service extension calls?  I didn't think we were...

devoncarew: > I think TrackRebuildWidgets is more like other boolean ext.flutter.* debugging extensions like
RepaintRainbows, RepaintBaselines, and ShowPerfOverlays

👍 makes sense

jacob314: We should start tracking the analytics for them VMServiceManager as that is a good centralized place to track them. With the CL @kenzieschmoll is working on, the VMServiceManager becomes the centralized place all these extensions set and automatically reapplied on a hot restart

pq: Sounds good.  I can follow-up with another round of analytics once that lands.

In the meantime, do we agree that this is we want to track?

That is, an `enableFoo` and `disableFoo` for each event type `Foo`?  Or is `enable` sufficient?

jacob314:
Disable is probably only useful to track if we can log with the disable event how long it was since the previous enable event or can infer that information from sessions. It is meaningful if users try a feature and turn it back off after 5 seconds out of frustration. otherwise, i'm not sure what actionable stats I could gather from disable events as users typically will disable these flags at some point in their session after they enable them unless the app crashed.

pq:
I've updated this PR to just include perf tip tracking and will follow-up w/ a more inclusive change that tackles all the VM Service analytics when @kenzieschmoll's other PR lands.

@jacob314  PTAL?

src/io/flutter/inspector/WidgetPerfPanel.java
```diff
@@ -102,6 +102,12 @@ public void selectionChanged(@NotNull FileEditorManagerEvent event) {
     Disposer.register(parentDisposable, perfTipComputeDelayTimer::stop);
   }
 
+  private static void handleTipSelection(@NotNull PerfTip tip) {
+    // Send analytics.
+    FlutterInitializer.getAnalytics().sendEvent("perf", "perfTipSelected");
```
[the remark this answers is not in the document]
pq: URL would be cumbersome since they're so long.  Could we consider adding a unique name to `PerfTipRule` that we could append to the event analytics?

So, something like:

    FlutterInitializer.getAnalytics().sendEvent("perf", "perfTipSelected." + tip.uniqueId);

?

pq:
Added analytics ids to distinguish perf tip events.

/cc @jacob314

PR #2840, opened 2018-11-19
Fixes Bugs.

terrylucas:


terrylucas:
Fixed a number of bugs:

- Initial selections in filter not primed.
- Cleanup close to match IntelliJ UI (CommandButton)
- Added RSS data (need to display properly either graph or text value).
- Correctly return instances allocated & total bytes (active and total in new/old heaps)
- Support AllocationProfile reset to track leaks of allocations
- Cleanup instance inspection.

flutter-studio/src/io/flutter/profiler/FlutterAllMemoryData.java
```diff
@@ -64,16 +73,24 @@ public ThreadSafeData() { }
...
+  // multiData[HEAP_CAPACITY] is heap capacity.
+  // multiData[EXTERNAL_MEMORY_USED] is total external in use.
+  // multiData[RSS_SIZE] is Resident Set Size of the application (process) in memory.
+  List<ThreadSafeData> multiData;
+
+  // Hookup VM listeners to collect all the VM memory information:
+  //    1. heap used
+  //    2. external space used
+  //    3. heap capacity
+  //    4. RSS.
```
jacob314: indicate that RSS is ResidentSetSize

flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
```diff
@@ -634,7 +639,10 @@ public void selectionCleared() {
       new RangedContinuousSeries("MemoryMax", getTimeline().getViewRange(), dataRanges, memoryMaxDataSeries);
     RangedContinuousSeries externalMemoryRange =
       new RangedContinuousSeries("MemoryExtern", getTimeline().getViewRange(), dataRanges, memoryExternalDataSeries);
+//    RangedContinuousSeries rssRange = new RangedContinuousSeries("RSS",
```
jacob314: add a quick comment on why this is commented out

jacob314: also I assume that "RSS" will show up in the UI. We'll need to use a longer name as users will think RSS means something completely different.

PR #2843, opened 2018-11-19
copy an icon from resources/flutter/material/ and update references to it

devoncarew:
Follow up from https://github.com/flutter/flutter-intellij/pull/2842, copy an icon from `resources/flutter/material/` and update references to it.

@stevemessick


devoncarew:
It's just that the icons in `resources/flutter/material/` are generated from the flutter material icons - they could change based on changes to the material font (a low probability situation, but possible), so I want to isolate us from any possible changes.

PR #2844, opened 2018-11-20
Add yellow icons.

jacob314:
Add yellow spinning icons matching the style of the red and grey spinning icons already provided.
A followup cl actually uses these icons.


kenzieschmoll:
LGTM!

PR #2848, opened 2018-11-20
Service Extension refactor.

kenzieschmoll:
- Use VMServiceManager to manage service extension states
- Create ServiceExtensionState class
- Remove unnecessary StreamSubscriptions


src/io/flutter/view/BoolServiceExtensionCheckbox.java
```diff
@@ -19,45 +16,36 @@
...
+      if (checkbox.isSelected() != state.isEnabled()) {
+        checkbox.setSelected(state.isEnabled());
       }
     }, true);
   }
 
   JCheckBox getComponent() {
     return checkbox;
   }
-
```
jacob314: add back this dispose method. you caught a big that we aren't calling this dispose method not that it isn't needed. otherwise we are leaking memory

src/io/flutter/view/TimeDilationAction.java
```diff
@@ -19,6 +19,7 @@
     super(app, "Enable Slow Animations", null, showIcon ? AllIcons.Vcs.History : null);
 
     setExtensionCommand("ext.flutter.timeDilation");
+    setEnabledStateValue(5.0);
```
jacob314: if you specified that the disabledStateValue was 1.0 then you could eliminate the override of the perform method.

kenzieschmoll: I overrode the perform method because timeDilation doesn't call a boolean service extension like FlutterViewToggleableAction's perform method, so we need to specify the params and call the service extension here.

PR #2852, opened 2018-11-20
Refactor of how metrics are calculated to make adding additional metics easier.

jacob314:
Refactor of FlutterWidgetPerf so that widget names are always available instead
of only available when displaying gutter icons. The problem was we were calculating widget names for given offsets
when generating the EditorPerfDecorations instead of keeping a global cache
of widget names. The new approach is more efficient and needed for
work to display a table of the most frequent widget rebuilds.

### DIFF
--- a/src/io/flutter/perf/EditorPerfDecorations.java
+++ b/src/io/flutter/perf/EditorPerfDecorations.java
@@ -5,6 +5,7 @@
  */
 package io.flutter.perf;
 
+import com.intellij.codeInsight.hint.HintManager;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.application.ApplicationManager;
@@ -23,9 +24,11 @@ import com.intellij.openapi.wm.ex.ToolWindowManagerEx;
 import com.intellij.ui.ColorUtil;
 import com.intellij.ui.JBColor;
 import icons.FlutterIcons;
+import com.intellij.xdebugger.XSourcePosition;
 import io.flutter.run.daemon.FlutterApp;
 import io.flutter.utils.AnimatedIcon;
 import io.flutter.view.FlutterPerfView;
+import io.flutter.utils.AsyncUtils;
 import io.flutter.view.InspectorPerfTab;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -161,7 +164,7 @@ class EditorPerfDecorations implements EditorMouseListener, EditorPerfModel {
 
   @Override
   public boolean isAnimationActive() {
-    return getStats().getCountPastSecond() > 0;
+    return getStats().getTotalValue(PerfMetric.peakRecent) > 0;
   }
 
   @Override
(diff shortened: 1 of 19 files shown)


jacob314:
A followup CL is coming with the actual UI using these model changes. I'm experimenting with whether splitting up this CL leads to easier to review code.

src/io/flutter/perf/EditorPerfDecorations.java
```diff
@@ -435,13 +448,14 @@ else if (stats.getKind() == PerfReportKind.repaint) {
...
-      sb.append("Since last hot reload/restart: " + stats.getTotal());
+      sb.append("For last frame: " + stats.getValue(PerfMetric.lastFrame) + "<br>");
+      sb.append("In past second: " + stats.getValue(PerfMetric.pastSecond) + "<br>");
+      sb.append("Since last route change: " + stats.getValue(PerfMetric.totalSinceRouteChange) + "<br>");
+      sb.append("Since last hot reload/restart: " + stats.getValue(PerfMetric.total));
       sb.append("</p>");
     }
     if (sb.length() == 0) {
-      sb.append("<p><b>No widget rebuilds or repaints detected for line.</p></b>");
+      sb.append("<p><b>No widget rebuilds detected for line.</p></b>");
```
terrylucas: End with </b></p> since it starts with <p><b>

terrylucas: `End with </b></p> since it starts with <p><b>`

src/io/flutter/perf/SummaryStats.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  public int getValue(PerfMetric metric) {
+    if (metric.timeIntervalMetric && !active) {
+      return 0;
+    }
+    return entry.getValue(metric);
+  }
+
+  public void markAppIdle() {
```
terrylucas: @override for the interface methods (and the rest)?

src/io/flutter/perf/EditorPerfDecorations.java
jacob314: done

src/io/flutter/perf/SummaryStats.java
jacob314: not sure what you mean. This class doesn't have an interface

src/io/flutter/perf/FilePerfInfo.java
```diff
@@ -66,12 +21,14 @@ public Location getLocation() {
...
   private final Multimap<TextRange, SummaryStats> stats = LinkedListMultimap.create();
   long maxTimestamp = -1;
-  private int totalPastSecond = 0;
+  private final int[] totalForMetric = new int[PerfMetric.values().length];
 
   public void clear() {
     stats.clear();
     maxTimestamp = -1;
-    totalPastSecond = 0;
+    for (int i = 0; i < totalForMetric.length; ++i) {
```
kenzieschmoll: nit: Could simplify with `Arrays.fill(totalForMetric, 0);`

jacob314: done. Fixed with a followup cleanup cl.

PR #2861, opened 2018-11-21
Make Flutter settings searchable

### LINKED ISSUE #2309: Search in preferences doesn't work for Flutter plugin settings
## Steps to Reproduce

1. Go to Android Studio preferences
2. Search for "widget"

This doesn't find the `Run applications with --track-widget-creation` checkbox.

intellij_preferences_bug mp4

This is the case for all options for the Flutter (and I think even the Dart) plugin — not just the option shown above. 

It's interesting that when I'm on that page and I start typing "widget", it'll highlight the option for a while, but then it will auto-update the list of relevant pages on the left panel, and it won't have that page there.

## Version info

[✓] Flutter (Channel beta, v0.4.4, on Mac OS X 10.13.4 17E202, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK 27.0.3)
[!] iOS toolchain - develop for iOS devices (Xcode 9.4)
    ! CocoaPods out of date (1.5.0 is recommended).
        CocoaPods is used to retrieve the iOS platform side's plugin code that responds to your plugin usage on the Dart side.
        Without resolving iOS dependencies with CocoaPods, plugins will not work on iOS.
        For more info, see https://flutter.io/platform-plugins
      To upgrade:
        brew upgrade cocoapods
        pod setup
[✓] Android Studio (version 3.1)
[!] IntelliJ IDEA Community Edition (version 2018.1)
    ✗ Flutter plugin not installed; this adds Flutter specific functionality.
    ✗ Dart plugin not installed; this adds Dart specific functionality.
[✓] VS Code (version 1.22.2)
[✓] Connected devices (1 available)

stevemessick:
Makes Flutter settings appear when a search term is entered in Settings/Preferences. As a side-effect it also moves strings from a form into a bundle.

Thanks to @alexander-doroshko for finding SearchableOptionContributor.

Fixes #2309

### DIFF
--- a/src/io/flutter/sdk/FlutterSearchableOptionContributor.java
+++ b/src/io/flutter/sdk/FlutterSearchableOptionContributor.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.sdk;
+
+import com.intellij.ide.ui.search.SearchableOptionContributor;
+import com.intellij.ide.ui.search.SearchableOptionProcessor;
+import io.flutter.FlutterBundle;
+import io.flutter.FlutterConstants;
+import org.jetbrains.annotations.NotNull;
+
+public class FlutterSearchableOptionContributor extends SearchableOptionContributor {
+  @Override
+  public void processOptions(@NotNull SearchableOptionProcessor processor) {
+    add(processor, FlutterBundle.message("settings.try.out.features.still.under.development"));
+    add(processor, FlutterBundle.message("settings.show.preview.area"));
+    add(processor, FlutterBundle.message("settings.experimental.flutter.logging.view"));
+    add(processor, FlutterBundle.message("settings.enable.android.gradle.sync"));
+    // For some reason the word "report" is ignored by the search feature, but the other words work.
+    add(processor, FlutterBundle.message("settings.report.google.analytics"));
+    add(processor, FlutterBundle.message("settings.enable.verbose.logging"));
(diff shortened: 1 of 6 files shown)


src/io/flutter/sdk/FlutterSearchableOptionContributor.java
```diff
@@ -0,0 +1,36 @@
...
+    add(processor, FlutterBundle.message("settings.format.code.on.save"));
+    add(processor, FlutterBundle.message("settings.organize.imports.on.save"));
+    add(processor, FlutterBundle.message("settings.flutter.version"));
+    add(processor, FlutterBundle.message("settings.open.inspector.on.launch"));
+    add(processor, FlutterBundle.message("settings.hot.reload.on.save"));
+    add(processor, FlutterBundle.message("settings.disable.tracking.widget.creation"));
+  }
+
+  private static void add(@NotNull SearchableOptionProcessor processor, @NotNull String key) {
+    processor.addOptions(key, null, null, FlutterConstants.FLUTTER_SETTINGS_PAGE_ID,
```
alexander-doroshko: @stevemessick @pq Note that if the 3rd argument (`hit`) is not null then the options will also become searchable via Find Action. That's the text that you'll see in Find Action results.

And one more Pro tip (not sure you need it, just FYI): `processor.addOptions()` allows setting aliases. For example, for checkbox called 'foo' you may set alias 'bar' by calling `processor.addOptions("foo bar", null, "You are lucky to find foo via Find Action", "page.id", null, false/true - no idea)` - you'll find 'foo' both by typing 'foo' or 'bar'.

stevemessick:
Good to know, thanks! I'll see about updating this after M31 is out.

PR #2863, opened 2018-11-22
various UI and layout tweaks to the perf tab

devoncarew:
Several and various UI and layout tweaks to the perf tab. A few of these changes could probably use discussion; happy to have feedback about them.

- have the `profile mode` text extend across the full with of the tool window (instead of just across the left hand component)
- have the splitter default to a 50/50 split (instead of 60/40)
- re-order and re-group the toolbar actions
- the FPS display now consistently displays one decimal point (`60.0`)
- change to always having the rebuild stats area displayed; I think this will facilitate discovery of the feature, and won't really negatively impact the area available to the other UI (FPS display)
- unify the use of titled borders around the various areas
- move the enablement of the rebuild stats table into the rebuild stats panel
- remove the title of the rebuild stats table (`Rebuild widgets stats for the current screen`); this took up some vertical space, and I'm banking on users being able to figure out the table from context
- fix an issue with the text color of selected rows
- choose to not display icons for rows that are not changing; the grey circle icon didn't provide a lot of info, and the addition and remove of the icon draws the eye to changing rows

@jacob314 @terrylucas @kenzieschmoll 

screen shot 2018-11-21 at 2 05 12 pm

screen shot 2018-11-21 at 2 20 12 pm

### DIFF
--- a/src/io/flutter/view/FlutterPerfView.java
+++ b/src/io/flutter/view/FlutterPerfView.java
@@ -193,11 +193,12 @@ public class FlutterPerfView implements Disposable {
                                            @NotNull FlutterApp app,
                                            Disposable parentDisposable) {
     final DefaultActionGroup toolbarGroup = new DefaultActionGroup();
-    toolbarGroup.add(registerAction(new DebugPaintAction(app)));
-    toolbarGroup.add(registerAction(new TogglePlatformAction(app)));
     toolbarGroup.add(registerAction(new PerformanceOverlayAction(app)));
+    toolbarGroup.add(registerAction(new TogglePlatformAction(app)));
     toolbarGroup.addSeparator();
+    toolbarGroup.add(registerAction(new DebugPaintAction(app)));
     toolbarGroup.add(registerAction(new ShowPaintBaselinesAction(app, true)));
+    toolbarGroup.addSeparator();
     toolbarGroup.add(registerAction(new TimeDilationAction(app, true)));
 
     return toolbarGroup;
@@ -257,7 +258,7 @@ public class FlutterPerfView implements Disposable {
   }
 
   public InspectorPerfTab showPerfTab(@NotNull FlutterApp app) {
-    PerfViewAppState appState = perAppViewState.get(app);
+    final PerfViewAppState appState = perAppViewState.get(app);
     if (appState != null) {
       final ToolWindow toolWindow = ToolWindowManager.getInstance(myProject).getToolWindow(TOOL_WINDOW_ID);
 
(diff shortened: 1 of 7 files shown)


src/io/flutter/view/WidgetPerfTable.java
```diff
@@ -149,6 +148,7 @@ public void onFrame() {
 
   private void updateIconUIAnimations() {
     int lastActive = -1;
+    // TODO(devoncarew): We've seen NPEs from here.
```
devoncarew: @jacob314 I wasn't able to track this NPE down. If I remember correctly, it happened early in the connection lifecycle.

jacob314: I have a fix. I'll send out the fix soon.

jacob314: Overall this looks really good. @InMatrix for thoughts on breaking out the Rebuild stats in their own column.
Can you add a screenshot of what the ui looks like if the user pins the perf tab to the right side of the screen?

src/io/flutter/view/FlutterPerfView.java
```diff
@@ -193,11 +193,12 @@ private DefaultActionGroup createToolbar(@NotNull ToolWindow toolWindow,
...
-    toolbarGroup.add(registerAction(new TogglePlatformAction(app)));
     toolbarGroup.add(registerAction(new PerformanceOverlayAction(app)));
+    toolbarGroup.add(registerAction(new TogglePlatformAction(app)));
     toolbarGroup.addSeparator();
+    toolbarGroup.add(registerAction(new DebugPaintAction(app)));
     toolbarGroup.add(registerAction(new ShowPaintBaselinesAction(app, true)));
+    toolbarGroup.addSeparator();
     toolbarGroup.add(registerAction(new TimeDilationAction(app, true)));
 
     return toolbarGroup;
```
kenzieschmoll: Perhaps the inspector toolbar actions should be reordered to mirror this change.

devoncarew:
> I think this warning only applies to the Frame Rendering Time graph, right? Making it extend across the window might make users believe the rebuild stats are not accurate, either.

This PR is more about lining things up aesthetically. We could certainly move performance disclosures closer into where they're relevant.

> Other than that, how about changing "Collect widget rebuild information" to "Show widget rebuild information" to better speak to what the user will get out of this feature?

The PR currently has the text as `Show Widget Rebuild Profiler`; happy to update this if you think we should.

> Before the rebuild information is enabled, we probably can show a description of the feature in place of the rebuild stats table, assuming that space is empty.

👍; will add an issue for that

src/io/flutter/view/FlutterPerfView.java
devoncarew: will do

devoncarew:
(updated to `Show widget rebuild information`)

Issue #2769, opened 2018-11-05
M31 release plan meta issue


devoncarew:
This is the plan for the M31 milestone.

(resurrecting a practice from earlier releases, I'd like to create a meta-issue to track work in M31)

## Plan items

- todo:

---

## Releasing

The release date is ~11/28.

- [ ] create release candidate, send out for testing; 11/26 @messick

---

devoncarew:
issue for testing feedback: https://github.com/flutter/flutter-intellij/issues/2866

devoncarew:
When starting up a flutter project on a fresh IDE install, the `Flutter Performance` tool window is open by default. It doesn't have any content (and isn't useful until an app is running). Other tool windows aren't open by default, so there's likely an xml flag we need to toggle.

devoncarew:
When running apps several times in a row, I notice that the 'enable widget recounts' checkbox doesn't always become enabled, even after the app starts. I suspect a race condition in detection of the availability of the service extension. I don't know if we also checking for the service extension added event or not.

cc @kenzieschmoll

PR #2865, opened 2018-11-26
Update change log for M31

stevemessick:
I condensed #2769 to focus on user-visible changes. Happy to add more. We did quite a lot of user-visible work this cycle and I want to make sure everything is mentioned. (But I'm happy to use 'bug fixes' as a catch-all for crashes.)

@devoncarew @pq @jacob314 @terrylucas @kenzieschmoll 
Did I include all the contributors for this release?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## 31.0
+- change FPS display to "Frame Rendering Time" and improve UI
+- reorganize inspector tools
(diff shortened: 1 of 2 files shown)


stevemessick:
@devoncarew do you think #2863 will be ready for M31?

devoncarew:
@stevemessick, yes, I would like it in, but it's not critical - it's polish, and not a bug fix.

PR #2871, opened 2018-11-26
Add service extensions after first frame event received.

kenzieschmoll:
Prevents a crash in the framework. 

Crash was occurring if the Performance Overlay or Hide Debug Mode Banner settings were enabled before performing a hot restart.

### DIFF
--- a/src/io/flutter/server/vmService/VMServiceManager.java
+++ b/src/io/flutter/server/vmService/VMServiceManager.java
@@ -246,22 +246,26 @@ public class VMServiceManager implements FlutterApp.FlutterAppListener {
   }
 
   private void maybeAddServiceExtension(String name) {
-    if (firstFrameEventReceived) {
-      addServiceExtension(name);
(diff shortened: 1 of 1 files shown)


src/io/flutter/server/vmService/VMServiceManager.java
```diff
@@ -238,6 +245,25 @@ private void onVmServiceReceived(String streamId, Event event) {
     }
   }
 
+  private void maybeAddServiceExtension(String name) {
+    if (firstFrameEventReceived) {
+      addServiceExtension(name);
+      if (pendingServiceExtensions.size() > 0) {
```
devoncarew: perhaps `!list.isEmpty()`

kenzieschmoll: Done.

kenzieschmoll:
@stevemessick Will this be able to make it into M31?

stevemessick:
Yep! I may push another build today, but certainly tomorrow.

PR #2872, opened 2018-11-26
Add instructions to test modules and inspector

stevemessick:
We probably need to add even more.

Note that the instructions for testing modules don't work right now because module support needs work.

@devoncarew @jacob314 @pq

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -119,7 +119,7 @@ In the newly created app:
 * open the inspector
 * change the display by clicking `Render Tree` and the refresh button
 * open the performance view
-* verify that the memory view is active and the performance view is not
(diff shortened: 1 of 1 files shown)


docs/testing.md
```diff
@@ -99,6 +116,10 @@ In the newly created app:
 * tap the `'+'` icon on the app
 * verify that the IDE pauses at the breakpoint, and that the `Variables` pane has
   the right value for `_counter`
+* open the inspector
+* change the display by clicking `Render Tree` and the refresh button
+* open the performance view
+* verify that the memory view is active and the performance view is not
```
devoncarew: prefix with: `(in Android Studio):`

Currently the memory view only displays in Android Studio

PR #2874, opened 2018-11-26
add url checking to the build

pq:
add url checking to the build

* moves action URLs into bundle
* adds grind task that checks URLs in bundle
* adds grinder call to travis build

Fixes: #2819 

a few things to consider:

* I’m using a heuristic to grab URL bundle properties; we could be more eager
* where in the build this happens is open to debate — perhaps we ultimately should have a separate bot that does this and `./bin/plugin lint` ? dunno.

/cc @devoncarew @stevemessick

### DIFF
--- a/tool/travis.sh
+++ b/tool/travis.sh
@@ -38,6 +38,10 @@ else
   # Run some validations on the repo code.
   ./bin/plugin lint
 
+  # Check plugin-referenced urls for liveness.
+  pub global activate grinder
+  grind check-urls
+
   # Run the build.
   ./bin/plugin build --only-version=$IDEA_VERSION
 fi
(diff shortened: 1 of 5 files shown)


tool/travis.sh
```diff
@@ -38,6 +38,9 @@ else
   # Run some validations on the repo code.
   ./bin/plugin lint
 
+  # Check plugin-referenced urls for liveness.
+  dart tool/grind.dart check-urls
```
devoncarew: These urls may be flaky outside of our control; perhaps as future work, move the check into a new travis job (so that when it fails, its easier to de-flake from the rest of the build).

pq: My thinking too.  I'll follow up with another PR that adds a bot.

PR #2881, opened 2018-11-27
Fix a couple NPE edge cases.

jacob314:
Polish InspectorPerfTab UI moving warning about debug mode closer to the FPS
display. Added description of what widget rebuild indicators are and moved some text around and made the fps warning a bit scarier.

The NPE fixes are targeting an error seen in a usability study but I haven't been able to reproduce the issue locally with either a device or an emulator so I am not 100% confident in the fix.

Tweaked UI: 
screen shot 2018-11-26 at 8 44 05 pm

### DIFF
--- a/src/io/flutter/view/InspectorPerfTab.java
+++ b/src/io/flutter/view/InspectorPerfTab.java
@@ -10,6 +10,7 @@ import com.intellij.openapi.util.Disposer;
 import com.intellij.openapi.wm.ToolWindow;
 import com.intellij.ui.JBColor;
 import com.intellij.ui.JBSplitter;
+import com.intellij.ui.ScrollPaneFactory;
 import com.intellij.ui.components.JBLabel;
 import com.intellij.ui.components.JBPanel;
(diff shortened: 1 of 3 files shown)


src/io/flutter/view/InspectorPerfTab.java
```diff
@@ -110,33 +93,56 @@ void updateUseSplitter(boolean force) {
...
 
     // Header
+    final JPanel footer = new JPanel(new VerticalLayout(0));
+    footer.add(new JSeparator());
     final JPanel labels = new JPanel(new BorderLayout(6, 0));
     labels.setBorder(JBUI.Borders.empty(0, 8));
-    add(labels, BorderLayout.NORTH);
 
     labels.add(
       new JBLabel("Running in " + app.getLaunchMode() + " mode"),
```
devoncarew: Is there a way to specify that the label contents should align to the top vertically?

jacob314: Good point. Done.
screen shot 2018-11-27 at 8 17 28 am

Issue #2875, opened 2018-11-26
investigate change in behavior for re-running apps for IntelliJ 2018.3


devoncarew:
Previously, we'd allowed the user to click on 'run' for an already running app (or 'debug' for an app launched in debug mode). That would signal us to perform a hot restart of the app.

It looks like IntelliJ's behavior here changed, in 2018.3. They now detect this case, and display a dialog, `''{0}'' is not allowed to run in parallel.\nWould you like to stop {1, choice, 1#the running one|2#{1, number} running instances}?`.

This looks to be implemented in a subclass of `ExecutionManager.restartRunProfile()`. And controlled by the result of calling `RunConfiguration.isAllowRunningInParallel()`. We may be able to get back to our previous behavior by overridding `RunConfiguration.isAllowRunningInParallel()` and returning true?

devoncarew:
I think in order to get this to work again, we'll need to create launch configurations with the 'allow running in parallel' flag set.

<img width="187" alt="screen shot 2018-11-26 at 2 57 22 pm" src="https://user-images.githubusercontent.com/1269969/49047233-a3767e00-f18b-11e8-8d72-347ba9db3175.png">

devoncarew:
Re-testing, this is an issue we'll have to solve soon. It affects IntelliJ 2018.3 users, and will affect Android Studio users once the platform version updates.

Manually going into the launch configuration and checking the 'Allow running in parallel' checkbox does fix the issue. We'll either need to figure out a way to do that programmatically for all launch configs, or enable that setting for each config as they are created.

PR #2895, opened 2018-11-28
force "allow running in parallel"

pq:
A band-aid to address the re-run behavior regression in 2018.3.  

Additionally I wonder if this couldn't be fixed when we create our initial launch configs?  (Although that doesn't help with migration.)

Fixes: #2875.

/cc @devoncarew @stevemessick

### DIFF
--- a/src/io/flutter/run/SdkRunner.java
+++ b/src/io/flutter/run/SdkRunner.java
@@ -28,10 +28,12 @@ public class SdkRunner extends LaunchState.Runner<SdkRunConfig> {
     return "FlutterRunner";
   }
 
+  @SuppressWarnings("Duplicates")
   @Override
   protected RunContentDescriptor doExecute(@NotNull RunProfileState state, @NotNull ExecutionEnvironment env) throws ExecutionException {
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/SdkRunner.java
```diff
@@ -22,6 +28,18 @@ public String getRunnerId() {
     return "FlutterRunner";
   }
 
+  @Override
+  protected RunContentDescriptor doExecute(@NotNull RunProfileState state, @NotNull ExecutionEnvironment env) throws ExecutionException {
+    if (FlutterUtils.is2018_3_or_higher()) {
+      // Force "allow running in parallel" (see: #2875).
+      final RunnerAndConfigurationSettings settings = env.getRunnerAndConfigurationSettings();
+      if (settings != null) {
+        settings.setSingleton(false);
```
devoncarew: Very cool! Does this address the issue for you - we're you then able to hot restart by clicking on the run button?

I'll patch this in locally.

pq:
> I think we should land this (and likely, add a todo: to switch to something like #2897 once our min. platform version is high enough)

TODOs added 👍

PR #2898, opened 2018-11-28
tweak to the perf urls

devoncarew:
- minor change to how we populate perf urls; this will allow the IntelliJ 'unused property file entry' lint help us notice when we are / aren't using url entries

@jacob314

### DIFF
--- a/src/io/flutter/FlutterBundle.properties
+++ b/src/io/flutter/FlutterBundle.properties
@@ -206,7 +206,6 @@ info.level.title=Info
 warning.level.title=Warning
 severe.level.title=Severe
 shout.level.title=Shout
-# End Flutter Log Color Page
 
 settings.try.out.features.still.under.development=Try out features still under development (a restart may be required)
 settings.show.preview.area=Show a live preview area in Flutter Outline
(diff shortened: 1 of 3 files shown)


src/io/flutter/FlutterBundle.properties
```diff
@@ -206,7 +206,6 @@ info.level.title=Info
 warning.level.title=Warning
 severe.level.title=Severe
 shout.level.title=Shout
-# End Flutter Log Color Page
```
jacob314: was deleting this line intentional?

src/io/flutter/perf/PerfTipRule.java
```diff
@@ -45,6 +44,7 @@
     String hackFileName,
     String message,
     String id,
+    String url,
```
jacob314: update the comment for id now that you have changed what the id is used for.

src/io/flutter/FlutterBundle.properties
devoncarew: Yes - there's a comment a few lines above mentioning the start of this section; I don't think we generally need to bookmark sections in this file (we don't other places).

src/io/flutter/perf/PerfTipRule.java
devoncarew: done

Issue #2896, opened 2018-11-28
Make preferences searchable via Find Action


stevemessick:
Follow up to #2861:

Using `Help > Find Action` to search Flutter preferences does not work. It should, with a little more work in the code added for #2861.

Currently, the `Find Action` command finds the `Hot Reload` button when searching for "hot reload" but it does not find the `Enable hot reload on restart` preference. It should find both.

PR #2899, opened 2018-11-28
Make Find Action search prefs

stevemessick:
Adds the Preference search strings to the database searched by `Help > Find Action`.

I didn't include this in M31 because I wasn't comfortable with it, but it looks safe enough. We'll have some time to check it out before M32.


stevemessick:
Fixes #2896

PR #2904, opened 2018-11-29
Release 31.1

stevemessick:
Updated change log for 31.1.

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -27,12 +27,6 @@
 
   <change-notes>
     <![CDATA[
-<h2>31.1</h2>
-<ul>
-  <li>perf table polish and fix links to tip docs</li>
-  <li>fix Split Mode resize issue</li>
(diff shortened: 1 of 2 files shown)


resources/META-INF/plugin.xml
```diff
@@ -10,7 +10,7 @@
   <vendor url="https://github.com/flutter/flutter-intellij">flutter.io</vendor>
 
   <category>Custom Languages</category>
-  
+  <version>31.1.1</version>
```
devoncarew: should this be contributed? You may just want a new commit with the same changelog updates as are in the `release_31` branch.

stevemessick: We're not currently including that line. The plugin tool was designed to do so and the process I follow keeps adding it, then I have to delete it like now. I'll go ahead and merge once the build is green.

stevemessick: And, of course, we can't use the --release option in the travis script. That means we can't do it the way I thought we would unless we do more work on the travis script.

PR #2909, opened 2018-12-03
Display legend again with line chart color/line style and last value.

terrylucas:
Show legend with proper line chart color or line style.
Legend displays the RSS value exposed under a flag (consider option to enable RSS stats).
Legend insert to not intersect with right side axis name.
Fix for Travis, AS 3.2 build.

### DIFF
--- a/flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
+++ b/flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
@@ -107,6 +107,12 @@ public class FlutterStudioMonitorStageView extends FlutterStageView<FlutterStudi
 
   private static final int AXIS_SIZE = 100000;
 
+  // TODO(terry): Currently don't display any RSS (resident set size) might be too confusing to users.
+  //              Check with asiva on why this is so important.  There are a number of set sizes e.g.,
+  //              Resident Set Size, Virtual Set Size, Proportial Set Size, and Unique Set Size in the Dalvik heap.
+  //              Also, consider exposing an a option to enable RSS statistics?
+  private static final Boolean displayRSSInformation = false;
+
   JBPanel classesPanel;                         // All classes information in this panel.
   JBScrollPane heapObjectsScoller;              // Contains the JTree of heap objects.
   private final JBLabel classesStatusArea;      // Classes status area.
@@ -419,7 +425,8 @@ public class FlutterStudioMonitorStageView extends FlutterStageView<FlutterStudi
 
           if (ref.getName().length() > 0) {
             rememberLibrary(ref.getName(), ref);
-          } else {
+          }
+          else {
             // No unique name to display use the Uri and don't display in list of known library/packages.
             rememberLibrary(PREFIX_LIBRARY_NAME_HIDDEN + ref.getUri(), ref);
           }
@@ -645,7 +652,10 @@ public class FlutterStudioMonitorStageView extends FlutterStageView<FlutterStudi
     FlutterAllMemoryData.ThreadSafeData memoryUsedDataSeries = stage.getUsedDataSeries();
     FlutterAllMemoryData.ThreadSafeData memoryMaxDataSeries = stage.getCapacityDataSeries();
     FlutterAllMemoryData.ThreadSafeData memoryExternalDataSeries = stage.getExternalDataSeries();
-    FlutterAllMemoryData.ThreadSafeData rssDataSeries = stage.getRSSDataSeries();
+    FlutterAllMemoryData.ThreadSafeData rssDataSeries = null;
+    if (displayRSSInformation) {
+      rssDataSeries = stage.getRSSDataSeries();
+    }
 
     Range dataRanges = new Range(0, 1024 * 1024 * 100);
     Range viewRange = getTimeline().getViewRange();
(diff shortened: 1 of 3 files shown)


tool/plugin/lib/plugin.dart
```diff
@@ -551,7 +551,7 @@ class BuildCommand extends ProductCommand {
             'flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java');
         source = processedFile.readAsStringSync();
         files[processedFile] = source;
-        source = source.replaceAll('new Range(100.0, 100.0)', '100');
+        source = source.replaceAll('usedMemoryRange.getXRange()', '100');
```
stevemessick: LGTM

flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
```diff
@@ -648,28 +648,30 @@ public void selectionCleared() {
...
-      new RangedContinuousSeries("MemoryMax", getTimeline().getViewRange(), dataRanges, memoryMaxDataSeries);
+      new RangedContinuousSeries("Capacity", viewRange, dataRanges, memoryMaxDataSeries);
     RangedContinuousSeries externalMemoryRange =
-      new RangedContinuousSeries("MemoryExtern", getTimeline().getViewRange(), dataRanges, memoryExternalDataSeries);
-    // TODO(terry): Temporary comment out working on displaying RSS properly in chart.
-    // RangedContinuousSeries rssRange = new RangedContinuousSeries("RSS",
-    //                                                              getTimeline().getViewRange(), dataRanges, rssDataSeries);
+      new RangedContinuousSeries("External", viewRange, dataRanges, memoryExternalDataSeries);
+    RangedContinuousSeries rssRange =
+      new RangedContinuousSeries("RSS", viewRange, dataRanges, rssDataSeries);
```
jacob314: Add a TODO to add a tooltip explaining what RSS is and why users should care. Perhaps the matching Android code already has a tooltip. I expect the term RSS used here is very confusing to users who will think of RSS feeds instead of Resident Set Size. @InMatrix any thoughts on what term our users might understand better. The goal for showing RSS is to indicate how close an app is to getting shutdown for using more memory than the device allows.
https://en.wikipedia.org/wiki/Resident_set_size

flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
```diff
@@ -724,6 +715,10 @@ public void selectionCleared() {
     final JBPanel legendPanel = new JBPanel(new BorderLayout());
     legendPanel.setOpaque(false);
     legendPanel.add(legendComponent, BorderLayout.EAST);
+    // Give the right-side a little space so it doesn't intersect with the right-side label.
+    Border border = legendPanel.getBorder();
+    Border margin = new EmptyBorder(0,0,0,30);
```
jacob314: run code format to fix up whitespace.

flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
```diff
@@ -724,6 +715,10 @@ public void selectionCleared() {
     final JBPanel legendPanel = new JBPanel(new BorderLayout());
     legendPanel.setOpaque(false);
     legendPanel.add(legendComponent, BorderLayout.EAST);
+    // Give the right-side a little space so it doesn't intersect with the right-side label.
```
jacob314: I assume you mean "get too close" not literally "intersect". Correct?

flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
terrylucas: I could change the name in the legend to Resident Set Size

flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
terrylucas: Right.

terrylucas: It's at the right-most edge w/o any gap, 30 pixels right border seems ok.

terrylucas: I'll change comment to say that.

flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
terrylucas: Done.

flutter-studio/src/io/flutter/profiler/FlutterStudioMonitorStageView.java
InMatrix: A tooltip explaining Resident Set Size would be useful.

terrylucas: I'm going to hide RSS for now (under debug flag) will re-visit after talking to Siva about why he cares so much about RSS for Flutter devs.

terrylucas:
Accidently closed.

PR #2910, opened 2018-12-03
Add Android libraries to Flutter project

stevemessick:
Enables code completion for Java and Kotlin. This is ready for review now.

- Gradle sync on start-up and whenever project roots are modified
- After sync, get Android libraries and add to Flutter project
- Add library dep to Android module for each library
- Add dep from Android module to Flutter module so the libraries can be found

Still has a couple TODOs, which I'll work on next. Some are noted in the code. Also:
- Make the Gradle sync run in the background
- See if other Android-like facets should be included (AndroidModuleLibraryManager line 153)

@devoncarew I had to refactor some of the library management code you added a while back.

### DIFF
--- a/flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
+++ b/flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
@@ -5,6 +5,9 @@
  */
 package io.flutter.android;
 
+import static io.flutter.android.AndroidModuleLibraryType.LIBRARY_KIND;
+import static io.flutter.android.AndroidModuleLibraryType.LIBRARY_NAME;
+
 import com.android.tools.idea.gradle.project.sync.GradleSyncInvoker;
 import com.android.tools.idea.gradle.project.sync.GradleSyncListener;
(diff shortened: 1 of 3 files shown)


flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
```diff
@@ -58,10 +65,57 @@ public AndroidModuleLibraryManager(@NotNull Project project) {
     super(project);
   }
 
-  public void update() {
+  public void updateOld() {
```
jacob314: can you delete this method or give it a clearer name?

stevemessick: Done, along with substantial additional changes. PTAL

stevemessick:
This is the one I mentioned. I'm not sure about merging it. It actually makes the feature usable, in my limited testing. Needs more testing, I think, but could possibly include it in the point release if testing succeeds.

stevemessick:
Merging, now that the point release is out.

Issue #2913, opened 2018-12-05
Create project throws null when creating module


navneetkarnani:
## What happened

Created a new Flutter project as a "Module" along with my kotlin server project

## Version information

IntelliJ IDEA `2018.3` • Flutter plugin `io.flutter 31.1.4` • Dart plugin `183.4733`

No Flutter sdk configured.

## Exception

null

```
java.lang.Throwable: Assertion failed: steps: [com.intellij.ide.projectWizard.ProjectTypeStep, com.intellij.ide.projectWizard.ChooseTemplateStep, com.intellij.openapi.externalSystem.service.project.wizard.ExternalModuleSettingsStep, com.intellij.openapi.externalSystem.service.project.wizard.ExternalModuleSettingsStep, org.jetbrains.idea.maven.wizards.MavenModuleWizardStep, org.jetbrains.idea.maven.wizards.SelectPropertiesStep, com.intellij.openapi.externalSystem.service.project.wizard.ExternalModuleSettingsStep, com.intellij.openapi.externalSystem.service.project.wizard.ExternalModuleSettingsStep, org.jetbrains.plugins.gradle.service.project.wizard.GradleModuleWizardStep, com.intellij.openapi.externalSystem.service.project.wizard.ExternalModuleSettingsStep, com.intellij.openapi.externalSystem.service.project.wizard.ExternalModuleSettingsStep, Project Settings] current: Project Settings
	at com.intellij.openapi.diagnostic.Logger.assertTrue(Logger.java:174)
	at com.intellij.ide.wizard.AbstractWizard.doNextAction(AbstractWizard.java:398)
	at com.intellij.ide.util.newProjectWizard.AbstractProjectWizard.doNextAction(AbstractProjectWizard.java:254)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at io.flutter.module.FlutterGeneratorPeer.requestNextStep(FlutterGeneratorPeer.java:217)
	at io.flutter.actions.InstallSdkAction$InstallAction.requestNextStep(InstallSdkAction.java:142)
[...]

PR #2915, opened 2018-12-05
require sdk for project creation (#2913)

pq:
image

(Note that **Next** is only enabled; it will not advance or throw errors.)

Fixes: #2913.

/cc @stevemessick


pq:
Yeah.  It's a real short-coming in the wizard framework.  If folks experience this a lot, we can look at doing something more dramatic.

Thanks!

PR #2916, opened 2018-12-05
misc. cleanups

pq:
* cleans up some unused imports
* does some lambda migrations
* adds missing annotations

Not complete; just opportunistic.

/cc @stevemessick @jacob314 @kenzieschmoll


src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -24,60 +24,44 @@
 import com.intellij.testFramework.LightVirtualFile;
 import com.intellij.util.BitUtil;
 import com.intellij.util.TimeoutUtil;
-import com.intellij.xdebugger.XDebugProcess;
-import com.intellij.xdebugger.XDebugSession;
-import com.intellij.xdebugger.XDebugSessionListener;
-import com.intellij.xdebugger.XDebuggerBundle;
-import com.intellij.xdebugger.XSourcePosition;
+import com.intellij.xdebugger.*;
```
stevemessick: This one is just going to keep changing, depending on who formats it. The AS rules use the expanded form.

pq: Ah, sorry.  The actual setting is a little further down:

image

pq: Thinking on this some more, I suggest I just land.  Until we all get unified behind a shared setting this churn will continue.  Once we unify, we could consider a sweeping PR to canonicalize...

pq: https://github.com/flutter/flutter-intellij/issues/2922 to follow-up.

stevemessick:
It's already approved for merging :)

Issue #2873, opened 2018-11-26
Ensure inspector/perf windows are WAI for flutter attach use case


kenzieschmoll:
As of this [PR](https://github.com/flutter/flutter-intellij/pull/2871), service extensions are added/restored after a Flutter.FirstFrame event has been received.

This may be problematic for the flutter attach use case if we do not get the Flutter.FirstFrame event.

jacob314:
Made this excessively robust with
https://github.com/flutter/flutter-intellij/pull/2920

PR #2920, opened 2018-12-06
Fix bug where the profile widget rebuilds service extension does

jacob314:
not show up as available on Flutter attach.

Fix race condition causing the profile rebuild extension not to be
reported when calling attach.

We now query the flutter framework directly to verify that a frame
has been drawn in the case of attach.

Also cleaned up locking logic for existingExtensions and
firstFrameEventReceived to be paranoid.

### DIFF
--- a/src/io/flutter/server/vmService/VMServiceManager.java
+++ b/src/io/flutter/server/vmService/VMServiceManager.java
@@ -6,10 +6,10 @@
 package io.flutter.server.vmService;
 
 import com.intellij.openapi.Disposable;
-import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.util.Disposer;
 import com.intellij.openapi.util.text.StringUtil;
 import gnu.trove.THashMap;
+import io.flutter.inspector.EvalOnDartLibrary;
 import io.flutter.run.daemon.FlutterApp;
 import io.flutter.server.vmService.HeapMonitor.HeapListener;
 import io.flutter.utils.EventStream;
@@ -21,10 +21,7 @@ import org.dartlang.vm.service.consumer.VMConsumer;
 import org.dartlang.vm.service.element.*;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.function.Consumer;
 
 public class VMServiceManager implements FlutterApp.FlutterAppListener {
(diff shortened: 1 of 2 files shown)


jacob314:
Also fixes https://github.com/flutter/flutter-intellij/issues/2873

src/io/flutter/server/vmService/VMServiceManager.java
```diff
@@ -252,21 +270,25 @@ private void onVmServiceReceived(String streamId, Event event) {
   }
 
   private void maybeAddServiceExtension(String name) {
-    synchronized (pendingServiceExtensions) {
+    synchronized (serviceExtensions) {
```
kenzieschmoll: Just curious as to why we need to synchronize on serviceExtensions here vs pendingServiceExtensions?

kenzieschmoll: Just curious as to why we need to synchronize on serviceExtensions here vs pendingServiceExtensions?

jacob314: to avoid future multi-threading bugs. Generally, the fact that we use synchronized a lot in this class isn't great. For this sort of code to be robust it is best to have some simple invariants. Specifically, the one I want is that we never acquire more than one "lock" at a time. Previously you could acquire a couple different locks at once. The simplest way to get a deadlock is if thread 1 locks object A and then tries to lock object B while thread 2 locks object B and then tries to lock object A.  If you only write code that locks one object at a time then that sort of case can never happen.

jacob314: Same as above. The goal is to just simplify the locking so that we are only acquiring 1 lock at a time instead of multiple locks. The existing code appeared alright in practice but I didn't want to have to audit to make sure there was never a competing case where the serviceExtensions lock could be acquired and then the pendingServiceExtensions lock was acquired.

jacob314:
High level, it would be nice to clean up this class to stop using synchronized at all and or at least only synchronize on one object.

src/io/flutter/server/vmService/VMServiceManager.java
kenzieschmoll: Gotcha makes sense. Sorry for the duplicate comment - was having github technical issues.

src/io/flutter/server/vmService/VMServiceManager.java
```diff
@@ -121,14 +120,32 @@ public void onError(RPCError error) {
...
+        // If there is an error we assume the first frame has been received.
+        boolean didSendFirstFrameEvent = e == null ||
+                                         v == null ||
+                                         Objects.equals(v.getValueAsString(), "true");
+        if (didSendFirstFrameEvent) {
+          onFrameEventReceived();
         }
+      });
+    }
+    if (isolate.getExtensionRPCs() != null) {
```
kenzieschmoll: Should this be else if?

jacob314: no.  I should add an extra line break to make it clearer.

PR #2940, opened 2018-12-10
commit project code style config

### LINKED ISSUE #2922: [code style] agree on class count for '*' imports (and enforce it)
Developers of the plugin have inconsistent import settings and so we see regular churn (e.g., https://github.com/flutter/flutter-intellij/pull/2916#discussion_r239308113).

In IDEA, the relevant setting looks like this in the code style preferences:

image

First off, we should 

- [ ] agree on a class count for '*' imports, and then we should
- [ ] enforce it in shared settings

/cc @stevemessick

pq:
Commits project-level code style configuration (previously git ignored).

These should be the IDEA defaults.  I'm not attached to any details besides the fact that we converge on some agreement to avoid future churn.

Fixes #2922.

/cc @devoncarew @stevemessick @jacob314

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,509 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <option name="OTHER_INDENT_OPTIONS">
+      <value>
+        <option name="INDENT_SIZE" value="2" />
+        <option name="CONTINUATION_INDENT_SIZE" value="4" />
+        <option name="TAB_SIZE" value="8" />
+      </value>
+    </option>
+    <option name="LINE_SEPARATOR" value="&#10;" />
+    <option name="RIGHT_MARGIN" value="140" />
+    <option name="FORMATTER_TAGS_ENABLED" value="true" />
+    <option name="LINE_COMMENT_AT_FIRST_COLUMN" value="false" />
+    <option name="BLOCK_COMMENT_AT_FIRST_COLUMN" value="false" />
+    <GroovyCodeStyleSettings>
+      <option name="STATIC_METHODS_ORDER_WEIGHT" value="5" />
+      <option name="METHODS_ORDER_WEIGHT" value="4" />
+    </GroovyCodeStyleSettings>
+    <JavaCodeStyleSettings>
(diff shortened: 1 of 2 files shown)


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="USE_PER_PROJECT_SETTINGS" value="true" />
```
devoncarew: 👍

devoncarew:
@pq, if we re-format the current project code w/ these settings, I assume that there is relatively little churn? That these settings closely match what we're doing in most of the files?

pq:
These are pretty close.  Doing a reformat of everything would improve a bunch of rogue whitespace and align everything.  Most of the changes are uncontroversial I think and should be consistent w/ vanilla IDEA.  If we did a bulk reformat we could consider any outliers in that review I guess and make tweaks with more conversation and context?

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,509 @@
...
+    <option name="FORMATTER_TAGS_ENABLED" value="true" />
+    <option name="LINE_COMMENT_AT_FIRST_COLUMN" value="false" />
+    <option name="BLOCK_COMMENT_AT_FIRST_COLUMN" value="false" />
+    <GroovyCodeStyleSettings>
+      <option name="STATIC_METHODS_ORDER_WEIGHT" value="5" />
+      <option name="METHODS_ORDER_WEIGHT" value="4" />
+    </GroovyCodeStyleSettings>
+    <JavaCodeStyleSettings>
+      <option name="PREFER_LONGER_NAMES" value="false" />
+      <option name="GENERATE_FINAL_LOCALS" value="true" />
```
stevemessick: I'd prefer not to have this. Currently, my code generation settings (from Android Studio) look like this:

screen shot 2018-12-10 at 4 41 49 pm

Highlighting can be used to easily identify variables that get reused. I see them underlined, which is enough to get me to use a fresh var when it makes sense.

screen shot 2018-12-10 at 4 36 34 pm

pq: Works for me.

PR #2956, opened 2018-12-13
trim unoptimized text from stack frame names

devoncarew:
- trim prefix text on dart stack frames (`[Unoptimized] `)
- move to using a non-deprecated icon; I'll need to see whether this compiles on our oldest IDEs before committing this PR
- unrelated, but remove the check for `supports.pausePostRequest` - this has been in VMs for a very long time now

@stevemessick for review


devoncarew:
@DanTup for FYI - I am just stripping off the new `Unoptimized` test...

DanTup:
> FYI - I am just stripping off the new Unoptimized test...

Ok, I'll do the same. Do we know if there are other prefixes that might show up?

devoncarew:
I don't think there are.

The `CodeRef` provides a very good, user friendly name apart from this. I think from the VM's perspective however, it's not really supported to be consumed for tooling. If we run into more issues, it would make sense to instead synthesize a new name from the `FuncRef` info.

Issue #2961, opened 2018-12-14
NullPointerException


fengqiangboy:
---

## What happened

1. flutter clean
2. debug on iOS similator

## Version information

Android Studio `3.2.1` • Flutter plugin `io.flutter 31.2.1` • Dart plugin `181.5656`

Flutter 1.0.0 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 5391447fae (2 weeks ago) • 2018-11-29 19:41:26 -0800
Engine • revision 7375a0f414
Tools • Dart 2.1.0 (build 2.1.0-dev.9.4 f9ebf21297)

## Exception

null

```
java.lang.NullPointerException
	at io.flutter.perf.FlutterWidgetPerfManager.debugActive(FlutterWidgetPerfManager.java:190)
	at io.flutter.perf.FlutterWidgetPerfManager.lambda$new$0(FlutterWidgetPerfManager.java:90)
	at sun.reflect.GeneratedMethodAccessor838.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.util.messages.impl.MessageBusConnectionImpl.deliverMessage(MessageBusConnectionImpl.java:117)
	at com.intellij.util.messages.impl.MessageBusImpl.doPumpMessages(MessageBusImpl.java:426)
	at com.intellij.util.messages.impl.MessageBusImpl.pumpWaitingBuses(MessageBusImpl.java:387)
	at com.intellij.util.messages.impl.MessageBusImpl.pumpMessages(MessageBusImpl.java:376)
	at com.intellij.util.messages.impl.MessageBusImpl.pumpMessages(MessageBusImpl.java:364)
	at com.intellij.util.messages.impl.MessageBusImpl.sendMessage(MessageBusImpl.java:357)
	at com.intellij.util.messages.impl.MessageBusImpl.access$200(MessageBusImpl.java:43)
	at com.intellij.util.messages.impl.MessageBusImpl$2.invoke(MessageBusImpl.java:208)
	at com.sun.proxy.$Proxy159.debugActive(Unknown Source)
	at io.flutter.view.FlutterViewMessages.sendDebugActive(FlutterViewMessages.java:45)
	at io.flutter.run.FlutterDebugProcess.onVmConnected(FlutterDebugProcess.java:63)
[...]

PR #2979, opened 2018-12-18
address an NPE in FlutterWidgetPerfManager.java

devoncarew:
- address an NPE in FlutterWidgetPerfManager.java
- fix https://github.com/flutter/flutter-intellij/issues/2961


src/io/flutter/perf/FlutterWidgetPerfManager.java
```diff
@@ -186,7 +186,10 @@ private boolean isProfilingEnabled() {
   private void debugActive(Project project, FlutterViewMessages.FlutterDebugEvent event) {
     debugIsActive = true;
 
-    assert (app != null);
```
jacob314: how is debug active without an app? It would be nice to get to the root cause.

devoncarew: I'm not sure, but from the stack trace, it seemed that the line after the assert (`app.addStateListener(this);`) was triggering the NPE.

Issue #3033, opened 2019-01-02
disable logging view auto-scroll on widget errors


pq:
The default scroll-to-end behavior is undesirable when an error occurs since it brings focus to the end of the stack-trace (which is the least useful).  Better to detect error entries and disable scroll-to-end.

Issue #2552, opened 2018-08-21
Respect ANSI color codes in log entries.


pq:
If log entries include ANSI color escapes, we should respect them.

Corresponding issue in the flutter repo: https://github.com/flutter/flutter/issues/20663.

(An open question is how this can/should interact with category colors.)

/cc @quangson91

jaumard:
Having Colorize package working with this plugin might be interesting for people/teams that are using both IDE and command line to run the app

Issue #2776, opened 2018-11-06
Flicker in logging view when window is narrow


jacob314:
If the log window is too small for the UI the # of items in the log flickers. It would be best to just hide the # of items or show it consistently on top of the other UI when the the log window is too narrow.

![flicker](https://user-images.githubusercontent.com/1226812/48080477-9ecd3400-e1a2-11e8-83c2-234095be8b89.gif)

devoncarew:
Closing as this is related to the experimental logging view.

PR #3034, opened 2019-01-02
Log entry ansi color support (and styling re-work)

pq:
Adds basic ANSI color escape handling and generally reworks when/how we calculate styles for log entry rendering.

High points:

* supports 8 bit colors and styles that correspond to the error reporting presentation improvements being exploring w/ @InMatrix (fixes: #2552)
* disables auto-scroll on widget errors (fixes: #3033)
* parses style information on entry receipt once (and not on every rendering)

Example output (w/ a version of Flutter patched to include ansi colors):

image

While the render fixes are a definite performance improvement, there's still some more work there (see #2776 for example) and the styling will evolve to track requirements driven by framework changes but this is complete enough I think to not hide behind a flag.  Happy to reconsider if anyone feels differently though.  👍 

/cc @jacob314 @devoncarew

### DIFF
--- a/src/io/flutter/logging/FlutterLogEntry.java
+++ b/src/io/flutter/logging/FlutterLogEntry.java
@@ -6,16 +6,19 @@
 package io.flutter.logging;
 
 import com.intellij.openapi.util.text.StringUtil;
+import io.flutter.logging.util.LineInfo;
+import io.flutter.logging.util.StyledText;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import static io.flutter.logging.FlutterLogEntryParser.TOOLS_CATEGORY;
+import java.util.List;
 
 public class FlutterLogEntry {
 
   public enum Kind {
     RELOAD,
     RESTART,
+    WIDGET_ERROR_START,
     UNSPECIFIED
   }
 
@@ -34,30 +37,24 @@ public class FlutterLogEntry {
 
   @NotNull
   private final Kind kind;
+  @NotNull private final List<StyledText> styledText;
 
-  public FlutterLogEntry(long timestamp, @NotNull String category, int level, @Nullable String message) {
+  public FlutterLogEntry(long timestamp,
+                         @NotNull String category,
+                         int level,
+                         @Nullable String message,
+                         @NotNull Kind kind,
+                         @NotNull List<StyledText> styledText) {
     this.timestamp = timestamp;
     this.category = category;
     this.level = level;
     this.message = StringUtil.notNullize(message);
-    this.kind = parseKind(category, this.message);
+    this.kind = kind;
+    this.styledText = styledText;
   }
 
-  private static Kind parseKind(@NotNull String category, @NotNull String message) {
-    if (category.equals(TOOLS_CATEGORY)) {
-      message = message.trim();
-      if (message.equals("Performing hot reload...") || message.equals("Initializing hot reload...")) {
-        return Kind.RELOAD;
-      }
-      if (message.equals("Performing hot restart...") || message.equals("Initializing hot restart...")) {
-        return Kind.RESTART;
-      }
-    }
-    return Kind.UNSPECIFIED;
-  }
-
-  public FlutterLogEntry(long timestamp, @NotNull String category, @Nullable String message) {
-    this(timestamp, category, UNDEFINED_LEVEL.value, message);
+  public FlutterLogEntry(long timestamp, @NotNull LineInfo info, int level) {
+    this(timestamp, info.getCategory(), level, info.getLine(), info.getKind(), info.getStyledText());
   }
 
   @Nullable
@@ -95,6 +92,11 @@ public class FlutterLogEntry {
   @NotNull
   public String getMessage() {
     return message;
+  }
+
+  @NotNull
+  public List<StyledText> getStyledText() {
+    return styledText;
   }
 
   /**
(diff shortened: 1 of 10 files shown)


src/io/flutter/logging/FlutterLogEntry.java
```diff
@@ -6,16 +6,19 @@
...
 
-import static io.flutter.logging.FlutterLogEntryParser.TOOLS_CATEGORY;
+import java.util.List;
 
 public class FlutterLogEntry {
 
   public enum Kind {
     RELOAD,
     RESTART,
+    WIDGET_ERROR_START,
```
jacob314: nit: a little bit of a surprising log entry kind but I don't see a much better option.

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -130,14 +152,26 @@ private static FlutterLogEntry parseGCEvent(@NotNull Event event) {
...
+  private static Kind parseKind(@NotNull String message, @NotNull String category) {
+    if (category.equals(TOOLS_CATEGORY)) {
+      message = message.trim();
+      if (message.equals("Performing hot reload...") || message.equals("Initializing hot reload...")) {
+        return Kind.RELOAD;
+      }
+      if (message.equals("Performing hot restart...") || message.equals("Initializing hot restart...")) {
+        return Kind.RESTART;
+      }
+      if (message.startsWith("══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY")) {
```
jacob314: in terms of technical debt for the quarter can you file a bug to remove this string matching and reference the bug with a TODO here.

testSrc/unit/io/flutter/logging/util/LineParserTest.java
```diff
@@ -0,0 +1,157 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.awt.*;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+
+@SuppressWarnings("UseJBColor")
```
jacob314: This suppress warnings about using JBColor brings up an interesting question. Are there reasonable ways to use JBColor here so that the colored output looks good in both the dark and light themes?

jaumard:
What is the `level` column ? because it's always info and doesn't really take the one of `logger` so it's basically useless (same as category with is alway flutter.tools). Or I miss something :) 
My log still look like this currently (so not with this PR): 
capture d ecran 2019-01-03 a 11 18 18
What it will be insteresting to have is `fine` as level and `CartRepository` as category for example

pq:
> What is the level column ? 

@jaumard: that corresponds to levels as defined by messages reported using the Dart `package:logging`.  If messages are reported by loggers, it can be useful for filtering --- you would see `fine` for example if someone specified it in the log call.

> What it will be insteresting to have is fine as level and CartRepository as category for example
That's why for my the current simple log as text is still better so I prefer not having this one other than with the the flag

Could you elaborate on this?  It would be great to know what we could do to line this view up with your imagined use cases. What does the simple text view allow you to do that this view doesn't?  Thanks in advance!

Feedback here or on one of the general issues (e.g., #2200) would be great. 💎

src/io/flutter/logging/FlutterLogEntry.java
pq: Yeah.  Not at all ideal.  But maybe so glaringly odd that it keeps the pressure on to fix this more properly at the framework level...

src/io/flutter/logging/FlutterLogEntryParser.java
pq: Totally, will do.

testSrc/unit/io/flutter/logging/util/LineParserTest.java
pq: I wondered that too and have it on my list to discuss w/ @InMatrix.  As it stands, the way colors are defined using ansi codes is a little low-level and doesn't map naturally to theming.  Definitely something to consider.

pq: FWIW: I just re-checked and I had already fixed things to not require this suppression so it's now removed.  The general issue still stands though.  Thanks for bringing it up!

jaumard:
@pq I've comment on the issue to not pollute your PR ^^ thanks

pq:
Thanks @jaumard !

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -189,13 +259,33 @@ public FlutterLogEntry parseDaemonEvent(@NotNull String eventText) {
         // Skip.
       }
       else {
-        return new FlutterLogEntry(timestamp(), TOOLS_CATEGORY, line);
+        // Trim "flutter: " prefix (made redundant by category).
+        if (line.startsWith("flutter:")) {
```
devoncarew: 👍 I see these in the output when running flutter apps - definitely adds noise.

firephreek:
This seems to still be an issue in the latest version of Android Studio.

Windows 10, Android Studio 3.5.3, Flutter v42.1.1

 Manually printing ANSI esc sequences doesn't do anything to change the color of the message in either Logcat or the Debug/Run consoles.

ex: 
```
    print('\x1b[101;93m STYLE $emoji - loggggg\x1b[0m');
    print('\u001B[101;93m STYLES \u001B[0m');
```

Gives me this in my debug console:

Capture

Issue #3039, opened 2019-01-03
Restore the log level combo selector


jaumard:
Really useful to me even if we can type the log level directly into the search field :) like this you can filter the log filter and apply a search more easily that way too :) 

<img width="231" alt="capture d ecran 2019-01-03 a 17 10 34" src="https://user-images.githubusercontent.com/2970449/50648082-931ba500-0f7a-11e9-99a8-bbb1e9717389.png">

pq:
Ah!  Did you see that we used to support this?  From conversations with people we got the impression that levels weren't as useful as categories and so we gave back the UI real estate.  It's great to get some countering feedback!

pq:
See also: https://github.com/flutter/flutter-intellij/issues/2624.

jaumard:
So I'm not crazy it was present before ! lol ^^ 
I prefer having it because in most general cases I don't care about fine / verbose and I might don't know the category at that point... so I'll let the filter to `info` level and once I figure it out I'll use the filter on the category. But I found this selector very handy ^^ 

Now we're talking about category a dynamic selector populate from existing log might also be very awesome to have !

freezy:
The thing about filtering instead of selecting a level is that I typically want to see my warnings and errors when selecting `info` as well.

I terms of real estate the filter options currently take about a third of the bar, so I don't think this is going to be an issue.

pq:
Thanks @freezy.  That's a good point.  Selecting **info** is really like filtering on **info|warning|error** (or similar) so not at all equivalent.  An alternative might be a a query language (#2688) where you might be able to type something like **"info+"** to filter/select but that has other issues (not the least of which discoverability).

jaumard:
Query language is also not really set friendly when you want to switch from `info+` to `severe+` and then `verbose+` in few minutes ^^ my point is that the selector allow you quick selection which is more useful than just type in search field each time if it append a lot.

PR #3050, opened 2019-01-07
restore log level combo

pq:
Restores work from @quangson91 based on user feedback.

image

(Level checking has been updated to prefer ordinal value checking over regexps for performance reasons but otherwise this is substantially the same as what was reverted in https://github.com/flutter/flutter-intellij/pull/2686.)

/cc @jacob314  @quangson91 

Fixes: #3039


pq:
Thanks Steve!

Issue #3052, opened 2019-01-08
Log entries truncated 


jaumard:
With the new logging view, long entries are truncate, which is annoying :( 
For example I print my push notification token on log and can't copy/paste it as it's truncated, same for long urls and json data... 

<img width="1378" alt="capture d ecran 2019-01-08 a 13 31 27" src="https://user-images.githubusercontent.com/2970449/50830974-be6f0d00-1349-11e9-9847-6917d0d145f7.png">

would be nice to have multiline support with auto wrap to not loose precious data. or maybe there already a trick I missed :)

pq:
To clarify, these truncated calls are happening when using `'dart:developer'` to log entries?

jaumard:
Yes :)

devoncarew:
Possibly this? https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/logging/FlutterLogEntryParser.java#L104

jaumard:
@devoncarew might be yes ! But why truncate at first place @pq ?

pq:
@devoncarew: 👍 exactly.

We'll want to do something similar to the data/payload case:

https://github.com/flutter/flutter-intellij/blob/7513a9d71e522cfdcf1ba27a4d6434e4ae196055/src/io/flutter/logging/FlutterLogEntryParser.java#L118-L136

The one tricky thing is that I've done some optimization around rendering to ensure that we don't do it gratuitously.  I need to see how best to mark truncated entry values as stale so they get re-rendered.  That's an investigation on my short list... 🤞 

Thanks for the continued feedback @jaumard!

PR #3057, opened 2019-01-09
Fill in truncated log entries

pq:
Adds support for truncated log entries.

* requests full entry from the debugProcess
* preserves any inherited / carried over color-styling

image

Fixes: #3052 

/cc @jacob314 @devoncarew

### DIFF
--- a/src/io/flutter/logging/FlutterLogEntryParser.java
+++ b/src/io/flutter/logging/FlutterLogEntryParser.java
@@ -16,6 +16,7 @@ import com.intellij.execution.ui.ConsoleViewContentType;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.Key;
+import com.intellij.ui.SimpleTextAttributes;
 import com.jetbrains.lang.dart.ide.runner.DartConsoleFilter;
 import io.flutter.console.FlutterConsoleFilter;
 import io.flutter.logging.util.LineInfo;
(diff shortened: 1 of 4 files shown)


src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -101,15 +102,29 @@ private FlutterLogEntry parseLoggingEvent(@NotNull Event event) {
...
+
     if (message.getValueAsStringIsTruncated()) {
-      messageStr += "...";
-    }
+      entry.setMessage(messageStr + "...");
+      final Isolate isolate = new Isolate(json.get("isolate").getAsJsonObject());
+      debugProcess.getVmServiceWrapper().getObject(isolate.getId(), message.getId(), new GetObjectConsumer() {
+        @Override
+        public void received(Obj response) {
+          entry.setMessage(((Instance)response).getValueAsString());
```
jacob314: FYI this can happen in a different thread. Make sure you trampoline back to the UI thread.

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -101,15 +102,29 @@ private FlutterLogEntry parseLoggingEvent(@NotNull Event event) {
...
+        }
 
-    final LineInfo lineInfo = parseLine(messageStr, category);
-    final FlutterLogEntry entry =  new FlutterLogEntry(timestamp(), lineInfo, level);
+        @Override
+        public void received(Sentinel response) {
+          // TODO(pq): handle?
```
jacob314: For this case and the error case I would just remove the ... and leave the original text perhaps adding <truncated> at the end of the text.

src/io/flutter/logging/FlutterLogEntryParser.java
jacob314: Additionally, we also need to schedule an event to repaint the UI. This code currently appears to be just assuming another UI repaint will happen.

jaumard:
So no wrapping here right @pq ?

pq:
That's correct.  I don't have a clear idea on how we want to handle that just yet... Feel free to open a targeted issue to track?

Thanks!

PR #3068, opened 2019-01-10
Log tree class refactor

pq:
A bit of a class re-org, isolated so the next (functional) change is easier to review.

Functionally there are no changes but:

* a new `tree` sub-package is introduced, and
* `util` is renamed to `text`

image

/cc @jacob314 @devoncarew


src/io/flutter/logging/tree/CategoryCellRenderer.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2019 The Chromium Authors. All rights reserved.
```
jacob314: nit: feel free to maintain the old copyright year when moving files like this.

pq: Ha!  Was just fixing. :)

pq:
> The refactor looks nice. Thanks for keeping the CLs simple!

👍

PR #3073, opened 2019-01-11
Add debugPaint and debugAllowBanner icons.

kenzieschmoll:
screen shot 2019-01-11 at 3 25 13 pm
screen shot 2019-01-11 at 3 25 02 pm


jacob314:
btw there is also an icon for repaint rainbow on the devtools side.

PR #3085, opened 2019-01-15
shorten message for debug mode perf disclaimer

devoncarew:
- shorten message for debug mode perf disclaimer (it was very long long and didn't lay out well in the UI)
- fix a likely bug in terms of when to show the perf table

older UI:
screen shot 2019-01-15 at 11 27 58 am

update:
screen shot 2019-01-15 at 12 45 16 pm


jacob314:
Thanks! The layout of the longer message was something I felt bad about.
Fyi @InMatrix who might have ideas on tweaking the message.

Issue #2893, opened 2018-11-28
make the inspector `recomputeTreeRoot` call tolerant to errors


YograjSolanke:
## What happened

(please describe what you were doing when this exception occurred)
I run the application, and then made some changes and re-run app then hot reloading applied and application crashed by throwing following error

## Version information

Android Studio `3.2.1` • Flutter plugin `io.flutter 30.0.1` • Dart plugin `181.5656`

Error getting Flutter sdk information.

## Exception

java.util.concurrent.CompletionException: java.io.IOException: Application terminated

```
java.lang.Throwable: java.util.concurrent.CompletionException: java.io.IOException: Application terminated
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:126)
	at io.flutter.view.InspectorPanel.lambda$recomputeTreeRoot$5(InspectorPanel.java:575)
	at io.flutter.inspector.InspectorService$ObjectGroup.lambda$null$69(InspectorService.java:880)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:315)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.doRun(LaterInvocator.java:447)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.runNextEvent(LaterInvocator.java:431)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:415)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:762)
	at java.awt.EventQueue.access$500(EventQueue.java:98)
	at java.awt.EventQueue$3.run(EventQueue.java:715)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:732)
	at com.intellij.ide.IdeEventQueue.defaultDispatchEvent(IdeEventQueue.java:822)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:763)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:423)
[...]

PR #3086, opened 2019-01-15
don't use LOG.error()

devoncarew:
- don't use `LOG.error()`
- fix https://github.com/flutter/flutter-intellij/issues/2893, and a number of related issues

Using `LOG.error(...)` will cause an exception dialog to be displayed to the user. We should prefer using LOG.warn / LOG.info, and / or surfacing any errors to the user in whatever mode is best for the situation (a toast, and error dialog, a failed operation, ...).


devoncarew:
> One idea: add a helper method we always call instead of LOG.warn and add an option to in debug builds log these as errors so people developing the plugin don't miss when exceptions were thrown/

Updating now to add an api we can hook into to know about issues (`FlutterUtils.warn`).

Issue #3088, opened 2019-01-16
Add popup for description to the project sample combo


pq:
Currently there is no context displayed for a selected sample.

![image](https://user-images.githubusercontent.com/67586/51252629-32647300-1951-11e9-8760-03c5bb931eba.png)

On selection, we should display a popup along-side the selection w/ sample description.  (Tooltips don't quite work for combo pulldowns so we'll want to render our own.)

Issue #2976, opened 2018-12-18
Banner notifications for source files containing runnable samples


pq:
Sub-task of #2968.

Show a banner with a clickable link saying something like “Open sample project in a new window…” for source files that have an embedded sample (for example, [Card](https://docs.flutter.io/flutter/material/Card-class.html)).  Clicking the link should create a new sample project.

(Note that we'll need we need to ensure that this works for both IntelliJ and Android Studio.)

Open questions:

1. should we pick a temporary location and project name, or
2. prompt for them (w/ a dialog), or
3. ~open the new project wizard, seeded w/ appropriate values~

I've included 3 for completeness but I think it would be cumbersome.

@InMatrix: thoughts on 1 vs. 2 (or some other variation?)

InMatrix:
I think #1 probably is sufficient, since those sample projects are likely to be thrown away after use.

pq:
Great.  Thanks @InMatrix!

@alexander-doroshko : is there a canonical location you'd suggest for temporary projects?   Maybe just fall back to `com.intellij.ide.impl.ProjectUtil.getBaseDir()`?

alexander-doroshko:
I don't know any entity close to 'temporary project' concept. Probably there aren't any.

The closest thing that came to my mind is [Scratch Files](https://www.jetbrains.com/help/idea/scratches.html) concept. They fit perfectly to try a toy single-file project. In some technologies (like Java) scratch files are runnable/debuggable. It should be technically possible to make Dart scratch files runnable/debuggable as well.

If it's possible to have just a single Dart scratch file to run a sample - then the concept is just great! Otherwise you may want to investigate how good scratch files are for multi-files toy projects, how to deal with auxiliary files like pubspec, .packages, etc.

pq:
Playing around with some mocked content.

![image](https://user-images.githubusercontent.com/67586/50566769-14143b80-0cf2-11e9-8993-ca677fc04260.png)

Word-smithing ideas welcome.

Also, it looks like some files host more than one sample.  In that case, we'll need to do more than provide a "Open sample project in a new window…" action.

@InMatrix @gspencergoog: thoughts?

fyi @DanTup (in case you want to do something similar for VS Code).

DanTup:
[...]

PR #3089, opened 2019-01-16
update sample json format

pq:
Updates to latest index format.

* adds `sourceLocation`
* updates display label

image

Follow-up to add a popup with description details tracked in https://github.com/flutter/flutter-intellij/issues/3088.

/cc @jacob314 @devoncarew

### DIFF
--- a/src/io/flutter/samples/FlutterSample.java
+++ b/src/io/flutter/samples/FlutterSample.java
@@ -17,22 +17,31 @@ public class FlutterSample {
   @NotNull
   private final String file;
   @NotNull
+  private final String sourcePath;
+  @NotNull
   private final String description;
 
-  private String relativePath;
-
   FlutterSample(@NotNull String element, @NotNull String library, @NotNull String id, @NotNull String file,
-                @NotNull String description) {
+                @NotNull String sourcePath, @NotNull String description) {
     this.element = element;
     this.library = library;
     this.id = id;
     this.file = file;
+    this.sourcePath = sourcePath;
     this.description = description;
   }
 
   @Override
   public String toString() {
-    return getElement() + " (" + getLibrary() + ")";
+    return getDisplayLabel();
+  }
+
+  /**
+   * Get a label suitable for display in a chooser (e.g., combobox).
+   */
+  public String getDisplayLabel() {
+    // TODO(pq): add disambiguation once it's needed.
+    return getElement();
   }
 
   @NotNull
@@ -60,15 +69,8 @@ public class FlutterSample {
     return file;
   }
 
-  public String getRelativePath() {
-    if (relativePath == null) {
-      // Chomp .dart suffix
-      relativePath = file.substring(file.length() - 5);
-      // material.sample => material/sample
-      relativePath = relativePath.replaceAll("\\.", "/");
-      // Re-add suffix.
-      relativePath += ".dart";
-    }
-    return relativePath;
+  @NotNull
+  public String getSourcePath() {
+    return sourcePath;
   }
 }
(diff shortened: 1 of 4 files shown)


src/io/flutter/samples/FlutterSample.java
```diff
@@ -17,22 +17,31 @@
...
   public String toString() {
-    return getElement() + " (" + getLibrary() + ")";
+    return getDisplayLabel();
+  }
+
+  /**
+   * Get a label suitable for display in a chooser (e.g., combobox).
+   */
+  public String getDisplayLabel() {
+    // TODO(pq): add disambiguation once it's needed.
```
jacob314: also looks like the existing labels aren't that user friendly.

src/io/flutter/samples/FlutterSampleManager.java
```diff
@@ -37,6 +37,10 @@
     final List<FlutterSample> samples = new ArrayList<>();
     try {
       // TODO(pq): replace w/ index read from repo (https://github.com/flutter/flutter/pull/25515).
+
+      //https://docs.flutter.io/snippets/index.json
```
jacob314: which of these locations is the right one to use?

src/io/flutter/samples/FlutterSample.java
pq: Agreed.  Context from the description helps but yeah, these aren't great.  VERY open to ideas, though we'll likely want to stay consistent w/ what we're doing for VSCode.

/fyi @DanTup

src/io/flutter/samples/FlutterSampleManager.java
pq: @DanTup : is this now reliably https://master-docs.flutter.io/snippets/index.json?

DanTup:
@pq Seems like I can't reply inline (maybe because it was merged), but here's my comments....

> labels

Here's how they currently show in VS Code:

screenshot 2019-01-15 at 2 57 16 pm


> is this now reliably https://master-docs.flutter.io/snippets/index.json?

There are two hosts:

- docs.flutter.io - this is tied to the stable version of Flutter and generally what we want to use - however, the Index file won't appear there until the next stable Flutter release (since the doc site is updated as part of the release)
- master-docs.flutter.io - this is a mirror of the site running from master. It's mainly for our own use (testing/previewing things). It's good for testing this, but less good to ship to users (at least, unless you know they're on the master branch - else they may see samples that then fail to create)

In VS Code, we currently have an empty default for the hostname, which results in this feature not showing. I have my local settings override this with the master-docs host which means it shows for me (for testing). When Flutter next has a stable release, I plan to update the default to docs.flutter.io and push a release (this means this feature generally won't be visible to users until the next stable release).

There was a little more discussion about this in https://github.com/Dart-Code/Dart-Code/pull/1391. If we're keen to enable the feature before a stable release, we may need to do some hacks (like just including a local registry of the snippets that are in the current live docs to use in the meantime).

pq:
> If we're keen to enable the feature before a stable release, we may need to do some hacks (like just including a local registry of the snippets that are in the current live docs to use in the meantime).

FWIW I'm doing just this in the IntelliJ plugin.  👍 

My thinking has been to first check the URL (docs.flutter.io I guess) and fall back to a local copy when we ship but I'm open to discussing.  (You certainly shouldn't feel compelled to do the same unless you think it's a good idea.)

As for how it presents in VSCode, I think what you've got looks great!

Out of curiosity, I'm curious if you've considered stripping out the dartdoc markup from the description text?  (I'm thinking about that over on #2976.)

Cheers and thanks for all the thoughtful follow-ups.

DanTup:
> My thinking has been to first check the URL (docs.flutter.io I guess) and fall back to a local copy

Sounds good to me - if you're planning on this, I'll do the same :-)

> Out of curiosity, I'm curious if you've considered stripping out the dartdoc markup from the description text? 

Yeah! I only noticed it when I posted the screenshot here so went to look up if we could use Markdown in those dropdowns.. we can't, do I think stripping the [brackets] off will make sense (I've also sorted it since that screenshot since it looked a bit weird in that order).

PR #3114, opened 2019-01-23
Basic Flutter Error extension event handling.

pq:
Basic Flutter Error extension event handling.

image

As a next step we'll want to do something sensible w/ the parsed `DiagnosticsNode` pending UI exploration in the devtools version.  In the meantime this proves out inspector service initialization and interaction.

/cc @jacob314

### DIFF
--- a/src/io/flutter/logging/FlutterLogEntryParser.java
+++ b/src/io/flutter/logging/FlutterLogEntryParser.java
@@ -14,18 +14,23 @@ import com.intellij.execution.filters.UrlFilter;
 import com.intellij.execution.process.ProcessEvent;
 import com.intellij.execution.ui.ConsoleViewContentType;
 import com.intellij.openapi.Disposable;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.Key;
 import com.intellij.ui.SimpleTextAttributes;
 import com.intellij.util.EventDispatcher;
 import com.jetbrains.lang.dart.ide.runner.DartConsoleFilter;
+import io.flutter.FlutterUtils;
 import io.flutter.console.FlutterConsoleFilter;
+import io.flutter.inspector.DiagnosticsNode;
+import io.flutter.inspector.InspectorService;
 import io.flutter.logging.text.LineInfo;
 import io.flutter.logging.text.LineParser;
 import io.flutter.logging.text.StyledText;
 import io.flutter.run.FlutterDebugProcess;
 import io.flutter.run.daemon.DaemonApi;
+import io.flutter.run.daemon.FlutterApp;
 import io.flutter.server.vmService.HeapMonitor;
 import io.flutter.utils.StdoutJsonParser;
 import org.dartlang.vm.service.VmService;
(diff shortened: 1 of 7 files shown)


src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -83,25 +91,37 @@ public void onError(RPCError error) {
     }
   }
 
-  private FlutterDebugProcess debugProcess;
+  static class InspectorHelper {
```
jacob314: Is this helper class useful given when operating you typically need to handle only having a CompletableFuture<InspectorService.ObjectGroup> due to timing?

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -126,47 +146,53 @@ private FlutterLogEntry parseLoggingEvent(@NotNull Event event) {
...
-          entry.setMessage(((Instance)response).getValueAsString());
-          contentChangedDispatcher.getMulticaster().onContentUpdate();
-        }
-      });
+      if (debugProcess != null) {
+        final Isolate isolate = new Isolate(json.get("isolate").getAsJsonObject());
+        debugProcess.getVmServiceWrapper().getObject(isolate.getId(), message.getId(), new GetObjectAdapter() {
+          @Override
+          public void received(Obj response) {
+            entry.setMessage(((Instance)response).getValueAsString());
```
jacob314: not a new issue but isn't this on a background thread updating UI code?

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -223,19 +249,57 @@ private static long timestamp(Event event) {
...
+      // TODO(pq): process children: diagnosticsNode.getChildren().getNow(null) and set data to something structured.
+      entry.setData(json.toString());
+      entries.add(entry);
+    }
+    return entries;
+  }
+
+  @Nullable
+  private DiagnosticsNode parseDiagnosticsNode(@NotNull JsonObject json) {
+    if (inspectorObjectGroup == null) {
```
jacob314: how can this happen? Should this be an assert as it should generally indicate a logic bug.

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -223,19 +249,57 @@ private static long timestamp(Event event) {
...
+      FlutterUtils.warn(LOG, "Attempt to create a diagnostics node before the object group is set.");
+      return null;
+    }
+    if (app == null) {
```
jacob314: looks like an assert as well.

src/io/flutter/logging/FlutterLogEntryParser.java
```diff
@@ -332,4 +396,9 @@ public FlutterLogEntry parseConsoleEvent(String text, ConsoleViewContentType typ
     // TODO(pq): handle else (errors, etc).
     return null;
   }
+
+  public void setupInspector(@NotNull FlutterApp app, @NotNull VmService vmService) {
```
jacob314: Nit: might be nice to match the naming convention of similar methods that take an app and vmService to initialize debugger state?

src/io/flutter/logging/FlutterLogEntryParser.java
pq: Good point.  This class did more once but now doesn't have much reason to exist.

Removing. 👍

src/io/flutter/logging/FlutterLogEntryParser.java
pq: On my list to look into.  Thanks for the nudge.

src/io/flutter/logging/FlutterLogEntryParser.java
pq: Updated.

src/io/flutter/logging/FlutterLogEntryParser.java
pq: You're right.  Updated.

src/io/flutter/logging/FlutterLogEntryParser.java
pq: Sure thing.  Done.

PR #3128, opened 2019-01-26
update log entry data badge

pq:
Swaps out the overloaded info icon.

Before:

image

After:

image

/cc @jacob314 @devoncarew


jacob314:
One idea is to use the standard expand/collapse icons even though the "expanded" data is shown in a separate panel for now to simplify the UI

pq:
@jacob314 I played around with a few options.  Do you mean the tree node expand?  

`right.png`:

image

Or the one associated w/ "expand all" actions?

`expandAll`:

image

jacob314:
I was thinking more right.png particularly if it is placed on the lhs of
the UI.

PR #3130, opened 2019-01-26
badge simplification

pq:
Follow-up from: https://github.com/flutter/flutter-intellij/pull/3128#issuecomment-457795950

image

/cc @jacob314 @devoncarew


jacob314:
What do you think of putting the expand collapse arrow on the LHS instead of the RHS. That could make it look more like an expand collapse block.

jacob314:
lgtm

PR #3138, opened 2019-01-29
structured error log data ui

pq:
An error data UI corresponding to the devtools prototype (https://github.com/flutter/devtools/pull/144).

image

Still a bunch to do but this proves us out end to end.

On the short-list:
* add hyperlink support to stack traces
* link node selections to inspector
* auto scroll-to-top of error pane
* dark-theme improvements / awareness
* styling in general

/cc @jacob314

### DIFF
--- a/src/io/flutter/logging/FlutterLogEntry.java
+++ b/src/io/flutter/logging/FlutterLogEntry.java
@@ -37,7 +37,10 @@ public class FlutterLogEntry {
   private final int level;
   @NotNull
   private String message;
-  // TODO(pq): consider making data an Instance or JsonElement
+
+  /**
+   * Associated data; may be a JSON string value or diagnostic node.
+   */
   @Nullable
   private Object data;
   private int sequenceNumber = -1;
@@ -46,7 +49,8 @@ public class FlutterLogEntry {
   private final Kind kind;
   private List<StyledText> styledText;
 
-  @NotNull private final List<Filter> filters;
+  @NotNull
+  private final List<Filter> filters;
 
   /**
    * Describes any style info that was inherited from previously parsed lines.
--- a/src/io/flutter/logging/FlutterLogEntry.java
+++ b/src/io/flutter/logging/FlutterLogEntry.java
@@ -39,7 +39,7 @@ public class FlutterLogEntry {
   private String message;
   // TODO(pq): consider making data an Instance or JsonElement
   @Nullable
-  private String data;
+  private Object data;
   private int sequenceNumber = -1;
 
   @NotNull
@@ -78,11 +78,11 @@ public class FlutterLogEntry {
   }
 
   @Nullable
-  public String getData() {
+  public Object getData() {
     return data;
   }
 
-  public void setData(@Nullable String data) {
+  public void setData(@Nullable Object data) {
     this.data = data;
   }
 
@@ -132,7 +132,6 @@ public class FlutterLogEntry {
     final LineInfo lineInfo = lineHandler.parseLineInfo(getMessage(), getCategory());
     return lineInfo.getStyledText();
   }
-
 
   /**
    * Return a sequence number, or -1 if unset.
--- a/src/io/flutter/logging/FlutterLogEntry.java
+++ b/src/io/flutter/logging/FlutterLogEntry.java
@@ -39,7 +39,7 @@ public class FlutterLogEntry {
   private String message;
 
   /**
-   * Associated data; may be a JSON string value or diagnostic node.
+   * Associated data; may be a JSON string value or DiagnosticsNode instance.
    */
   @Nullable
   private Object data;
--- a/src/io/flutter/logging/FlutterLogView.java
+++ b/src/io/flutter/logging/FlutterLogView.java
@@ -7,8 +7,6 @@ package io.flutter.logging;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonParser;
 import com.intellij.execution.filters.Filter;
 import com.intellij.execution.filters.HyperlinkInfo;
 import com.intellij.execution.process.ProcessHandler;
@@ -40,8 +38,8 @@ import com.intellij.util.ui.JBUI;
 import com.intellij.util.ui.UIUtil;
 import io.flutter.FlutterUtils;
 import io.flutter.logging.FlutterLog.Level;
+import io.flutter.logging.tree.DataPanel;
 import io.flutter.run.daemon.FlutterApp;
-import io.flutter.utils.JsonUtils;
 import io.flutter.utils.UIUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -70,7 +68,7 @@ public class FlutterLogView extends JPanel implements ConsoleView, DataProvider,
   @NotNull
   private static final Logger LOG = Logger.getInstance(FlutterLogView.class);
 
-  private static final float DATA_PANEL_SPLITTER_PROPORTION_DEFAULT = 0.75f;
+  private static final float DATA_PANEL_SPLITTER_PROPORTION_DEFAULT = 0.60f;
 
   @NotNull
   private static final Map<FlutterLog.Level, TextAttributesKey> LOG_LEVEL_TEXT_ATTRIBUTES_KEY_MAP;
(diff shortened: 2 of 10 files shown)


src/io/flutter/logging/FlutterLogEntry.java
```diff
@@ -39,7 +39,7 @@
   private String message;
   // TODO(pq): consider making data an Instance or JsonElement
```
jacob314: remove this TODO and document what types data is allowed to have.

src/io/flutter/logging/FlutterLogView.java
```diff
@@ -663,7 +631,7 @@ protected void onScrollToEnd() {
     dataPane = ScrollPaneFactory.createScrollPane(
       dataPanel,
       ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED,
-      ScrollPaneConstants.HORIZONTAL_SCROLLBAR_AS_NEEDED);
+      ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER);
```
jacob314: is it possible to keep the HORIZONTAL_SCROLLBAR_AS_NEEDED and still have the UI lay out as expected in IntelliJ? If not it is fine. Could be nice if users were able to scroll horizontally to see lines that intentionally do not wrap.

src/io/flutter/logging/tree/DataPanel.java
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  /*
+   * Swing's support for CSS is limited and doesn't allow for multiple classes so we cons
+   * together css styling as a single style attribute.
+   *
+   * These values are loosely derived from devtools/inspector.css.
+   *
+   * TODO(pq): add theme-awareness.
+   */
```
jacob314: I like this use of CSS to keep the two tools consistent.

src/io/flutter/logging/tree/DataPanel.java
```diff
@@ -0,0 +1,282 @@
...
+      switch (style) {
+        case error:
+          return "background-color: #f97c7c; padding: 12px; margin:5px 0; ";
+        default:
+          return "";
+      }
+    }
+  }
+
+  class NodeRenderer extends ColoredTreeCellRenderer {
```
jacob314: you should use the multi icon tree cell renderer from the inspector instead. It allows multiple icons per row which is important to accurately display inspector diagnostics.

src/io/flutter/logging/tree/DataPanel.java
```diff
@@ -0,0 +1,282 @@
...
+          final Icon icon = node.getIcon();
+          if (icon != null) {
+            setIcon(icon);
+          }
+        }
+      }
+    }
+  }
+
+  class NodeModel extends BaseTreeModel<Object> {
```
jacob314: It would be nice to use similar concepts as the similar tree of diagnostics in the inspector.
Should the inspector be using this node model instead of the DefaultMutableTreeNode model or should this code be simplified to just use DefaultMutableTreeNode and friends?

src/io/flutter/logging/tree/DataPanel.java
```diff
@@ -0,0 +1,282 @@
...
+
+    // Body.
+    final List<DiagnosticsNode> properties = data.getInlineProperties();
+    for (DiagnosticsNode node : properties) {
+      String contents = "";
+      if (node.getName() != null) {
+        contents += node.getName();
+      }
+
+      final ArrayList<DiagnosticsNode> children = getChildren(node);
```
jacob314: This code is missing a number of edge cases. For example, you shouldn't assume that all "name" and "description" fields should be separated by a colon. It would be best to refactor the code that renders a DiagnosticsNode as a list of icons and styled text blocks. That can work fine with your clever CSS scheme. Add a TODO and we can work on the refactor later.

src/io/flutter/logging/tree/DataPanel.java
```diff
@@ -0,0 +1,282 @@
...
+      }
+    });
+
+    final NonOpaquePanel panel = new NonOpaquePanel();
+    panel.add(tree);
+    add(panel);
+  }
+
+  private JEditorPane createEditorPane() {
+    final JEditorPane editorPane = new JEditorPane();
```
jacob314: Interesting use of editor pane.

src/io/flutter/utils/HtmlBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public static String span(String style, String contents) {
+    return "<span style = \"" + style + "\"> " + contents + "</span>";
+  }
+
+  public static String span(String contents) {
+    return tag("span", contents);
+  }
+
+  public static String attrs(String attribute, String value) {
```
jacob314: nit: why the s? This looks like a single attribute.

src/io/flutter/utils/HtmlBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public static String body(String... contents) {
+    return join(contents);
+  }
+
+  private static String join(String... contents) {
+    final StringBuilder sb = new StringBuilder();
+    for (String c : contents) {
+      sb.append(c);
+      sb.append(java.lang.System.lineSeparator());
```
jacob314: surprising the line separator has any impact given this is html.
Also minor nit: Don't we always want just '\n' regardless of the platform?

src/io/flutter/logging/FlutterLogEntry.java
pq: Good idea. Done

src/io/flutter/logging/FlutterLogView.java
pq: I couldn't get that to work (though I tried). I'll add it to my list of things to revisit.

src/io/flutter/logging/tree/DataPanel.java
pq: 👍 

I started trying to use it even more faithfully but quickly hit the limits of the Swing implementation.  I may explore doing a bit more again but for now this is where I back-peddled to.

src/io/flutter/logging/tree/DataPanel.java
pq: Good call.  Done.

src/io/flutter/logging/tree/DataPanel.java
pq: I started out thinking we should share a model but there's a lot going on in the inspector proper and I'm not sure...  I'll add a TODO and we can discuss in person.

src/io/flutter/logging/tree/DataPanel.java
pq: Yeah.  It handles HTML and reflows pretty well.

src/io/flutter/logging/tree/DataPanel.java
```diff
@@ -0,0 +1,282 @@
...
+        linkSelected(e.getURL());
+      }
+    });
+
+    return editorPane;
+  }
+
+  private static ArrayList<DiagnosticsNode> getChildren(DiagnosticsNode node) {
+    final ArrayList<DiagnosticsNode> children = node.getChildren().getNow(new ArrayList<>());
+    if (!children.isEmpty()) {
```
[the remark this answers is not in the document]
pq: The result of getChildren() was giving me surprising (and duplicated) results when I merged it w/ properties...  Maybe I just need to de-dup?

src/io/flutter/utils/HtmlBuilder.java
pq: Yeah.  It was plural.  Fixing.

src/io/flutter/logging/tree/DataPanel.java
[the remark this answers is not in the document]
pq: Hmmmm....  I'm not seeing this now actually so I guess I made a mistake elsewhere.  Will fix.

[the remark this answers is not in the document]
pq: Ok.  Fixed. 👍

pq:
@jacob314: I think the blockers are all addressed.  PTAL!

src/io/flutter/logging/FlutterLogEntry.java
```diff
@@ -37,7 +37,10 @@
   private final int level;
   @NotNull
   private String message;
-  // TODO(pq): consider making data an Instance or JsonElement
+
+  /**
+   * Associated data; may be a JSON string value or diagnostic node.
```
jacob314: nit: DiagnosticsNode instead of diagnostic node to make clear what the actual union type is.

pq: 👍 fixed.

src/io/flutter/logging/tree/DataPanel.java
```diff
@@ -273,12 +276,8 @@ private JEditorPane createEditorPane() {
...
 
-  private static ArrayList<DiagnosticsNode> getChildren(DiagnosticsNode node) {
+  private static List<DiagnosticsNode> getChildren(DiagnosticsNode node) {
     final ArrayList<DiagnosticsNode> children = node.getChildren().getNow(new ArrayList<>());
-    if (!children.isEmpty()) {
-      return children;
-    }
-
-    return node.getInlineProperties();
+    return Stream.of(children, node.getInlineProperties()).flatMap(Collection::stream).collect(Collectors.toList());
```
jacob314: nit: is this really easier to read then just adding the inline properties and children to an ArrayList?
I'm at a loss to why there is a flagMap call in the middle of this.

pq: Yeah.  Idiomatic Java is hard to read.  I'll sweep this on the next pas..

pq:
Thanks for the review!

PR #3152, opened 2019-02-01
Correct the bazel output for debugging bazel tests.

DaveShuckerow:
Bazel package roots start with :// not : .


jwren:
LGTM

DaveShuckerow:
We will have to redesign the bazel runners to be able to test this properly and prevent a regression.  I'll plan resources for this one internally.

For now, I'm going to land this fix.

:cry:

PR #3154, opened 2019-02-04
organize imports

devoncarew:
- organize imports for newly committed code (including removing unused imports from package we don't want to depend on, like `com.sun.org.apache.xpath...` classes)
- have our build system check and fail if we see any `com.sun. ...` imports
- also format code


devoncarew:
cc @DaveShuckerow

DaveShuckerow:
LGTM.

What tool are we running to reformat the code?

devoncarew:
Just IntelliJ itself - ctrl-option-o to organize imports, option-command-l to format code.

PR #3177, opened 2019-02-11
change the icon used for paint baselines

devoncarew:
- change the icon used for paint baselines

screen shot 2019-02-10 at 6 11 50 pm


jacob314:
Nice. Are you updating the devtools icon as well?

devoncarew:
Yup, in progress :)

PR #3181, opened 2019-02-11
Hot reload control

### LINKED ISSUE #3168: Error check before hot reload will cause reloads to fail if a clean app is in a bigger project with separate failures
1. Introduce an error to test code (say a line that has no semicolon).
2. Run the app
3. Reloading the app fails.

The lines below cause this to happen:
https://github.com/flutter/flutter-intellij/blob/a1748e99fd093697946b4cbbfa975bc33cc57d52/src/io/flutter/run/FlutterReloadManager.java#L199-L205

I'm looking into removing those and letting the flutter daemon tell us if it's safe to reload the app.  The parent to this issue is https://github.com/flutter/flutter/issues/27618

DaveShuckerow:
Disabling the check that the project has no analysis errors.

Fixes #3168 pending an upstream fix to the Flutter daemon (flutter/flutter#27618).


devoncarew:
I think we'll want the flutter change in place, and in a more stable release (beta? stable?) before we land something like this.

If we do need to don something short-term, we could add a user preference.

DaveShuckerow:
Agreed.  I wanted to have this up so that we can see the change that has to be made.

I'll close this PR until we're ready to actually submit the fix.

passsy:
I have flutter as submodule in my project. Whenever I jump into flutter source, the analysis server discovers the flutter submodule and scans all source. It finds errors in `/packages/*` because the pub packages aren't pulled. The analysis server then prevents my 100% error/warning free app from starting.
I have to close all flutter source files and restart the server.

The whole procedure takes ~30s because the analysis server is so busy scanning the sources.

I'd appreciate a user preference 👍

DaveShuckerow:
Commit history is somewhat broken; I'm going to open a new pr off of this.

Issue #3173, opened 2019-02-08
NPE from WidgetPerfSummaryView.<init>(WidgetPerfSummaryView.java:48)


zeeshan-ali:
Thanks for the feedback! If your issue is related to the Flutter framework itself,
please open an issue at
[github.com/flutter/flutter](https://github.com/flutter/flutter/issues/new).

## Steps to Reproduce

_Please tell us what you were doing and what went wrong_

## Version info

Please paste the output of running `flutter doctor -v` here (available from the command
line or from `Tools > Flutter > Flutter Doctor`). It will provide the version of the
Flutter framework as well as of the IntelliJ plugin.

zeeshan-ali:
java.lang.NullPointerException
	at io.flutter.view.WidgetPerfSummaryView.<init>(WidgetPerfSummaryView.java:48)
	at io.flutter.view.InspectorPerfTab.buildUI(InspectorPerfTab.java:252)
	at io.flutter.view.InspectorPerfTab.<init>(InspectorPerfTab.java:80)
	at io.flutter.view.FlutterPerfView.addPerformanceTab(FlutterPerfView.java:217)
	at io.flutter.view.FlutterPerfView.addPerformanceViewContent(FlutterPerfView.java:174)
	at io.flutter.view.FlutterPerfView.debugActive(FlutterPerfView.java:101)
	at io.flutter.view.FlutterPerfViewFactory.lambda$initPerfView$1(FlutterPerfViewFactory.java:27)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:315)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.doRun(LaterInvocator.java:447)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.runNextEvent(LaterInvocator.java:431)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:415)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:762)
	at java.awt.EventQueue.access$500(EventQueue.java:98)
	at java.awt.EventQueue$3.run(EventQueue.java:715)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.security.AccessController.doPrivileged(Native Method)
[...]

PR #3182, opened 2019-02-11
Avoid NPE if addPerfListener is called out of order with creating perfManager.getCurrentStats()

jacob314:
This should fix
https://github.com/flutter/flutter-intellij/issues/3173

### DIFF
--- a/src/io/flutter/perf/FlutterWidgetPerfManager.java
+++ b/src/io/flutter/perf/FlutterWidgetPerfManager.java
@@ -62,6 +62,8 @@ public class FlutterWidgetPerfManager implements Disposable, FlutterApp.FlutterA
   private boolean trackRepaintWidgets = trackRepaintWidgetsDefault;
   private boolean debugIsActive;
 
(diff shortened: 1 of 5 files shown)


jacob314:
Hold off on reviewing. I think I found one more edge case.

jacob314:
Implemented a more robust fix. Now listeners are disposed when the view is destroyed but not when the app is just rerun.

src/io/flutter/perf/FlutterWidgetPerfManager.java
```diff
@@ -318,6 +324,21 @@ public void dispose() {
     if (currentStats != null) {
       currentStats.dispose();
       currentStats = null;
+      listeners.clear();
```
devoncarew: Should we instead call `removePerfListener` for each listener?

jacob314: I don't think we need  to as the stats object the listeners are added to is disposed a couple lines earlier.

PR #3183, opened 2019-02-11
Enable support for detaching from Flutter apps on exit instead of stopping

DaveShuckerow:
Call "app.detach" on the Flutter daemon instead of "app.stop".

Questions:

* [ ] This API was added a few months ago to the Flutter daemon.  How do we check versions on that for API compatibility?
* [ ] This change uses the simplest approach of just letting the user decide globally whether they want to do this.  Another possibility would be to surface a 'detach' button in the UI that users can click instead of the big stop button.  Is there precedent for this?

I know IntelliJ occasionally demonstrates understanding a distinction between 'terminate' and 'disconnect'.  Is there a way we can use that logic for this workflow instead?


jwren:
change LGTM

devoncarew:
> This change uses the simplest approach of just letting the user decide globally whether they want to do this.  Another possibility would be to surface a 'detach' button in the UI that users can click instead of the big stop button.  Is there precedent for this?

If it's important to get a change for this is soon, we can land this. However - from talking to @keertip - another solution here is to:

- enhance flutter_tools so that the flutter attach command passes the fact that we connected to the app via attach to the daemon library
- when the daemon app start event is sent to the IDE, it includes the launch mode (in this case, something like `"attach"`)
- when stopping an app, if we determine that it had been started via flutter attach (based on the daemon info), we instead call the detach command

I'd prefer this impl over the user preference one. If we do land the user pref one for the short term, we should plan to remove once we've enhanced flutter_tools enough to support knowldge about how an app was launched.

DaveShuckerow:
I also feel better about determining from how the app was run what the appropriate termination is.

I'd like to ship a fix to this issue in the release this month, and there is very little that can go wrong with the userpreferences approach, so I think it's safest to do this and launch, then add a more nuanced detection algorithm without time pressure.

devoncarew:
> I'd like to ship a fix to this issue in the release this month, and there is very little that can go wrong with the use rpreferences approach, so I think it's safest to do this and launch, then add a more nuanced detection algorithm without time pressure.

Got it. Let me review this PR then; can you (or @keertip) file a flutter_tools issue to enhance the daemon protocol's app.start event to include the method that the app was started from (`"attach"`)?

keertip:
Will create an issue for enhancing flutter_tools

devoncarew:
Given https://github.com/flutter/flutter-intellij/pull/3185 and the associated flutter_tools change, do we still want to land this?

DaveShuckerow:
Closing because #3185 is in.

PR #3185, opened 2019-02-12
More intelligently enable support for detaching from Flutter apps on exit

DaveShuckerow:
This change depends on flutter/flutter#27812 and uses it the additional information passed by the daemon to decide whether the app was attached-to or run, and then detach or stop when terminating.

This change works with Jonah's changes in g3.

cc: @jonahwilliams @keertip


DaveShuckerow:
I generally create new branches off of upstream/master.  In this case, I started by forking the other pending change to pick the API additions to the flutter daemon.  If you have any suggestions about cherrypicking strategies, Devon, I'm happy to hear.

PR #3188, opened 2019-02-13
read samples from remote / update local cache

pq:
* looks for remote before falling back to local cache
* updates local cache to be consistent w/ current stable

/cc @stevemessick @DanTup


DanTup:
I'm not a great reviewer for this since I don't know this code (or Java), but the logic does read good and seems to match what we're doing in VS Code. I had to look up `IOUtils.toString` as it wasn't obvious from the call that it's fetching from the web 🙃

pq:
Thanks for taking a look @DanTup .  Mainly I just wanted to confirm that we're both doing the same thing.

PR #3189, opened 2019-02-13
Build 33

stevemessick:
@pq @devoncarew 

Update the build to make artifacts for Android Studio 3.3.1, 3.4, 3.5, and IntelliJ 2018.2.5, 2018.3, 2019.1. Might not need 2018.2.5 but if so we can simply not publish it and remove it later.


stevemessick:
Forgot to push new artifacts! Doing it now.

PR #3201, opened 2019-02-16
update the product matrix

devoncarew:
- update the product matrix

@stevemessick, I noticed that we were shipping nearly the same version ranges for `IntelliJ 2018.2.5` and `AS 3.3.1` (`182.5107.7` - `182.*`). I don't know that the jetbrains plugin server would know which version to distribute, and assume that we no longer need to support IntelliJ 2018.2. I assume this change is correct, but would like your eyes on it.

We may want to add some automation to the bots here - perhaps a dart script which will parse this file, and fail with error if there are any overlapping ranges, or unintentional gaps in ranges.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ script: ./tool/travis.sh
 env:
   - DART_BOT=true
   - CHECK_BOT=true
-  - IDEA_VERSION=2018.2.5
   - IDEA_VERSION=3.3.1
   - IDEA_VERSION=2018.3
   - IDEA_VERSION=2019.1
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -1,16 +1,6 @@
 {
   "list": [
     {
-      "comments": "IntelliJ 2018.2.5",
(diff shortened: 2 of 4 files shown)


stevemessick:
I don't have 2018.2.5 installed to easily check that it is working. I'll get it as soon as I can. All the others work, but that's a good point that this one needs to be verified. I had a lot of trouble getting the 2018.3 builds figured out, so I'm pretty sure I don't fully understand how the Jetbrains server decides which file to provide -- meaning that automatic checking should wait until we have better understanding. BTW I'm thinking about adding beta and canary channels to our plugin distro so we can just side-step the whole version range issue for Android Studio.

stevemessick:
BTW the redness is caused by a bug in the travis script. I'll look at it later.

devoncarew:
> I don't have 2018.2.5 installed to easily check that it is working.

I think we can drop support for IntelliJ 2018.2 - stable is at 2018.3. I assume we just need support for this platform version range for AS 3.3.1 / AS 3.3.

PR #3238, opened 2019-02-28
Port 'inferPubRootDirectoryIfNeeded' from devtools to plugin

keertip:


src/io/flutter/inspector/InspectorService.java
```diff
@@ -255,6 +257,80 @@ private void onVmServiceReceived(String streamId, Event event) {
     return invokeServiceMethodDaemonNoGroup(methodName, params);
   }
 
+  /*
+   * We make a best guess for the pub directory based on the the root directory
+   * of the first non artifical widget in the tree.
+  */
+  public CompletableFuture<String> inferPubRootDirectoryIfNeeded(){
```
keertip: Do we need to return String here?

PR #3239, opened 2019-02-28
Turn the new bazel test runner to enabled-by-default

DaveShuckerow:
We've seen no instability in it in g3, so we'll turn it on by default for new users.


devoncarew:
The change generally looks good; also adding @lambdabaa for review.

@DaveShuckerow, what release do you anticipate use removing the user preference entirely? Perhaps open an issue to track that.

DaveShuckerow:
> @DaveShuckerow, what release do you anticipate use removing the user preference entirely? Perhaps open an issue to track that.

Done.  I've cc'd you on the internal tracking issue.

jwren:
LGTM.  No objection here.

PR #3242, opened 2019-02-28
Port inferPubRootDirectoryIfNeeded from devtools

keertip:


src/io/flutter/inspector/InspectorService.java
```diff
@@ -255,6 +257,80 @@ private void onVmServiceReceived(String streamId, Event event) {
     return invokeServiceMethodDaemonNoGroup(methodName, params);
   }
 
+  /*
+   * We make a best guess for the pub directory based on the the root directory
+   * of the first non artifical widget in the tree.
+  */
+  public CompletableFuture<String> inferPubRootDirectoryIfNeeded(){
```
keertip: Do we need to return String here?

jacob314: I would continue to return a CompleteableFuture<String> 
could be useful if you want to display what the inferred root project directory was or as an enhancement pair it with what the project directory defined by the user's actual project was. 
A v2 refinement would be to use this directory to rewrite the user's actual open package roots to match the same path scheme. E.g. what happened was the dirs used by the kernel transformer are probably in a directory with a different file prefix.

src/io/flutter/inspector/InspectorService.java
```diff
@@ -255,6 +257,80 @@ private void onVmServiceReceived(String streamId, Event event) {
...
+    ObjectGroup group = createObjectGroup("temp");
+    DiagnosticsNode root = group.getRoot(FlutterTreeType.widget).join();
+
+    if (root == null) {
+      // No need to do anything as there isn't a valid tree (yet?).
+      group.dispose();
+      return null;
```
jacob314: Return CompletableFuture.completedFuture(null) instead of null.

src/io/flutter/inspector/InspectorService.java
```diff
@@ -255,6 +257,80 @@ private void onVmServiceReceived(String streamId, Event event) {
...
+    if (children != null && children.isEmpty() == false) {
+      // There are already widgets identified as being from the summary tree so
+      // no need to guess the pub root directory.
+      return null;
+    }
+
+    ArrayList<DiagnosticsNode> allChildren =
+      group.getChildren(root.getDartDiagnosticRef(), false, null).join();
+
+    if (allChildren == null || allChildren.isEmpty() == true){
```
jacob314: make sure you run the java code formatter befrore checkin. Looks like there is a missing space here, etc.

src/io/flutter/inspector/InspectorService.java
```diff
@@ -255,6 +257,80 @@ private void onVmServiceReceived(String streamId, Event event) {
...
+    ArrayList<DiagnosticsNode> children = root.getChildren().join();
+
+    if (children != null && children.isEmpty() == false) {
+      // There are already widgets identified as being from the summary tree so
+      // no need to guess the pub root directory.
+      return null;
+    }
+
+    ArrayList<DiagnosticsNode> allChildren =
+      group.getChildren(root.getDartDiagnosticRef(), false, null).join();
```
jacob314: avoid using join. Likely you are freezing the IntelliJ UI thread which is where this method was probably called from. Use the async methods on the CompletableFuture class instead to make this method Async like the other methods on this class. That helps keep the logic consistent on the Dart and Java sides although it makes the Java code a bit painful to write.

src/io/flutter/inspector/InspectorService.java
jacob314: nit: == true is not needed.

src/io/flutter/inspector/InspectorService.java
```diff
@@ -255,6 +257,80 @@ private void onVmServiceReceived(String streamId, Event event) {
...
+    if (location == null) {
+      return null;
+    }
+    String path = location.getPath();
+    if (path == null) {
+      group.dispose();
+      return null;
+    }
+    // TODO(jacobr): it would be nice to use Isolate.rootLib similar to how
+    // debugger.dart does but we are currently blocked by the
```
jacob314: Update comment to not reference debugger.dart or reference debugger.dart in the devtools.

src/io/flutter/inspector/InspectorService.java
```diff
@@ -255,6 +257,80 @@ private void onVmServiceReceived(String streamId, Event event) {
...
+    // able to use package: paths. Temporarily all tools tracking widget
+    // locations will need to support both path formats.
+    // TODO(jacobr): use the list of loaded scripts to determine the appropriate
+    // package root directory given that the root script of this project is in
+    // this directory rather than guessing based on url structure.
+    ArrayList<String> parts = new ArrayList<String>(Arrays.asList(path.split("/")));
+    String pubRootDirectory = null;
+    for (int i = parts.size() - 1; i >= 0; i--) {
+      String part = parts.get(i);
+      if (part == "lib" || part == "web") {
```
jacob314: you need to use equals in java instead of == for string equality.

jacob314: same elsewhere as well. 
Wow porting from Java to Dart is a bit easier than the other way around. :)

src/io/flutter/inspector/InspectorService.java
```diff
@@ -255,6 +257,80 @@ private void onVmServiceReceived(String streamId, Event event) {
...
+    }
+    if (pubRootDirectory == null) {
+      parts.remove( parts.size() - 1 );
+      pubRootDirectory = String.join("/", parts);
+    }
+
+    setPubRootDirectories(new ArrayList<String>(Arrays.asList(pubRootDirectory))).join();
+    group.dispose();
+    String roorDir = pubRootDirectory;
+    return CompletableFuture.supplyAsync(()-> roorDir);
```
jacob314: I don't think this supplyAsync call does what you want.  What is it trying to do?
What I would expect you want is to create
CompletableFuture<String> ret = new CompletableFuture<>();

and then when you have the root directory, complete with the root directory.

src/io/flutter/view/InspectorPanel.java
```diff
@@ -546,6 +546,7 @@ void maybeLoadUI() {
     if (flutterAppFrameReady) {
       // We need to start by quering the inspector service to find out the
       // current state of the UI.
+      inspectorService.inferPubRootDirectoryIfNeeded();
```
jacob314: you need to wait for the pub root directory to be inferred before continuing.
Otherwise you will fetch the selection before you have the pub root directory.

src/io/flutter/inspector/InspectorService.java
keertip: Done

src/io/flutter/inspector/InspectorService.java
keertip: removed

src/io/flutter/inspector/InspectorService.java
keertip: done

src/io/flutter/inspector/InspectorService.java
keertip: Fixed. Found it really tough to write Java after Dart.

src/io/flutter/inspector/InspectorService.java
keertip: changed.

src/io/flutter/inspector/InspectorService.java
keertip: changed to thenComposeAsync.

src/io/flutter/view/InspectorPanel.java
keertip: done

src/io/flutter/inspector/InspectorService.java
keertip: ack.

keertip:
@jacob314, ptal

jacob314:
Lgtm!

Issue #3237, opened 2019-02-28
Error when opening project in GoLand (Missing IOUtils)


Michael77:
## What happened

I opened a preexisting flutter project in GoLand.

## Version information

GoLand `2018.3.4` • Flutter plugin `io.flutter 33.3.2` • Dart plugin `183.5901`

Flutter 1.2.1 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 8661d8aecd (2 weeks ago) • 2019-02-14 19:19:53 -0800
Engine • revision 3757390fa4
Tools • Dart 2.1.2 (build 2.1.2-dev.0.0 0a7dcf17eb)

## Exception

null

```
java.lang.NoClassDefFoundError: org/apache/commons/io/IOUtils
	at io.flutter.samples.FlutterSampleManager.readSampleIndex(FlutterSampleManager.java:57)
	at io.flutter.samples.FlutterSampleManager.readSampleIndex(FlutterSampleManager.java:64)
	at io.flutter.samples.FlutterSampleManager.loadSamples(FlutterSampleManager.java:79)
	at io.flutter.samples.FlutterSampleManager.getSamples(FlutterSampleManager.java:51)
	at io.flutter.samples.FlutterSampleNotificationProvider.getSamplesForFile(FlutterSampleNotificationProvider.java:60)
	at io.flutter.samples.FlutterSampleNotificationProvider.createNotificationPanel(FlutterSampleNotificationProvider.java:42)
	at io.flutter.samples.FlutterSampleNotificationProvider.createNotificationPanel(FlutterSampleNotificationProvider.java:24)
	at com.intellij.ui.EditorNotificationsImpl$3.performInReadAction(EditorNotificationsImpl.java:136)
	at com.intellij.openapi.progress.util.ReadTask.lambda$runBackgroundProcess$0(ReadTask.java:66)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:977)
	at com.intellij.openapi.application.ReadAction.compute(ReadAction.java:57)
	at com.intellij.openapi.progress.util.ReadTask.runBackgroundProcess(ReadTask.java:66)
	at com.intellij.openapi.progress.util.ProgressIndicatorUtils.lambda$runUnderProgress$4(ProgressIndicatorUtils.java:227)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$3(CoreProgressManager.java:180)
[...]

PR #3243, opened 2019-03-01
add org.apache.commons.io. to the list of imports to avoid

devoncarew:
- add org.apache.commons.io. to the list of imports to avoid

@pq - this is not yet something we can land - I don't know how to circumlocute around `org.apache.commons.io.IOUtils`.


pq:
Looks good.  I'll fix up #3237 and get rid of the dep so we can land this.

pq:
OK, my change has landed and kicked this build.  We should be good to go.

devoncarew:
Thanks! Will merge master into this and land the PR.

PR #3246, opened 2019-03-01
Gui tests

stevemessick:
@devoncarew @pq 

Add a simple test, just to confirm that the framework is usable.

Unfortunately, it doesn't work as-is. I have to copy the `io.flutter.tests.gui` package into another module (`intellij.idea.community.guiTests`)  to get it to work. I'm working with one of the Jetbrains engineers to understand what's wrong.


flutter-gui-tests/testSrc/io/flutter/tests/gui/ProjectCreator.kt
```diff
@@ -0,0 +1,61 @@
...
+        GuiTestUtilKt.waitProgressDialogUntilGone(robot = robot(), progressTitle = "Loading Templates",
+                                                  timeoutToAppear = Timeouts.seconds02)
+        dialog("New Project") {
+          jList("Flutter").clickItem("Flutter")
+          button("Next").click()
+          typeText(projectName)
+          button("Finish").click()
+          GuiTestUtilKt.waitProgressDialogUntilGone(robot = robot(), progressTitle = "Creating Flutter Project",
+                                                    timeoutToAppear = Timeouts.seconds03)
+        }
```
pq: aside: these wait conditions scare me a little (and bring me back to the kind of test flakes that haunted windowtester)

flutter-gui-tests/testSrc/io/flutter/tests/gui/ProjectCreator.kt
```diff
@@ -0,0 +1,61 @@
...
+
+  private val defaultProjectName = "untitled"
+  private val LOG = Logger.getInstance(this.javaClass)
+
+  fun createProject(projectName: String = defaultProjectName, needToOpenMain: Boolean = true) {
+    with(guiTestCase) {
+      welcomeFrame {
+        actionLink("Create New Project").click()
+        GuiTestUtilKt.waitProgressDialogUntilGone(robot = robot(), progressTitle = "Loading Templates",
+                                                  timeoutToAppear = Timeouts.seconds02)
```
pq: ~this feels like a race condition waiting to happen.~

**EDIT:** I'm guessing this actually waits for the dialog to appear and then disappear?

flutter-gui-tests/testSrc/io/flutter/tests/gui/SmokeTest.kt
```diff
@@ -0,0 +1,62 @@
...
+        // Notice that the counter didn't reset back to zero; the application
+        // is not restarted.
+        primarySwatch: Colors.blue,
+      ),
+      home: MyHomePage(title: 'Flutter Demo Home Page'),
+    );
+  }
+}
+"""
+}
```
pq: trailing newline?

flutter-gui-tests/testSrc/io/flutter/tests/gui/SmokeTest.kt
```diff
@@ -0,0 +1,62 @@
...
+  @Test
+  fun testStartup() {
+    ProjectCreator.createProject(projectName = "guitest")
+    ideFrame {
+      editor {
+        // Wait until current file has appeared in current editor and set focus to editor.
+        moveTo(1)
+      }
+      val editorCode = editor.getCurrentFileContents(false)
+      Assert.assertTrue(editorCode!!.unifyCode().startsWith(codeText.unifyCode()))
```
pq: I guess the templates don't change often but my inclination would be to make this a little fuzzier in case they do.

(On the other hand, we can just cross this bridge if/when we need to.)

flutter-gui-tests/testSrc/io/flutter/tests/gui/ProjectCreator.kt
stevemessick: That was my first reaction, too! It isn't that bad, though. The time-out serves as a safety catch. If the dialog never shows we just continue after a couple seconds waiting for it. If it does show, the function waits until it disappears.

flutter-gui-tests/testSrc/io/flutter/tests/gui/ProjectCreator.kt
stevemessick: Exactly, and the timer limits the wait for appearance.

flutter-gui-tests/testSrc/io/flutter/tests/gui/SmokeTest.kt
stevemessick: Good catch.

flutter-gui-tests/testSrc/io/flutter/tests/gui/SmokeTest.kt
stevemessick: Yeah, I could probably just check that the content is not empty. We don't really care what it is, do we?

pq: Right.  You'd hope this is being well tested elsewhere.

PR #3262, opened 2019-03-07
Add the 'all' device to the devices dropdown

DaveShuckerow:
Only adds this when there are multiple devices to choose from.

This is blocked on flutter/flutter#29011.

### DIFF
--- a/src/io/flutter/run/daemon/DeviceDaemon.java
+++ b/src/io/flutter/run/daemon/DeviceDaemon.java
@@ -87,7 +87,14 @@ class DeviceDaemon {
    * <p>This is calculated based on add and remove events seen since the process started.
    */
   ImmutableList<FlutterDevice> getDevices() {
-    return devices.get();
+    ImmutableList<FlutterDevice> attachedDevices = devices.get();
+    if (attachedDevices.size() > 1) {
+      return new ImmutableList.Builder<FlutterDevice>()
+        .addAll(attachedDevices)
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/daemon/DeviceSelection.java
```diff
@@ -15,7 +15,8 @@
 /**
  * An immutable snapshot of the list of connected devices and current selection.
  */
-class DeviceSelection {
+class
+DeviceSelection {
```
devoncarew: reformat?

src/io/flutter/run/daemon/DeviceDaemon.java
```diff
@@ -87,7 +87,14 @@ boolean isRunning() {
    * <p>This is calculated based on add and remove events seen since the process started.
    */
   ImmutableList<FlutterDevice> getDevices() {
-    return devices.get();
+    ImmutableList<FlutterDevice> attachedDevices = devices.get();
+    if (attachedDevices.size() > 1) {
```
devoncarew: We may want `>= 1`, in order for consistency in the UI when devices are being connected and disconnected.

DaveShuckerow:
> Some comments; I'll look at the other IntelliJ PR now (and is sounds like we'll want to delay this PR anyway until the 'all' device is supported by flutter_tools).

Correct; this will be blocked until the daemon has support for `--machine -d all`

Issue #3263, opened 2019-03-08
File descriptor leak in PluginConfig.load()


rsauciuc:
Android Studio 3.5 Canary 6 with Flutter plugin v33.3.2 runs out of file descriptors on Mac, subsequently failing every I/O call with "Too many open files".

A quick lsof inspection of the process showed ~10K entries for `config/intellij-plugins/flutter.json`.

One idea: does PluginConfig.load() close the InputStreamReader, or should that be a try-with-resources in 
https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/bazel/PluginConfig.java#L75 
(I don't know if GSON.fromJson() is supposed to close it.)

devoncarew:
Thanks for the (very precise!) report. It does look like we're leading resources there.

pq:
Super report.  That you @rsauciuc!

Tywholland:
Nice! Thanks for the quick fix!
What's a good way to track which version of the Flutter plugin this fix will be in? I'd like to verify this is the root cause of another issue I've been seeing.

pq:
It should be in the next release which will be 34.0.

You can see versions with changelogs here:

https://plugins.jetbrains.com/plugin/9212-flutter

devoncarew:
Depending on @DaveShuckerow eval of the fix, if may be worth doing a patch release to capture this fix.

DaveShuckerow:
I'm checking open files with `lsof -p <aswb pid> | wc -l` to count the number of open files.

Here's my results with the fix pulled down:

```
➜  google3 lsof -p 71205 | wc -l     
     140
➜  google3 lsof -p 71205 | wc -l
     140
➜  google3 lsof -p 71205 | wc -l
     145
➜  google3 lsof -p 71205 | wc -l
     145
➜  google3 lsof -p 71205 | wc -l
     142
➜  google3 lsof -p 71205 | wc -l
     140
```

I'm running ASWB 3.4 Beta 4.

DaveShuckerow:
I think we should cherrypick this.

devoncarew:
@DaveShuckerow, do you mind driving the cherry-pick and publishing process? I think it's documented enough, and you should have sufficient access; if not, please let me know and we can correct; thanks!

PR #3264, opened 2019-03-08
address a file handle leak

devoncarew:
- address a file handle leak
- fix https://github.com/flutter/flutter-intellij/issues/3263

@DaveShuckerow

### DIFF
--- a/src/io/flutter/bazel/PluginConfig.java
+++ b/src/io/flutter/bazel/PluginConfig.java
@@ -68,11 +68,12 @@ public class PluginConfig {
   /**
    * Reads plugin configuration from a file, if possible.
    */
-  public static @Nullable
-  PluginConfig load(@NotNull VirtualFile file) {
+  @Nullable
+  public static PluginConfig load(@NotNull VirtualFile file) {
     final Computable<PluginConfig> readAction = () -> {
-      try {
-        final InputStreamReader input = new InputStreamReader(file.getInputStream(), "UTF-8");
+      try (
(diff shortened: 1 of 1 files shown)


jwren:
LGTM

src/io/flutter/bazel/PluginConfig.java
```diff
@@ -68,11 +68,12 @@ public int hashCode() {
...
    */
-  public static @Nullable
-  PluginConfig load(@NotNull VirtualFile file) {
+  @Nullable
+  public static PluginConfig load(@NotNull VirtualFile file) {
     final Computable<PluginConfig> readAction = () -> {
-      try {
-        final InputStreamReader input = new InputStreamReader(file.getInputStream(), "UTF-8");
+      try (
+        final InputStreamReader input = new InputStreamReader(file.getInputStream(), "UTF-8")
```
lambdabaa: I think it could help to add a comment here to describe what this try parentheses block is achieving since it's not a very common structure and this looks to be a very subtle change. I'm unclear on what behavior is changing in this patch.

devoncarew: Thanks - good point; I'll do that in a follow up.

devoncarew: It's a try-with-resources statement (https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html), which I myself hadn't used before.

Issue #3219, opened 2019-02-21
Logger performance


jaumard:
In my app I'm doing a synchro of a lot of data, and each http request and sql command is logged, which grow a lot (like 4500 lines in few seconds). During the time the log panel show all those entries, I can't do anything, no refactoring, no open file as it doesn't do the search. On log panel I can't filter by level or search a keyword :/  

```
[✓] Android Studio (version 3.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 33.0.1
    • Dart plugin version 182.5215
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1248-b01)

```

jaumard:
Same after the update: 
```
[✓] Android Studio (version 3.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 33.1.1
    • Dart plugin version 182.5215
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1248-b01)

```

stevemessick:
@pq

pq:
Thanks for the report @jaumard!

pq:
@jaumard: I'm looking into this.  Do you have a shareable repro?  (I've got some ideas but nothing beats the real thing.)

Thanks in advance!

jaumard:
I'm having this at work on our private project, let me see if I can create a smaller project to reproduce this

pq:
Thanks!

pq:
I'd be curious if you're seeing any exceptions when things bog down as well.

jaumard:
The fastest and easiest way I was able to get close to what I get in my project is by using: 

```
import 'dart:developer';
import 'package:logging';
import 'package:flutter/material.dart';
void main() => runApp(MyApp());
class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: MyHomePage(title: 'Flutter Logger test'),
    );
  }
}

class MyHomePage extends StatelessWidget {
  MyHomePage({Key key, this.title}) : super(key: key);
  final String title;
  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        title: Text(title),
      ),
      body: Center(
        child: RaisedButton(
          onPressed: () {
            var test= 9;
[...]

PR #3280, opened 2019-03-12
throttle logger updates

pq:
Fixes: #3219.

* puts logging view updates on a throttled timer (updates limited to once per 100ms)
* simplifies append
* unifies UI exec calls and adds guards

Show's liveness using a version of @jaumard's stress test.

logging_perf

(Previously this would have hung the UI.)

/cc @stevemessick @jacob314

### DIFF
--- a/src/io/flutter/logging/FlutterLogTree.java
+++ b/src/io/flutter/logging/FlutterLogTree.java
@@ -22,10 +22,10 @@ import org.apache.commons.lang.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import javax.swing.Timer;
 import javax.swing.*;
 import javax.swing.table.*;
 import javax.swing.tree.DefaultMutableTreeNode;
-import javax.swing.tree.MutableTreeNode;
 import java.awt.*;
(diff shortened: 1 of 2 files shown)


src/io/flutter/logging/FlutterLogTree.java
```diff
@@ -207,6 +210,7 @@ private TreeModel(@NotNull ColumnModel columns, @NotNull Disposable parent, @Not
       setShowSequenceNumbers(false);
 
       uiThreadAlarm = new Alarm(Alarm.ThreadToUse.SWING_THREAD, parent);
+      updateTimer = new Timer(100, e -> uiExec(this::update, 0));
```
jacob314: consider using AsyncRateLimiter instead of manually creating the timer. It is designed for the problem of ensuring you don't overdo UI updates

pq: ah!  Crazy I overlooked that.  I'm not super familiar with the nuances here... Can you explain what tangible difference it makes over a Swing Timer?

Thanks!

pq: Nevermind.  I get it now.  Will migrate.

🍻

PR #3288, opened 2019-03-15
Unpack canned projects

stevemessick:
@devoncarew @pq 

New features for GUI tests
- Unpack canned projects from testData
- Update pub dependencies
- New test to exercise both

Canned projects can be saved as a single src.zip, which is unzipped before running the test, or as a runnable project, which is copied before running the test. 

This also includes the missing module files. They are generated by Gradle. I'm not sure this is the right way to handle them, but we can revisit when we know more.


flutter-gui-tests/testSrc/io/flutter/tests/gui/ProjectCreator.kt
```diff
@@ -1,60 +1,131 @@
...
+    val projectPath: File = guiTestCase.guiTestRule.importProject(projectDirFile)
+    with(guiTestCase) {
+      openMainInProject()
+      waitForFirstIndexing()
+      ideFrame {
+        editor {
+          try {
+            val note = notificationPanel()
+            // Best practice is to avoid reusing strings defined in bundles.
+            if (note?.getLabelText() == "Pubspec has been edited") {
```
pq: I'd be interested to see if we ever hit race conditions here.  (In my experience there can be a lag in the banner appearing.) But maybe `waitForFirstIndexing()` normalizes this away?

flutter-gui-tests/testSrc/io/flutter/tests/gui/ProjectCreator.kt
```diff
@@ -1,60 +1,131 @@
...
+              log.info("Get dependencies")
+              note.clickLink("Get dependencies")
+              robot().waitForIdle() // TODO(messick) Check Messages view for completion of 'flutter packages get'
+            }
+          }
+          catch (ex: WaitTimedOutError) {
+            // TODO(messick) Check that pubspec is valid.
```
pq: 👍

pq: Maybe at least a log entry here in the meantime?

flutter-gui-tests/testSrc/io/flutter/tests/gui/ProjectCreator.kt
stevemessick: I wish. Sometimes the banner is delayed so long the catch is triggered. I suspect there is some way to cause all background tasks to complete but I haven't found it yet.

flutter-gui-tests/testSrc/io/flutter/tests/gui/ProjectCreator.kt
stevemessick: I probably should have done that. I'm writing on that code now so it will be handled in the next PR.

PR #3291, opened 2019-03-18
Provides a better display if the variable has a `toStringDeep()` method defined.

jingbian:
This is a similar piece of effort as [this one](https://github.com/JetBrains/intellij-plugins/pull/642) to IntelliJ Dart plugin.

### DIFF
--- a/src/io/flutter/server/vmService/frame/DartVmServiceValue.java
+++ b/src/io/flutter/server/vmService/frame/DartVmServiceValue.java
@@ -162,6 +162,7 @@ public class DartVmServiceValue extends XNamedValue {
       .evaluateInTargetContext(myIsolateId, myInstanceRef.getId(), "toStringDeep()", new VmServiceConsumers.EvaluateConsumerWrapper() {
         @Override
         public void received(final InstanceRef toStringInstanceRef) {
+          // TODO: Get actual DiagnosticsNode object instead of calling toStringDeep here.
           if (toStringInstanceRef.getKind() == InstanceKind.String) {
             String content = toStringInstanceRef.getValueAsString();
             int firstLineBreak = content.indexOf('\n');
--- a/src/io/flutter/server/vmService/frame/DartVmServiceValue.java
+++ b/src/io/flutter/server/vmService/frame/DartVmServiceValue.java
@@ -149,8 +149,42 @@ public class DartVmServiceValue extends XNamedValue {
     if (computeRegExpPresentation(node)) return;
     if (computeMapPresentation(node)) return;
     if (computeListPresentation(node)) return;
-    computeDefaultPresentation(node);
+
+    // computeDefaultPresentation is called internally here when no result is got.
+    // The reason for this is that the async method used cannot be properly waited.
+    computeDebugPresentation(node);
+
     // todo handle other special kinds: Type, TypeParameter, Pattern, may be some others as well
+  }
+
+  private void computeDebugPresentation(final XValueNode node) {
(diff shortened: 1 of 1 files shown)


src/io/flutter/server/vmService/frame/DartVmServiceValue.java
```diff
@@ -149,10 +149,44 @@ public void computePresentation(@NotNull final XValueNode node, @NotNull final X
...
+    // computeDefaultPresentation is called internally here when no result is got.
+    // The reason for this is that the async method used cannot be properly waited.
+    computeDebugPresentation(node);
+
     // todo handle other special kinds: Type, TypeParameter, Pattern, may be some others as well
   }
 
+  private void computeDebugPresentation(final XValueNode node) {
+    myDebugProcess.getVmServiceWrapper()
+      .evaluateInTargetContext(myIsolateId, myInstanceRef.getId(), "toStringDeep()", new VmServiceConsumers.EvaluateConsumerWrapper() {
```
jacob314: Add a TODO here to get the actual DiagnosticsNode objects instead of calling toStringDeep.

src/io/flutter/server/vmService/frame/DartVmServiceValue.java
```diff
@@ -149,10 +149,44 @@ public void computePresentation(@NotNull final XValueNode node, @NotNull final X
...
+        @Override
+        public void received(final InstanceRef toStringInstanceRef) {
+          if (toStringInstanceRef.getKind() == InstanceKind.String) {
```
jacob314: Add a TODO that this value may be truncated.

jingbian: I think the truncated case is handled around Line 170. Would you mind take a look whether that's right?

src/io/flutter/server/vmService/frame/DartVmServiceValue.java
jingbian: Added. Please take a look at the latest version.
Just curious. How big difference it will make to get the actual object?
I know you mentioned to me about the performance thing here, but I wonder how big is that and how useful is that compared to allowing people to implement a simple toStringDeep method.

jacob314: Getting the actual objects also lets you provide custom rendering of common property types. For example, based on the structured object view, the inspector window in intellij displays colors like colors and icons like icons.
Additionally, the structured view lets you expand and collapse subtrees for cases where an element has a large # of children.

src/io/flutter/server/vmService/frame/DartVmServiceValue.java
```diff
@@ -149,10 +149,45 @@ public void computePresentation(@NotNull final XValueNode node, @NotNull final X
     if (computeRegExpPresentation(node)) return;
     if (computeMapPresentation(node)) return;
     if (computeListPresentation(node)) return;
-    computeDefaultPresentation(node);
+
+    // computeDefaultPresentation is called internally here when no result is got.
```
jacob314: is got --> is received.

src/io/flutter/server/vmService/frame/DartVmServiceValue.java
jingbian: I see. Thanks!
Though it may be in another location of the code.

devoncarew:
I assume this is good to land? In which case, we should do that on behalf of @jingbian.

PR #3310, opened 2019-03-23
Support package:flutter_web as well as package:flutter libraries.

jacob314:
Here is the matching change for devtools:

https://github.com/flutter/devtools/commit/c458df0dd6254b0e89f86777479ac7da33dda713


src/io/flutter/inspector/InspectorService.java
```diff
@@ -710,15 +713,20 @@ DiagnosticsNode parseDiagnosticsNodeHelper(JsonElement jsonElement) {
         }));
     }
```
jacob314: I accidentally ran the code formatter. All the rest of the changes in this file are whitespace only.

jwren:
LGTM

PR #3316, opened 2019-03-26
Add inspector dependency to test

stevemessick:
@pq @devoncarew 

With these changes the integration test now is able to work with classes defined by the Flutter plugin.
- Refactor InspectorTest
- Add functionality to FlutterInspectorFixture and allow it to refer to plugin classes
- Change plugin architecture to use the Flutter plugin as a plugin (to enable above)
- Change plugin tool to create a valid plugin for testing (needs version tag)

There is a manual set-up step that is documented in build.gradle. It would be easy to add to the plugin tool, or just use a shell script.

### DIFF
--- a/flutter-gui-tests/build.gradle
+++ b/flutter-gui-tests/build.gradle
@@ -10,13 +10,16 @@ repositories {
   mavenLocal()
   mavenCentral()
   jcenter()
+  // unpack flutter-intellij.zip and rename the jar file to io.flutter.jar in this directory
+  flatDir dirs: ['../releases/release_master/2019.1']
 }
 
 dependencies {
   compile "org.jetbrains.kotlin:kotlin-reflect"
   testCompile group: 'junit', name: 'junit', version: '4.12'
+  testCompile fileTree(include: ['*.jar'], dir: '../releases/release_master/2019.1')
+  testImplementation fileTree(include: ['*.jar'], dir: '../releases/release_master/2019.1')
 }
-
 
 sourceSets {
   main {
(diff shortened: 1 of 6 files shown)


flutter-gui-tests/build.gradle
```diff
@@ -10,14 +10,17 @@ repositories {
...
   mavenCentral()
   jcenter()
+  // unpack flutter-intellij.zip and rename the jar file to io.flutter.jar in this directory
+  flatDir dirs: ['../releases/release_master/2019.1']
 }
 
 dependencies {
   compile "org.jetbrains.kotlin:kotlin-reflect"
   testCompile group: 'junit', name: 'junit', version: '4.12'
+  testCompile fileTree(include: ['*.jar'], dir: '../releases/release_master/2019.1')
```
jacob314: can we make 2019.1 a variable so we don't have to update it here when intellij gets a new release?

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
```diff
@@ -32,20 +33,30 @@ class InspectorTest : GuiTestCase() {
...
-      }, Timeouts.seconds10)
-      val inspector = flutterInspectorFixture()
-      inspector.activate()
-      pause(1000)
-      runner.stop()
+      launchFlutterApp()
+      val inspector = flutterInspectorFixture(this)
+      inspector.populate()
+      val widgetTree = inspector.widgetTreeFixture()
+      println("DEBUG")
```
jacob314: nit: strip these print statements before commit?

stevemessick:
I will make those changes in the next PR.

Issue #3096, opened 2019-01-17
Flutter Performance tool window shows up in non-Flutter projects


alexander-doroshko:
When I work with a pure Dart or pure Java project I don't expect to have Flutter Performance tool window available.

pq:
Good catch.  Thanks @alexander-doroshko!

jacob314:
This also applies to the Inspector window. Working on a quick fix.

PR #3320, opened 2019-03-27
Make Perf and Inspector views only display when a Flutter app is being debugged.

jacob314:
If the views were open and the flutter app is closed and then reopened,
the views are also automatically reopened.
Consistent with before, neither view is opened by default unless the users
opts to.

This fixes.
https://github.com/flutter/flutter-intellij/issues/3096

smaller_demo


jacob314:
Fixed a bug in where I was calling setAvailable.

PR #3321, opened 2019-03-27
Update build for M34

stevemessick:
@pq @devoncarew 

The AS GUI test framework is in flux. I disabled that source (`testSrc`) directory; there's no way to make it compile right now.

### DIFF
--- a/flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
+++ b/flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
@@ -65,7 +65,7 @@ public class FlutterProjectSystem implements AndroidProjectSystem {
     return gradleProjectSystem.getModuleSystem(module);
   }
 
-  @Override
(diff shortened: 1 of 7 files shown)


flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
```diff
@@ -65,7 +65,7 @@ public AndroidModuleSystem getModuleSystem(@NotNull Module module) {
     return gradleProjectSystem.getModuleSystem(module);
   }
 
-  @Override
+  @SuppressWarnings("override")
```
jacob314: I'm guessing this is an override only on one platform, right?

stevemessick: It is on several, but yeah, not all. This method got removed in the latest AS sources but is still required to compile older versions.

Issue #2538, opened 2018-08-08
NPE in DartVmServiceEvaluator


goodmang:
Please file this bug report at https://github.com/flutter/flutter-intellij/issues/new.

---

## What happened

(please describe what you were doing when this exception occurred)

## Version information

Android Studio `3.1.4` • Flutter plugin `io.flutter 27.1.1` • Dart plugin `173.4700`

Flutter 0.5.1 • channel beta • https://github.com/flutter/flutter.git
Framework • revision c7ea3ca377 (2 months ago) • 2018-05-29 21:07:33 +0200
Engine • revision 1ed25ca7b7
Tools • Dart 2.0.0-dev.58.0.flutter-f981f09760

## Exception

null

```
java.lang.NullPointerException
	at io.flutter.server.vmService.frame.DartVmServiceEvaluator.evaluate(DartVmServiceEvaluator.java:108)
	at com.intellij.xdebugger.impl.evaluate.quick.XValueHint.evaluateAndShowHint(XValueHint.java:186)
	at com.intellij.xdebugger.impl.evaluate.quick.common.AbstractValueHint.invokeHint(AbstractValueHint.java:154)
	at com.intellij.xdebugger.impl.evaluate.quick.common.ValueLookupManager.lambda$null$3(ValueLookupManager.java:190)
	at com.intellij.util.ui.UIUtil.invokeLaterIfNeeded(UIUtil.java:2795)
	at com.intellij.xdebugger.impl.evaluate.quick.common.ValueLookupManager.lambda$doShowHint$4(ValueLookupManager.java:182)
	at org.jetbrains.concurrency.AsyncPromise.setHandler(AsyncPromise.kt:201)
	at org.jetbrains.concurrency.AsyncPromise.done(AsyncPromise.kt:43)
	at com.intellij.xdebugger.impl.evaluate.quick.common.ValueLookupManager.doShowHint(ValueLookupManager.java:176)
	at com.intellij.xdebugger.impl.evaluate.quick.common.ValueLookupManager.lambda$showHint$1(ValueLookupManager.java:155)
	at com.intellij.psi.impl.PsiDocumentManagerBase.performWhenAllCommitted(PsiDocumentManagerBase.java:546)
	at com.intellij.psi.impl.PsiDocumentManagerBase.performWhenAllCommitted(PsiDocumentManagerBase.java:537)
[...]

PR #3324, opened 2019-03-27
Fix NPE issue sometimes hit evaluating expressions.

jacob314:
Matches https://github.com/JetBrains/intellij-plugins/commit/ce51ac028ff97ae0191b476e4d9a7bdecd629ca0

This fixes https://github.com/flutter/flutter-intellij/issues/2538

### DIFF
--- a/src/io/flutter/server/vmService/frame/DartVmServiceEvaluator.java
+++ b/src/io/flutter/server/vmService/frame/DartVmServiceEvaluator.java
@@ -20,6 +20,7 @@ import io.flutter.server.vmService.DartVmServiceDebugProcess;
 import com.jetbrains.lang.dart.psi.*;
 import com.jetbrains.lang.dart.util.DartResolveUtil;
 import gnu.trove.THashSet;
+import io.flutter.server.vmService.VmServiceWrapper;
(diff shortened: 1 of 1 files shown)


src/io/flutter/server/vmService/frame/DartVmServiceEvaluator.java
```diff
@@ -56,17 +57,23 @@ public void evaluate(@NotNull final String expression,
       callback.errorOccurred("Device disconnected");
       return;
     }
+    final VmServiceWrapper vmService = myDebugProcess.getVmServiceWrapper();
+    if (vmService == null) {
+      // Not connected to the VM yet.
+      callback.errorOccurred("No connection to the Dart VM");
```
stevemessick: This won't be visible to users, right?

jacob314: Not on hover. For hover if there is an error, no hover is displayed. This might be visible if you managed to hit this (race?) condition evaluating an expression in the watch window.
In any case, this is the same message the regular Dart IntelliJ plugin displays.

Issue #2135, opened 2018-04-24
Experiment: don't show a background square in the inspector summary tree


jacob314:
Explore not using the background (square) color changes, but instead of use the secondary selection highlight on nodes - the primary highlight would show for the selected node, and the secondary highlight for nodes that are the ones displayed in the details panel.

brianegan:
I like the idea behind this experiment as well -- I've found the highlighting a bit odd personally.

PR #3326, opened 2019-03-27
Cleanup inspector ui background

jacob314:


jacob314:
Screenshot of the cleaned up UI.
This fixes https://github.com/flutter/flutter-intellij/issues/2135

Screen Shot 2019-03-27 at 2 35 56 PM

src/io/flutter/view/InspectorTreeUI.java
```diff
@@ -67,8 +68,6 @@
 
   boolean leftToRight = true; /// TODO(jacobr): actually support RTL mode.
 
-  static final JBColor SUBTREE_BOUNDS_COLOR = new JBColor(Color.WHITE, Gray._43);
```
jacob314: The only thing that actually changed in this file is that SUBTREE_BOUNDS_COLOR was removed 
and 
the custom paint method in this class was removed.

Issue #3225, opened 2019-02-23
Highlight class names in light mode


InMatrix:
## Steps to Reproduce

I noticed that in the light theme, IntelliJ doesn't highlight Dart class names. This makes it harder to see the widget hierarchy, because widget names look similar to the "child" before them.

![image](https://user-images.githubusercontent.com/348942/53279582-8425b900-36c6-11e9-81c0-3f968fe396d9.png)

However, IntelliJ does highlight class names in the dark theme:

![image](https://user-images.githubusercontent.com/348942/53279593-9c95d380-36c6-11e9-9557-8aba1674e5ac.png)

So as in VS Code: 
![image](https://user-images.githubusercontent.com/348942/53279544-37da7900-36c6-11e9-9d5a-4d5c939c6d6d.png)

## Version info
[✓] Flutter (Channel dev, v1.2.2, on Mac OS X 10.14.1 18B75, locale en-US)
    • Flutter version 1.2.2 at /Users/taodong/Development/flutter/flutter
    • Framework revision 007a415c2a (21 hours ago), 2019-02-21 20:22:47 -0800
    • Engine revision f1f19bba8f
    • Dart version 2.2.0 (build 2.2.0-dev.2.1 c92d5ca288)

[✓] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
    • Android SDK at /Users/taodong/Library/Android/sdk
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-28, build-tools 28.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1248-b01)
    • All Android licenses accepted.

[✓] iOS toolchain - develop for iOS devices (Xcode 10.0)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 10.0, Build version 10A255
    • ios-deploy 2.0.0
    • CocoaPods version 1.5.3

[✓] Android Studio (version 3.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 33.1.1
    • Dart plugin version 182.5215
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1248-b01)

[✓] IntelliJ IDEA Community Edition (version 2018.3.4)
    • IntelliJ at /Applications/IntelliJ IDEA CE.app
[...]

PR #3327, opened 2019-03-27
Make Dart constructor calls pop out in light mode. 

jacob314:
Constructors are already nicely visible in Darkula mode.

Having this separate color style format will be handy for future additional UI as Code related
color tweaks. Here are screenshots of what this looks like in action. The Dark theme isn't
impacted but the light theme is.

Screen Shot 2019-03-27 at 3 53 15 PM
Screen Shot 2019-03-27 at 3 53 40 PM

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -909,6 +909,7 @@
     <projectStructureDetector implementation="io.flutter.project.FlutterProjectStructureDetector"/>
     <colorSettingsPage implementation="io.flutter.logging.FlutterLogColorPage"/>
     <additionalTextAttributes scheme="Default" file="colorSchemes/FlutterLogColorSchemeDefault.xml"/>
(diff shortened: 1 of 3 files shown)


jacob314:
This fixes
https://github.com/flutter/flutter-intellij/issues/3225

FYI @InMatrix in case there is a different color we should use.

Long term it would be nice but not crucial if this color default is added to the Dart Plugin but it makes sense to add it to the Flutter plugin so we can get quick feedback and iterate if need be.
For example, this color is quite aggressively light.

InMatrix:
Can we try this color: #2196f3. It's blue[500] in material design and it feels slightly more solid than the one used now.

jacob314:
That new color looks really nice and does make me wonder about creating a full material design colors only color scheme for Dart/Flutter. :)
Screen Shot 2019-03-29 at 11 04 17 AM

PR #3332, opened 2019-03-28
Defined fixed sample index

### LINKED ISSUE #3330: Creating a plugin sample fails confusingly
Repro steps:
Open IntelliJ 2019 CE
Create a new project selecting the card Sample
You get this error
Screen Shot 2019-03-28 at 12 58 46 PM
And then rather being prompted to select a different sample, you get an empty project with no content.
Screen Shot 2019-03-28 at 1 00 45 PM

Fyi @pq

stevemessick:
@pq @devoncarew @jacob314 

Fixes #3330 

This is a hack that restores the ability to generate sample projects. The index file at the URL is bad. I couldn't find it online but I got the content from the command line tool.

I'd be happy to close this without merging if someone wants to track down the proper URL.

With this new index we now see "Card" twice in the list. I think fixing that can wait, until the first patch or even M35. I want to have #3330 fixed for M34, though.


src/io/flutter/samples/FlutterSampleManager.java
```diff
@@ -38,7 +38,7 @@
 
 public class FlutterSampleManager {
 
-  private static final String SNIPPETS_REMOTE_INDEX_URL = "https://docs.flutter.io/snippets/index.json";
+  private static final String SNIPPETS_REMOTE_INDEX_URL = "BROKEN_https://docs.flutter.io/snippets/index.json";
```
jacob314: Can you clean this up? I think we may display this URL in the UI?
Also, does this patch work for Stable as well as master?

stevemessick: I'm pretty sure the URL isn't displayed.

I don't know if this works for stable.

After reading create.dart (flotter_tools) I'm thinking we might have to live with this problem for this release. The right way to fix it is to invoke `flutter create --list-samples`.

No one has hit this problem even though it has existed for weeks. I'm no longer convinced fixing it is important.

PR #3333, opened 2019-03-29
disables samples until creation issue is fixed

devoncarew:
- disables samples until creation issue is fixed

see https://github.com/flutter/flutter-intellij/issues/3330

From my understanding of the issue, the data about avialable samples we have in the flutter-intellij repo is out of date with the actual samples in the flutter/flutter repo.

@pq @stevemessick @jacob314


pq:
Maybe consider doing this in `io.flutter.samples.FlutterSampleManager#getSamples`?

```java
  public static List<FlutterSample> getSamples() {
    return Collections.emptyList();
  }
```

That will make sure the new project wizard doesn't see samples either...

devoncarew:
👍 makes sense, will move the check

PR #3315, opened 2019-03-25
Some initial work for FlutterWeb apps

jwren:

Issue #3293, opened 2019-03-18
P0: IntelliJ can create, launch, connect and inspect a FlutterWeb app (a webdev based workflow)


devoncarew:
@jwren, this covers a few more items then just launching the apps. We can add additional issues if we want more granularity.

Current status (as of March 27th):

- [x] Add code to determine if some Flutter project is a Flutter Web project, done with https://github.com/flutter/flutter-intellij/commit/aafadf8c2a32ff3ad6d89e49d6e5b45d2fd88dd2
- [x] Modify the flutter run configuration to work with a null device, if a FlutterWeb- done with https://github.com/flutter/flutter-intellij/pull/3315- [ ] Have the WebSocket connect, currently blocked by https://github.com/dart-lang/webdev/issues/233
- [x] After the WebSocket connection is established, test that any protocol calls that aren't supported (or ignored) aren't sent from Flutter IntelliJ
- [x] remove the `FlutterCommand.localWebDevExe` field and associated code, test that the command line from FlutterSdk.flutterRunWeb works, https://github.com/flutter/flutter-intellij/issues/3348
- [x] Stagehand support- test that the generated samples for FlutterWeb work in the Flutter IntelliJ plugin
- [ ] provision webdev on the user's machine
- [x] ensure that the inspector works

jwren:
https://github.com/flutter/flutter-intellij/pull/3315

jwren:
Current status (March 27th):

- [x] Add code to determine if some Flutter project is a Flutter Web project, done with https://github.com/flutter/flutter-intellij/commit/aafadf8c2a32ff3ad6d89e49d6e5b45d2fd88dd2
- [x] Modify the flutter run configuration to work with a null device, if a FlutterWeb- done with https://github.com/flutter/flutter-intellij/pull/3315- [ ] Have the WebSocket connect, currently blocked by https://github.com/dart-lang/webdev/issues/233
- [ ] After the WebSocket connection is established, test that any protocol calls that aren't supported (or ignored) aren't sent from Flutter IntelliJ
- [ ] After the webdev lands in the Dart SDK in the Flutter SDK, remove the `FlutterCommand.localWebDevExe` field and associated code, test that the command line from FlutterSdk.flutterRunWeb works.
- [ ] Stagehand support- test that the generated samples for FlutterWeb work in the Flutter IntelliJ plugin

devoncarew:
Awesome, thanks for the update and checklist!

jwren:
[...]

PR #3342, opened 2019-04-01
Take 2 on "Some initial work for FlutterWeb apps" change

jwren:
The fix to the `flutter run` regression is in DartVmServiceDebugProcess.java

Original, reverted, PR: https://github.com/flutter/flutter-intellij/pull/3315
Tracking issue: https://github.com/flutter/flutter-intellij/issues/3293

### DIFF
--- a/src/io/flutter/run/LaunchState.java
+++ b/src/io/flutter/run/LaunchState.java
@@ -121,7 +121,7 @@ public class LaunchState extends CommandLineState {
     final Project project = getEnvironment().getProject();
     final FlutterDevice device = DeviceService.getInstance(project).getSelectedDevice();
 
-    // If the device is null and the project is not a flutter web project, show a message that a device is required and return null.
+    // Flutter web does not yet support device objects yet, hence the null as the indicator that this might be a flutter web project.
     if (device == null) {
       boolean isFlutterWeb = false;
       final String filePath = ((SdkRunConfig)runConfig).getFields().getFilePath();
--- a/src/io/flutter/run/LaunchState.java
+++ b/src/io/flutter/run/LaunchState.java
@@ -45,6 +45,7 @@ import io.flutter.actions.RestartFlutterApp;
 import io.flutter.dart.DartPlugin;
 import io.flutter.logging.FlutterLog;
 import io.flutter.logging.FlutterLogView;
+import io.flutter.pub.PubRoot;
 import io.flutter.run.bazel.BazelRunConfig;
 import io.flutter.run.daemon.*;
 import org.jetbrains.annotations.NotNull;
@@ -120,9 +121,21 @@ public class LaunchState extends CommandLineState {
     final Project project = getEnvironment().getProject();
     final FlutterDevice device = DeviceService.getInstance(project).getSelectedDevice();
 
+    // If the device is null and the project is not a flutter web project, show a message that a device is required and return null.
     if (device == null) {
-      showNoDeviceConnectedMessage(project);
-      return null;
+      boolean isFlutterWeb = false;
+      final String filePath = ((SdkRunConfig)runConfig).getFields().getFilePath();
+      if (filePath != null) {
+        final MainFile main = MainFile.verify(filePath, project).get();
+        final PubRoot root = PubRoot.forDirectory(main.getAppDir());
+        if (root != null) {
+          isFlutterWeb = FlutterUtils.declaresFlutterWeb(root.getPubspec());
+        }
+      }
+      if (!isFlutterWeb) {
+        showNoDeviceConnectedMessage(project);
+        return null;
+      }
     }
     final FlutterApp app = myCreateAppCallback.createApp(device);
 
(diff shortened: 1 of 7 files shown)


src/io/flutter/sdk/FlutterCommand.java
```diff
@@ -48,13 +48,28 @@
   @NotNull
   private final List<String> args;
 
+  // TODO(github.com/flutter/flutter-intellij/issues/3293) This is a temporary "<pub-cache>/bin/webdev" that can be provided to test some
```
jacob314: please update these to bugs related to the problem being solved with the workaround not the feature being implemented.
its fine if the bug descriptions are short but that will make more sense to users viewing the TODO. For example, the bug you linked to will be closed but the TODOs will still exist. In general the bugs for the TODO should not be closed until the TODO is removed.

src/io/flutter/sdk/FlutterCommand.java
```diff
@@ -245,23 +260,57 @@ public OSProcessHandler startProcessOrShowError(@Nullable Project project) {
...
+      }
+      line.addParameters(type.subCommand);
+      line.addParameters(args);
+    }
+    return line;
+  }
 
-    line.setExePath(FileUtil.toSystemDependentName(sdk.getHomePath() + "/bin/" + FlutterSdkUtil.flutterScriptName()));
+  /**
+   * Creates a FlutterWeb command line to run.
```
jacob314: remove this comment as it doesn't add any value beyond the method name.

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -269,6 +269,18 @@ else if (flutterLaunchMode == FlutterLaunchMode.RELEASE) {
...
   }
 
+  public FlutterCommand flutterRunWeb(@NotNull PubRoot root, @NotNull RunMode mode) {
+    // flutter packages pub global run webdev serve [--debug] [--hot-reload]
+    final List<String> args = new ArrayList<>();
+    args.add("global");
+    args.add("run");
+    args.add("webdev");
+    args.add("daemon");
+    // TODO After debug is supported by webdev, this should be modified to check for debug and add any additional needed flags:
```
jacob314: add a bug related to this TODO or add your username to the todo.

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
```diff
@@ -262,17 +263,22 @@ public void scheduleConnect() {
...
-      try {
-        vmService = VmService.connect(url);
-        vmOpenSourceLocationListener = VmOpenSourceLocationListener.connect(url);
-      }
-      catch (IOException | RuntimeException e) {
-        onConnectFailed("Failed to connect to the VM observatory service at: " + url + "\n"
-                        + e.toString() + "\n" +
-                        formatStackTraces(e));
-        return;
+      // TODO(github.com/flutter/flutter-intellij/issues/3293, github.com/dart-lang/webdev/issues/233) The following check disables the
```
jacob314: same as other cases. add a bug for the TODO

jacob314: actually in this case only list the one bug related to the issue and not the bug describing the feature you are implementing.

src/io/flutter/run/LaunchState.java
```diff
@@ -120,9 +121,21 @@ protected RunContentDescriptor launch(@NotNull ExecutionEnvironment env) throws
     final Project project = getEnvironment().getProject();
     final FlutterDevice device = DeviceService.getInstance(project).getSelectedDevice();
 
+    // If the device is null and the project is not a flutter web project, show a message that a device is required and return null.
```
jacob314: instead of adding a comment, add an isFlutterWeb() method. That way it is self documenting at it will be clear when the hack needs to be removed.

jacob314: Nit: make the comment explain that flutter web doesn't yet support devices so this tweak is needed. As is the comment explains what happens which is visible from the code but not why that is a good thing.

src/io/flutter/sdk/FlutterCommand.java
jwren: Will do, I suppose it will make it more explicit, right now the issue, #3293, has a list of tasks associated with it... I will create some additional more explicit issues to track the specific tasks (4th bullet point, in this case.)

jwren: Done, https://github.com/flutter/flutter-intellij/issues/3348

src/io/flutter/sdk/FlutterCommand.java
jwren: done

src/io/flutter/sdk/FlutterSdk.java
jwren: done, https://github.com/flutter/flutter-intellij/issues/3349

src/io/flutter/server/vmService/DartVmServiceDebugProcess.java
jwren: done

src/io/flutter/run/LaunchState.java
jwren: done.

PR #3345, opened 2019-04-02
Define test_target repo for GUI testing

stevemessick:
@pq @devoncarew @jacob314 

Define the `test_target` local Gradle repository that contains the testable Flutter plugin built by the plugin tool. This automates the manual step introduced previously. (It also implements a couple suggestions @jacob314 made)

The `isTestTarget` flag is used to identify which builds in the product matrix are used for GUI testing. In a local build, only one will be kept if multiple plugin versions are built at once. Use the `-o` flag to choose which to build. On the CI system only one is built at a time (it always uses `-o`) so that won't be a concern.

The testing plugin requires the version to be SNAPSHOT so the `test_target` repo is not updated during a release build.


resources/META-INF/plugin.xml
```diff
@@ -13,7 +13,7 @@
 
   <category>Custom Languages</category>
   <version>SNAPSHOT</version>
-  <idea-version since-build="182.5107.16" until-build="192.SNAPSHOT"/>
+  <idea-version since-build="182.5107.16" until-build="191.*"/>
```
pq: Is this correct?

pq: (I mean, I'm surprised that we're bounding to 191.*.)

stevemessick: You're right! I should not have deleted the final build spec. That broke the debugger. Restored. PTAL.

Issue #3341, opened 2019-04-01
update the debugger to support the use of a service protocol url


devoncarew:
We should update the debugger to support the use of a service protocol url (instead of a VM service protocol port.

PR #3351, opened 2019-04-04
update how we manipulate the service protocol url

devoncarew:
- update how we manipulate the service protocol url
- this is in prep for handling urls in the form `http://xxx:yyy/token/` (and `ws://xxx:yyy/token/ws`)
- fix https://github.com/flutter/flutter-intellij/issues/3341


src/io/flutter/run/test/FlutterTestRunner.java
```diff
@@ -132,10 +133,11 @@ public void processWillTerminate(@NotNull ProcessEvent event, boolean willBeDest
     @Nullable
     @Override
     public String getWebSocketUrl() {
```
jacob314: nit: make this one static helper that gets a wsUrl from an observatoryUri used in both places.

devoncarew: 👍 will do

PR #3356, opened 2019-04-07
address reported Java lints

devoncarew:
- address reported Java lints
- address lints in java vm service library (as a separate commit, to simplify review)

The goal here is to move us slightly closer to being able to rely on these in our regular workflow in order to catch issues (like potential NPEs).


src/io/flutter/inspector/WidgetPerfTipsPanel.java
```diff
@@ -146,6 +146,7 @@ private void updateTip() {
         }
       }
 
+      // TODO(devoncarew): We never query changedEditors.
```
jacob314: go ahead and delete. This is code that should have been removed in a refactor.

devoncarew: done

PR #3362, opened 2019-04-08
Update for AS canary 10

stevemessick:
- Define a project type for Flutter projects
- Set it when a project is first opened (handles `flutter create ...`)
  - In Android Studio the default project type is "Android"
  - In IntelliJ the default is null
- Use it via a chameleon action, defined by XML,  to choose the proper Project Structure dialog
- Remove the hack that replaced the original with the custom action

@pq


stevemessick:
Oops! I forgot to update the plugin tool to work with the latest changes.

PR #3366, opened 2019-04-10
Adjust build to make plugin for testing

stevemessick:
Adjust build to produce a plugin suitable for integration testing.

### DIFF
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -26,7 +26,6 @@
       "name": "Android Studio 3.5",
       "version": "3.5",
       "ijVersion": "2019.1",
-      "isTestTarget": "true",
       "ideaProduct": "android-studio",
       "ideaVersion": "191.5416148",
       "dartPluginVersion": "191.6183.87",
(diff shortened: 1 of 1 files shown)


product-matrix.json
```diff
@@ -37,10 +36,11 @@
       "comments": "IntelliJ 2019.2 EAP",
       "name": "IntelliJ 2019.2 EAP",
       "version": "2019.2",
+      "isTestTarget": "true",
       "ideaProduct": "ideaIU",
       "ideaVersion": "2019.1",
       "dartPluginVersion": "191.4899",
-      "sinceBuild": "192.0",
+      "sinceBuild": "191.0",
```
devoncarew: I think we'll over overlap issue wrt the version ranges that match each plugin (2019.1 vs 2019.2). Is this something we'd want to land? Something to enable testing?

stevemessick:
I see your point. If this build was published it would collide with the
released version for 2019.1. But this one doesn't get published since the
EAP for 2019.2 isn't out yet. When the EAP is out this spec will be updated
to work with it. Meanwhile we need a 192.* version to work with the
debugger, and it also can be used with the integration tests. A bit hacky,
but only for internal use.

devoncarew:
Landing sounds good, but we'll just need to be careful to make the supported versions above disjoint before we next publish.

Related but separate from this PR, it would be nice to have some validation on the bots that our version ranges don't overlap.

PR #3374, opened 2019-04-16
Integration test update

stevemessick:
This includes a test for hot reload that uses the inspector to verify it actually worked. Pretty cool, I think :)

I added a bunch of `step` calls that wrap a block and log a message. Generally, if the first line of the diff starts with "step" then the only change is indentation.

@pq @jacob314

### DIFF
--- a/flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
+++ b/flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
@@ -20,6 +20,7 @@ import org.fest.swing.edt.GuiActionRunner.execute
 import org.fest.swing.edt.GuiQuery
 import org.fest.swing.fixture.JButtonFixture
 import org.fest.swing.timing.Condition
+import org.fest.swing.timing.Pause
 import org.fest.swing.timing.Pause.pause
 import org.junit.Assert.assertNotNull
 import org.junit.Test
@@ -32,11 +33,21 @@ class InspectorTest : GuiTestCase() {
   @Test
   fun importSimpleProject() {
     ProjectCreator.importProject()
+    println("DEBUG 1")
     ideFrame {
       launchFlutterApp()
+      println("DEBUG 2")
       val inspector = flutterInspectorFixture(this)
       inspector.populate()
+      println("DEBUG 3 $inspector")
       val widgetTree = inspector.widgetTreeFixture()
+      println("DEBUG 4 $widgetTree")
+      val inspectorTree = widgetTree.inspectorTreeFixture()
+      println("DEBUG 5 $inspectorTree")
+      inspectorTree.selectRow(1)
+      println("DEBUG 6 ${inspectorTree.selection()}")
+      Pause.pause(2000)
+      println("DEBUG 7 ${inspectorTree.selection()}")
       runner().stop()
     }
   }
@@ -70,8 +81,8 @@ class InspectorTest : GuiTestCase() {
     assertNotNull(actionToolbarContainer)
 
     // These next two lines look like the right way to select the simulator, but it does not work.
-    //    val comboBoxActionFixture = ComboBoxActionFixture.findComboBoxByText(robot(), actionToolbarContainer!!, "<no devices>")
-    //    comboBoxActionFixture.selectItem(devName)
+//    val comboBoxActionFixture = ComboBoxActionFixture.findComboBoxByText(robot(), actionToolbarContainer!!, "<no devices>")
+//    comboBoxActionFixture.selectItem(devName)
     // Need to get focus on the combo box but the ComboBoxActionFixture.click() method is private, so it is inlined here.
     val selector = button("<no devices>")
     val comboBoxButtonFixture = JButtonFixture(robot(), selector.target())
--- a/flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
+++ b/flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
@@ -13,35 +13,27 @@ import com.intellij.testGuiFramework.fixtures.IdeFrameFixture
 import com.intellij.testGuiFramework.framework.RunWithIde
 import com.intellij.testGuiFramework.framework.Timeouts
 import com.intellij.testGuiFramework.impl.GuiTestCase
-import com.intellij.testGuiFramework.impl.GuiTestUtilKt
-import com.intellij.testGuiFramework.impl.button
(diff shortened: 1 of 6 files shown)


flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
```diff
@@ -6,94 +6,163 @@
...
-    robot().pressAndReleaseKey(KeyEvent.VK_DOWN)
-    robot().pressAndReleaseKey(KeyEvent.VK_ENTER)
-    robot().waitForIdle()
+  private fun IdeFrameFixture.runner(): ExecutionToolWindowFixture.ContentFixture {
+    return runToolWindow.findContent("main.dart")
   }
 
-  private fun IdeFrameFixture.findActionButtonByActionId(actionId: String): ActionButtonFixture {
+  private fun findActionButtonByActionId(actionId: String): ActionButtonFixture {
+    // This seems to be broken, but finding by simple class name works.
```
pq: Bummer.  But I guess since we own all the classes under test, not a big deal?

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
```diff
@@ -6,94 +6,163 @@
...
+          return runner.isExecutionInProgress
+        }
+      }, Timeouts.seconds30)
+    }
   }
 
-  fun IdeFrameFixture.selectSimulator() {
-    selectDevice("Open iOS Simulator")
+  private fun findHotReloadButton(): ActionButtonFixture {
+    return findActionButtonByClassName("ReloadFlutterAppRetarget")
```
pq: To make this slightly more robust could we consider something like `ReloadFlutterAppRetarget.class.getName()` (or whatever it is in Java)?

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
```diff
@@ -6,94 +6,163 @@
...
+            return reload.isEnabled
+          }
+        }, Timeouts.seconds05)
+        step("Work around #3370") {
+          // https://github.com/flutter/flutter-intellij/issues/3370
+          inspector.renderTreeFixture().show() // The refresh button is broken so force tree update by switching views.
+          inspector.populate()
+          widgets.show() // And back to the one we want
+        }
+        widgetsTree.selectRow(6) // Text widget
```
pq: Out of curiosity, is there a more semantic way of identifying the row to select?

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
```diff
@@ -6,94 +6,163 @@
...
+        val key = if (isMac) KeyEvent.VK_BACK_SPACE else KeyEvent.VK_DELETE
+        for (n in 1..4) typeKey(key)
+        typeText("that")
+        //typeKey(KeyEvent.VK_ESCAPE) // Dismiss completion popup -- not needed with "that" but is needed with "so"
+      }
+
+      step("Trigger Hot Reload and wait for it to finish") {
+        val reload = findHotReloadButton()
+        reload.click()
+        editor.clickCenter() // Need to cycle the event loop to get button enabled on Mac.
```
pq: I bet this was fun to figure out. 😬

flutter-gui-tests/testSrc/io/flutter/tests/gui/fixtures/FlutterInspectorFixture.kt
```diff
@@ -54,16 +91,92 @@ class FlutterInspectorFixture(project: Project, robot: Robot, private val ideFra
...
+  // This fixture is used to access the tree for both Widgets and Render Objects.
+  inner class InspectorTreeFixture(private val inspectorTree: InspectorTree) {
+
+    fun treeFixture(): JTreeFixture {
+      return JTreeFixture(ideFrame.robot(), inspectorTree)
+    }
+
+    fun selectRow(number: Int, expand: Boolean = true) {
+      waitForContent()
+      treeFixture().clickRow(number) // This should not collapse the tree, but it does.
```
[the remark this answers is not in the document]
stevemessick: It looks like an off-by-one error. The click occurs exactly on the left-side boundary of the row bounding box. That rectangle apparently overlaps the bounding box of the triangle icon that expands and collapses the sub-tree. I don't think it is a framework error because I can, with sufficient care, reproduce the problem manually. OTOH if the click was in the center of the bounding box there'd be no problem, but that's a choice made by fest, a library used by the testing framework.

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
stevemessick: I'm not sure what the problem is. I haven't even ruled out pilot error (hmm, that may not be a great idiom these days). It might be a bug in the framework. I keep this around to remind myself not to use it.

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
stevemessick: Great idea. I'll look into that. Next PR?

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
stevemessick: I considered a named constant here, but didn't bother, mostly because all the other embedded constants don't really need names. The sample code isn't generated by Flutter, so it won't change unless we want it to. If we change it, this test will break, and the comment should at least point to a solution.

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
stevemessick: Oh yes, as someone with an extensive background in window testing would understand :)

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
pq: Sounds good.

flutter-gui-tests/testSrc/io/flutter/tests/gui/InspectorTest.kt
pq: Great!

Issue #3303, opened 2019-03-19
P1: Migrate Sample index to read from `flutter create`


pq:
Corresponding to: https://github.com/Dart-Code/Dart-Code/issues/1529.

pq:
See also: https://github.com/flutter/flutter/pull/28938.

PR #3379, opened 2019-04-18
read sample index from flutter_tool call

pq:
Fixes: #3303 .

/cc @stevemessick

### DIFF
--- a/src/io/flutter/samples/FlutterSample.java
+++ b/src/io/flutter/samples/FlutterSample.java
@@ -42,7 +42,23 @@ public class FlutterSample {
    * Get a label suitable for display in a chooser (e.g., combobox).
    */
   public String getDisplayLabel() {
-    // TODO(pq): add disambiguation once it's needed.
+    // TODO(pq): come up with disambiguated labels.
+    // TODO(pq): consider adding (package suffix too).
+    // Index isn't quite enough for disambiguation (there are still dups like DeletableChips).
+    //final String[] parts = id.split("\\.");
+    //final String lastPart = parts[parts.length-1];
+    //
+    //String suffix = "";
+    //try {
+    //  final int index = Integer.parseInt(lastPart);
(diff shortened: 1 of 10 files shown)


src/io/flutter/samples/FlutterSample.java
```diff
@@ -42,7 +42,23 @@ public String toString() {
...
+    //try {
+    //  final int index = Integer.parseInt(lastPart);
+    //  if (index != 1) {
+    //    suffix = " (" + index + ")";
+    //  }
+    //} catch (NumberFormatException e) {
+    //  // ignore
+    //}
+    //
+    //return getElement() + suffix;
```
stevemessick: I'm assuming you want to keep all the above in case it is useful solving the TODOs. If you just forgot about it, then here's a reminder to delete it. :)

src/io/flutter/samples/FlutterSampleManager.java
```diff
@@ -32,31 +36,112 @@
...
+        LOG.warn(e);
+        // On ExecutionException, short circuit checking again.
+        flutterSamples = Collections.emptyList();
+      }
     }
-    return SAMPLES;
+    catch (IOException e) {
+      LOG.warn(e);
+      // On IOException, short circuit checking again.
+      flutterSamples = Collections.emptyList();
```
stevemessick: If the network is unreliable, will the IDE have to be restarted to reset flutterSamples?

src/io/flutter/samples/FlutterSample.java
pq: Intentional.  But thanks!

src/io/flutter/samples/FlutterSampleManager.java
pq: I'm going to look into this some more.  There are a few wrinkles.

Thanks!

PR #3381, opened 2019-04-19
Event test

stevemessick:
@jacob314 

This makes `EventStreamTest` runnable, although two tests fail. Previously, the two tests would cause it to hang, so I added a timer to fail the test.

I think the two failing tests (`duplicateValues()` and `nullInitialValue()`) don't run correctly because in unit test mode `AsyncUtils.invokeLater()` uses the UI thread. That is called by `EventStream.setValue()`.


jacob314:
lgtm

PR #3388, opened 2019-04-19
Fix MainFileTest

stevemessick:
@pq 

This is a hack. But the test runs. The failure had nothing to do with any test code we have control over. I think it is due to something that was added to the Kotlin test framework, but I don't know for sure.

### DIFF
--- a/testSrc/unit/io/flutter/testing/AdaptedFixture.java
+++ b/testSrc/unit/io/flutter/testing/AdaptedFixture.java
@@ -5,6 +5,8 @@
  */
 package io.flutter.testing;
 
+import com.intellij.openapi.diagnostic.ExceptionWithAttachments;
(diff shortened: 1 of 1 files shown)


stevemessick:
This also fixes `FlutterSdkUtilTest.shouldInstallFlutterSDK()`

pq:
Neat!  Is there any way we could make sure we're only ignoring the desired runtime exception?  Or maybe that's not worth it?

stevemessick:
It is a private class. I used it's super class.

pq:
Got it.  If you think this might miss genuine failures, you could do something to check the class name but that's totally up to you.  Thanks for following up!

stevemessick:
Thanks for the tip. I'll do that in a follow-up PR.

PR #3389, opened 2019-04-20
more work to support running Flutter web apps

devoncarew:
Various work in support of https://github.com/flutter/flutter-intellij/issues/3293:
- ensure that the hot reload action is disabled for flutter web app runs
- disable hot reload on save for flutter web app runs
- fix a bug where some text (like `"Build: Running build completed, took 191ms"`) was erroneously hyperlinked to in the run console
- change to connecting to service protocol apps via urls (not ports)
- remove no longer used 'restart recommended...' text
- fix an issue where a device was being associated with a flutter web launch; this would cause the iOS simulator to be brought to the front when launching flutter web apps
- fix an issue where already running flutter web apps weren't reloaded when the run button was hit
- add support for a new `daemon.log` event, to better support showing output from running flutter web apps
- refactor the initial flutter web support - add a new `FlutterCommand.FLUTTER_WEB_RUN` command

Things that are not yet supported:
- there's a race condition - which we generally lose - which means that the inspector is not aware of running flutter web apps
- we do not yet provision webdev locally for users (we rely on a manually pub globally activated version existing)

Both of the above issues will need to be resolved for flutter web support (not necessarily in this PR).

### DIFF
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -106,16 +106,16 @@ public class DevToolsManager {
   }
 
   public void openBrowser() {
-    openBrowserImpl(-1);
+    openBrowserImpl(null);
   }
 
-  public void openBrowserAndConnect(int port) {
-    openBrowserImpl(port);
+  public void openBrowserAndConnect(String uri) {
+    openBrowserImpl(uri);
   }
 
-  private void openBrowserImpl(int port) {
+  private void openBrowserImpl(String uri) {
     if (devToolsInstance != null) {
-      devToolsInstance.openBrowserAndConnect(port);
+      devToolsInstance.openBrowserAndConnect(uri);
       return;
     }
 
@@ -133,7 +133,7 @@ public class DevToolsManager {
     DevToolsInstance.startServer(project, sdk, pubRoots.get(0), instance -> {
       devToolsInstance = instance;
 
-      devToolsInstance.openBrowserAndConnect(port);
+      devToolsInstance.openBrowserAndConnect(uri);
     }, instance -> {
       // Listen for closing, null out the devToolsInstance.
       devToolsInstance = null;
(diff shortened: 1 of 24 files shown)


src/io/flutter/run/LaunchState.java
```diff
@@ -119,25 +119,31 @@ protected RunContentDescriptor launch(@NotNull ExecutionEnvironment env) throws
...
-          isFlutterWeb = FlutterUtils.declaresFlutterWeb(root.getPubspec());
-        }
-      }
-      if (!isFlutterWeb) {
-        showNoDeviceConnectedMessage(project);
-        return null;
+    @Nullable FlutterDevice device = DeviceService.getInstance(project).getSelectedDevice();
+
+    boolean isFlutterWeb = false;
+    final String filePath = ((SdkRunConfig)runConfig).getFields().getFilePath();
```
jwren: Do you need to normalize here to ensure windows is supported, or has that already happened in getFilePath?

devoncarew:
Some updates:

- fix an npe in the inspector (to do with null devices)
- fix an npe in the perf view (to do with null devices)
- handle exceptions from the service protocol when trying to get the source location of widgets

With the latest updates, the inspector now works.

src/io/flutter/run/LaunchState.java
devoncarew: This just restructures the code on the left to remove a level of nesting - it's fundamentally the same as the previous code.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -214,9 +216,15 @@ public void openBrowserAndConnect(String serviceProtocolUri) {
...
     else {
-      // TODO: do we need to url encode uri?
+      String urlParam = serviceProtocolUri;
+      try {
+        urlParam = URLEncoder.encode(serviceProtocolUri, "UTF-8");
+      }
+      catch (UnsupportedEncodingException ignored) {
+      }
+
       BrowserLauncher.getInstance().browse(
```
jwren: Do you want `browse()` to be called even if there is a caught exception in the try block above?  Reading the code here makes it look like `browse()` here is supposed to be inside the try-block.

src/io/flutter/inspector/DiagnosticsNode.java
```diff
@@ -834,7 +842,8 @@ public boolean identicalDisplay(DiagnosticsNode node) {
         }
         group.safeWhenComplete(future, action);
       });
-    } catch (Exception e) {
+    }
+    catch (Exception e) {
       // Nothing to do if the service can't be acquired.
     }
```
jwren: e -> ignored for IntelliJ inspections.

devoncarew: done

src/io/flutter/devtools/DevToolsManager.java
devoncarew: good point; this shouldn't happen - the UTF-8 encoding should be available - but I'll move the browser call up to within the try block in any case

PR #3401, opened 2019-04-23
Build for canary 12

stevemessick:
@devoncarew @pq

### DIFF
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -1,17 +1,7 @@
 {
   "list": [
     {
-      "comments": "AS 3.3.2",
-      "name": "Android Studio 3.3.2",
-      "version": "3.3.2",
-      "ideaProduct": "android-studio",
-      "ideaVersion": "182.5314842",
-      "dartPluginVersion": "182.5215",
-      "sinceBuild": "182.5107.16",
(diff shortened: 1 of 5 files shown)


product-matrix.json
```diff
@@ -22,14 +12,14 @@
...
+      "comments": "IntelliJ 2019.1.1, AS 3.5 canary",
       "name": "Android Studio 3.5",
       "version": "3.5",
-      "ijVersion": "2019.1",
+      "ijVersion": "2019.1.1",
       "ideaProduct": "android-studio",
-      "ideaVersion": "191.5416148",
-      "dartPluginVersion": "191.6183.87",
-      "sinceBuild": "191.5416148",
+      "ideaVersion": "191.5471097",
```
devoncarew: Is rev'ing this intentional (and, will it leave behind users)? We should just verify what version people are using internally (2019.1, or 2019.1.1).

stevemessick: That's the build artifact, and it is based on 2019.1.

stevemessick:
PTAL. There were some version mis-matches that prevented the plugin from loading in some products.

stevemessick:
TBR, to get the build going.

PR #3407, opened 2019-04-24
validate that our supported platform version ranges don't overlap

devoncarew:
- validate that our supported platform version ranges don't overlap

In this PR, we check that the upper bounds of an earlier supported platform doesn't overlap with the lower bounds of the next one.

I'm not 100% on what we were using `SNAPSHOT` for - it looks like it was temporary in order to support a recent EAP?


devoncarew:
Thanks for the feedback! Let's talk in person to see if this PR is solving the right problems.

PR #3408, opened 2019-04-24
remove the option to disable the memory view from the settings

devoncarew:
- remove the option to disable the memory view from the settings

@terrylucas

### DIFF
--- a/src/io/flutter/settings/FlutterSettings.java
+++ b/src/io/flutter/settings/FlutterSettings.java
@@ -12,9 +12,7 @@ import com.intellij.openapi.util.registry.Registry;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.util.EventDispatcher;
 import io.flutter.analytics.Analytics;
-import io.flutter.bazel.Workspace;
 import io.flutter.sdk.FlutterSdk;
-import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
 import java.util.EventListener;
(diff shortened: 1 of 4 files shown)


src/io/flutter/settings/FlutterSettings.java
```diff
@@ -32,8 +30,6 @@
   private static final String legacyTrackWidgetCreationKey = "io.flutter.trackWidgetCreation";
   private static final String disableTrackWidgetCreationKey = "io.flutter.disableTrackWidgetCreation";
   private static final String useFlutterLogView = "io.flutter.useLogView";
-  private static final String memoryProfilerKey = "io.flutter.memoryProfiler";
-  private static final String newBazelTestRunnerKey = "io.flutter.bazel.legacyTestBehavior";
```
terrylucas: Something else that's outside of memory profiler?

src/io/flutter/view/FlutterPerfView.java
```diff
@@ -207,7 +206,7 @@ private void addPerformanceViewContent(FlutterApp app, ToolWindow toolWindow) {
 
       addPerformanceTab(runnerTabs, app, toolWindow, true);
 
-      if (FlutterUtils.isAndroidStudio() && !FlutterSettings.getInstance().isMemoryProfilerDisabled()) {
```
terrylucas: Should we silently remove the memory tab? Or would you like to wait until we either launch DevTool automatically or a link on how to use DevTools for profiling?

src/io/flutter/settings/FlutterSettings.java
devoncarew: yup, this was an unused field (from code removed by a previous PR)

src/io/flutter/view/FlutterPerfView.java
devoncarew: We should improve the way we launch devtools slightly. Perhaps open an issue to remove the memory view for the next release?

PR #3411, opened 2019-04-24
provision webdev for flutter_web projects

devoncarew:
- provision webdev for flutter_web projects
- fix https://github.com/flutter/flutter-intellij/issues/3293
- add a `WebDevManager` manager class to provision `webdev`
- on startup for flutter web projects, check if there's a flutter sdk, and if so provision webdev
- allow the user to pass additional args into `webdev daemon` in the launch configs (similar to what we do for `flutter run`)
- show webdev (and devtools) installation progress in IntelliJ (instead of using the `Messages` tool window)


src/io/flutter/devtools/WebDevManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Manage installing the webdev cli.
+ */
+public class WebDevManager {
+  public static WebDevManager getInstance(@NotNull Project project) {
+    return ServiceManager.getService(project, WebDevManager.class);
+  }
+
+  private final Project project;
```
jwren: `@NotNull` can be put on this field.

devoncarew: done

PR #3414, opened 2019-04-24
Add flag to skip the dart analyzer error check before a hot reload

DaveShuckerow:
Currently, if the analysis server detects any errors in the project, it will not perform a hot reload.
This can cause hot reload to stop working needlessly when, eg, there is an analysis error in a test file.
 
Take two on #3181.

### DIFF
--- a/src/io/flutter/FlutterBundle.properties
+++ b/src/io/flutter/FlutterBundle.properties
@@ -228,4 +228,4 @@ settings.open.inspector.on.launch=Open Flutter Inspector view on app launch
 settings.hot.reload.on.save=Perform hot reload on save
 settings.disable.tracking.widget.creation=Disable tracking widget creation locations
 settings.enable.bazel.test.runner=Enable new Bazel test runner
-settings.hot.reload.with.error=Perform hot reload even if there are analysis errors in the project
+settings.hot.reload.with.error=Perform hot reload even if there are analysis errors
--- a/src/io/flutter/FlutterBundle.properties
+++ b/src/io/flutter/FlutterBundle.properties
@@ -228,3 +228,4 @@ settings.open.inspector.on.launch=Open Flutter Inspector view on app launch
 settings.hot.reload.on.save=Perform hot reload on save
 settings.disable.tracking.widget.creation=Disable tracking widget creation locations
 settings.enable.bazel.test.runner=Enable new Bazel test runner
+settings.hot.reload.with.error=Perform hot reload even if there are analysis errors in the project
--- a/src/io/flutter/run/FlutterReloadManager.java
+++ b/src/io/flutter/run/FlutterReloadManager.java
@@ -190,6 +190,16 @@ public class FlutterReloadManager {
         return;
       }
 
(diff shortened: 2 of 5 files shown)


src/io/flutter/FlutterBundle.properties
```diff
@@ -228,3 +228,4 @@ settings.open.inspector.on.launch=Open Flutter Inspector view on app launch
 settings.hot.reload.on.save=Perform hot reload on save
 settings.disable.tracking.widget.creation=Disable tracking widget creation locations
 settings.enable.bazel.test.runner=Enable new Bazel test runner
+settings.hot.reload.with.error=Perform hot reload even if there are analysis errors in the project
```
devoncarew: Perhaps, `Perform hot reload even if there are analysis errors`?

src/io/flutter/run/FlutterReloadManager.java
```diff
@@ -190,14 +190,16 @@ private void handleSaveAllNotification(@Nullable Editor editor) {
         return;
       }
 
-      if (hasErrors(app.getProject(), app.getModule(), editor.getDocument())) {
+      // If the analysis server detects any errors in the project, it will not perform a hot reload.
+      // This can cause hot reload to stop working needlessly when, eg, there is an analysis error in a test file.
+      // The reloadWithError option in settings is a workaround.
+      if (hasErrors(app.getProject(), app.getModule(), editor.getDocument()) && !mySettings.isReloadWithError()) {
```
devoncarew: Is there an issue tracking a better fix (in flutter_tools)? If so, we could in-line the reference here.

src/io/flutter/sdk/FlutterSettingsConfigurable.form
```diff
@@ -196,13 +196,22 @@
...
               <toolTipText value="Allows the Flutter Inspector to track source locations where widgets are created. This setting is required for advanced Flutter Inspector functionality and performance tools."/>
             </properties>
           </component>
+          <component id="ff6b2" class="javax.swing.JCheckBox" binding="myHotReloadIgnoreErrorCheckBox">
+            <constraints>
+              <grid row="1" column="0" row-span="1" col-span="2" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+            </constraints>
+            <properties>
+              <text resource-bundle="io/flutter/FlutterBundle" key="settings.hot.reload.with.error"/>
+              <toolTipText value="Hot reload changes into running Flutter apps even with analysis errors."/>
```
devoncarew: `Hot reload changes into running Flutter apps even if there are analysis errors`?

DaveShuckerow: Done.

src/io/flutter/run/FlutterReloadManager.java
DaveShuckerow: Done.

src/io/flutter/FlutterBundle.properties
DaveShuckerow: Done.

PR #3434, opened 2019-04-30
Create "Editor" section of Flutter Settings

jacob314:
Moved the UI Guides settings to this section along with the format on save settings.
This lets us use the shorter "UI Guides" name while still making it clear they are an Editor feature instead of some separate tool window. The UI Guides are still unchecked by default for this release with a plan to turn them on by default next release.
Screen Shot 2019-04-30 at 12 58 59 PM


jacob314:
Screenshot of the entire settings menu.
Screen Shot 2019-04-30 at 12 57 09 PM

jacob314:
At first I did have just such a layout bug but I fixed it. The problem was I'd missed updating the spacer needed at the end of the list to keep the final item in the list from being stretched.

PR #3435, opened 2019-04-30
Setting to ignore pubspec.yaml in bazel projects

DaveShuckerow:
This makes the Flutter plugin obey the existing registry setting "dart.projects.without.pubspec" from the Dart plugin.

### DIFF
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.form
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.form
@@ -51,7 +51,7 @@
           </component>
         </children>
       </grid>
-      <grid id="e885c" layout-manager="GridLayoutManager" row-count="5" column-count="2" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
+      <grid id="e885c" layout-manager="GridLayoutManager" row-count="6" column-count="2" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
         <margin top="0" left="0" bottom="0" right="0"/>
         <constraints>
           <grid row="2" column="0" row-span="1" col-span="1" vsize-policy="3" hsize-policy="3" anchor="0" fill="3" indent="0" use-parent-layout="false"/>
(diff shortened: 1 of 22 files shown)


src/io/flutter/settings/FlutterSettings.java
```diff
@@ -217,6 +227,25 @@ public void setOpenInspectorOnAppLaunch(boolean value) {
...
+    fireEvent();
+  }
+
+  /**
+   * Determines whether bazel should be used for a project based on the user's preferences and the structure of the project.
+   */
+  public boolean shouldUseBazel(Project project) {
+    final boolean isBazelProject = FlutterModuleUtils.isFlutterBazelProject(project);
+    final boolean isNotPubProject = PubRoots.forProject(project).isEmpty();
+    return isBazelProject && (useBazelByDefault() || isNotPubProject);
```
jwren: Why not just `return Registry.is(ignorePubspecRegistryKey, false);`  Is there some usecase where this complexity will help?

src/io/flutter/settings/FlutterSettings.java
```diff
@@ -26,6 +30,9 @@
   private static final String syncAndroidLibrariesKey = "io.flutter.syncAndroidLibraries";
   private static final String disableTrackWidgetCreationKey = "io.flutter.disableTrackWidgetCreation";
   private static final String useFlutterLogView = "io.flutter.useLogView";
+  // The Dart plugin uses this registry key to keep bazel users from getting their settings overridden on projects that include a
+  // pubspec.yaml.
+  private static final String ignorePubspecRegistryKey = "dart.projects.without.pubspec";
```
jwren: Rename to something like `DART_PROJECTS_WITHOUT_PUBSPEC_KEY`, the current name makes it sound like you are going to ignore the value stored in `dart.projects.without.pubspec`.

DaveShuckerow: Good idea, done.

src/io/flutter/settings/FlutterSettings.java
DaveShuckerow: Yes, this is useful:

 - If we ask for the value of the setting in a pub-only project, then we should not attempt to use bazel.
 - Conversely, if this is not set in a bazel-only project, then we should still always use bazel.

src/io/flutter/sdk/FlutterSettingsConfigurable.form
```diff
@@ -88,6 +88,15 @@
               <toolTipText value="Enables verbose logging (this can be useful for diagnostic purposes)."/>
             </properties>
           </component>
+          <component id="984f6" class="javax.swing.JCheckBox" binding="myUseBazelByDefaultCheckBox">
+            <constraints>
+              <grid row="3" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+            </constraints>
+            <properties>
+              <text value="Use Bazel instead of Pub by default"/>
```
devoncarew: Do we need a UI setting for this? Should this be controlled on a per app basis, or just via the registry setting?

devoncarew:
Dave, why don't you land this first, and then I can land https://github.com/flutter/flutter-intellij/pull/3436 (we'll have merge conflicts with these two PRs).

src/io/flutter/sdk/FlutterSettingsConfigurable.form
DaveShuckerow: I think ideally, we will be able to determine this and other sensible defaults for Bazel at startup time and apply them in bulk.

Until adding that, this is useful.  I filed #3437 to do this, and I'll submit this change for now.

PR #3438, opened 2019-05-01
Add an option to turn off closing labels in Dart source code when UI guides are on.

jacob314:
The option is enabled by default as UI Guides serve a lot of the same purpose as the closing tags.

We have analytics for when users turn the closing labels back on so we will be able to see how many users actually want both options at the same time. With very large build methods it could definitely make sense to want both options.

Settings dialog with one additional option:
When "Turn off closing labels in Dart source code when UI guides are on" is checked, checking the "UI Guides" checkbox toggles the closing labels checkbox in the separate settings panel. When the box is unchecked, checking the UI Guides checkbox has no impact.
Screen Shot 2019-04-30 at 6 32 07 PM

### DIFF
--- a/src/io/flutter/editor/WidgetIndentsHighlightingPassFactory.java
+++ b/src/io/flutter/editor/WidgetIndentsHighlightingPassFactory.java
@@ -26,6 +26,7 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiFile;
+import com.jetbrains.lang.dart.analyzer.DartClosingLabelManager;
 import io.flutter.FlutterUtils;
 import io.flutter.dart.FlutterDartAnalysisServer;
 import io.flutter.dart.FlutterOutlineListener;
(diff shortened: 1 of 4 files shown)


src/io/flutter/editor/WidgetIndentsHighlightingPassFactory.java
```diff
@@ -26,6 +26,7 @@
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiFile;
+import com.jetbrains.lang.dart.analyzer.DartClosingLabelManager;
```
devoncarew: based on this PR building through, it looks like this class exists in all Dart plugin versions we support

src/io/flutter/sdk/FlutterSettingsConfigurable.form
```diff
@@ -179,9 +179,17 @@
               <toolTipText value="When this option is checked lines are drawn on the scrollbar indicating where build methods are."/>
             </properties>
           </component>
+          <component id="23423483" class="javax.swing.JCheckBox" binding="myDisableDartClosingLabels">
+            <constraints>
+              <grid row="3" column="0" row-span="1" col-span="1" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="2" use-parent-layout="false"/>
+            </constraints>
+            <properties>
+              <text value="Turn off closing labels in Dart source code when UI guides are on"/>
```
devoncarew: `Hide closing labels...`?

stevemessick: Either "hide" or "do not display"

jacob314: Switched to hide.
Done

PR #3440, opened 2019-05-01
Release 35.1

stevemessick:
@pq @devoncarew @jacob314 

This has all the cherry picks. It also has the change log update. And, because I had already started on a branch named release_35.1 I modified the plugin tool to allow the ".1" in the branch name.


stevemessick:
To be clear, this PR is not meant to be merged. There are conflicts with PR's in master that are not included here.

devoncarew:
lgtm

after we publish - given the size of the dot release - we should get people to try it out

Issue #3444, opened 2019-05-02
add a toast promoting the flutter user study


pq:
target dates TBD.  might need to be in an M35 dot release...

pq:
(Targeting a dot-release.)

PR #3459, opened 2019-05-06
Q2 survey notifcation

pq:
Fixes: #3444.

/cc @DanTup @stevemessick @devoncarew

### DIFF
--- a/src/io/flutter/survey/FlutterSurveyNotifications.java
+++ b/src/io/flutter/survey/FlutterSurveyNotifications.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright 2019 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.survey;
+
+import com.intellij.ide.BrowserUtil;
+import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationType;
+import com.intellij.notification.Notifications;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.fileEditor.FileEditorManager;
+import com.intellij.openapi.fileEditor.FileEditorManagerEvent;
+import com.intellij.openapi.fileEditor.FileEditorManagerListener;
+import com.intellij.openapi.project.Project;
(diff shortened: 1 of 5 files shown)


src/io/flutter/survey/FlutterSurveyNotifications.java
```diff
@@ -0,0 +1,158 @@
...
+    this.myProject = project;
+  }
+
+  private void checkForDisplaySurvey() {
+    // Limit display to the survey window.
+    if (!isSurveyOpen()) return;
+
+    final PropertiesComponent properties = PropertiesComponent.getInstance();
+
+    // Don't prompt more often than every 40 hours.
```
stevemessick: 40 hours seems a bit random. Why not 48, or 168?

src/io/flutter/survey/FlutterSurveyNotifications.java
```diff
@@ -0,0 +1,158 @@
...
+
+      private void check(@NotNull VirtualFile file) {
+        if (PubRoot.isPubspec(file) || file.getFileType() == DartFileType.INSTANCE) {
+          new FlutterSurveyNotifications(project).checkForDisplaySurvey();
+        }
+      }
+    });
+  }
+
+  @NotNull final Project myProject;
```
stevemessick: If I open two projects will I get notified twice?

DanTup:
The logic seems good to me (and IIUC it matches what I've done in VS Code), but I'm not all that familiar with Java or this codebase, so it's worth having a second opinion too :-)

src/io/flutter/survey/FlutterSurveyNotifications.java
DanTup: Heh, that's my fault :D I started using 20 hours for things like this so that they won't show again "the same day", but will tend to appear at the start of the day rather than end if you'd first seen them at the end of the day. We wanted around two days with this one, and I lazily doubled it, and Phil copied it.

No real thought went into it though - if anyone has a preference I don't mind tweaking it in VS Code :-)

stevemessick: That's as good a reason as any :)

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: Nope.   That state is cached in a global property.  (Thanks for asking!)

pq:
Thanks @DanTup!  Added the missing file. 😬 

Cheers @stevemessick for the quick review!

PR #3468, opened 2019-05-07
remove a println in WidgetIndentsHighlightingPass.java

devoncarew:
- remove a System.out.println in WidgetIndentsHighlightingPass.java

### DIFF
--- a/src/io/flutter/editor/WidgetIndentsHighlightingPass.java
+++ b/src/io/flutter/editor/WidgetIndentsHighlightingPass.java
@@ -6,6 +6,7 @@
 
 package io.flutter.editor;
 
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.*;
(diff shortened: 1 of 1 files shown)


src/io/flutter/editor/WidgetIndentsHighlightingPass.java
```diff
@@ -748,7 +751,7 @@ private RangeHighlighter createHighlighter(MarkupModel mm, TextRangeDescriptorPa
     final TextRange range = entry.range;
     final FlutterSettings settings = FlutterSettings.getInstance();
     if (range.getEndOffset() >= myDocument.getTextLength() && DEBUG_WIDGET_INDENTS) {
-      System.out.println("Warning: highlighter extends past the end of document.");
+      LOG.info("Warning: highlighter extends past the end of document.");
```
jacob314: This bug shouldn't impact users as `DEBUG_WIDGET_INDENTS  == false`

Issue #3478, opened 2019-05-08
remove the preview option in the outline view


devoncarew:
This experiment is not generally used, and I think we get good mileage from the UI Guides feature, and building your UI up progressively via hot reload.

PR #3481, opened 2019-05-08
remove the preview view from the flutter outline view

devoncarew:
- remove the preview view from the flutter outline view
- fix https://github.com/flutter/flutter-intellij/issues/3478

@scheglov for fyi (and happy to discuss in person as well)

We can remove the `flutter.getChangeAddForDesignTimeConstructor` call in the analysis server api in a month or two, after the next milestone of the Flutter IntelliJ plugin is released.


scheglov:
LGTM

PR #3487, opened 2019-05-10
Fix errors in AS 3.5 beta 1

### LINKED ISSUE #3473: Problem creating projects in Android Studio 3.5 beta 1
We've lost the ability to customize the project in the third page of the new project wizard. It is just blank.
Screen Shot 2019-05-07 at 6 27 38 PM

The Android Studio devs have been adding then deleting some debugging code that causes this problem. Up until now, that debugging code has been reverted before making a beta release, and even prior to canary releases before 10. I was hoping the problem would not appear in the beta, but here we are.

stevemessick:
This fixes a number of problems that surfaced with Android Studio 3.5 beta 1.

1. The third page of the new project wizard is visible now.
2. The File menu has _New Flutter Project_ next to _New Project_ again.
3. The Welcome screen has the proper label for creating Flutter projects.
4. The Welcome screen has a spiffy new icon for Flutter.
5. The plugin tool allows resource files to be editing during the build (for # 2).

Here's the new icon, in context:
Screen Shot 2019-05-10 at 3 40 13 PM

Fixes #3473 

@pq @devoncarew

### DIFF
--- a/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
+++ b/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
@@ -9,14 +9,22 @@ package io.flutter.actions;
 import com.android.tools.idea.ui.wizard.StudioWizardDialogBuilder;
 import com.android.tools.idea.wizard.model.ModelWizard;
 import com.android.tools.idea.wizard.model.ModelWizardDialog;
+import com.intellij.icons.AllIcons;
 import com.intellij.openapi.actionSystem.AnAction;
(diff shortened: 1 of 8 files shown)


stevemessick:
Well, I'm not exactly sure if it needs to be updated. The script doesn't specifically ask people to check that the menu items are in the right place, or that the labels are correct, or that the wizard pages all show. It leaves that to the tester, I guess. If it had to call out every single thing it would be too verbose, and people would start skipping. Still, maybe we should add something to remind people that little things can change and should be noted.

Issue #3477, opened 2019-05-08
remove the memory usage view


devoncarew:
This is better supported by Dart DevTools now. We should still show general memory awareness, just shell out to DevTools in order to do diagnostics.

cc @terrylucas

devoncarew:
I've started on some of the work here.

PR #3497, opened 2019-05-17
Remove the android studio specific memory view

devoncarew:
- remove the android studio specific memory view (@terrylucas)
- fix https://github.com/flutter/flutter-intellij/issues/3477
- refactor the `Performance` view in light of the above removable
  - show three tabs in the view for FPS, memory, and widget rebuilds
  - use the older memory chart in order to provide basic memory usage awareness

We'll want to make if clear how to open devtools from the Performance view in order for users to dig into performance issues (IntelliJ can be used for basic awareness); that work is not in this PR.

Screen Shot 2019-05-16 at 7 48 01 PM

Screen Shot 2019-05-16 at 7 47 51 PM

Screen Shot 2019-05-16 at 7 47 44 PM


src/io/flutter/view/PerfMemoryTab.java
```diff
@@ -0,0 +1,53 @@
...
+    this.app = app;
+
+    setLayout(new BorderLayout());
+    setBorder(JBUI.Borders.empty(3));
+
+    final JPanel memoryPanel = new JPanel(new BorderLayout());
+    final JPanel heapDisplay = HeapDisplay.createJPanelView(parentDisposable, app);
+    memoryPanel.add(heapDisplay, BorderLayout.CENTER);
+    memoryPanel.setBorder(BorderFactory.createTitledBorder(BorderFactory.createEtchedBorder(), "Memory usage"));
+
```
jacob314: Can we add text to open in the devtools here even if we don't hook it up till a followup cl so we don't forget.

devoncarew: Actually, I updated the PR to have a hyperlink here to open devtools.

devoncarew:
Updated to add hyperlinks to open devtools in the FPS and memory tabs:

Screen Shot 2019-05-17 at 8 32 42 AM

PR #3505, opened 2019-05-20
Make the save dialog refer to the save all action, not the save action

DaveShuckerow:
We had a user who was confused when he ran 'save document' and didn't get a hot reload.


src/io/flutter/FlutterBundle.properties
```diff
@@ -219,7 +219,7 @@ settings.format.code.on.save=Format code on save
 settings.organize.imports.on.save=Organize imports on save
 settings.flutter.version=Version:
 settings.open.inspector.on.launch=Open Flutter Inspector view on app launch
-settings.hot.reload.on.save=Perform hot reload on save
+settings.hot.reload.on.save=Perform hot reload on Save All
```
devoncarew: Can we instead keep this as `save` (saving generally, and not the specific IntelliJ action), and call out the `Save All` action in tooltip text?

DaveShuckerow: SGTM.

src/io/flutter/sdk/FlutterSettingsConfigurable.form
```diff
@@ -121,7 +121,7 @@
             </constraints>
             <properties>
               <text resource-bundle="io/flutter/FlutterBundle" key="settings.hot.reload.on.save"/>
-              <toolTipText value="On save, hot reload changes into running Flutter apps."/>
+              <toolTipText value="On save all, hot reload changes into running Flutter apps."/>
```
devoncarew: `On a 'Save All' action, hot reload...`?

DaveShuckerow: Done.

PR #3507, opened 2019-05-21
Add an inline run menu option to run or watch Bazel Flutter tests

DaveShuckerow:
Image: image

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -872,6 +872,7 @@
 
     <configurationType implementation="io.flutter.run.bazelTest.FlutterBazelTestConfigurationType"/>
     <runConfigurationProducer implementation="io.flutter.run.bazelTest.BazelTestConfigProducer"/>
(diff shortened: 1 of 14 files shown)


devoncarew:
In `BazelTestConfigProducer`, should we change the name we set to `Run + config.getName()`? That would parallel the `Watch counter_test.dart` menu item in the picture above. I'm not 100% sure of this, as we might also need to make the returned name dependent on the execution mode (run vs debug). Thoughts? I suppose we could just use `Run ...` for both run and debug.

DaveShuckerow:
> In `BazelTestConfigProducer`, should we change the name we set to `Run + config.getName()`? That would parallel the `Watch counter_test.dart` menu item in the picture above. I'm not 100% sure of this, as we might also need to make the returned name dependent on the execution mode (run vs debug). Thoughts? I suppose we could just use `Run ...` for both run and debug.

That sounds good to me.  I also took the opportunity to make the names easier to read when the test name or file name is long and the configuration name gets truncated.

PR #3509, opened 2019-05-21
Configure default options for running Bazel

DaveShuckerow:
Add a section to settings with a button bazel users can press to configure everything they need.

image

@jwren You have a plugin with similar settings, so I'd like to make sure that I submit this in a state that's consistent with what you have.


jwren:
Sorry for the delay here.

I like that it has its own group below the other settings now.

It feels strange to have a button and a check box.  It feels that the checkbox should be sufficient.  Users might and will accidentally press that button, potentially mis- configuring their workspace if they are an external user.

On startup, the FlutterInitializer should observe the checkbox setting, modifying the workspace as needed.  If the user modifies the setting in the dialog (in either direction) appropriate actions should be made, right now it looks like only one direction (enabling) is handled.

When it is enabled, I believe that we need to also delete output from `pub *` actions, and then restart the analysis server, especially if there were files that were generated.

DaveShuckerow:
Thoughts inline, then my proposal at the bottom.

> It feels strange to have a button and a check box. It feels that the checkbox should be sufficient. Users might and will accidentally press that button, potentially mis- configuring their workspace if they are an external user.

I set this section up so that it only shows when the user is inside of a Bazel workspace.  The button turns on additional settings, like preconfiguring the Dart SDK/Flutter SDK paths to save time for bazel mac-based users.

> On startup, the FlutterInitializer should observe the checkbox setting, modifying the workspace as needed. If the user modifies the setting in the dialog (in either direction) appropriate actions should be made, right now it looks like only one direction (enabling) is handled.

There are two pieces here that are reversible here: the registry setting for suggesting multiple run configurations, and the setting for using bazel or pub by default.  I can add a toggle box for suggesting multiple configurations.

The other pieces don't have a well-defined pub default: It makes sense to enable the Dart SDK for both pub and bazel projects.  We don't know where the Dart SDK lives by default for a non-bazel project, so we can't set it by default for a pub user.

> When it is enabled, I believe that we need to also delete output from `pub *` actions, and then restart the analysis server, especially if there were files that were generated.

That's an interesting idea.

How about this:

It seems like I'm putting multiple separate features in with this single PR.  I will simplify this to just provide the bazel section with two configurable flags, one for each user.  In follow-ups, we can add other features like customizing the SDK path by default or auto-deleting output from pub actions.

I want to keep the focus of this PR small.  My goal is to land this in M36 so that we can make it easier to see all run configurations for a context (https://github.com/flutter/flutter-intellij/pull/3507 uses this feature).

PR #3511, opened 2019-05-23
Enable devtools launching from Bazel

DaveShuckerow:
Involves setting up a Bazel-compatible FlutterSDK and delegating to it.

Because the Bazel SDK does not surface the same behavior as the pub-based SDK, I made this an explicit step that we have to take when it is necessary.

### DIFF
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -30,6 +30,7 @@ import io.flutter.sdk.FlutterSdk;
 import io.flutter.settings.FlutterSettings;
 import io.flutter.utils.JsonUtils;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
@@ -65,8 +66,9 @@ public class DevToolsManager {
     }
 
     final CompletableFuture<Boolean> result = new CompletableFuture<>();
-    // We don't need a pubroot to call pub global.
-    final FlutterCommand command = sdk.flutterPackagesPub(null, "global", "activate", "devtools");
+    // TODO(https://github.com/flutter/flutter/issues/33324): We shouldn't need a pubroot to call pub global.
+    @Nullable final PubRoot pubRoot = PubRoots.forProject(project).stream().findFirst().orElse(null);
+    final FlutterCommand command = sdk.flutterPackagesPub(pubRoot, "global", "activate", "devtools");
 
     final ProgressManager progressManager = ProgressManager.getInstance();
     progressManager.run(new Task.Backgroundable(project, "Installing DevTools...", true) {
@@ -162,8 +164,9 @@ class DevToolsInstance {
     Callback<DevToolsInstance> onSuccess,
     Callback<DevToolsInstance> onClose
   ) {
-    // We don't need a pubroot to call pub global.
-    final FlutterCommand command = sdk.flutterPackagesPub(null, "global", "run", "devtools", "--machine", "--port=0");
+    // TODO(https://github.com/flutter/flutter/issues/33324): We shouldn't need a pubroot to call pub global.
+    @Nullable final PubRoot pubRoot = PubRoots.forProject(project).stream().findFirst().orElse(null);
+    final FlutterCommand command = sdk.flutterPackagesPub(pubRoot, "global", "run", "devtools", "--machine", "--port=0");
 
     // TODO(devoncarew): Refactor this so that we don't use the console to display output - this steals
     // focus away from the Run (or Debug) view.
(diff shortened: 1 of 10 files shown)


src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -165,11 +161,11 @@ private void openBrowserImpl(String uri, String page) {
   public static void startServer(
     Project project,
     FlutterSdk sdk,
-    PubRoot pubRoot,
     Callback<DevToolsInstance> onSuccess,
     Callback<DevToolsInstance> onClose
   ) {
-    final FlutterCommand command = sdk.flutterPackagesPub(pubRoot, "global", "run", "devtools", "--machine", "--port=0");
+    // We don't need a pubroot to call pub global.
```
devoncarew: good catch

src/io/flutter/sdk/FlutterCommand.java
```diff
@@ -254,7 +254,7 @@ public GeneralCommandLine createGeneralCommandLine(@Nullable Project project) {
     if (workDir != null) {
       line.setWorkDirectory(workDir.getPath());
     }
-    if (!isDoctorCommand()) {
+    if (!isDoctorCommand() && !(sdk instanceof FlutterSdk.BazelSdk)) {
```
devoncarew: I don't know why the existing code is conditionally passing in `--no-color` (and only not passing it in for the doctor command). I wonder if we couldn't just remove the condition and always pass in `--no-color`.

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -568,4 +578,83 @@ public void onTextAvailable(@NotNull ProcessEvent event, @NotNull Key outputType
...
+    /**
+     * The Bazel workspace for this project.
+     */
+    @NotNull final Workspace workspace;
+
+    @NotNull final Project project;
+
+    private BazelSdk(@NotNull Project project, @NotNull Workspace workspace) {
+      super(
+        Objects.requireNonNull(workspace.getRoot().findFileByRelativePath("mobile/flutter/tools/ide/gflutter")),
```
devoncarew: Should we move these paths to `dart/config/intellij-plugins/flutter.json` (and loaded from io.flutter.bazel.Workspace.java`)?

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -56,18 +59,15 @@ public boolean hasInstalledDevTools() {
   }
 
   public CompletableFuture<Boolean> installDevTools() {
-    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
+    final FlutterSdk sdk =
+      FlutterSettings.getInstance().shouldUseBazel() ? FlutterSdk.forBazel(project) : FlutterSdk.getFlutterSdk(project);
```
devoncarew: This expression appears twice:

```
FlutterSettings.getInstance().shouldUseBazel() ? FlutterSdk.forBazel(project) : FlutterSdk.getFlutterSdk(project)
```

Should we instead in-line this into `FlutterSdk`? I'm wondering if it's going to be rare or common. `FlutterSdk.getFlutterOrBazelSdk(project)`? I'm not sure there's an easy name for the function.

DaveShuckerow: I can pull this out into a separate method, however, we need to only use the bazelsdk in places where we are sure we want to: without changes to the internal bazel tooling, the only command that will work right now is the packages pub command.

devoncarew: Gotcha. Having duplicates make sense if there are only a few (in this case 2), and if we do want to be explicit about what we're doing. It sounds like in this case we do - that a bazel flutter sdk is not a general replacement for a flutter sdk.

src/io/flutter/sdk/FlutterCommand.java
DaveShuckerow: Doctor command output becomes very hard to read when IntelliJ passes the --no-color flag.

See gpaste/5549831424573440 for a sample.

The Flutter bazel commands do not all accept --no-color.  The only one that does right now is the test command.

I'll leave this section as-is for now, and if we want to tell IJ to not color doctor command output in the future, we can file an issue for it.

DaveShuckerow: I filed https://github.com/flutter/flutter/issues/33321 about this.

src/io/flutter/devtools/DevToolsManager.java
DaveShuckerow: Interestingly, on paper we don't need it, but the pubspec.yaml check in the Flutter tool is too aggressive:

https://github.com/flutter/flutter/issues/33324

src/io/flutter/sdk/FlutterSdk.java
DaveShuckerow: Yes.  Done.

PR #3513, opened 2019-05-23
Change DeviceDaemon to show a detailed error when it fails too many times.

DaveShuckerow:


DaveShuckerow:
A slow daemon startup does not trigger the warning.  I specifically tested the case where Flutter has to rebuild / refetch its dependencies after an upgrade.

I looked into just showing the warning, logging the exception, and stopping, but I noticed that this would create two notifications at once, and one or more would stop showing if a third one came up.  I decided that showing first that an error occurred and then giving the user actionable information about what is wrong worked better.

Another option would be to add the error to the IntelliJ EventLog instead of popping up a notification, but I think this failure mode is unlikely enough of an occurrence that switching between different implementations of the behavior would be unproductive churn.

This PR works properly, so I'm going to submit for now. We can make additional behavior changes if need be in follow-up PRs.

PR #3520, opened 2019-05-27
remove survey notifications

pq:
/cc @devoncarew @stevemessick


pq:
There likely will be more but my thinking was we could just add a variation on this back if and when.

Having said that, if you'd prefer I can add a comment and just remove the call to initialize leaving this as a breadcrumb for the future?

Issue #3514, opened 2019-05-24
Error detected in Android Studio after running a prog on iPhone.


donaldww:
What happened

Sending a program to run directly on an iPhone, via USB

## Version information

Android Studio `3.4.1` • Flutter plugin `io.flutter 35.3.1` • Dart plugin `183.6270`

Flutter 1.5.4-hotfix.2 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 7a4c33425d (3 weeks ago) • 2019-04-29 11:05:24 -0700
Engine • revision 52c7a1e849
Tools • Dart 2.3.0 (build 2.3.0-dev.0.5 a1668566e5)

## Exception

null

```
java.lang.NullPointerException
	at io.flutter.editor.NativeEditorNotificationProvider.findRootDir(NativeEditorNotificationProvider.java:82)
	at io.flutter.editor.NativeEditorNotificationProvider.createNotificationPanel(NativeEditorNotificationProvider.java:45)
	at io.flutter.editor.NativeEditorNotificationProvider.createNotificationPanel(NativeEditorNotificationProvider.java:24)
	at com.intellij.ui.EditorNotificationsImpl$3.performInReadAction(EditorNotificationsImpl.java:136)
	at com.intellij.openapi.progress.util.ReadTask.lambda$runBackgroundProcess$0(ReadTask.java:66)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:977)
	at com.intellij.openapi.application.ReadAction.compute(ReadAction.java:57)
	at com.intellij.openapi.progress.util.ReadTask.runBackgroundProcess(ReadTask.java:66)
	at com.intellij.openapi.progress.util.ProgressIndicatorUtils.lambda$runUnderProgress$4(ProgressIndicatorUtils.java:227)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$3(CoreProgressManager.java:180)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$2(CoreProgressManager.java:164)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:582)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:532)
[...]

PR #3524, opened 2019-05-28
guard against null project basedir

pq:
Fixes: #3514.

(Still trying to figure out why the basedir would be `null` but crashing on banner creation isn't good regardless 😬)

/cc @stevemessick @devoncarew

### DIFF
--- a/src/io/flutter/editor/NativeEditorNotificationProvider.java
+++ b/src/io/flutter/editor/NativeEditorNotificationProvider.java
@@ -76,7 +76,8 @@ public class NativeEditorNotificationProvider extends EditorNotifications.Provid
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/editor/NativeEditorNotificationProvider.java
```diff
@@ -77,6 +77,9 @@ else if (root.getName().equals("ios")) {
   }
 
   private static VirtualFile findRootDir(@NotNull VirtualFile file, VirtualFile projectDir) {
+    if (projectDir == null) {
```
devoncarew: 👍 

Perhaps also mark the param as `@Nullable` (and the method return as well)?

pq: Good call.  Done.

PR #3558, opened 2019-06-07
Revise Android Studio version down 

DaveShuckerow:
This makes the plugin compatible with the g3 version of Android Studio


stevemessick:
This change looks fine. You'll need to do `bin/plugin generate` and commit those changes to get a clean build. After that, LGTM.

DaveShuckerow:
Done.

PR #3560, opened 2019-06-07
update the plugin tool to run tests via gradle

devoncarew:
- update the plugin tool to run tests via `./gradlew test`

@stevemessick


stevemessick:
You might want to check that JAVA_HOME is set. Gradle blows up if it isn't.

devoncarew:
> You might want to check that JAVA_HOME is set. Gradle blows up if it isn't.

👍 will do

PR #3564, opened 2019-06-10
Re-enable tests to run on Travis

DaveShuckerow:


tool/travis.sh
```diff
@@ -4,9 +4,6 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-# Fast fail the script on failures.
-set -e
```
stevemessick: Need to keep this. Drop "set +e" and "return 0" at line 44, 46.

DaveShuckerow: Done.

PR #3566, opened 2019-06-10
Integration test update

stevemessick:
Various things relating to integration tests. With this, we have a stable platform from which real work can get done.

- Inspector tests pass.
- Perf tests started but not working yet.
- Use Java 11 for flutter-gui-tests module to keep Gradle sync working.
- Update build numbers for 2019.2.
- Clean up module definition.

The Java 11 bit is surprisingly important and hard to track down. Gradle sync just fails with Java 8, and there are a whole bunch of comments about that on the net. Lots of cargo-cult techniques to fix it; I tried them all, and sometimes they did work. Switching to Java 11 just eliminated the problem. You can get Java 11 bundled with IntelliJ 2019.2 EAP.

@jacob314 @devoncarew

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,10 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel>
-      <module name="flutter-gui-tests.main" target="1.8" />
-      <module name="flutter-gui-tests.test" target="1.8" />
-    </bytecodeTargetLevel>
     <resourceExtensions />
     <wildcardResourcePatterns>
       <entry name="!?*.java" />
@@ -22,5 +18,9 @@
         <processorPath useClasspath="true" />
       </profile>
     </annotationProcessing>
+    <bytecodeTargetLevel>
+      <module name="flutter-gui-tests.main" target="11" />
+      <module name="flutter-gui-tests.test" target="11" />
+    </bytecodeTargetLevel>
   </component>
 </project>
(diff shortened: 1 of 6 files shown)


devoncarew:
> Gradle sync just fails with Java 8, and there are a whole bunch of comments about that on the net. Lots of cargo-cult techniques to fix it; I tried them all, and sometimes they did work. Switching to Java 11 just eliminated the problem. You can get Java 11 bundled with IntelliJ 2019.2 EAP.

I assume that java 11 is needed to perform the build, but the plugin will continue to work on older javas (like the one for 2018.3)?

.idea/compiler.xml
```diff
@@ -22,5 +18,9 @@
         <processorPath useClasspath="true" />
       </profile>
     </annotationProcessing>
+    <bytecodeTargetLevel>
+      <module name="flutter-gui-tests.main" target="11" />
```
devoncarew: Can we target java 11 if we still support 2018.3? I'm not sure what jvm it runs on.

stevemessick:
Thanks for the review. The Flutter plugin is not changing here. The flutter-gui-tests plugin is separate from the Flutter plugin, and it is what needs to use Java 11. Most people won't even see it as a plugin, since it has to be loaded separately when setting up the project. When the integration tests run, IntelliJ loads both the Flutter plugin and the flutter-gui-tests plugin.

devoncarew:
> Thanks for the review. The Flutter plugin is not changing here. The flutter-gui-tests plugin is separate from the Flutter plugin, and it is what needs to use Java 11.

Ah, gotcha, makes sense.

PR #3574, opened 2019-06-13
fix an offset issue with the UI guides code

devoncarew:
- fix an offset issue with the UI guides code (we could get startOffsets that were greater than endOffsets)
- fix https://github.com/flutter/flutter-intellij/issues/3572

@jacob314 

I'm less familiar with this code; this fix will avoid the error, but I don't know if we should back the fix up to the calling code.

### DIFF
--- a/src/io/flutter/editor/OutlineLocation.java
+++ b/src/io/flutter/editor/OutlineLocation.java
@@ -105,15 +105,15 @@ public class OutlineLocation implements Comparable<OutlineLocation> {
 
     final int delta = Math.max(column - indent, 0);
     final int markerEnd = offset;
+
     // Create a range marker that goes from the start of the indent for the line
     // to the column of the actual entity.
     final int startOffset = Math.max(markerEnd - delta, 0);
-    final int endOffset = Math.min(markerEnd + 1, document.getTextLength());
+    int endOffset = Math.min(markerEnd + 1, document.getTextLength());
+    endOffset = Math.max(startOffset, endOffset);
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/editor/OutlineLocation.java
```diff
@@ -96,19 +96,24 @@ public void dispose() {
...
+    final int delta = Math.max(column - indent, 0);
     final int markerEnd = offset;
     // Create a range marker that goes from the start of the indent for the line
-    // to the column of the actual entity;
-    marker = document.createRangeMarker(max(markerEnd - delta, 0), min(markerEnd + 1, document.getTextLength()));
-    nodeStartingWord = getCurrentWord(document, markerEnd);
+    // to the column of the actual entity.
+    final int startOffset = Math.max(markerEnd - delta, 0);
+    final int endOffset = Math.min(markerEnd + 1, document.getTextLength());
+
```
jacob314: consider instead making
endOffset = max(startOffset, endOffset)
or
flagging the outline location as invalid and avoiding displaying it in the UI guides.
The current option will fix the crash but would have the unintended consequence that we use an untracked location for getColumn and related methods even though `track` has been called. This could result in nonsense UI guides after the user makes a couple local edits.

devoncarew: OK, makes sense. I updated this code to make sure that `startOffset` isn't past the end of the file (which I think was the underlying problem w/ the reported bug). And, we don't conditionally create the range marker now.

PR #3577, opened 2019-06-13
Perf view

stevemessick:
More clean up of integration tests. They were super flakey, and not usable for a while. This makes it so the tests pass almost always. The random `pause()` calls are not ideal and I'm looking for a way to be more deterministic.

@jacob314 @devoncarew

### DIFF
--- a/flutter-gui-tests/testSrc/io/flutter/tests/gui/PerfTest.kt
+++ b/flutter-gui-tests/testSrc/io/flutter/tests/gui/PerfTest.kt
@@ -6,6 +6,7 @@
 
 package io.flutter.tests.gui
 
+import com.intellij.testGuiFramework.fixtures.ActionButtonFixture
 import com.intellij.testGuiFramework.fixtures.ExecutionToolWindowFixture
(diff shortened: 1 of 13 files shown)


flutter-gui-tests/testSrc/io/flutter/tests/gui/PerfTest.kt
```diff
@@ -28,24 +31,10 @@ class PerfTest : GuiTestCase() {
...
-  }
+      pause()
 
-  // TODO Share with InspectorTest
-  private fun IdeFrameFixture.runner(): ExecutionToolWindowFixture.ContentFixture {
-    return runToolWindow.findContent("main.dart")
+      runner().stop()
+    }
   }
 }
```
jacob314: nit: have you run the kotlin formatter on these files? I would expect it would fix the missing trialing newline, etc.

stevemessick: Yes, I did

stevemessick:
Finally green

Issue #3591, opened 2019-06-17
Add IDE to DevTools querystring


DanTup:
When launching DevTools, the `ide` querystring variable should be set for better stats reported.

VS Code change that does this is here:

https://github.com/Dart-Code/Dart-Code/commit/52b1c3cbf13838314e2dd9b0bf631d53956628a7

Might be slightly different here if you're not using launchDevTools though.

PR #3592, opened 2019-06-17
Add IDE query param to DevTools url.

kenzieschmoll:
Adds "ide=IntelliJ" or "ide=AndroidStudio" to the DevTools url upon launch. Fixes https://github.com/flutter/flutter-intellij/issues/3591.

Also removed the `hide=debugger` param from the `serviceProtocolUrl == null` case to be consistent with what we do when the vm service uri is not null (the else path in this if statement). Discussion for this [here](https://github.com/flutter/devtools/issues/731). 
@jacob314 @terry @DanTup

### DIFF
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -23,6 +23,7 @@ import com.intellij.openapi.util.Key;
 import com.intellij.openapi.vfs.VirtualFile;
 import io.flutter.bazel.Workspace;
 import io.flutter.console.FlutterConsoles;
+import io.flutter.FlutterUtils;
 import io.flutter.pub.PubRoot;
 import io.flutter.pub.PubRoots;
 import io.flutter.sdk.FlutterCommand;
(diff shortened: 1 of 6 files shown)


src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -224,16 +225,20 @@ public void processTerminated(@NotNull ProcessEvent event) {
   }
 
   public void openBrowserAndConnect(String serviceProtocolUri, String page) {
+    final String ideParam = "ide=" + (FlutterUtils.isAndroidStudio() ? "AndroidStudio&" : "IntelliJ&");
```
jacob314: why the trailing `&`?
I will instead joint the "ideParams" with '&'

DanTup:
The `ide` change LGTM, though I'm not familiar with the reasoning for `hide=debugger` previously only being in one code path (if anything, it seems like the wrong one... I would expect you'd want to avoid overlapping debuggers when you have a VM service, but maybe not when just launching it as a standalone app?).

I've added some notes to the other issue about why we had the debugger for VS Code anyway, I'm not sure if it applies for IntelliJ/AS.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -224,16 +225,20 @@ public void processTerminated(@NotNull ProcessEvent event) {
...
+      );
     }
     else {
       try {
         final String urlParam = URLEncoder.encode(serviceProtocolUri, "UTF-8");
         final String pageParam = page == null ? "" : ("#" + page);
 
         BrowserLauncher.getInstance().browse(
-          "http://" + devtoolsHost + ":" + devtoolsPort + "/?uri=" + urlParam + pageParam,
+          "http://" + devtoolsHost + ":" + devtoolsPort + "/?" + ideParam + "uri=" + urlParam + pageParam,
```
jacob314: create a list of params and join instead of manually adding `&`
That will avoid bugs where we end up with extra of missing `&`

Add a unit test for the URL generation code. It should be pretty easy to add a test for this which would catch the trailing `&` case.

src/io/flutter/devtools/DevToolsManager.java
kenzieschmoll: done.

PR #3599, opened 2019-06-19
Treat FlutterApp as a long-running process

DaveShuckerow:
- switch it to a mostly silent process
- refactor the mostly silent process handler into a shared util

Justification: Flutter apps that take a long time to start up in debug mode (for example, they have native logic that they invoke before the Flutter app itself starts) can make IntelliJ upset.

### DIFF
--- a/src/io/flutter/run/daemon/DeviceDaemon.java
+++ b/src/io/flutter/run/daemon/DeviceDaemon.java
@@ -25,6 +25,7 @@ import io.flutter.bazel.WorkspaceCache;
 import io.flutter.sdk.FlutterSdk;
 import io.flutter.sdk.FlutterSdkUtil;
 import io.flutter.utils.FlutterModuleUtils;
+import io.flutter.utils.MostlySilentOsProcessHandler;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/daemon/DeviceDaemon.java
```diff
@@ -203,17 +204,8 @@ DeviceDaemon start(Supplier<Boolean> isCancelled,
...
-         * of the device daemon process (this also address a log message in the IntelliJ log).
-         */
-        @NotNull
-        @Override
-        protected BaseOutputReader.Options readerOptions() {
-          return BaseOutputReader.Options.forMostlySilentProcess();
-        }
-      };
+      // The mostly silent process handler reduces CPU usage of the daemon process.
+      final ProcessHandler process = new MostlySilentOsProcessHandler(toCommandLine());
```
devoncarew: Do we think that this reduces the CPU usage by reducing the frequency of polling, or by doing something like running a process at a lower priority?

DaveShuckerow: I am not sure; this is the explanation that the original comment & IntelliJ's warnings gave.

Looking at the implementation for the two readers, it looks like the mostly silent process configuration reduces the polling frequency; I didn't see any reference to process priority.

DaveShuckerow: I've updated the comments with a more detailed explanation of how this works.

devoncarew: > Looking at the implementation for the two readers, my understanding is the mostly silent process configuration reduces the polling frequency

Thanks for investigating. Out of curiosity, what are the two polling frequencies?

DaveShuckerow: The normal mode will run blocking, waiting for all input from the process before it continues.

The mostly silent process will read whatever data is available when it polls, sleeping every 1 or 5 seconds in between.

Implementation is here: https://github.com/JetBrains/intellij-community/blob/4c8ecb4e7d3e908182527a127a70ebe4f4f7804d/platform/util/src/com/intellij/util/io/BaseDataReader.java#L90

PR #3608, opened 2019-06-21
Update installation for libraries

stevemessick:
Unit tests can now use mockito through the PowerMock libraries. The instructions are updated to include installation of them.

You'll need to do this if you haven't already.

@kenzieschmoll @jacob314 @terrylucas @pq @devoncarew


stevemessick:
I forgot to say, thanks @kenzieschmoll for setting this up :)

PR #3609, opened 2019-06-21
make the area for the flutter version smaller

devoncarew:
- reduce the screen area we use to display the Flutter version in the preferences view. We've added several new preferences recently - this change will make it less likely for the Flutter preference page to need to scroll.
- also, add a 'copy tp clipboard' button for the version text

Screen Shot 2019-06-21 at 1 08 19 PM

### DIFF
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.java
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.java
@@ -335,6 +335,7 @@ public class FlutterSettingsConfigurable implements SearchableConfigurable {
   private void onVersionChanged() {
     final FlutterSdk sdk = FlutterSdk.forPath(getSdkPathText());
     if (sdk == null) {
+      // Clear the label out with a non-empty string, so that the layout doesn't give this element 0 height.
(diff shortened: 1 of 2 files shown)


src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -345,7 +363,7 @@ private void onVersionChanged() {
   private void updateVersionTextIfCurrent(@NotNull FlutterSdk sdk, @NotNull String value) {
     final FlutterSdk current = FlutterSdk.forPath(getSdkPathText());
     if (current == null) {
-      myVersionLabel.setText("");
+      myVersionLabel.setText(" ");
```
jacob314: maybe comment why a " " is used instead of the empty string?

jacob314:
lgtm

src/io/flutter/sdk/FlutterSettingsConfigurable.java
devoncarew: I'll add a comment (it's so the layout doesn't collapse to 0 height).

Issue #3508, opened 2019-05-21
support launching on desktop, web


devoncarew:
This would depend on some implementation work in flutter tools and on the daemon protocol.

cc @DanTup @stuartmorgan @jonahwilliams

PR #3618, opened 2019-06-25
support showing desktop and web devices

devoncarew:
- show web and desktop devices in the device selector
- add a user preference to enable showing the above devices
- fix https://github.com/flutter/flutter-intellij/issues/3508

Note that running with those devices won’t work unless the projects have been created with the rightt support (flutter web, or desktop).

Screen Shot 2019-06-25 at 8 03 22 AM

Screen Shot 2019-06-24 at 4 24 48 PM

### DIFF
--- a/src/io/flutter/run/daemon/FlutterDevice.java
+++ b/src/io/flutter/run/daemon/FlutterDevice.java
@@ -18,11 +18,26 @@ public class FlutterDevice {
   private final @Nullable String myPlatform;
   private final boolean myEmulator;
 
+  private final @Nullable String myCategory;
+  private final @Nullable String myPlatformType;
+  private final boolean myEphemeral;
+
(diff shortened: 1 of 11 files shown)


src/io/flutter/run/daemon/FlutterDevice.java
```diff
@@ -44,6 +59,20 @@ public boolean emulator() {
...
+  public String category() {
+    return myCategory;
+  }
+
+  @Nullable
+  public String platformType() {
+    return myPlatformType;
+  }
+
+  public boolean ephemeral() {
```
kenzieschmoll: just a question: what does ephemeral mean in the context of devices?

devoncarew: I'll add some javadoc; ephemeral here means a device that isn't persistent on the machine (Chrome, or the 'MacOs' device are persistent; an iPhone would be an ephemeral device).

Issue #3615, opened 2019-06-25
the Flutter outline 'remove widget' button is always disabled


yizems:
![image](https://user-images.githubusercontent.com/16949474/60061585-4e521900-9728-11e9-82d7-d710dcb6bbdf.png)

![image](https://user-images.githubusercontent.com/16949474/60061629-6a55ba80-9728-11e9-9fed-146a41941b9c.png)

remove widget button disable

devoncarew:
cc @scheglov

scheglov:
https://github.com/flutter/flutter-intellij/pull/3622

stevemessick:
This will be fixed in M37, due out next week.

PR #3622, opened 2019-06-26
Issue 3615. Fix message for 'Remove widget'.

scheglov:
Ideally we should use `id` for matching Quick Assists and actions.
I think we postponed because at that time IDs were added only recently.

fix https://github.com/flutter/flutter-intellij/issues/3615


yizems:
can you  give me an artifacts
@scheglov I really need it.

email: yizeliang@126.com

devoncarew:
@yizems this fix will likely be out with our next (monthly) plugin release - currently tracking as M37

PR #3623, opened 2019-06-26
Remove netty and gnu.trove imports

DaveShuckerow:
Less dependencies = healthier code.

### DIFF
--- a/src/io/flutter/perf/FlutterWidgetPerf.java
+++ b/src/io/flutter/perf/FlutterWidgetPerf.java
@@ -20,7 +20,6 @@ import com.intellij.openapi.fileEditor.TextEditor;
 import com.intellij.openapi.util.TextRange;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.ui.EdtInvocationManager;
-import gnu.trove.TIntObjectHashMap;
 import io.flutter.utils.AsyncUtils;
 
 import javax.swing.Timer;
@@ -48,7 +47,7 @@ public class FlutterWidgetPerf implements Disposable, WidgetPerfListener {
   public static final long IDLE_DELAY_MILISECONDS = 400;
 
   class StatsForReportKind {
-    final TIntObjectHashMap<SlidingWindowStats> data = new TIntObjectHashMap<>();
+    final Map<Integer, SlidingWindowStats> data = new HashMap<>();
     private int lastStartTime = -1;
     private int lastNonEmptyReportTime = -1;
   }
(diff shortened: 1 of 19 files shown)


src/io/flutter/perf/FlutterWidgetPerf.java
```diff
@@ -20,7 +20,6 @@
 import com.intellij.openapi.util.TextRange;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.ui.EdtInvocationManager;
-import gnu.trove.TIntObjectHashMap;
```
devoncarew: I believe this is a performance optimization (`TIntObjectHashMap` - optimized when the keys of your map are an int). I don't know if it's actually required in the places of the code where we're using `TIntObjectHashMap`.

cc @jacob314 for this use site

Have you looked to see if there's an equivalent in the IntelliJ codebase, or our other deps?

DaveShuckerow: What frequency of operations are we expecting that the hashing algorithm used might be a performance hit?

DaveShuckerow: Looking at this, I believe we may get better worst-case performance on java.util.HashMap than gnu.trove.THashMap.

Looking at the hashing strategy used in [TIntObjectHashMap](https://github.com/palantir/trove-3.0.3/blob/master/generated-src/gnu/trove/impl/hash/TIntHash.java#L256), it reads like a simple double-hashing algorithm.

Java HashMaps use [balanced tree buckets](http://openjdk.java.net/jeps/180) in degenerate hashing cases.  This approach reduces the number of hash collisions to handle and gives asymptotically better big-O performance.  Both, naturally, have amortized O(1) performance.  However, the tree bucketing algorithm gives un-amortized O(log(n)) put and get operations, while a double-hash gives un-amortized O(n).

DaveShuckerow:
> Looks good.
> 
> You could also consider adding the netty import to the list of proscribed imports: https://github.com/flutter/flutter-intellij/blob/master/tool/plugin/lib/lint.dart#L75. That lint will fail our bot on usages of those imports. It’ll protect against these imports getting accidentally re-added.

That looks like a great lint to have.  I've added these imports to it.

src/io/flutter/view/MultiIconSimpleColoredComponent.java
```diff
@@ -117,7 +114,7 @@ public PositionedIcon(Icon icon, int index) {
 
   private int myMainTextLastIndex = -1;
 
-  private final TIntIntHashMap myFragmentPadding;
+  private final Map<Integer, Integer> myFragmentPadding;
```
devoncarew: Note: it looks like this changed behavior - before this, querying for an entry that doesn’t exist would return `0`; now, it throws an NPE.

devoncarew:
Given the differences in behavior from `gnu.trove.TIntObjectHashMap` and Java’s Map (`TIntObjectHashMap` return’s zero when queried for an entry that doesn’t exist, and Java’s Map will throw an NPE - trying to convert the null Integer object to an `int` literal), I think we should revert the gnu portion of this PR. I’m seeing several issues with the Flutter Inspector after this change.

devoncarew:
https://github.com/flutter/flutter-intellij/pull/3643

From looking at the InttelliJ community sources, they make liberal use of `gnu.trove.TIntObjectHashMap`.

DaveShuckerow:
I am OK with rolling back the gnu.trove changes.

Issue #3590, opened 2019-06-17
NPE when closing browser opened by IntelliJ + flutter web


mfichman:
## What happened

Closed browser opened by IntelliJ when using flutter web

## Version information

IntelliJ IDEA `2019.1.3` • Flutter plugin `io.flutter 36.1.4` • Dart plugin `191.7830`

Flutter 1.5.4-hotfix.2 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 7a4c33425d (7 weeks ago) • 2019-04-29 11:05:24 -0700
Engine • revision 52c7a1e849
Tools • Dart 2.3.0 (build 2.3.0-dev.0.5 a1668566e5)

## Exception

null

```
java.lang.NullPointerException
	at io.flutter.server.vmService.HeapMonitor$HeapObject.getAsJsonObject(HeapMonitor.java:55)
	at io.flutter.server.vmService.HeapMonitor$HeapObject.getEntries(HeapMonitor.java:59)
	at io.flutter.server.vmService.HeapMonitor$IsolateObject.getHeaps(HeapMonitor.java:92)
	at io.flutter.inspector.HeapState.handleIsolatesInfo(HeapState.java:104)
	at io.flutter.inspector.HeapDisplay$1.lambda$handleIsolatesInfo$0(HeapDisplay.java:64)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(Unknown Source)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(Unknown Source)
	at java.desktop/java.awt.EventQueue$4.run(Unknown Source)
	at java.desktop/java.awt.EventQueue$4.run(Unknown Source)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(Unknown Source)
	at java.desktop/java.awt.EventQueue.dispatchEvent(Unknown Source)
	at com.intellij.ide.IdeEventQueue.defaultDispatchEvent(IdeEventQueue.java:729)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:678)
	at com.intellij.ide.IdeEventQueue.dispatchEvent(IdeEventQueue.java:373)
	at java.desktop/java.awt.EventDispatchThread.pumpOneEventForFilters(Unknown Source)
[...]

PR #3625, opened 2019-06-26
guard against null json heap entries

pq:
Handle null JSON entries more gracefully.  This happens, for example, if we ask an Isolate for it's heap data (as for a web app in https://github.com/flutter/flutter-intellij/issues/3590) and there isn't any.

/cc @devoncarew

### DIFF
--- a/src/io/flutter/inspector/HeapState.java
+++ b/src/io/flutter/inspector/HeapState.java
@@ -15,6 +15,7 @@
  */
 package io.flutter.inspector;
 
+import com.google.gson.JsonElement;
 import io.flutter.vmService.HeapMonitor;
 import org.dartlang.vm.service.element.IsolateRef;
 import org.dartlang.vm.service.element.VM;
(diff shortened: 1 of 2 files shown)


src/io/flutter/vmService/HeapMonitor.java
```diff
@@ -52,11 +49,13 @@ String getAsString(String memberName) {
     }
 
     JsonObject getAsJsonObject(String memberName) {
-      return json.get(memberName).getAsJsonObject();
+      final JsonElement element = json.get(memberName);
```
devoncarew: From some slight perusing of the source code, we may also fall down when trying to read `"_currentRSS"` out of the `VM` object? I assume none of this info is populated for flutter web.

pq: Good catch!

src/io/flutter/inspector/HeapState.java
```diff
@@ -110,7 +111,8 @@ public void handleIsolatesInfo(VM vm, List<HeapMonitor.IsolateObject> isolates)
       }
     }
 
-    rssBytes = vm.getJson().get("_currentRSS").getAsInt();
+    final JsonElement rss = vm.getJson().get("_currentRSS");
```
pq: /cc @devoncarew 👍

pq:
> Have you tried running this with a flutter web app?

Yep.  Seems fixed.  👍

PR #3628, opened 2019-06-27
Update build ranges for new releases

stevemessick:
Build a single file for all 2019.1.3 products: IJ, AS beta & canary.

After this (ie M37), switch to using beta and canary distro channels.

@devoncarew @DaveShuckerow


devoncarew:
I think you need to regenerate the .travis.yaml file.

stevemessick:
Thanks! I updated the tests, too. If this is green I'll finish up quick.

PR #3640, opened 2019-07-03
Fix per SDK Stream<Uint8List> breaking changes

tvolkert:
This updates call sites with forward-compatible fixes for
a recent Dart SDK change that changed the following APIs
to be `Stream<Uint8List>` rather than `Stream<List<int>>`:

* HttpClientResponse
* File.openRead()

https://github.com/dart-lang/sdk/issues/36900


tvolkert:
Fixed

PR #3641, opened 2019-07-04
show truncated logging messages

devoncarew:
- show truncated logging messages


src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -144,4 +172,54 @@ private void printStackTraceToConsole(
   private String stringValueFromStringRef(InstanceRef ref) {
     return ref.getValueAsStringIsTruncated() ? ref.getValueAsString() + "..." : ref.getValueAsString();
   }
+
+  private String stringValueFromStringRef(Instance instance) {
+    return instance.getValueAsStringIsTruncated() ? instance.getValueAsString() + "..." : instance.getValueAsString();
+  }
+
+  private String getFullStringValue(@NotNull VmService service, String isolateId, @Nullable InstanceRef ref) {
```
lambdabaa: Would it make sense to add a couple unit tests given the different cases for what gets written to the `GetObjectConsumer`?

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -144,4 +172,54 @@ private void printStackTraceToConsole(
...
+    if (ref == null) return null;
+
+    if (!ref.getValueAsStringIsTruncated()) {
+      return ref.getValueAsString();
+    }
+
+    final CountDownLatch latch = new CountDownLatch(1);
+    final String[] result = new String[1];
```
lambdabaa: Why a length one array instead of a simple string?

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -144,4 +172,54 @@ private void printStackTraceToConsole(
...
+    service.getObject(isolateId, ref.getId(), 0, ref.getLength(), new GetObjectConsumer() {
+      @Override
+      public void onError(RPCError error) {
+        result[0] = ref.getValueAsString() + "...";
+        latch.countDown();
+      }
+
+      @Override
+      public void received(Obj response) {
+        if (response instanceof Instance && ((Instance)response).getKind() == InstanceKind.String) {
```
lambdabaa: Does it need to be cast to an `Instance` if it passes the `instanceof` check?

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -144,4 +172,54 @@ private void printStackTraceToConsole(
...
+        else {
+          result[0] = ref.getValueAsString() + "...";
+        }
+
+        latch.countDown();
+      }
+
+      @Override
+      public void received(Sentinel response) {
+        result[0] = ref.getValueAsString() + "...";
```
lambdabaa: nit: Optional cleanup to refactor these into something like `String formatTruncatedString(InstanceRef ref) {}`

src/io/flutter/logging/FlutterConsoleLogManager.java
pq: I think because `result` gets assigned in the lambda.  (There is probably a more modern idiom for this though...)

src/io/flutter/logging/FlutterConsoleLogManager.java
devoncarew: 👍 yes, good point. I’ll see about where these can slot into our existing tests

src/io/flutter/logging/FlutterConsoleLogManager.java
devoncarew: Yup, it’s a Java idiom. Hopefully not too out of date…

src/io/flutter/logging/FlutterConsoleLogManager.java
devoncarew: It does, through I could also assign it to a new local variable.

src/io/flutter/logging/FlutterConsoleLogManager.java
devoncarew: Yup, I’ll make a common utility function.

devoncarew:
Updated to add tests (based on a mocked vm service, mocked ConsoleView for output, and some canned json for the tests): 4bb11469677a8057489d7fbc36b7867b0a7a08fb.

PR #3644, opened 2019-07-07
initial work on displaying Flutter.Error events

devoncarew:
- initial work on displaying Flutter.Error events

This PR recursively renders `Flutter.Error` events, uses ansi colors for the IntelliJ console, and performs folding of long lines in order to present the error events in a concise manner.

We'll want to adjust the output to account for previous investigations in this area as well as iterate on the look in general. We can also add hyperlinks from text in the console to widgets in the inspector.

before this PR (note that some error events are several hundred lines long):

Screen Shot 2019-07-05 at 6 07 57 AM

and after:

Screen Shot 2019-07-07 at 2 12 40 PM

Screen Shot 2019-07-07 at 2 12 53 PM

cc @InMatrix @jacob314 @pq


InMatrix:
Wow, I didn't know you can fold lines in the IntelliJ console! Now I wonder what else it can do.

devoncarew:
The current latest; will update the PR.

Screen Shot 2019-07-10 at 2 46 04 PM

src/io/flutter/inspector/DiagnosticsTreeStyle.java
```diff
@@ -72,13 +72,15 @@
   singleLine,
 
   /**
-   * Render the tree on a single line with the name and value on separate
-   * lines.
+   * Render the tree on a single line with the name and value on separate lines.
```
jacob314: this name is obsolete. as part of the code review of the matching Flutter code it changed to errrorProperty.
https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/foundation/diagnostics.dart#L163

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -6,62 +6,246 @@
...
+        });
+      }
+    }
+    catch (Throwable t) {
+      LOG.warn(t);
+    }
+  }
+
+  private static final String errorSeparator =
+    "◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤◢◤";
```
jacob314: FWIW, this is the separator that on the flutter side was intended to indicate "overflow" I believe. 
I don't have a better idea for a generic separator to use. I think this is at least as reasonable as the other unicode line art used in flutter to display error boxes.

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -6,62 +6,246 @@
...
+        return;
+      }
+    }
+
+    if (contentType == null) {
+      if (property.getLevel() == DiagnosticLevel.error) {
+        contentType = ERROR_CONTENT_TYPE;
+      }
+      else if (property.getLevel() == DiagnosticLevel.summary) {
+        contentType = ERROR_CONTENT_TYPE; // NORMAL_CONTENT_TYPE;
```
jacob314: nit: remove commented out line. Perhaps change ERROR_CONTENT_TYPE to ERROR_SUMMARY_TYPE

src/io/flutter/inspector/DiagnosticsTreeStyle.java
devoncarew: gotcha, I’ll update this file

src/io/flutter/logging/FlutterConsoleLogManager.java
devoncarew: OK, I hadn’t put that together - that is was used for overflow, I thought it was just a separator.

I do like the look of it as a separator however…

src/io/flutter/logging/FlutterConsoleLogManager.java
devoncarew: 👍

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -6,62 +6,240 @@
...
 
 /**
  * Handle displaying dart:developer log messages and Flutter.Error messages in the Run and Debug
  * console.
  */
 public class FlutterConsoleLogManager {
-  public static final boolean SHOW_STRUCTURED_ERRORS = false;
+  private static final Logger LOG = Logger.getInstance(FlutterConsoleLogManager.class);
+
+  public static final boolean SHOW_STRUCTURED_ERRORS = true;
```
devoncarew: Ah, note that this PR turns error display on. I could also change this to an opt-in user setting.

src/io/flutter/logging/FlutterConsoleLogManager.java
jacob314: here's the style that Flutter uses for errors by default. I agree I'm not sure which style is actually better.
https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/foundation/diagnostics.dart#L557

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -40,134 +43,197 @@
...
-        final String[] lines = wrap(description, COL_WIDTH - indent.length(), col);
+    if (property.hasChildren()) {
+      try {
+        final CompletableFuture<ArrayList<DiagnosticsNode>> future = property.getChildren();
+        final ArrayList<DiagnosticsNode> children = future.get();
 
-        if (lines.length == 1) {
-          console.print(lines[0] + "\n", contentType);
+        if (!isInChild && children.stream().noneMatch(DiagnosticsNode::hasChildren)) {
+          final String childIndent = indent + "  ";
```
jacob314: if the parent has style DiagnosticsTreeStyle.flat 
then the child indent should just be the indent. Somehow your indents in the screenshot seem flat for the outer children. Not sure how that is happening.
https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/foundation/diagnostics.dart#L1674

devoncarew: OK, I'll adjust the indent when I see `DiagnosticsTreeStyle.flat`. The initial set of children - the immediate properties on the root diagnostics node - do have some special casing in terms of indent and padding blank lines.

devoncarew: Things like exception traces now print out with less indent (seems legit):

Screen Shot 2019-07-10 at 4 21 46 PM

devoncarew:
Updated a bit to adjust the colors for the light IntelliJ theme, and to respect the `DiagnosticsTreeStyle.shallow` tree style.

shinayser:
I just updated to plugin version 38 and still the same colors as before.  
Is that an opt-in preference should I switch on somewhere?

InMatrix:
Yes, this is an opt-in feature for now. Please turn on this setting to see the effect.

image

PR #3658, opened 2019-07-11
fix an issue with some daemon json output appearing in the console

devoncarew:
- fix an issue with some daemon json output appearing in the console
- unrelated, rename a `System` class to `SystemUtils` so it doesn't have the same name as `java.lang.System`

```
[{"event":"app.progress","params":{"appId":"3afb1f88-db66-490d-bce7-f30050cb5d2a","id":"7","progressId":"hot.reload","message":"Performing hot reload..."}}]Performing hot reload...
```

### DIFF
--- a/src/io/flutter/utils/StdoutJsonParser.java
+++ b/src/io/flutter/utils/StdoutJsonParser.java
@@ -32,7 +32,7 @@ public class StdoutJsonParser {
       if (!bufferIsJson && buffer.length() == 2 && buffer.charAt(0) == '[' && c == '{') {
         bufferIsJson = true;
       }
-      else if (bufferIsJson && c == ']' && possiblyTerminatesJson(string, i, buffer)) {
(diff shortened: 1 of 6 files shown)


src/io/flutter/utils/StdoutJsonParser.java
```diff
@@ -41,25 +41,27 @@ else if (bufferIsJson && c == ']' && possiblyTerminatesJson(string, i, buffer))
...
     }
 
-    // Eagerly flush if we are not within JSON so regular log text is written
-    // as soon as possible.
+    // Eagerly flush if we are not within JSON so regular log text is written as soon as possible.
     if (!bufferIsJson) {
       flushLine();
     }
+    else if (buffer.toString().endsWith("}]")) {
+      flushLine();
```
devoncarew: this is the bulk of the fix

devoncarew:
(will address the travis failure)

PR #3661, opened 2019-07-11
init default settings for the run console text wrapping

devoncarew:
- init default settings for the run console text wrapping

`Flutter.Error` events will render better with text wrapping enabled for the Run and Debug consoles. We turn that setting on for the user (once).


src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -60,6 +63,22 @@
   final private CompletableFuture<InspectorService.ObjectGroup> objectGroup;
   private static QueueProcessor<Runnable> queue;
 
+  /**
+   * Set our preferred settings for the run console.
+   */
+  public static void initConsolePreferences() {
+    final String consolePreferencesSet = "io.flutter.console.preferencesSet";
```
lambdabaa: Maybe it's a constant?

devoncarew: done - moved to a static var

PR #3662, opened 2019-07-11
Use DAS test annotations to flag runnable tests

DaveShuckerow:
Working proof of concept.

Some questions I want to verify as part of testing this:
* [x] proper disposal of the open editors service
* [x] test/group identification in sample dart files
* [x] FlutterSettings listener interaction with the open editor listener in the widget indentations factory

Quirks:
* The CommonTestConfig doesn't know about what project it's part of, and it's consumed by other objects that are also project-agnostic (specifically RunConfigurationProducer, FlutterTestLocationProvider, FlutterTestLineMarkerContributor)
  * This means that it doesn't have a way to inject the outline service.
  * Both the current and previous implementations identified runnable tests with pull-based instead of push-based logic, so this is OK, at least locally.
* In plugin.xml, I don't know what the line is between a projectService and a component.

Difficulties:
* We didn't have an easy way of feeding FlutterOutlines to run the test config tests over them.  To get the tests running again, I added a golden file system for reading in the outlines from disk.  If the feature isn't explained well enough, I can put together a doc describing in detail.


src/io/flutter/editor/WidgetIndentsHighlightingPassFactory.java
```diff
@@ -107,128 +77,37 @@ public void caretPositionChanged(@NotNull CaretEvent event) {
...
       }
-    }
+    });
   }
 
   private void syncSettings(FlutterSettings settings) {
     if (isShowBuildMethodGuides != settings.isShowBuildMethodGuides()) {
       isShowBuildMethodGuides = settings.isShowBuildMethodGuides();
-      updateActiveEditors();
+      // QUESTION: Do we need this?
```
jacob314: I would leave this.  This is a case where the active editors haven't changed but we need to force refresh
 our display as our settings have changed internally.

src/io/flutter/editor/WidgetIndentsHighlightingPassFactory.java
```diff
@@ -295,6 +168,8 @@ void runWidgetIndentsPass(EditorEx editor, @NotNull FlutterOutline outline) {
       return;
     }
 
+    // QUESTION: why doesn't this check against isShowBuildMethodGuides ?
```
jacob314: Now that you have refactored, I agree it is safest to check against the flag here.
It is possible it wasn't needed before as this wouldn't trigger unless the guides were enabled.

src/io/flutter/editor/WidgetIndentsHighlightingPassFactory.java
```diff
@@ -324,9 +199,32 @@ void runWidgetIndentsPass(EditorEx editor, @NotNull FlutterOutline outline) {
 
   @Override
   public void dispose() {
-    clearListeners();
     FlutterSettings.getInstance().removeListener(settingsListener);
   }
+
+  private class Listener implements FlutterSettings.Listener {
```
jacob314: nit: what is the advantage of defining this as an inner class vs a closure like was done previously?

src/io/flutter/editor/outline/OpenEditorOutlineService.java
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * Service that watches the Flutter Outlines for all open Dart files.
+ *
+ * <p>
+ * This service works by listening to the {@link Project}'s MessageBus for {@link FileEditor}s that are
+ * added or removed. Using the set of currently active editor windows ({@link EditorEx}s), this service
+ * then subscribes to the {@link FlutterDartAnalysisServer} for updates to the {@link FlutterOutline} of each file.
+ */
+public class OpenEditorOutlineService implements Disposable {
```
jacob314: The word open is a bit confusing to me as it made me thing this service opens editor outlines not that it is showing outlines for the visible editors. How about
ActiveEditorOutlineService or SelectedEditorOutlineService?

src/io/flutter/editor/outline/OpenEditorOutlineService.java
```diff
@@ -0,0 +1,232 @@
...
+import io.flutter.dart.FlutterOutlineListener;
+import org.dartlang.analysis.server.protocol.FlutterOutline;
+import org.jetbrains.annotations.NotNull;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+import java.util.*;
+
+/**
+ * Service that watches the Flutter Outlines for all open Dart files.
```
jacob314: Note this isn't really all open files. Just all Dart files that are visible to the user. Files that are open but in tabs that are are not selected are not included which is a good thing for performance.

src/io/flutter/editor/outline/OpenEditorOutlineService.java
```diff
@@ -0,0 +1,232 @@
...
+
+
+  /**
+   * The cached {@link FlutterOutline} for some file.
+   *
+   * <p>
+   * Provides useful metadata for test identification.
+   */
+  @Immutable
+  public static class OutlineCache {
```
jacob314: super nit: maybe OutlineCacheEntry as OutlineCache makes me think this is an overall cache vs a cache for a single file.

src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -26,45 +32,76 @@
...
 public abstract class CommonTestConfigUtils {
   /**
-   * Regex that matches customized versions of the Widget test function from package:flutter_test/src/widget_tester.dart.
-   * <p>
-   * This will match all test methods with names that start with 'test', optionally
-   * have additional text in the middle, and end with 'Widgets'.
+   * How the Dart Analysis Server identifies runnable tests in the {@link Outline} it generates.
    */
-  public static final Pattern WIDGET_TEST_REGEX = Pattern.compile("test([A-Z][A-Za-z0-9_$]*)?Widgets");
+  private static final String UNIT_TEST_TEST = "UNIT_TEST_TEST";
```
jacob314: does the analysis server not include UNIT_TEST_TEST as a constant?

src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -26,45 +32,76 @@
...
   }
 
+  public abstract TestType asTestCall(@NotNull PsiElement element);
+
+  /**
+   * Gets the elements from the outline that are runnable tests.
+   */
+  protected Map<DartCallExpression, TestType> getTestsFromOutline(@NotNull PsiFile file) {
+    final Project project = file.getProject();
+    final FlutterOutline outline = OpenEditorOutlineService.getInstance(project).get(file.getVirtualFile().getPath());
```
jacob314: consider adding a method to the outline editor that goes from a VirtualFile to an outline instead of a String.
Also, do we care about cases where the PsiFile and the Outline are out of sync? For example, you can check that their links are consistent. There was some logic in the outline rendering code that makes sure the lengths are the same or can be made the same by transforming the locations. It could make sense to add that logic to the OpenEditorOutlineService so we are consistent about the cases where the outline is slightly out of date so shouldn't be used. The outlines become consistent a second after an edit but if you are displaying visualizations of where tests occur in the source editors in the future then the mismatch will be noticeable.

src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -26,45 +32,76 @@
...
 
   @Nullable
   protected TestType findNamedTestCall(@NotNull PsiElement element) {
     if (element instanceof DartCallExpression) {
       final DartCallExpression call = (DartCallExpression)element;
-      for (TestType type : TestType.values()) {
-        if (type.matchesFunction(call)) return type;
+      final Map<DartCallExpression, TestType> callToTestType = getTestsFromOutline(element.getContainingFile());
+
+      if (callToTestType.containsKey(call)) {
```
jacob314: the call to containsKey isn't needed. Just return the value regardless of if it is null.

src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -26,45 +32,76 @@
...
     return StringUtil.trimTrailing(
       url.replaceFirst("http:", "ws:"), '/') + "/ws";
   }
 
+  public abstract TestType asTestCall(@NotNull PsiElement element);
+
+  /**
+   * Gets the elements from the outline that are runnable tests.
+   */
+  protected Map<DartCallExpression, TestType> getTestsFromOutline(@NotNull PsiFile file) {
```
jacob314: We're creating a map and only using it to look up a single item as far as I can tell. Either simplify to just lookup the one item you care about directly or cache the map invalidating when outlines change.

testSrc/unit/io/flutter/run/common/TestTypeTest.java
```diff
@@ -35,7 +35,7 @@
                                              "  test('test 1', () {});\n" +
                                              "  testingWidgets('does not test widgets');\n" +
                                              "}";
-
+/*
```
jacob314: is it intentional that this case is commented out? If so, indicate why.

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: IntelliJ appears to be taking care of consistency here under the hood.  In the current state of this PR, behavior is the same as we get at head.  

Reading the [details](https://www.jetbrains.org/intellij/sdk/docs/tutorials/custom_language_support/line_marker_provider.html#even-more-technical-details) of how the line markers pass works, it looks like it is possible for the pass it takes to take longer than the outline update.

However, there's currently no way to ask IntelliJ to check for runnable tests again after getting the new outline.  The fix for that is to enable the line marker pass to listen to the outline service the same way that the indent highlighting pass does.  Any logic involving outdated outlines will need to go there where we can make another pass instead of here where we can't.

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: The map call is used in findNamedTestCall below.  IntelliJ might give us an arbitrary PsiElement and ask us to find the test call that is parent to it.

This means we need to look up the tree to find an element that is in the collection of all test calls.

src/io/flutter/editor/ActiveEditorsOutlineService.java
```diff
@@ -0,0 +1,256 @@
...
+ * <p>
+ * This service works by listening to the {@link Project}'s MessageBus for {@link FileEditor}s that are
+ * added or removed. Using the set of currently active {@link EditorEx} editor windows, this service
+ * then subscribes to the {@link FlutterDartAnalysisServer} for updates to the {@link FlutterOutline} of each file.
+ *
+ * <p>
+ * This class provides a {@link Listener} that notifies consumers when
+ * <ul>
+ * <li>The collection of currently active editors has changed</li>
+ * <li>Each outline for a currently active editor has updated.</li>
```
jacob314: nit:
Each outline --> The outline

src/io/flutter/editor/WidgetIndentsHighlightingPassFactory.java
```diff
@@ -107,135 +86,57 @@ public void caretPositionChanged(@NotNull CaretEvent event) {
...
-    synchronized (outlineListeners) {
-      final Iterator<Map.Entry<String, FlutterOutlineListener>> entries = outlineListeners.entrySet().iterator();
-
-      while (entries.hasNext()) {
-        final Map.Entry<String, FlutterOutlineListener> entry = entries.next();
-
-        if (entry.getValue() != null) {
-          flutterDartAnalysisService.removeOutlineListener(entry.getKey(), entry.getValue());
+  // Updates the indent guides in a specific editor.
+  private void updateEditors(@NotNull final String path) {
```
jacob314: comment says editor but method name says editors. Unify presumably removing the "s".

testData/sample_tests/test/simple_test.dart
```diff
@@ -0,0 +1,21 @@
...
+
+/// Test data for test line marker tests.
+/// The outline for this file is pretty-printed in simple_outline.txt.
+///
+/// To update the outline, open this file in a debugging IntelliJ instance, then
+/// set a breakpoint in the ActiveEditorsOutlineService's OutlineListener call
+/// to `notifyOutlineUpdated`, then look at the Flutter Outline for this file.
+/// This will break at the outline and allow you to copy its json value.
+///
+/// You can pretty-print the json using the pretty_printer.dart script.
```
jacob314: provide a full path to where to find the pretty_printer.dart script.

testSrc/unit/io/flutter/editor/ActiveEditorsOutlineServiceTest.java
```diff
@@ -0,0 +1,214 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.*;
+import static org.junit.Assert.assertThat;
+
+public class ActiveEditorsOutlineServiceTest {
```
jacob314: Fyi @stevemessick to make sure the test code is following the best practices for the IntelliJ repo.

src/io/flutter/editor/ActiveEditorsOutlineService.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+    synchronized (pathToOutline) {
+      pathToOutline.clear();
+    }
+  }
+
+  /**
+   * Listener for changes to the active editors or open outlines.
+   */
+  public interface Listener {
```
jacob314: High level it would be cleaner if these events included the payload of editors and outlines rather than requiring clients to call methods back on the ActiveEditorsOutlineService object. That would reduce the risk of race conditions and perhaps simplify the public api for ActiveEditorsOutlineService.

testSrc/unit/io/flutter/editor/ActiveEditorsOutlineServiceTest.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @Override
+    public void onOutlineChanged(String path) {
+      final Integer changes = outlineChanged.get(path);
+      outlineChanged.put(path, changes == null ? 1 : changes + 1);
+    }
+  }
+
+  private static class Fixture extends AdaptedFixture<CodeInsightTestFixture> {
```
stevemessick: Did you try changing ProjectFixture to inherit from CodeInsightTestFixture instead of IdeaProjectTestFixture? If that causes breakage then I'd like to see this Fixture class moved to a sibling of ProjectFixture, perhaps named CodeInsightProjectFixture (I don't care too much what it is named). It would include getProject() and getModule().

DaveShuckerow:
There is an additional quirk to work around before this can be merged: we need to be sure that the test utils can wait until it has an up-to-date outline before generating run configurations.

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: Done.

src/io/flutter/editor/WidgetIndentsHighlightingPassFactory.java
DaveShuckerow: Done.

testData/sample_tests/test/simple_test.dart
DaveShuckerow: Done.

testSrc/unit/io/flutter/editor/ActiveEditorsOutlineServiceTest.java
DaveShuckerow: Done.

src/io/flutter/editor/ActiveEditorsOutlineService.java
DaveShuckerow: Done.  I was able to make the get method protected instead of public with this.

src/io/flutter/editor/ActiveEditorsOutlineService.java
```diff
@@ -186,8 +189,61 @@ public FlutterOutline get(String path) {
...
+   * If the cached outline is not updated yet, will request an updated outline from the analysis service.
+   */
+  public FlutterOutline request(@NotNull VirtualFile file) {
+    if (isUpToDate(file)) {
+      return get(file);
+    }
+    final String path = file.getCanonicalPath();
+    // If we haven't gotten any data, set up a temporary listener for this file, and wait for a response.
+    // We block to wait with a Lock.
+    final Lock lock = new ReentrantLock();
```
jacob314: We need to avoid blocking the UI thread in IntelliJ. How do we ensure this API isn't called in a blocking way on the UI thread?
An alternate design would just be to assume that an updated outline will arrive at some point if the current outline is out of date at which point the correct code flow can trigger based on the existing event driven API.

src/io/flutter/editor/ActiveEditorsOutlineService.java
```diff
@@ -186,8 +189,61 @@ public FlutterOutline get(String path) {
...
+    final DartAnalysisServerService das = DartAnalysisServerService.getInstance(project);
+    return psiFile != null && outline != null && (
+      psiFile.getTextLength() == outline.getLength()
+      || psiFile.getTextLength() == das.getConvertedOffset(file, outline.getLength())
+    );
+  }
+
+  @VisibleForTesting
+  protected PsiFile getPsiFile(VirtualFile file) {
+    return PsiManager.getInstance(project).findFile(file);
```
jacob314: Make this a utility method on a different class if you need it.

src/io/flutter/editor/ActiveEditorsOutlineService.java
DaveShuckerow: I have finally found a way of requesting a new line markers pass:

```java
DaemonCodeAnalyzer.getInstance(project).restart();
```

The IntelliJ Dart plugin does this in cases where it receives new data from the dart analyzer already.  I've updated this PR with that data.

src/io/flutter/editor/ActiveEditorsOutlineService.java
DaveShuckerow: It's no longer necessary, but it was originally intended for use during tests to inject custom files.

src/io/flutter/editor/ActiveEditorsOutlineService.java
```diff
@@ -171,18 +166,28 @@ public void removeListener(@NotNull Listener listener) {
...
-   * Gets the most up-to-date {@link FlutterOutline} for {@param file}.
+   * Gets the {@link FlutterOutline} for {@param file} if and only if the outline is up to date with the file.
+   *
+   * <p>
+   * Returns null if the file is out of date.
    */
   @Nullable
-  public FlutterOutline get(@NotNull VirtualFile file) {
-    return get(file.getCanonicalPath());
+  public FlutterOutline getIfUpdated(@NotNull PsiFile file) {
```
jacob314: nit: slightly clumsy name. Maybe getIfLatest
or perhaps only support getting outlines if they are compatible. There is no point in serving out of date outlines.

DaveShuckerow: Reasoning here explained in an earlier comment.

I'd prefer to leave the name at getIfUpdated.  I'm not sure 'latest' is less clumsy than 'updated', and the surrounding logic and comments all discuss an up-to-date or out-of-date outline, so the name fits closer with the other language.

Issue #3423, opened 2019-04-29
revise VmOpenSourceLocationListener to share the existing service protocol connection


devoncarew:
The VmOpenSourceLocationListener class implements the IDE's `'open source'` service protocol extension. It's implementation is older, and looks like it doesn't share our existing connection to the VM (it opens its own). We should update this code.

stevemessick:
I believe VmOpenSourceLocationListener was recently removed.

PR #3669, opened 2019-07-15
remove use of the older _registerService API

devoncarew:
- remove use of the older _registerService API
- partial work towards https://github.com/flutter/flutter-intellij/issues/3423

We can restore this functionality in a future PR (and version detect to decide whether to call `_registerService` or `registerService`).

cc @bkonyi


jacob314:
I think this is dead code from work done by a DartVM team intern around 3 years ago.

devoncarew:
> I think this is dead code from work done by a DartVM team intern around 3 years ago.

Yes; ironically, this was the very first use of the `_registerService` call. We can restore this if we think that having IDEs expose an `openSourceLocation` service extension is something we would use from other tools.

Issue #3680, opened 2019-07-18
ship error message presentation


devoncarew:
- add an opt-in flag
- update the ansi chars used to title and suffix the error message

PR #3682, opened 2019-07-18
update the flutter error display

devoncarew:
- update the flutter error display wrt the header and footer ansi chars
- fix https://github.com/flutter/flutter-intellij/issues/3680

We output a header and footer string that are >= 100 chars long, with the error message centered 1/3 of the way on the header line.

cc @DanTup for the VS Code work

Here is the current display from the 3 error messages generated by the `flutter_error_studies` app:

Screen Shot 2019-07-18 at 9 55 32 AM

Screen Shot 2019-07-18 at 9 55 41 AM

Screen Shot 2019-07-18 at 9 55 52 AM

### DIFF
--- a/src/io/flutter/logging/FlutterConsoleLogManager.java
+++ b/src/io/flutter/logging/FlutterConsoleLogManager.java
@@ -173,7 +173,7 @@ public class FlutterConsoleLogManager {
       }
     }
 
-    console.print(StringUtil.repeat(errorSeparatorChar, errorSeparatorLength) + "\n\n", TITLE_CONTENT_TYPE);
+    console.print(StringUtil.repeat(errorSeparatorChar, errorSeparatorLength) + "\n", TITLE_CONTENT_TYPE);
   }
 
   private void printTerseNodeProperty(ConsoleView console, String indent, DiagnosticsNode property) {
@@ -256,7 +256,11 @@ public class FlutterConsoleLogManager {
     console.print(description + "\n", contentType);
 
     if (property.hasInlineProperties()) {
-      final String childIndent = getChildIndent(indent, property);
+      String childIndent = getChildIndent(indent, property);
+      if (property.getStyle() == DiagnosticsTreeStyle.shallow && !indent.startsWith("...")) {
+        // Render properties of shallow nodes as collapesed.
+        childIndent = "...  " + indent;
+      }
       for (DiagnosticsNode childProperty : property.getInlineProperties()) {
         printDiagnosticsNodeProperty(console, childIndent, childProperty, contentType, isInChild);
       }
--- a/src/io/flutter/logging/FlutterConsoleLogManager.java
+++ b/src/io/flutter/logging/FlutterConsoleLogManager.java
@@ -34,7 +34,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -130,6 +129,8 @@ public class FlutterConsoleLogManager {
   private static final int errorSeparatorLength = 100;
   private static final String errorSeparatorChar = "═";
 
+  private static final ArrayList<DiagnosticsNode> emptyList = new ArrayList<>();
+
   /**
    * Pretty print the error using the available console syling attributes.
    */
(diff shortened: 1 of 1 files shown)


DanTup:
Ok, I think I've got these the same in VS Code (besides some tweaks, like our clickable paths end up on the right, and we highlights the users frames in a stack in red). However, the "missing material" one looks kinda bed - it's *very* long and noisy. I'm not sure if we should exclude some of these nodes when rendering here?

Do we have a central location for people to report feedback on the display of the errors here? It'd be good to link to there in the release notes so we can collect it in one place (for ex. if people find specific types of errors that aren't reported well).

Screenshot 2019-07-22 at 2 56 55 pm

Screenshot 2019-07-22 at 2 58 42 pm

Screenshot 2019-07-22 at 2 56 13 pm

DanTup:
@devoncarew I compared the top screenshots, and I think we are both displaying the same info - I think the "Ancestor widgets" section is really noisy though. LMK if you think there's a way we could reasonable cut this down (or format it better).

devoncarew:
- this has been updated to use the new `errorsSinceReload` field
- also, some very slight tweaks in the display

tweaks:
Screen Shot 2019-07-23 at 2 19 06 PM

and, the terse display of secondary errors:
Screen Shot 2019-07-23 at 2 19 20 PM

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -283,6 +346,21 @@ public void handleLoggingEvent(@NotNull Event event) {
     });
   }
 
+  private ConsoleViewContentType getContentTypeFor(DiagnosticLevel level) {
+    if (level == DiagnosticLevel.error) {
```
jacob314: nit: looks like a switch statement.

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -126,95 +127,157 @@ public void handleFlutterErrorEvent(@NotNull Event event) {
...
+  private void printTerseNodeProperty(ConsoleView console, String indent, DiagnosticsNode property) {
+    boolean skip = true;
+
+    if (property.getLevel() == DiagnosticLevel.summary) {
+      skip = false;
+    }
+    else if (property.hasChildren()) {
+      try {
+        final CompletableFuture<ArrayList<DiagnosticsNode>> future = property.getChildren();
+        final ArrayList<DiagnosticsNode> children = future.get();
```
jacob314: you could all
future.getNow(null)
the children are already present for the error message payload case.
Calling .get() could lock. This will eliminate catching the threading exceptions.

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -126,95 +127,157 @@ public void handleFlutterErrorEvent(@NotNull Event event) {
...
+
+        final String childIndent = getChildIndent(indent, property);
+        for (DiagnosticsNode child : children) {
+          printDiagnosticsNodeProperty(console, childIndent, child, contentType, false);
+        }
+
+        if (property.hasInlineProperties()) {
```
jacob314: by convention properties should always be displayed before children instead of after children.

jacob314:
Looks good!

src/io/flutter/logging/FlutterConsoleLogManager.java
devoncarew: done

src/io/flutter/logging/FlutterConsoleLogManager.java
devoncarew: ah, great, will make this change

src/io/flutter/logging/FlutterConsoleLogManager.java
devoncarew: OK, I'll re-order this (and post new screen shots if things change).

devoncarew: Re-ordering did end up eliminating a bunch of duplicate code.

devoncarew:
And the latest screen shots:

missing material (which needs work on the framework side):

Screen Shot 2019-07-24 at 10 40 14 AM

the 2nd error send by missing material (due to the overflow):

Screen Shot 2019-07-24 at 10 40 22 AM

the overflow error:

Screen Shot 2019-07-24 at 10 40 38 AM

and the scaffold.of() error:

Screen Shot 2019-07-24 at 10 41 31 AM

Note that the overflow error used to have the last 8-9 lines collapsed, due to a bug in how we where displaying properties. This has been fixed, but I think the display looks worse. If we do choose to collapse, we'd loose the stack traces on other errors. I think this needs more iteration in the IDE side and on the framework side (to decide how we should be sending some of the data). I think this iteration is best done after the initial launch.

devoncarew:
Screen Shot 2019-07-24 at 2 41 52 PM

PR #3684, opened 2019-07-18
Make the event stream tests pass and re-enable them.

DaveShuckerow:
If we're shipping with the event stream's current behavior, then it seems to me the Event Stream is working the way we intend it to.

If that's the case, the best fix is to update the test and merge.

If that's not the case and we actually want to the behavior to match the tests, I think the best fix is to add a TODO for fixing the behavior and then merge these tests in a working state.  There are 9 tests in the file, and only 2 seem to be failing.  So long as this test is off, we aren't getting the value of the test cases that do work.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,7 +85,6 @@ sourceSets {
 }
 
 test {
-  exclude "io/flutter/utils/EventStreamTest.class"
   ignoreFailures true
 }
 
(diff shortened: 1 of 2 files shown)


stevemessick:
I think @jacob314 should have a look at this.

PR #3687, opened 2019-07-19
update the widgets.json catalog file

devoncarew:
- update the widgets.json catalog file


devoncarew:
> Not actually reviewing widgets.json as I assume it is copied from the web.

Yup, just copied in from the current location of record.

PR #3688, opened 2019-07-22
Add target platform selector for togglePlatform service extension.

kenzieschmoll:
Removes the toggle platform button from the service extension button bars (Inspector View and Performance View), and replaces it with a dropdown selector. We now include Fuchsia as an option and can add additional platforms in the future.

Screen Shot 2019-07-23 at 8 42 36 AM
Screen Shot 2019-07-23 at 8 42 49 AM

Confirmed that states stay synchronized between IntelliJ and DevTools.
Matching CL from DevTools is here: https://github.com/flutter/devtools/pull/783.

cc: @sfshaza2 in case this changes any docs or screenshots.

### DIFF
--- a/src/io/flutter/view/TogglePlatformAction.java
+++ b/src/io/flutter/view/TogglePlatformAction.java
@@ -5,12 +5,10 @@
  */
 package io.flutter.view;
 
-import com.intellij.execution.ui.ConsoleViewContentType;
 import com.intellij.openapi.Disposable;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.actionSystem.DefaultActionGroup;
 import com.intellij.openapi.actionSystem.Toggleable;
-import io.flutter.FlutterBundle;
 import io.flutter.run.daemon.FlutterApp;
 import io.flutter.utils.StreamSubscription;
 import io.flutter.vmService.ServiceExtensionState;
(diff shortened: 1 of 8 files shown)


src/io/flutter/view/TogglePlatformAction.java
```diff
@@ -6,23 +6,65 @@
...
 import org.jetbrains.annotations.NotNull;
 
-import java.util.concurrent.CompletableFuture;
+import javax.swing.*;
 
-class TogglePlatformAction extends FlutterViewAction {
-  private static final ToggleableServiceExtensionDescription extensionDescription = ServiceExtensions.togglePlatformMode;
-  private Boolean isCurrentlyAndroid;
-  CompletableFuture<Boolean> cachedHasExtensionFuture;
+class TogglePlatformAction extends ToolbarComboBoxAction {
```
jacob314: maybe add a TODO to show the current value on selection change.

src/io/flutter/view/TogglePlatformAction.java
```diff
@@ -36,34 +78,58 @@ public void update(@NotNull AnActionEvent e) {
...
-        });
-      });
+      app.togglePlatform(platformTarget.name());
+
+      app.getVMServiceManager().setServiceExtensionState(
+        extensionDescription.getExtension(),
+        true,
+        platformTarget.name());
+
+      app.getConsole().print(
```
jacob314: Does this print still make sense now that it is unambiguous what the platform override got set to?

src/io/flutter/vmService/ServiceExtensions.java
```diff
@@ -55,22 +61,23 @@
...
-    new ToggleableServiceExtensionDescription<>(
+  public static final ServiceExtensionDescription<String> togglePlatformMode =
+    new ServiceExtensionDescription<>(
       "ext.flutter.platformOverride",
-      "iOS",
-      "android",
-      "Toggle Platform",
-      "Toggle Platform");
+      "Override Target Platform",
+      new ArrayList<>(Arrays.asList("iOS", "android", "fuchsia")),
```
jacob314: change the type of the API to List from ArrayList and you can avoid the ArrayList constructor calls.

src/io/flutter/vmService/VMServiceManager.java
```diff
@@ -303,13 +308,13 @@ else if (StringUtil.equals(streamId, VmService.LOGGING_STREAM_ID) || StringUtil.
   }
 
   private Object getExtensionValueFromEventJson(String name, String valueFromJson) {
-    final Object enabledValue =
-      ServiceExtensions.toggleableExtensionsWhitelist.get(name).getEnabledValue();
+    final Object sampleValue =
```
jacob314: nit:
sampleValue --> firstValue?

jacob314: alternately better Java style would be to add a
getValueClass() method to the ServiceExtension class that returns a Class object consistent with this.
The class object could be generated from just getting the first value. https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html

kenzieschmoll: done

PR #3691, opened 2019-07-23
q3 survey

pq:
/cc @jayoung-lee @DanTup to double-check URLs, etc.

@devoncarew

### DIFF
--- a/src/io/flutter/survey/FlutterSurveyNotifications.java
+++ b/src/io/flutter/survey/FlutterSurveyNotifications.java
@@ -44,15 +44,15 @@ public class FlutterSurveyNotifications {
   private static final String SURVEY_TAKEN = "io.flutter.survey.2019.q3.alreadyTaken";
   private static final String SURVEY_URL = "https://google.qualtrics.com/jfe/form/SV_3kiGXYfYOfXUjB3?Source=IntelliJ&ClientID=";
 
-  private static long SURVEY_START;
-  private static long SURVEY_END;
+  private static long SURVEY_START_MS_EPOCH;
+  private static long SURVEY_END_MS_EPOCH;
 
   static {
     final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd HH:mm");
     dateFormat.setTimeZone(TimeZone.getTimeZone("PST"));
     try {
-      SURVEY_START = dateFormat.parse("2019/08/12 9:00").getTime();
-      SURVEY_END = dateFormat.parse("2019/08/24 18:00").getTime();
+      SURVEY_START_MS_EPOCH = dateFormat.parse("2019/08/12 9:00").getTime();
+      SURVEY_END_MS_EPOCH = dateFormat.parse("2019/08/24 18:00").getTime();
     }
     catch (ParseException e) {
       // Shouldn't happen.
@@ -61,7 +61,7 @@ public class FlutterSurveyNotifications {
 
   private static boolean isSurveyOpen() {
     final long now = System.currentTimeMillis();
-    return now >= SURVEY_START && now <= SURVEY_END;
+    return now >= SURVEY_START_MS_EPOCH && now <= SURVEY_END_MS_EPOCH;
   }
 
   interface FlutterSurveyNotifier {
--- a/src/io/flutter/survey/FlutterSurveyNotifications.java
+++ b/src/io/flutter/survey/FlutterSurveyNotifications.java
@@ -38,11 +38,11 @@ public class FlutterSurveyNotifications {
   private static final String FLUTTER_LAST_SURVEY_CHECK_KEY = "FLUTTER_LAST_SURVEY_CHECK_KEY";
   private static final long CHECK_INTERVAL_IN_MS = TimeUnit.HOURS.toMillis(40);
 
-  private static final String SURVEY_ACTION_TEXT = "Take Flutter Q3 Survey";
+  private static final String SURVEY_ACTION_TEXT = "Take survey";
   private static final String SURVEY_TITLE = "Help improve Flutter! Take our Q3 survey.";
 
   private static final String SURVEY_TAKEN = "io.flutter.survey.2019.q3.alreadyTaken";
-  private static final String SURVEY_URL = "https://google.qualtrics.com/jfe/form/SV_3kiGXYfYOfXUjB3?Source=IntelliJ&ClientID=";
+  private static final String SURVEY_URL = "https://google.qualtrics.com/jfe/form/SV_3kiGXYfYOfXUjB3?Source=IntelliJ";
 
   private static long SURVEY_START_MS_EPOCH;
   private static long SURVEY_END_MS_EPOCH;
@@ -136,7 +136,7 @@ public class FlutterSurveyNotifications {
         // Add a client ID if analytics have been opted into.
         if (reportAnalytics) {
           final String clientId = FlutterInitializer.getAnalytics().getClientId();
-          url += clientId;
+          url += ("&ClientID=" + clientId);
         }
 
         BrowserUtil.browse(url);
--- a/src/io/flutter/survey/FlutterSurveyNotifications.java
+++ b/src/io/flutter/survey/FlutterSurveyNotifications.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright 2019 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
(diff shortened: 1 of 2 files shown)


src/io/flutter/survey/FlutterSurveyNotifications.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.concurrent.TimeUnit;
+
+// TODO(pq): remove for M39 / post: 2019/08/24 18:00 PST.
+public class FlutterSurveyNotifications {
+  private static final int NOTIFICATION_DELAY_IN_SECS = 3;
+
+  private static final String FLUTTER_LAST_SURVEY_CHECK_KEY = "FLUTTER_LAST_SURVEY_CHECK_KEY";
+  private static final long CHECK_INTERVAL_IN_MS = TimeUnit.HOURS.toMillis(40);
+
+  private static final String SURVEY_ACTION_TEXT = "Take Flutter Q3 Survey";
```
devoncarew: `"Take Flutter Q3 Survey"` ==> `"Take Flutter's Q3 Survey"`? Not sure where or if this string is ever displayed to users

src/io/flutter/survey/FlutterSurveyNotifications.java
```diff
@@ -0,0 +1,162 @@
...
+  private static final String SURVEY_TITLE = "Help improve Flutter! Take our Q3 survey.";
+
+  private static final String SURVEY_TAKEN = "io.flutter.survey.2019.q3.alreadyTaken";
+  private static final String SURVEY_URL = "https://google.qualtrics.com/jfe/form/SV_3kiGXYfYOfXUjB3?Source=IntelliJ&ClientID=";
+
+  private static long SURVEY_START;
```
devoncarew: Perhaps `SURVEY_START_MS_EPOCH` or similar?

src/io/flutter/survey/FlutterSurveyNotifications.java
```diff
@@ -0,0 +1,162 @@
...
+    // Limit display to the survey window.
+    if (!isSurveyOpen()) return;
+
+    final PropertiesComponent properties = PropertiesComponent.getInstance();
+
+    // Don't prompt more often than every 40 hours.
+    final long lastCheckedMillis = properties.getOrInitLong(FLUTTER_LAST_SURVEY_CHECK_KEY, 0);
+    if (System.currentTimeMillis() - lastCheckedMillis < CHECK_INTERVAL_IN_MS) return;
+
+    properties.setValue(FLUTTER_LAST_SURVEY_CHECK_KEY, String.valueOf(System.currentTimeMillis()));
```
devoncarew: I assume that `properties` values can only be strings (in terms of why you're not storing as a long)?

src/io/flutter/survey/FlutterSurveyNotifications.java
```diff
@@ -0,0 +1,162 @@
...
+      FlutterMessages.FLUTTER_NOTIFICATION_GROUP_ID,
+      FlutterIcons.Flutter,
+      SURVEY_TITLE,
+      null,
+      notificationContents,
+      NotificationType.INFORMATION,
+      null
+    );
+
+    notification.addAction(new AnAction(SURVEY_ACTION_TEXT) {
```
devoncarew: Ah, here's where this string is displayed. Perhaps something like `Sure` or `Sure!` (or just something more plain but terser than the current text, like `Take survey`)? That would go well with the `No thanks` below.

DanTup:
On the surface it all looks reasonable to me, but there's so much I don't know about Java and this project that I'm not sure that's all that useful :-)

src/io/flutter/survey/FlutterSurveyNotifications.java
```diff
@@ -0,0 +1,162 @@
...
+  private static final int NOTIFICATION_DELAY_IN_SECS = 3;
+
+  private static final String FLUTTER_LAST_SURVEY_CHECK_KEY = "FLUTTER_LAST_SURVEY_CHECK_KEY";
+  private static final long CHECK_INTERVAL_IN_MS = TimeUnit.HOURS.toMillis(40);
+
+  private static final String SURVEY_ACTION_TEXT = "Take Flutter Q3 Survey";
+  private static final String SURVEY_TITLE = "Help improve Flutter! Take our Q3 survey.";
+
+  private static final String SURVEY_TAKEN = "io.flutter.survey.2019.q3.alreadyTaken";
+  private static final String SURVEY_URL = "https://google.qualtrics.com/jfe/form/SV_3kiGXYfYOfXUjB3?Source=IntelliJ&ClientID=";
```
devoncarew: You may want to remove the `&ClientID=` suffix, and only add it if also appending a client ID.

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: I'm great with these text improvements but should defer to @jayoung-lee.  This would also diverge from VS Code...

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: This was the format @jayoung-lee suggested.  Not sure what's most desirable for easy processing...

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: 👍

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: @jayoung-lee ?

devoncarew:
> This was the format @jayoung-lee suggested. Not sure what's most desirable for easy processing...

Gotcha, but this:

> https://google.qualtrics.com/jfe/form/SV_3kiGXYfYOfXUjB3?Source=IntelliJ&ClientID=

has a hanging GET variable.

src/io/flutter/survey/FlutterSurveyNotifications.java
jayoung-lee: Does this get displayed to the users? I thought we are using "Help improve Flutter! Take our Q3 survey." for the pop up message.

src/io/flutter/survey/FlutterSurveyNotifications.java
jayoung-lee: Either way is fine :)

src/io/flutter/survey/FlutterSurveyNotifications.java
jayoung-lee: Either way (&ClientID="" or nothing) will give me the same result to me. So please do whatever works better for you.

jayoung-lee:
Please see my comments above! Thanks for getting me involved.

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: @jayoung-lee: you're correct.  The visible text is "Help improve Flutter! Take our Q3 survey."   (The other text is an internal label for the action.)

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: 👍

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: Sounds good.  Thanks!

pq:
Thanks for all the feedback!

I updated the dialog text and removed the dangling GET.

image

👍

jayoung-lee:
> Thanks for all the feedback!
> 
> I updated the dialog text and removed the dangling GET.
> 
> image
> 
> 👍

This will show “By clicking on this link you agree to share feature usage along with the survey responses.” when there's clientID to append, right?

pq:
Yep.  Thanks for checking @jayoung-lee !

image

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: EDIT: it does.  Trimmed to "Take survey"

image

src/io/flutter/survey/FlutterSurveyNotifications.java
pq: Trimmed.  Thanks!

DanTup:
I've tweaked the action text in VS Code to just "Take Survey" to match 👍

pq:
Awesome.  Thanks @DanTup!

PR #3692, opened 2019-07-23
add a user preference to opt-into showing structured errors

devoncarew:
- add a user preference to opt-into showing structured errors
- address part of https://github.com/flutter/flutter-intellij/issues/3680

Screen Shot 2019-07-22 at 9 16 40 PM


src/io/flutter/settings/FlutterSettings.java
```diff
@@ -219,6 +223,16 @@ public void setShowWebDesktopDevices(boolean value) {
     fireEvent();
   }
 
+  public boolean isShowStructuredErrors() {
+    return getPropertiesComponent().getBoolean(showStructuredErrors, false);
```
lambdabaa: Should the default here be `true` to match the previous behavior?

devoncarew: We didn't have this setting previously; the `false` will mean that the setting is not enabled unless the user explicitly toggles it. We'll keep it like that for a month or two, then change the default to `true` (and eventually, possibly remove the setting completely).

Issue #3697, opened 2019-07-24
java.lang.Throwable: Synchronous execution on EDT: emulator -list-avds (on 2019.2)


devoncarew:
Also in 2019.2:

```
java.lang.Throwable: Synchronous execution on EDT: /Users/devoncarew/Library/Android/sdk/emulator/emulator -list-avds
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:145)
	at com.intellij.execution.process.OSProcessHandler.checkEdtAndReadAction(OSProcessHandler.java:117)
	at com.intellij.execution.process.OSProcessHandler.waitFor(OSProcessHandler.java:62)
	at io.flutter.android.AndroidSdk.getEmulators(AndroidSdk.java:105)
	at io.flutter.actions.OpenEmulatorAction.getEmulatorActions(OpenEmulatorAction.java:27)
	at io.flutter.actions.DeviceSelectorAction.updateActions(DeviceSelectorAction.java:139)
	at io.flutter.actions.DeviceSelectorAction.lambda$update$2(DeviceSelectorAction.java:74)
	at com.intellij.openapi.application.impl.ApplicationImpl.invokeAndWait(ApplicationImpl.java:630)
	at io.flutter.FlutterUtils.invokeAndWait(FlutterUtils.java:63)
	at io.flutter.actions.DeviceSelectorAction.update(DeviceSelectorAction.java:73)
	at io.flutter.actions.DeviceSelectorAction.lambda$update$1(DeviceSelectorAction.java:67)
	at io.flutter.run.daemon.DeviceService.lambda$fireChangeEvent$4(DeviceService.java:151)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at com.intellij.ide.IdeEventQueue.defaultDispatchEvent(IdeEventQueue.java:873)
[...]

Issue #3696, opened 2019-07-24
java.lang.Throwable: Synchronous execution on EDT: flutter config (on 2019.2)


devoncarew:
I see this at startup in 2019.2:

```
java.lang.Throwable: Synchronous execution on EDT: /Users/devoncarew/projects/flutter/flutter/bin/flutter --no-color config --machine
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:145)
	at com.intellij.execution.process.OSProcessHandler.checkEdtAndReadAction(OSProcessHandler.java:117)
	at com.intellij.execution.process.OSProcessHandler.waitFor(OSProcessHandler.java:62)
	at io.flutter.sdk.FlutterSdk.queryFlutterConfigImpl(FlutterSdk.java:557)
	at io.flutter.sdk.FlutterSdk.queryFlutterConfig(FlutterSdk.java:521)
	at io.flutter.android.IntelliJAndroidSdk.chooseAndroidHome(IntelliJAndroidSdk.java:118)
	at io.flutter.android.AndroidSdk.createFromProject(AndroidSdk.java:36)
	at io.flutter.actions.OpenEmulatorAction.getEmulatorActions(OpenEmulatorAction.java:22)
	at io.flutter.actions.DeviceSelectorAction.updateActions(DeviceSelectorAction.java:139)
	at io.flutter.actions.DeviceSelectorAction.lambda$update$2(DeviceSelectorAction.java:74)
	at com.intellij.openapi.application.impl.ApplicationImpl.invokeAndWait(ApplicationImpl.java:630)
	at io.flutter.FlutterUtils.invokeAndWait(FlutterUtils.java:63)
	at io.flutter.actions.DeviceSelectorAction.update(DeviceSelectorAction.java:73)
	at io.flutter.actions.DeviceSelectorAction.update(DeviceSelectorAction.java:68)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performDumbAwareUpdate(ActionUtil.java:176)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.doUpdate(ActionUpdater.java:378)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.lambda$null$1(ActionUpdater.java:78)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.callAction(ActionUpdater.java:106)
[...]

PR #3700, opened 2019-07-25
address some issues with blocking the EDT thread in 2019.2

devoncarew:
- address some issues with blocking the EDT thread in 2019.2
- fix https://github.com/flutter/flutter-intellij/issues/3697, fix https://github.com/flutter/flutter-intellij/issues/3696

2019.2 changed to complaining if we use a synchronous process execution library call from the EDT thread (and request to block for greater than 10ms).

Depending on whether we receive lots of user reports of this or not, we may choose to cherry-pick this into current stable. If we see few reports, we can assume that its not affecting users, and it can wait until the next stable (eta, ~1 week).

This does have the nice side benefit that the startup of the plugin is likely faster by 1-3 seconds.

### DIFF
--- a/src/io/flutter/actions/OpenEmulatorAction.java
+++ b/src/io/flutter/actions/OpenEmulatorAction.java
@@ -9,23 +9,25 @@ import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.project.Project;
 import io.flutter.android.AndroidEmulator;
-import io.flutter.android.AndroidSdk;
+import io.flutter.sdk.AndroidEmulatorManager;
 import org.jetbrains.annotations.NotNull;
 
-import java.util.Collections;
 import java.util.List;
 
 import static java.util.stream.Collectors.toList;
 
 public class OpenEmulatorAction extends AnAction {
+  /**
+   * Retrieve a list of {@link OpenEmulatorAction}s.
+   *
+   * This list of based of of cached information from the {@link AndroidEmulatorManager}. Callers who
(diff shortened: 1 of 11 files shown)


src/io/flutter/actions/OpenEmulatorAction.java
```diff
@@ -9,23 +9,25 @@
...
-import java.util.Collections;
 import java.util.List;
 
 import static java.util.stream.Collectors.toList;
 
 public class OpenEmulatorAction extends AnAction {
+  /**
+   * Retrieve a list of {@link OpenEmulatorAction}s.
+   *
+   * This list of based of of cached information from the {@link AndroidEmulatorManager}. Callers who
```
kenzieschmoll: nit: typo in first sentence

src/io/flutter/android/AndroidEmulator.java
```diff
@@ -76,4 +76,19 @@ public void processTerminated(ProcessEvent event) {
...
+
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof AndroidEmulator)) {
+      return false;
+    }
+
+    final AndroidEmulator other = (AndroidEmulator)obj;
+    return other.id.equals(id);
+  }
```
kenzieschmoll: [optional]: this could be condensed to 
`return obj instanceof AndroidEmulator && (AndroidEmulator)obj.id.equals(id);`

src/io/flutter/actions/OpenEmulatorAction.java
devoncarew: done

src/io/flutter/android/AndroidEmulator.java
devoncarew: done

PR #3714, opened 2019-07-30
Release 38

stevemessick:
Update for M38.


CHANGELOG.md
```diff
@@ -1,3 +1,36 @@
+## 38.0
```
kenzieschmoll: nit: some of these have periods at the end and some do not. No preference on which standard we should use, but we should be consistent.

same below in xml file.

stevemessick:
Thanks! I removed them. We're gradually evolving to more consistency. Now, if we could just get everyone to use the same style in their commit messages it would be even easier :)

PR #3717, opened 2019-07-30
rev to the latest vm service protocol library

devoncarew:
- rev to the latest vm service protocol library

This brings in the latest generated Java version of the library (supporting `3.24`). Note that there are several changes here as we no longer generate code for private APIs, and some private APIs have migrated to public APIs in the spec.

(not for landing in the current release)


devoncarew:
- updated to support `3.25` (cc @bkonyi)

bkonyi:
RSLGTM!

PR #3719, opened 2019-07-31
Fix androidx modules

stevemessick:
This fixes two problems discovered during testing.

Flutter modules can use AndroidX. I thought they couldn't when I wrote the code.

The selected SDK should be used determine visibility of the AndroidX section. It wasn't getting updated properly.

### DIFF
--- a/flutter-studio/src/io/flutter/project/FlutterSettingsStep.java
+++ b/flutter-studio/src/io/flutter/project/FlutterSettingsStep.java
@@ -74,6 +74,7 @@ public class FlutterSettingsStep extends ModelWizardStep<FlutterProjectModel> {
     FlutterProjectModel model = getModel();
     myKotlinCheckBox.setText(FlutterBundle.message("module.wizard.language.name_kotlin"));
     mySwiftCheckBox.setText(FlutterBundle.message("module.wizard.language.name_swift"));
+    myBindings.bindTwoWay(new SelectedProperty(myUseAndroidxCheckBox), getModel().useAndroidX());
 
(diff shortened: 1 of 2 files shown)


flutter-studio/src/io/flutter/project/FlutterSettingsStep.java
```diff
@@ -140,6 +141,14 @@ protected ObservableBool canGoForward() {
 
   @Override
   protected void onEntering() {
+    Project project = getModel().project().getValueOrNull();
+    FlutterSdk sdk = FlutterSdk.forPath(getModel().flutterSdk().get());
+    boolean enableAndroidX = project == null && sdk != null && sdk.getVersion().isAndroidxSupported();
+    if (myAndroidXLabel.isVisible() != enableAndroidX) {
+      myAndroidXLabel.setVisible(enableAndroidX);
+      myUseAndroidxCheckBox.setVisible(enableAndroidX);
+      myRootPanel.doLayout();
```
jacob314: is the call to doLayout really needed here? I would expect you could invalidate instead of that the previous calls in this method would trigger invalidation.

stevemessick: Yes, it is needed. I tried invalidate() instead but that had no effect. If the AndroidX section isn't visible when the initial layout is done then no vertical space is allocated to it. Changing visibility doesn't force re-layout (I guess -- not a Swing expert). And if it is visible initially then there's a big gap when it becomes invisible.

stevemessick:
Thanks for the quick review! Turns out I need to make another change. Forgot to get() the value of a StringProperty.

PR #3722, opened 2019-08-02
Guard against null sdk

stevemessick:
It is possible for the Flutter SDK to be not set when the first project is created. This change allows the wizard to operate.


stevemessick:
@jacob314 gave me a lgtm in chat

Issue #3765, opened 2019-08-12
perf issue with FlutterDartAnalysisServer.processString() 


devoncarew:
Related to https://github.com/flutter/flutter-intellij/issues/3760, it looks like FlutterDartAnalysisServer.processString() is slow, or does too much work; we should investigate. More data available here: https://youtrack.jetbrains.com/issue/WEB-40646.

devoncarew:
cc @lambdabaa

PR #3772, opened 2019-08-13
only parse analysis server events we're interested in

devoncarew:
- only parse analysis server events we're interested in
- fix https://github.com/flutter/flutter-intellij/issues/3765

cc @lambdabaa


devoncarew:
Thanks!

Do you think we should change the Dart plugin to send us the already processed JSON object (`processResponse(JsonObject response)`)? I'm not sure if that's too much coupling, or a simpler solution.

Issue #3767, opened 2019-08-12
perf issue with FlutterIconProvider.getIcon()


devoncarew:
Related to https://github.com/flutter/flutter-intellij/issues/3760, it looks like FlutterIconProvider.getIcon() is slow, or does too much work; we should investigate. More data available here: https://youtrack.jetbrains.com/issue/WEB-40646.

PR #3773, opened 2019-08-13
cache the results of parsing the pubspec file

devoncarew:
- cache the results of parsing the pubspec file
- fix https://github.com/flutter/flutter-intellij/issues/3767

### DIFF
--- a/src/io/flutter/FlutterUtils.java
+++ b/src/io/flutter/FlutterUtils.java
@@ -331,6 +331,7 @@ public class FlutterUtils {
         // Check the dependencies.
         final Object dependencies = yamlMap.get("dependencies");
         if (dependencies instanceof Map) {
+          // We use `|=` for assigning to 'flutter' below as it might have been assigned to true above.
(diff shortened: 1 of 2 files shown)


src/io/flutter/FlutterUtils.java
```diff
@@ -278,129 +307,65 @@ public static PluginId getPluginId() {
...
-      // Special case the 'flutter' package itself - this allows us to run their unit tests from IntelliJ.
-      final Object name = yamlMap.get("name");
-      if ("flutter".equals(name)) {
-        return true;
-      }
+        // Check the dependencies.
+        final Object dependencies = yamlMap.get("dependencies");
+        if (dependencies instanceof Map) {
+          info.flutter |= ((Map)dependencies).containsKey("flutter");
+          info.flutterWeb = ((Map)dependencies).containsKey("flutter_web");
```
kenzieschmoll: why does only one of these use `|=`

devoncarew: The `flutter` field can sometimes be set to true above; I'll add a comment about that.

Issue #3766, opened 2019-08-12
perf issue with TestLineMarkerContributor.getInfo()


devoncarew:
Related to https://github.com/flutter/flutter-intellij/issues/3760, it looks like TestLineMarkerContributor.getInfo() is slow, or does too much work; we should investigate. More data available here: https://youtrack.jetbrains.com/issue/WEB-40646.

DaveShuckerow:
My understanding is that getInfo should only be called on LeafPsiElements, and right now we are calling it on more than that.

On large files this can slow down the file analysis.  The fix should be to get info for only the leaf nodes.

Currently, we look at TestCallExpressions, for example:

```dart
test('foo test', () {
  // All the contents of the test.
});
```

Ideally, we should only look at the leaf elements, such as the method name:

```
  test
```

Changing to support this will be difficult because we'll need to look at two leaves on the tree: the test call and the test name.  It's also necessary to do this without breaking support for right clicking inside of a test file.

devoncarew:
I think this may be slow because it's called a lot, and is calling some APIs which are themselves slow (not optimized or not cached):

<img width="1329" alt="Screen Shot 2019-08-12 at 5 31 02 PM" src="https://user-images.githubusercontent.com/1269969/62907244-23974080-bd27-11e9-9449-4b9ad9f7a464.png">

devoncarew:
I may have a solution to this, based on optimizations to `FlutterUtils.isInTestDir()`.

devoncarew:
I'm going to close this on the assumption that the primary slowdown was in FlutterUtils.isInTestDir(). We'll re-profile after the next milestone release, and open new perf issues if we see problems.

PR #3774, opened 2019-08-13
optimize FlutterUtils.isInTestDir

devoncarew:
- optimize FlutterUtils.isInTestDir
- address some or most of https://github.com/flutter/flutter-intellij/issues/3766

`FlutterUtils.isInTestDir` is called very frequently by editors when rendering, and is called from the EDT. This may have been the cause of the bulk of the precent performance issues. From some profiling, `FlutterRunConfigurationProducer.getFlutterEntryFile()` is fairly slow; this PR re-writes `isInTestDir()` to not use it.


devoncarew:
cc @DaveShuckerow

Issue #3777, opened 2019-08-13
Use the VMService instead of the daemon for all ext.flutter.inspector requests from the inspector


jacob314:
Using the VMService directly is faster and can help avoid bugs if there are problems parsing the daemon protocol output.

devoncarew:
fixed by https://github.com/flutter/flutter-intellij/pull/3790

PR #3786, opened 2019-08-15
work on supporting the flutter web inspector

### LINKED ISSUE #3770: I keep getting JSON in my logs that looks like IDE data structures
I'm occasionally seeing a lot of extra output in the log console that appears in the middle of my regular log output. I suspect that it is some kind of IDE data structure that is going to the console instead of being read by the IDE.

## Steps to Reproduce
I just launched my app, and it filled the log console with this output (note the one REALLY long line).  See this file: [console.log](https://github.com/flutter/flutter-intellij/files/3494246/console.log)

## Version info

```
[✓] Flutter (Channel route_focus, v1.8.5-pre.122, on Linux, locale en_US.UTF-8)
    • Flutter version 1.8.5-pre.122 at /usr/local/google/home/gspencer/code/flutter
    • Framework revision efd111d96e (5 weeks ago), 2019-07-09 16:30:58 -0700
    • Engine revision ff49ca1c6e
    • Dart version 2.5.0 (build 2.5.0-dev.1.0 f29f41f1a5)

[!] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
    • Android SDK at /usr/local/google/home/gspencer/Android/Sdk
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-28, build-tools 28.0.3
    • ANDROID_HOME = /usr/local/google/home/gspencer/Android/Sdk
    • Java binary at: /usr/bin/X11/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_222-8u222-b10-1-b10)
    ✗ Android license status unknown.
      Try re-installing or updating your Android SDK Manager.
      See https://developer.android.com/studio/#downloads or visit https://flutter.dev/setup/#android-setup for detailed instructions.

      (or visit https://flutter.dev/setup/#android-setup for detailed instructions).

[✓] IntelliJ IDEA Community Edition (version 2019.2)
    • IntelliJ at /opt/my-intellij-ce-current
    • Flutter plugin version 38.2.4
    • Dart plugin version 192.6262.13

[✓] Connected device (1 available)
    • Nexus 6P • 84B7N15A07007628 • android-arm64 • Android 8.1.0 (API 27)

! Doctor found issues in 2 categories.
```

devoncarew:
Work on supporting the flutter web inspector:
- move to no longer send inspector traffic over the daemon protocol - just use the vm service protocol (fix https://github.com/flutter/flutter-intellij/issues/3777; fix #3770)
- work around an issue with package:dwds involving the object id of strings; https://github.com/dart-lang/webdev/issues/575; done as a 2nd commit for easier reviewability

With this PR, the IntelliJ Inspector works with the unforked version of Flutter for web.

### DIFF
--- a/src/io/flutter/inspector/EvalOnDartLibrary.java
+++ b/src/io/flutter/inspector/EvalOnDartLibrary.java
@@ -132,33 +132,31 @@ public class EvalOnDartLibrary implements Disposable {
   public CompletableFuture<InstanceRef> eval(String expression, Map<String, String> scope, InspectorService.ObjectGroup isAlive) {
     return addRequest(isAlive, () -> {
       final CompletableFuture<InstanceRef> future = new CompletableFuture<>();
-      libraryRef.thenAcceptAsync((LibraryRef ref) -> {
-        vmService.evaluate(
-          getIsolateId(), ref.getId(), expression,
-          scope, true,
-          new EvaluateConsumer() {
-            @Override
-            public void onError(RPCError error) {
-              future.completeExceptionally(new RuntimeException(error.getMessage()));
-            }
-
-            @Override
-            public void received(ErrorRef response) {
-              future.completeExceptionally(new RuntimeException(response.toString()));
-            }
-
-            @Override
-            public void received(InstanceRef response) {
-              future.complete(response);
-            }
-
-            @Override
-            public void received(Sentinel response) {
-              future.completeExceptionally(new RuntimeException(response.toString()));
-            }
+      libraryRef.thenAcceptAsync((LibraryRef ref) -> vmService.evaluate(
+        getIsolateId(), ref.getId(), expression,
+        scope, true,
+        new EvaluateConsumer() {
+          @Override
+          public void onError(RPCError error) {
+            future.completeExceptionally(new RuntimeException(error.getMessage()));
           }
-        );
-      });
+
+          @Override
+          public void received(ErrorRef response) {
+            future.completeExceptionally(new RuntimeException(response.toString()));
+          }
+
+          @Override
+          public void received(InstanceRef response) {
+            future.complete(response);
+          }
+
+          @Override
+          public void received(Sentinel response) {
+            future.completeExceptionally(new RuntimeException(response.toString()));
+          }
+        }
+      ));
       return future;
     });
   }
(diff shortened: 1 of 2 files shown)


src/io/flutter/inspector/EvalOnDartLibrary.java
```diff
@@ -132,33 +132,31 @@ public void dispose() {
...
-            @Override
-            public void received(InstanceRef response) {
-              future.complete(response);
-            }
-
-            @Override
-            public void received(Sentinel response) {
-              future.completeExceptionally(new RuntimeException(response.toString()));
-            }
+      libraryRef.thenAcceptAsync((LibraryRef ref) -> vmService.evaluate(
```
jacob314: this change block is whitespace only, right?

devoncarew: Yeah, it's just the built-in IntelliJ refactoring to prefer expression lambdas.

src/io/flutter/inspector/InspectorService.java
```diff
@@ -240,43 +232,29 @@ private void onVmServiceReceived(String streamId, Event event) {
    * new frames will be triggered to draw unless something changes in the UI.
    */
   public CompletableFuture<Boolean> isWidgetTreeReady() {
-    if (useDaemonApi()) {
```
jacob314: the invokeServiceMethodObservatory code path is unfortunately using 
eval instead of invoking the vm service extensions via the right api. That code path should only be used when paused at a breakpoint when we cannot invoke the service extensions or for cases where we really need to be using eval due to needing to get back actual VM Service api instance refs for objects.
What I think we want instead of this CL is to invoke the VM service extensions via the VMService's api to invoke vm service extensions on an isolate instead of invoking them on a daemon protocol. :(

devoncarew: OK, I didn't follow that completely, but I gather we still want to avoid using the daemon protocol for the inspector traffic (it'll fix some existing issues, and likely be faster).

I think I'll close this PR; we still have the existing issue to avoid use of the daemon protocol. I'll re-create a smaller version of this PR with the workaround for the package:dwds issue (it'll still be correct even after the dwds fix lands).

jacob314: yes. we want to avoid the daemon protocol but continue to call the ext.flutter.inspector service extensions. Just use the VMService to call them directly instead of the daemon.
the invokeServiceMethodObservatory code path is weird mainly legacy code using eval to invoke the InspectorService.

PR #3790, opened 2019-08-16
Switch to using the VM Service directly instead of the Daemon protoco…

jacob314:
…l when invoking service extensions.

Changed the names of method calls using eval vs the VM Service extensions to be clear eval is used.
Previous the names were horrible as the name Observatory was used for Eval and Daemon was used for the VMService.
Devtools already used the VMService Extensions directly as the daemon protocol isn't available in devtools so mistakes couldn't be made.

Fyi @DaveShuckerow this might solve some of the slow inspector issues you were seeing.

### DIFF
--- a/src/io/flutter/inspector/InspectorService.java
+++ b/src/io/flutter/inspector/InspectorService.java
@@ -20,6 +20,7 @@ import io.flutter.vmService.ServiceExtensions;
 import io.flutter.vmService.VmServiceConsumers;
 import io.flutter.vmService.frame.DartVmServiceValue;
 import org.dartlang.vm.service.VmService;
+import org.dartlang.vm.service.consumer.ServiceExtensionConsumer;
 import org.dartlang.vm.service.element.*;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/inspector/InspectorService.java
```diff
@@ -352,19 +355,48 @@ private void onVmServiceReceived(String streamId, Event event) {
...
-    return getApp().callServiceExtension(ServiceExtensions.inspectorPrefix + methodName, params)
-      .thenApply((JsonObject json) -> {
-        if (json.has("errorMessage")) {
-          String message = json.get("errorMessage").getAsString();
-          throw new RuntimeException(methodName + " -- " + message);
+  CompletableFuture<JsonElement> invokeServiceExtensionNoGroup(String methodName, JsonObject params) {
+    return invokeServiceExtensionHelper(methodName, params);
+  }
+
+  private CompletableFuture<JsonElement> invokeServiceExtensionHelper(String methodName, JsonObject params) {
```
jacob314: this is the one real method added. The rest is just changing Map<String, Object> to JsonObject and renaming methods.

PR #3798, opened 2019-08-19
fix an issue parsing BoundVariable

devoncarew:
- fix an issue parsing BoundVariable

Not for landing - we'll instead need to address this upstream in the code generator (cc @bkonyi). https://github.com/dart-lang/webdev/issues/586 for context.


devoncarew:
`BoundVariable.declarationTokenPos` is a required field, so we likely don't want this change anyway.

PR #3802, opened 2019-08-20
Adapt to API changes in ASc6

stevemessick:
There were a couple changes in the Android Studio API with 3.6 canary 6.


flutter-studio/src/io/flutter/android/AndroidModuleLibraryManager.java
```diff
@@ -204,7 +204,7 @@ private void doGradleSync(Project flutterProject, Function<Project, Void> callba
     if (dir == null) dir = flutterProject.getBaseDir().findChild(".android"); // For modules.
     assert (dir != null);
     EmbeddedAndroidProject androidProject = new EmbeddedAndroidProject(FileUtilRt.toSystemIndependentName(dir.getPath()), null);
-    androidProject.init();
+    androidProject.init(null);
```
devoncarew: I assume this param is ok to be `null`?

stevemessick: It is @Nullable.

Thanks for the quick review!

PR #3809, opened 2019-08-22
Fix ArrayIndexOutOfBounds for target platform selector.

kenzieschmoll:
Added a print so that we can catch cases where we are getting an unknown platform target.

Fixes https://github.com/flutter/flutter-intellij/issues/3771.

### DIFF
--- a/src/io/flutter/view/TogglePlatformAction.java
+++ b/src/io/flutter/view/TogglePlatformAction.java
@@ -55,8 +55,13 @@ class TogglePlatformAction extends ToolbarComboBoxAction {
       }
 
       final int platformIndex = extensionDescription.getValues().indexOf(selectedPlatform.name());
-      System.out.println("Unknown platform: " + selectedPlatform.name());
-      selectorText = platformIndex == -1 ? "Platform: Unknown" : (String)extensionDescription.getTooltips().get(platformIndex);
+      try {
+        selectorText = (String)extensionDescription.getTooltips().get(platformIndex);
+      }
+      catch (ArrayIndexOutOfBoundsException ex) {
+        selectorText = "Platform: Unknown";
+        System.out.println("Unknown platform: " + selectedPlatform.name());
+      }
     }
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/view/TogglePlatformAction.java
```diff
@@ -55,7 +55,13 @@ public final void update(AnActionEvent e) {
...
 
       final int platformIndex = extensionDescription.getValues().indexOf(selectedPlatform.name());
-      selectorText = (String)extensionDescription.getTooltips().get(platformIndex);
+      try {
+        selectorText = (String)extensionDescription.getTooltips().get(platformIndex);
+      }
+      catch (ArrayIndexOutOfBoundsException ex) {
+        selectorText = "Platform: Unknown";
+        System.out.println("Unknown platform: " + selectedPlatform.name());
+      }
```
pq: Alternatively, maybe just check for `-1` rather than trapping the exception? 

Dunno if that's much better...  🤷‍♂

kenzieschmoll: done.

src/io/flutter/view/TogglePlatformAction.java
```diff
@@ -55,7 +55,13 @@ public final void update(AnActionEvent e) {
       }
 
       final int platformIndex = extensionDescription.getValues().indexOf(selectedPlatform.name());
-      selectorText = (String)extensionDescription.getTooltips().get(platformIndex);
+      if (platformIndex == -1) {
+        selectorText = "Platform: Unknown";
+        System.out.println("Unknown platform: " + selectedPlatform.name());
```
devoncarew: Perhaps, instead:

```
private static final Logger LOG = Logger.getInstance(TogglePlatformAction.class);
```

and

```
LOG.warn("Unknown platform: " + selectedPlatform.name());
```

That will show up in the IntelliJ log.

PR #3817, opened 2019-08-26
[WIP] Support co-editing Flutter and Android in a single project

stevemessick:
**Obsolete.** Replaced by https://github.com/flutter/flutter-intellij/pull/3850

Caveat: only works for Android Studio; IntelliJ support will be added later.

TODO: Find a better way to manage Maven repositories in IntelliJ. The project structure dialog is way too slow.
TODO: Clean up module dependencies after Android Q is released.

There is a comment in AndroidUtils about Add-to-app notes. It will be deleted before the co-edit work is finalized (after Android Q) but I want it to stay for now. Likewise, #triggerSyncIfFlutterModuleFoundAfterBuild has been commented-out and will probably be removed, but there's a (rare?) use case that might require it.


stevemessick:
Close due to replacement by #3850

PR #3839, opened 2019-08-31
Cache values that cause Edt-sync warnings

stevemessick:
Finding Android Studio can cause a warning about synchronous execution on EDT. This will (probably) fix that. It needs more testing and should be queued for M40.

### DIFF
--- a/src/io/flutter/ProjectOpenActivity.java
+++ b/src/io/flutter/ProjectOpenActivity.java
@@ -10,6 +10,7 @@ import com.intellij.notification.NotificationType;
 import com.intellij.notification.Notifications;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.DumbAware;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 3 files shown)


src/io/flutter/ProjectOpenActivity.java
```diff
@@ -55,6 +55,9 @@ public void runActivity(@NotNull Project project) {
       return;
     }
 
+    sdk.queryFlutterConfig("android-sdk", false);
```
devoncarew: This will delay the items after this in runActivity() method. I don't know if they're time sensitive, or if the return from runActivity() is (if slowing it down will impact other things).

Perhaps, run these two queries on a separate thread via something like `JobScheduler.getScheduler().schedule(...)`?

src/io/flutter/actions/OpenInAndroidStudioAction.java
```diff
@@ -198,7 +198,10 @@ private static String findAndroidStudio(@Nullable Project project) {
 
     final FlutterSdk flutterSdk = FlutterSdk.getFlutterSdk(project);
     if (flutterSdk != null) {
-      String androidSdkLocation = flutterSdk.queryFlutterConfig("android-studio-dir", false);
+      String androidSdkLocation = flutterSdk.queryFlutterConfig("android-studio-dir", true);
```
devoncarew: The `useCachedValue` flag will return a value if one is cached; if there's no cached value, it'll call through to querying `flutter config`; I don't think we need the 2nd call to `queryFlutterConfig()` here.

src/io/flutter/ProjectOpenActivity.java
stevemessick: Finding the Android SDK happens later during project opening, so it isn't needed here. Finding Android Studio can certainly happen in the background, so I changed that.

src/io/flutter/actions/OpenInAndroidStudioAction.java
stevemessick: The File.exists() is key here. The internet is full of advice about using `flutter config --android-studio-dir` which will invalidate the cache if done while the IDE is active.

devoncarew: ack

PR #3842, opened 2019-09-02
add a reload/restart delay for flutter web

devoncarew:
- add a reload/restart delay for flutter web

When reloading (or restarting), we save all files before we reload. We then issue a reload (or a restart) command to the flutter process. However, it looks like the package:build based compilation service used in flutter web only knows about file changes when notified by its own watcher. So, there's a race condition:

- IntelliJ saves file contents
- tells flutter daemon to restart a flutter web app
- it restarts
- it gets notified of changes top the file system (from the save at step 1)
- the changes get compiled

We nearly always loose this race condition. This PR introduces a 125 ms delay, which means we now generally win this race condition. I'll open an issue on flutter/flutter to address the underlying issue.

### DIFF
--- a/src/io/flutter/run/FlutterReloadManager.java
+++ b/src/io/flutter/run/FlutterReloadManager.java
@@ -104,6 +104,7 @@ public class FlutterReloadManager {
     this.myProject = project;
     this.mySettings = FlutterSettings.getInstance();
 
+    // TODO(devoncarew): Prefer using AnActionListener.TOPIC.
     ActionManagerEx.getInstanceEx().addAnActionListener(new AnActionListener.Adapter() {
       private @Nullable Project eventProject;
       private @Nullable Editor eventEditor;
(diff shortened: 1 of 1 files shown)


jonahwilliams:
I think we should add this fix to the flutter tool instead of intellij, because it will be easier to both roll out and remove

devoncarew:
> I think we should add this fix to the flutter tool instead of intellij, because it will be easier to both roll out and remove

Do we think a fix is to hand? In terms of where to put a workaround, if there's a handy place in the flutter tool, that would work. We are able to iterate faster in the IntelliJ plugin; the IntelliJ plugin releases monthly and flutter every 3 months; w/ IntelliJ, we can cherry-pick changes into a release at relatively low cost, where the flutter cherry-pick process is fairly heavyweight.

jonahwilliams:
Well for the time being we require users of flutter for web to be on dev/master. Also, other editors would likely require this workaround too

devoncarew:
Closing in favor of https://github.com/flutter/flutter/pull/39752.

Issue #3256, opened 2019-03-05
Provide shortcuts for Theme.of() and similar


feinstein:
Flutter doesn't have many boilerplate code, but still sometimes we need to do a lot of repetitive code such as `Theme.of(context)`, and many other with this same pattern.

It would be nice to have some templates ready for this, such as `stless` does for creating a new `StatelessWidget` class.

pq:
Great idea.  Thanks @feinstein!

feinstein:
My pleasure.

Thinking more about it, I think it would be useful to have several templates for the framework's `of(context)`, that would expand something like `thof` to `Theme.of(context)` and one for facilitating user implemented `InheritedWidget` that would expand `ihof` to `[complete-here].of`.

PR #3846, opened 2019-09-03
inherited theme.of live template

pq:
Fixes: #3256 

image

/cc @stevemessick 

/fyi @feinstein


feinstein:
Thanks for the update :)

PR #3864, opened 2019-09-10
Fix a couple AS issues

stevemessick:
A couple changes:
 - Reset default for AndroidX to true
 - Ensure Flutter projects opened in Android Studio have project-type = "io.flutter"

### DIFF
--- a/src/io/flutter/ProjectOpenActivity.java
+++ b/src/io/flutter/ProjectOpenActivity.java
@@ -59,6 +59,9 @@ public class ProjectOpenActivity implements StartupActivity, DumbAware {
     ApplicationManager.getApplication().executeOnPooledThread(() -> {
       sdk.queryFlutterConfig("android-studio-dir", false);
     });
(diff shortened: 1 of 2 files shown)


src/io/flutter/ProjectOpenActivity.java
```diff
@@ -59,6 +59,9 @@ public void runActivity(@NotNull Project project) {
     ApplicationManager.getApplication().executeOnPooledThread(() -> {
       sdk.queryFlutterConfig("android-studio-dir", false);
     });
+    if (FlutterUtils.isAndroidStudio() && !FLUTTER_PROJECT_TYPE.equals(ProjectTypeService.getProjectType(project))) {
```
devoncarew: We already have similar code to this at the end of the method.

Is this code different? Or, should we unify the two cases somehow?

stevemessick: Those lines got deleted by the co-edit PR. I should probably remove them here too.

PR #3865, opened 2019-09-10
remove support for AS 3.4

devoncarew:
- remove support for AS 3.4

@stevemessick @DaveShuckerow


devoncarew:
From discussion with @stevemessick, there may be some follow on work to remove AS 3.4 specific build customization.

PR #3866, opened 2019-09-10
more normalization of flutter error codes

devoncarew:
- more normalization of flutter error codes

Ensure that we normalize doubles as well as ints.

@jacob314

### DIFF
--- a/src/io/flutter/logging/FlutterErrorHelper.java
+++ b/src/io/flutter/logging/FlutterErrorHelper.java
@@ -13,7 +13,7 @@ import io.flutter.inspector.DiagnosticsNode;
 import java.util.regex.Pattern;
 
 public class FlutterErrorHelper {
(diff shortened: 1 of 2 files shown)


src/io/flutter/logging/FlutterErrorHelper.java
```diff
@@ -13,7 +13,7 @@
 import java.util.regex.Pattern;
 
 public class FlutterErrorHelper {
-  private static final Pattern numberPattern = Pattern.compile("[0-9]+");
+  private static final Pattern numberPattern = Pattern.compile("[0-9][0-9.]*");
```
jacob314: nit: 
`[0-9]+[.][0-9]*`
would be a little more precise and avoid catching 10.1.0.33

devoncarew: updated to `[0-9]+(\\.[0-9]+)?`

PR #3868, opened 2019-09-11
Add build actions

stevemessick:
This is needed for add-to-app integration (aar) and convenient for the others.

Also the `Flutter > Open in Android Studio` has been flakey for a while. I think this is due to recently-added timeouts in the event handling code. There's a change that should help improve performance of our `update()` method for that action, hopefully getting past the flakiness.

### DIFF
--- a/src/io/flutter/ProjectOpenActivity.java
+++ b/src/io/flutter/ProjectOpenActivity.java
@@ -59,6 +59,9 @@ public class ProjectOpenActivity implements StartupActivity, DumbAware {
     ApplicationManager.getApplication().executeOnPooledThread(() -> {
       sdk.queryFlutterConfig("android-studio-dir", false);
     });
+    if (FlutterUtils.isAndroidStudio() && !FLUTTER_PROJECT_TYPE.equals(ProjectTypeService.getProjectType(project))) {
+      ProjectTypeService.setProjectType(project, FLUTTER_PROJECT_TYPE);
+    }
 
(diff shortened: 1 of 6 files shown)


devoncarew:
Can you add a screenshot of the UI additions (I assume just an additional submenu on the `Build` menu?)?

src/io/flutter/ProjectOpenActivity.java
```diff
@@ -59,6 +59,9 @@ public void runActivity(@NotNull Project project) {
     ApplicationManager.getApplication().executeOnPooledThread(() -> {
       sdk.queryFlutterConfig("android-studio-dir", false);
     });
+    if (FlutterUtils.isAndroidStudio() && !FLUTTER_PROJECT_TYPE.equals(ProjectTypeService.getProjectType(project))) {
```
devoncarew: Was this change intended for the PR?

src/io/flutter/utils/FlutterModuleUtils.java
```diff
@@ -324,6 +329,22 @@ public static boolean isDeprecatedFlutterModuleType(@NotNull Module module) {
     return FlutterModuleUtils.declaresFlutter(module);
   }
 
+  public static boolean isInFlutterAndroidModule(@NotNull Project project, @NotNull VirtualFile file) {
```
devoncarew: Is this impl. the same as the private method of the same name in `FlutterExternalIdeActionGroup.java`?

PR #3877, opened 2019-09-13
convert to using the `getMemoryUsage()` API

devoncarew:
- convert the memory monitor in IntelliJ to use the service protocol's `getMemoryUsage()` API
- refactor the code that manages the list of current isolates to make that information available to the heap monitor (we need fewer RPC calls that way)

We do lose the RSS (resident set size - total process memory) info with this PR (we weree reading this from a private field from the VM object - `_currentRSS`; we'll likely want to find a way to add this info back, as the total memory used by the Flutter process will be useful to users.


devoncarew:
@terrylucas, we'll likely want to make a similar change to DevTools - to use the new memory API instead of polling the VM object and Isolates.

PR #3890, opened 2019-09-17
Query display refresh rate and use in Performance window.

kenzieschmoll:
@jacob314

### DIFF
--- a/src/io/flutter/vmService/VMServiceManager.java
+++ b/src/io/flutter/vmService/VMServiceManager.java
@@ -5,12 +5,17 @@
  */
 package io.flutter.vmService;
 
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.intellij.openapi.Disposable;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.util.Disposer;
 import com.intellij.openapi.util.text.StringUtil;
 import gnu.trove.THashMap;
 import io.flutter.inspector.EvalOnDartLibrary;
+import io.flutter.run.daemon.DaemonApi;
 import io.flutter.run.daemon.FlutterApp;
 import io.flutter.utils.EventStream;
 import io.flutter.utils.StreamSubscription;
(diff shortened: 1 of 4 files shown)


devoncarew:
You may want to re-merge master into this PR's branch to see if it addresses the issues on travis.

src/io/flutter/vmService/VMServiceManager.java
```diff
@@ -355,6 +367,12 @@ private void onFrameEventReceived() {
         addServiceExtension(extensionName);
       }
       pendingServiceExtensions.clear();
+
+      // Query for display refresh rate and add the value to the stream.
```
jacob314: it might have also been a deadlock because this code is within a synchronized block.
My rule of thumb is to avoid thinking of complex locking cases by never having code that performs any actions that might trigger a second lock while already inside one lock.
You could try moving this code after the end on the `synchronized (serviceExtensions)` block to see if that fixes the issue.

src/io/flutter/vmService/VMServiceManager.java
```diff
@@ -560,6 +578,120 @@ private void unregisterService(String serviceName) {
     }
   }
 
+  public int getTargetMicrosPerFrame() {
+    Double fps = getCurrentDisplayRefreshRateRaw();
+    if (fps == null) {
+      fps = defaultRefreshRate;
+    }
+    return Math.round(((float) Math.floor(1000000.0f / fps)));
```
jacob314: nit: cast to an int at the very end instead of rounding the double to a float.

src/io/flutter/vmService/VMServiceManager.java
```diff
@@ -560,6 +578,120 @@ private void unregisterService(String serviceName) {
...
+  private CompletableFuture<String> getFlutterViewId() {
+    return getFlutterViewsList().thenApplyAsync((JsonElement element) -> {
+      final JsonArray viewsList = element.getAsJsonObject().get("views").getAsJsonArray();
+      for (JsonElement jsonElement : viewsList) {
+        final JsonObject view = jsonElement.getAsJsonObject();
+        if (view.get("type").getAsString().equals("FlutterView")) {
+          return view.get("id").getAsString();
+        }
+      }
+      throw new RuntimeException("No Flutter views to query: " + viewsList.toString());
```
jacob314: will this exception get surfaced to users? it could make sense to fail gracefully if flutter happens to not really be running any views yet.

kenzieschmoll: No it won't. We will log it as a warning and then return the default refresh rate.

src/io/flutter/vmService/VMServiceManager.java
kenzieschmoll: hmm except I don't want `getDisplayRefreshRate()` called on every frame. We need to short circuit if `firstFrameEventReceived == true`.

kenzieschmoll: I tried a few variations of pulling the call outside the synchronized block, but all resulted in the same situation where we don't ever get a response. Leaving as is.

jacob314: sg. fine to leave as is. you can also set a boolean variable indicating where to short circuit outside the synchronization block.

PR #3895, opened 2019-09-19
Make devtools run from bazel instead of pub in a bazel workspace.

DaveShuckerow:


src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -122,19 +132,63 @@ public void openBrowserAndConnect(String uri, String page) {
...
+    @Nullable final PubRoot pubRoot = PubRoots.forProject(project).stream().findFirst().orElse(null);
+    final FlutterCommand command = sdk.flutterPackagesPub(pubRoot, "global", "run", "devtools", "--machine", "--port=0");
+
+    // TODO(devoncarew): Refactor this so that we don't use the console to display output - this steals
+    // focus away from the Run (or Debug) view.
+    return command.startInConsole(project);
+  }
+
+  /**
+   * Gets the process handler that wills tart DevTools in bazel.
```
devoncarew: sp: start

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -122,19 +132,63 @@ public void openBrowserAndConnect(String uri, String page) {
...
+    final GeneralCommandLine commandLine = new GeneralCommandLine().withWorkDirectory(workspace.getRoot().getPath());
+    commandLine.setExePath(FileUtil.toSystemDependentName(workspace.getRoot().getPath() + "/" + workspace.getLaunchScript()));
+    commandLine.setCharset(CharsetToolkit.UTF8_CHARSET);
+    commandLine.addParameters(workspace.getDevtoolsScript(), "--", "--machine", "--port=0");
+    OSProcessHandler handler;
+    try {
+      handler = new OSProcessHandler(commandLine);
+    }
+    catch (ExecutionException e) {
+      e.printStackTrace();
```
devoncarew: Should this be reported to users? Will returning null for the handler cause an issue to be reported for the failure to launch?

We should `Log.` this here, likely at the warn level.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -122,19 +132,63 @@ public void openBrowserAndConnect(String uri, String page) {
...
     if (devToolsInstance != null) {
       devToolsInstance.openBrowserAndConnect(uri, page);
       return;
     }
 
-    final FlutterSdk sdk = FlutterSdk.forPubOrBazel(project);
-    if (sdk == null) {
-      return;
-    }
+    final OSProcessHandler handler = isBazel(project) ? getProcessHandlerForBazel() : getProcessHandlerForPub();
```
devoncarew: Here, I would annotate `handler` as Nullable, since getProcessHandlerForBazel() can return null.

Do we handle null in the code below?

src/io/flutter/devtools/DevToolsManager.java
DaveShuckerow: Done.

src/io/flutter/devtools/DevToolsManager.java
DaveShuckerow: Done.

src/io/flutter/devtools/DevToolsManager.java
DaveShuckerow: Annotated.

We do handle nulls in the code that starts DevTools.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -166,6 +173,7 @@ private OSProcessHandler getProcessHandlerForBazel() {
       handler = new OSProcessHandler(commandLine);
     }
     catch (ExecutionException e) {
+      FlutterUtils.warn(LOG, e);
       e.printStackTrace();
```
devoncarew: I'd remove the `e.printStackTrace();` call here.

DaveShuckerow: Done.

PR #3914, opened 2019-09-29
address a compilation problem against 2019.2

devoncarew:
- address a compilation problem against 2019.2

I wasn't able to compile locally against 2019.2; I'm not sure if user's would see this issue or not, but in-lining a small bit of the android `SdkConstants` seems low risk.

@stevemessick


devoncarew:
I'll validate that this is or isn't necessary locally - this could have been a symptom of a setup issue.

Issue #3898, opened 2019-09-22
Bring back "Remove widget" refactor


InMatrix:
It seems that the "Remove widget" refactor was renamed to "Replace widget with its children" at some point. The new name doesn't seem to register in users' mind, as I have seen several users asking where the refactor went:
* https://twitter.com/meAndAlfred/status/1161044962856882183
* https://www.reddit.com/r/FlutterDev/comments/d7jahb/is_there_a_shortcut_to_remove_a_whole_widget_from/

Can we consider bringing the old name back?

DanTup:
I think the name change may have come about due to confusion with the old name (I think it came up in the Dart-Code issue tracker). There's also a comment on the reddit thread linked above:

> When I hear remove widget it sounds like it should remove the whole subtree at given widget. Then replace should well replace the widget with its children.

There are some additional comments here:

https://www.reddit.com/r/FlutterDev/comments/d4l8qt/action_of_remove_widget_is_gone_in_vs_code/

including:

> 'Remove child' heavily implied that the whole widget would get removed (including it's child/children) but it only replaced the parent with it's child/children, so it made sense to change the name.

and

> Edit: My reasoning is actually wrong here, since all the descendants are also part of that widget. My bad.

My interpretation is that maybe neither name is clear. I think if you're looking to remove a widget (but leave it's children) looking for a "remove widget" refactor in isolation makes sense. However when you're not in that situation, being presented with "Remove widget" is a big ambigious.

Maybe we could have a name that contains "Remove widget" but is less vague (for example something along the line of "Remove widget (keep child widgets)" might suit both needs, though could probably be worded better). Having an action to remove the whole sub-tree ("Remove children (including children)") might also remove ambiguity?

escamoteur:
What about 'remove this widget?

devoncarew:
Sounds like something like renaming to:

- `Remove widget (keep child widgets)`
- or, `Remote this widget`

would increase clarity and discoverability.

Perhaps, we could add another refactoring option to help draw a distinction between what it does and around what the `Replace widget with its children` refactoring does?

- `Remove this widget` and `Remove widget and children`

InMatrix:
> Remove this widget and Remove widget and children

This sounds reasonable, though I'm not sure how often the latter will be used. Maybe exposing `Extract widget and children` is more useful than `Remove widget and children`?

PR #3915, opened 2019-09-29
rename the remove widget action

devoncarew:
- rename the "Replace widget with its children" to "remove widget action"
- fix https://github.com/flutter/flutter-intellij/issues/3898
- remove the feedback button from the Flutter Outline view - this was added to get initial feedback on the view
- move an action (the show only widgets filter) to the 'title actions' toolbar for the view; this gives more horizontal space for the build method manipulation actions

### DIFF
--- a/src/io/flutter/preview/PreviewView.java
+++ b/src/io/flutter/preview/PreviewView.java
@@ -9,7 +9,10 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.Uninterruptibles;
 import com.intellij.icons.AllIcons;
-import com.intellij.ide.*;
+import com.intellij.ide.CommonActionsManager;
+import com.intellij.ide.DataManager;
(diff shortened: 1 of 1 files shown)


InMatrix:
@DanTup The action's name in VS Code probably needs to be updated, too.

DanTup:
In almost all places in VS Code, this text comes straight from the server. We did have "Remove widget" on the context menu for the Flutter Outline though, so I've changed that to "Remove this widget" to match.

@devoncarew is this being done in the server too?

devoncarew:
@scheglov - does this action have a text name in the server, or just the ID (`dart.assist.flutter.removeWidget`)?

scheglov:
It has both id and text.
https://github.com/dart-lang/sdk/blob/ad7374fcdb9f2af9fd75edde86efc5c2893d4975/pkg/analysis_server/lib/src/services/correction/assist.dart#L136

devoncarew:
https://dart-review.googlesource.com/c/sdk/+/119480

Issue #3874, opened 2019-09-13
ConcurrentModificationException while editing source code


ksilz:
## What happened

Changing source code and reloading. Application was running in debug mode in iOS simulator.

## Version information

IntelliJ IDEA `2019.2.2` • Flutter plugin `io.flutter 39.0.5` • Dart plugin `192.6603.23`

Flutter 1.9.1+hotfix.2 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 2d2a1ffec9 (6 days ago) • 2019-09-06 18:39:49 -0700
Engine • revision b863200c37
Tools • Dart 2.5.0

## Exception

null

```
java.util.ConcurrentModificationException
	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1445)
	at java.util.HashMap$ValueIterator.next(HashMap.java:1474)
	at io.flutter.perf.FlutterWidgetPerf.clearModels(FlutterWidgetPerf.java:461)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:312)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.doRun(LaterInvocator.java:433)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.runNextEvent(LaterInvocator.java:416)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:399)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:764)
	at java.awt.EventQueue.access$500(EventQueue.java:98)
	at java.awt.EventQueue$3.run(EventQueue.java:715)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:734)
	at com.intellij.ide.IdeEventQueue.defaultDispatchEvent(IdeEventQueue.java:878)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:827)
	at com.intellij.ide.IdeEventQueue.lambda$dispatchEvent$8(IdeEventQueue.java:466)
[...]

PR #3916, opened 2019-09-29
fix a ConcurrentModificationException

devoncarew:
- fix a ConcurrentModificationException from FlutterWidgetPerf.java (reported twice from users)
- fix https://github.com/flutter/flutter-intellij/issues/3874

cc @jacob314

### DIFF
--- a/src/io/flutter/perf/FlutterWidgetPerf.java
+++ b/src/io/flutter/perf/FlutterWidgetPerf.java
@@ -101,8 +101,10 @@ public class FlutterWidgetPerf implements Disposable, WidgetPerfListener {
   }
 
   private void onFrame(ActionEvent event) {
-    for (EditorPerfModel decorations : editorDecorations.values()) {
-      decorations.onFrame();
+    synchronized (editorDecorations) {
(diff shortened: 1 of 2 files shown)


src/io/flutter/perf/FlutterWidgetPerf.java
```diff
@@ -101,8 +101,10 @@ private void setRequestInProgress(boolean value) {
   }
 
   private void onFrame(ActionEvent event) {
-    for (EditorPerfModel decorations : editorDecorations.values()) {
-      decorations.onFrame();
+    synchronized (editorDecorations) {
```
jacob314: this is fine as a workaround. The right fix though is probably to  find which of these methods is being called on a background thread. It looks like onFrame is probably the culprit. Could make sense to add asserts for all the methods that we are on the UI thread to avoid the issue and add  `AsyncUtils.invokeLater(() -> {`
calls for any ones that are unexpectedly actually sometimes called on a background thread.

devoncarew:
Closing in favor of #3939.

Issue #3925, opened 2019-10-02
NPE from VMServiceManager.getFlutterViewsList() (VMServiceManager.java:662)


lookfirst:
Please file this bug report at https://github.com/flutter/flutter-intellij/issues/new.

---

## What happened

I have no idea.

## Version information

IntelliJ IDEA `2019.3 EAP` • Flutter plugin `io.flutter 40.0.5` • Dart plugin `193.3793.14`

No Flutter sdk configured.
(Untrue... it just isn't being detected here?)

## Exception

null

```
java.lang.NullPointerException
	at io.flutter.vmService.VMServiceManager.getFlutterViewsList(VMServiceManager.java:662)
	at io.flutter.vmService.VMServiceManager.getFlutterViewId(VMServiceManager.java:647)
	at io.flutter.vmService.VMServiceManager.getDisplayRefreshRate(VMServiceManager.java:607)
	at io.flutter.vmService.VMServiceManager.lambda$onFrameEventReceived$1(VMServiceManager.java:372)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:309)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.doRun(LaterInvocator.java:441)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.runNextEvent(LaterInvocator.java:424)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:407)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:313)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:776)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:727)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:721)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:85)
	at java.desktop/java.awt.EventQueue.dispatchEvent(EventQueue.java:746)
	at com.intellij.ide.IdeEventQueue.defaultDispatchEvent(IdeEventQueue.java:906)
	at com.intellij.ide.IdeEventQueue._dispatchEvent(IdeEventQueue.java:779)
[...]

PR #3927, opened 2019-10-02
Fix NPE in when querying display refresh rate.

kenzieschmoll:
Fail gracefully when calling getFlutterViewsList if current isolate is null.

Fixes https://github.com/flutter/flutter-intellij/issues/3925.

### DIFF
--- a/src/io/flutter/vmService/VMServiceManager.java
+++ b/src/io/flutter/vmService/VMServiceManager.java
@@ -571,7 +571,7 @@ public class VMServiceManager implements FlutterApp.FlutterAppListener, Disposab
     if (fps == null) {
       fps = defaultRefreshRate;
     }
-    return (int) Math.round((Math.floor(1000000.0f / fps)));
+    return (int)Math.round((Math.floor(1000000.0f / fps)));
   }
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/vmService/VMServiceManager.java
```diff
@@ -644,7 +644,9 @@ public void received(JsonObject object) {
   }
 
   private CompletableFuture<String> getFlutterViewId() {
-    return getFlutterViewsList().thenApplyAsync((JsonElement element) -> {
+    return getFlutterViewsList().exceptionally(exception -> {
+      throw new RuntimeException(exception.getMessage());
```
devoncarew: Just confirming, when we complete exceptionally here (and below on the `ret.completeExceptionally()` line) - is that exception displayed to users? Or do we fall back on assuming 60hz?

(I'd want to make sure that these issues are not surfaced to the users as exceptions)

kenzieschmoll: The exception is not displayed to users - we fall back to 60 and log what happened as a warning.

PR #3981, opened 2019-10-16
various bits of code cleanup

devoncarew:
various bits of code cleanup:
- remove the use of icon references scheduled for deletion in 2020.1
- clean up unused imports
- silence a warning about duplicate switch statement cases


src/io/flutter/actions/InstallSdkAction.java
```diff
@@ -38,9 +38,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-@SuppressWarnings("ComponentNotRegistered")
```
stevemessick: Please leave this annotation. Removing it causes a warning in my dev env (Android Studio).

stevemessick: Please leave this annotation. Removing it causes a warning for me (using Android Studio settings).

src/io/flutter/vmService/DartVmServiceListener.java
```diff
@@ -40,6 +40,7 @@ public void connectionOpened() {
 
   }
 
+  @SuppressWarnings("DuplicateBranchesInSwitch")
```
stevemessick: Why is this needed? I don't see any dups, and I don't have a warning about it. I do have a warning about missing branches (Logging, VMFlagUpdate).

devoncarew: Interesting. I don't see the missing branches warnings. The duplicated case statements refer to the cases that all just end in `break;`.

src/io/flutter/actions/InstallSdkAction.java
devoncarew: will do

devoncarew: will do

PR #4004, opened 2019-10-22
Co-edit module created in Android Studio

stevemessick:
Add the ability to create an add-to-app module directly to Android Studio.

Needs work: Gradle sync generates an error claiming that a module cannot be compiled. That error is clean up in a later pass, but for now I can't find a way to prevent it.

`File > New > New Module...` looks like this for an Android app with no Flutter components, after scrolling to the bottom of the list:

Screen Shot 2019-10-22 at 9 38 22 AM

And, yes, there is a TODO to differentiate the icons.

`Import Flutter Module` already exists and is how existing modules are linked into Android apps. The new `Flutter Module` is for creating add-to-app modules.

After a Flutter module has been added it has a few more options:

Screen Shot 2019-10-22 at 9 41 48 AM

This is quite a lengthy process. In order to link the Flutter module into the Android app we first have to build it. Running `flutter build aar` takes about 5 minutes on my machine. In addition we have to run Gradle sync twice: once after the AAR is built to pick up the Flutter Android Gradle module, and again after that to fully link the module into the app to enable co-editing. The two sync's can run in just a few seconds, though, depending on project complexity.

Switching the Project view to the Project Files option provides good visibility of both Flutter and Android sources.


flutter-studio/flutter-studio.iml
```diff
@@ -48,5 +48,6 @@
     <orderEntry type="library" name="com.android.tools:common" level="project" />
     <orderEntry type="library" name="com.android.tools:repository" level="project" />
     <orderEntry type="library" name="studio-analytics-proto" level="project" />
+    <orderEntry type="module" module-name="intellij.groovy.psi" />
```
devoncarew: Is this used? I don't see any references in the code to a groovy package.

flutter-studio/src/io/flutter/module/FlutterModuleModel.java
```diff
@@ -5,27 +5,215 @@
...
+      // Link the new module into the Gradle project, which also enables co-editing.
+      new FlutterGradleLinker(this).linkNewModule();
     }
     else {
-      assert (!projectName().get().isEmpty());
-      assert (!flutterSdk().get().isEmpty());
-      new FlutterProjectCreator(this).createModule();
+      if (projectType().get().isPresent() && projectType().get().get() == FlutterProjectType.IMPORT) {
+        String location = projectLocation().get();
+        assert (!location.isEmpty());
```
devoncarew: Do we want to assert here? If the location is empty, that will throw an exception.

I don't see where the `location` local variable is used.

devoncarew:
So, after this PR:

File > New > New Flutter Project will so:
- Flutter Application
- Flutter Plugin
- Flutter Package
- Flutter Module

File > New > New Module will show:
- Flutter Module
- Import Flutter Module

And the Flutter Module screen will show the additional four options?

It sounds like most users will still go through the `File > New > New Flutter Project` route?

flutter-studio/src/io/flutter/module/FlutterModuleModel.java
```diff
@@ -5,27 +5,215 @@
...
+  }
+
   @Override
   protected void handleFinished() {
     // Do not call the superclass method.
-    if (projectType().get().isPresent() && projectType().get().get() == FlutterProjectType.IMPORT) {
-      String location = projectLocation().get();
-      assert (!location.isEmpty());
+    if (isModule()) {
+      // The FlutterProjectType.MODULE is used in two places. In FlutterProjectModel it is used to create a new top-level
```
devoncarew: Is `handleFinished()` long running? Can or should anything here be done in a background task, w/ progress reported to the user?

stevemessick:
After this PR the New Flutter Project wizard does not change. The New Module wizard does not change if the host project is Flutter. If the host project is Android without Flutter then the New Module wizard shows two items for Flutter (first screen shot). If the host project is Android that has an add-to-app Flutter module then the New Module wizard shows four items for Flutter (second screen shot).

> It sounds like most users will still go through the File > New > New Flutter Project route?

I expect that to be true. This is only for those people who need to add Flutter to an existing Android app and want to have the IDE do it for them. IIRC there is a relatively low number of people interested today, but a recent survey indicated growing interest. If the tool makes it easy to experiment then I'd expect the number to grow, but still be significantly lower that pure-Flutter users.

One idea for dealing with the Gradle error is to wrap the long-running code in a modal progress indicator. Doing that would prevent making any changes to the code while the analysis and conversion is completing. It might be possible to "clean up" the error message, too. I haven't found any way to prevent it, though.

flutter-studio/flutter-studio.iml
stevemessick: Yes. I tried taking it out and got a compile error.

flutter-studio/src/io/flutter/module/FlutterModuleModel.java
stevemessick: It's been a long time since I wrote that and don't remember why I wanted that assert. I think it is last-chance error checking since without a location we can't do an import. But the wizard verification code shouldn't let the user get to this point without a valid location.

flutter-studio/src/io/flutter/module/FlutterModuleModel.java
stevemessick: Yes, it runs a while. Progress is reported via the console and the status bar at the bottom of the window. It is in here that the error is generated. I'm thinking it might be better to wrap a modal progress indicator around all this, but that's another issue.

stevemessick:
There's no need to add another background task; that stuff is already running in the background. The user is free to edit or whatever while the Flutter build and Gradle sync are running.

Thanks for the review!

Issue #3952, opened 2019-10-10
ArrayIndexOutOfBounds exception from FlutterLogView.java


devoncarew:
We've seen dozens of these from our internal crash reporting:

```
javax.swing.DefaultRowSorter.convertRowIndexToModel(), DefaultRowSorter.java:518
javax.swing.JTable.od_name: "convertRowIndexToModel(), JTable.java:2645
io.flutter.logging.FlutterLogTree.lectedNodes(), FlutterLogTree.java:447
io.flutter.logging.FlutterLogView.eDataPanel(), FlutterLogView.java:614
sun.reflect.GeneratedMethodAccessor884. "invoke(), : lct.DelegatingMethodAccessorImpl. "invoke(), Djava.lang.reflect.Method.e: "invoke(), Method.java:498
com.intellij.util.EventDispatcher.tchVoidMethod(), EventDispatcher.java:131
com.intellij.util.EventDispatcher.cess$000(), EventDispatcher.java:25
com.intellij.util.EventDispatcher$1. "invoke(), EventDispatcher.java:89
io.flutter.logging.$Proxy175.selectionChanged(), : tr.logging.FlutterLogTree.da$new$1(), Fjavax.swing.JTree.hod_name: "fireValueChanged(), JTree.java:2934
javax.swing.JTree$TreeSelectionRedirector.eChanged(), JTree.java:3398
javax.swing.tree.DefaultTreeSelectionModel.eChanged(), DefaultTreeSelectionModel.java:635
javax.swing.tree.DefaultTreeSelectionModel.thChange(), DefaultTreeSelectionModel.java:1093
javax.swing.tree.DefaultTreeSelectionModel.ionPaths(), DefaultTreeSelectionModel.java:412
com.intellij.ui.treeStructure.treetable.TreeTable$ListToTreeSelectionModelWrapper.ctedRows(), TreeTable.java:330
com.intellij.ui.treeStructure.treetable.TreeTable$ListToTreeSelectionModelWrapper$ListSelectionHandler.eCha(), TreeTable.java:347
javax.swing.DefaultListSelectionModel.eChanged(), DefaultListSelectionModel.java:184
javax.swing.DefaultListSelectionModel.eChanged(), DefaultListSelectionModel.java:164
javax.swing.DefaultListSelectionModel.eChanged(), DefaultListSelectionModel.java:211
javax.swing.DefaultListSelectionModel.election(), DefaultListSelectionModel.java:405
[...]

PR #4005, opened 2019-10-22
hide the custom logging view setting

devoncarew:
- hide the custom logging view setting

From analytics, 1.8% of people have this enabled.

@pq 

This will address https://github.com/flutter/flutter-intellij/issues/3952 (as well as the longer tail of issues - https://github.com/flutter/flutter-intellij/labels/logging).

### DIFF
--- a/src/io/flutter/logging/FlutterLog.java
+++ b/src/io/flutter/logging/FlutterLog.java
@@ -20,7 +20,6 @@ import com.intellij.util.EventDispatcher;
 import io.flutter.FlutterUtils;
 import io.flutter.run.FlutterDebugProcess;
 import io.flutter.run.daemon.FlutterApp;
-import io.flutter.settings.FlutterSettings;
 import io.flutter.utils.VmServiceListenerAdapter;
 import io.flutter.vmService.ServiceExtensions;
 import org.dartlang.vm.service.VmService;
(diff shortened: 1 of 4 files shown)


robertodoering:
I really liked the custom logging view. If it is only hidden, is there another way to enable it?

jaumard:
I was searching a reason why the experimental logging view was gone but wasn't expecting this ^^
 
I find this logging view far more productive that the default one for big project. Would like to see it (or something similar) come back and replace the default one.

devoncarew:
So:

- we weren't actively maintaining this view, and it had accumulated some https://github.com/flutter/flutter-intellij/issues?q=is%3Aissue+is%3Aopen+label%3Alogging over time
- there were not a lot of users - ~1.8% developers had enable the option to display the view
- we've enhanced the output of the regular console view, specifically to better display errors from the flutter framework and to display dart:developer `log()` messages

We have disabled the view for now.

> I find this logging view far more productive that the default one for big project.

Can you talk a bit about the main reasons you used it? I think that would help us either improve the regular console output more, or choose to re-invest in the alternative logging view.

jaumard:
I think I already tell those points few time ago, probably in one of the accumulated issues :p lol 
But basically what's help me on a day to day basis with the experimental view is:

- Filtering by log level (I have an app that log a lot of stuff between http calls and SQL requests just on app start I already have hundreds or more logs) so filtering is a huge plus 
- Search box (when it was working lol) allowing me to narrow the filtering by logger name (or names with regex usage), again super useful when debugging a specific feature(s)
- Color syntax on logs level, to visually see what log is more important (errors, warnings...)

Basically this give the equivalent of the Logcat view for Android, which is super nice.

fquintan:
I agree with @jaumard that the ability to filter logs either by log level or regex will be sorely missed.
The color syntax on logs level I believe can be achieved with some packages like [logger](https://pub.dev/packages/logger) though I haven't tried it myself

PR #4006, opened 2019-10-22
remove the 'organize imports on save' setting

devoncarew:
- remove the 'organize imports on save' setting

From analytics, this is used by 2.7% of users.


stevemessick:
I am one of the apparently few who do use this feature. I wonder how many others (like me) turn off analytics?

devoncarew:
We can keep this feature, but we should look at simplifying (and hardening?) the implementation.

stevemessick:
We could simplify it by extending the analyzer API to do both operations at once. I think most of the complexity comes from having to make two sets of edits, sequentially, which require thread management.

devoncarew:
cc @stereotype441, @bwilkerson, @scheglov re: the idea of invoking more than one operation at once (or a new, combined custom operation).

PR #3993, opened 2019-10-19
replace the weberknecht websocket library with Java-WebSocket-1.4.0.jar

devoncarew:
- replace the weberknecht websocket library with Java-WebSocket-1.4.0.jar

This replaces the `weberknecht-0.1.5.jar` websocket library with the `Java-WebSocket-1.4.0.jar` one. The Java-WebSocket library is likely better maintained and may address some of the longer tail issues we see with weberknecht (like the issue below, where a websocket message is split at char 51604).

If we land this, we'll want to update the source `org/dartlang/vm/service/VmServiceBase.java` library in the dart-lang/sdk repo, as well as the Dart plugin.

I verified that use cases like the inspector, and source level debugging, work with this new library.

```
com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Unterminated object at line 1 column 51604 path $.result.asyncCausalFrames[9].function.owner
  com.google.gson.internal.Streams.parse(), Streams.java:60
  com.google.gson.JsonParser.parse(), JsonParser.java:84
  com.google.gson.JsonParser.parse(), JsonParser.java:59
  com.google.gson.JsonParser.parse(), JsonParser.java:45
  org.dartlang.vm.service.VmServiceBase.processMessage(), VmServiceBase.java:412
  org.dartlang.vm.service.VmServiceBase$1.onMessage(), VmServiceBase.java:86
  de.roderick.weberknecht.WebSocketReceiver.run(), WebSocketReceiver.java:65

Parse message failed: {"jsonrpc":"2.0", "result":{"type":"Stack","frames":[{"type":"Frame","kind":"Regular","location":{"type":"SourceLocation","script":{"type":"@Script","fixedId":true,"id":"libraries\/@1321134198\/scripts\/p...
```

PR #4011, opened 2019-10-24
switch to using the org.java-websocket:Java-WebSocket library

devoncarew:
- switch to using the org.java-websocket:Java-WebSocket library

This brings the TooTallNate `Java-WebSocket` library in via maven. This is an alternative to https://github.com/flutter/flutter-intellij/pull/3993; we'll likely want to update the Dart plugin as well.


devoncarew:
Hmm, some build problems here:
- https://travis-ci.org/flutter/flutter-intellij/jobs/602434675#L681
- https://travis-ci.org/flutter/flutter-intellij/jobs/602434668#L653

stevemessick:
I think it is already approved. LGTM

devoncarew:
@alexander-doroshko, I'm thinking about converting the service protocol library over to use a different websocket library impl. Do you have any thoughts or concerns here? We're using maven to bring the library in for the flutter plugin, but I imagine for the dart plugin we'd just use the straight jar file.

alexander-doroshko:
@devoncarew I'm fine with that. In the Dart plugin it should be a module level maven-based library (just like `"org.apache.commons:commons-lang3:3.9"` in `intellij.dart.iml` and `Dart-community.iml`).

devoncarew:
tabling for now

Issue #3583, opened 2019-06-14
list of flaky tests


devoncarew:
The list of tests we've disabled due to flakes:
-  [ ] WorkspaceCacheTest.shouldDetectConfigFileChanges
-  [ ] WorkspaceCacheTest.shouldDetectModuleRootChange
-  [ ] FlutterWidgetPerfTest.testFileStatsCalculation
-  [x] FlutterWidgetPerfTest.testOverallStatsCalculation

stevemessick:
This one fails frequently on travis, but usually passes locally:
```java
  io.flutter.run.bazelTest.LaunchCommandsTest
    ✘ producesCorrectCommandLineForTestName (10.3s)
      junit.framework.AssertionFailedError: Not disposed java.swing.Timer: 
        Timer (listeners: [io.flutter.perf.FlutterWidgetPerf$$Lambda$5/1164268997@6c32798]) (delayed for 1ms); queue:TimerQueue (javax.swing.Timer@4becde8d)
```
The failure has nothing to do with that test, though. It looks to me like a Timer created in FlutterWidgetPerf was not properly disposed. FlutterWidgetPerf is often the first test class that runs. In this case, we have a lot of tests running after it:
```java
  io.flutter.perf.FlutterWidgetPerfTest
    ✔ testFileStatsCalculation
    ✔ testOverallStatsCalculation
  io.flutter.perf.SlidingWindowStatsTest
    ✔ duplicateSlidingWindowStatTimestamps
    ✔ simpleSlidingWindowStats
    ✔ totalSinceNavigation
    ✔ clearSlidingWindowStats
  io.flutter.samples.FlutterSampleTest
    ✔ testScaffoldDescription
    ✔ testParseShortHtmlDescription
    ✔ testParseCardDescription
    ✔ testParseHtmlDescription
    ✔ testParseDocLink
    ✔ testParseEscapes
  io.flutter.devtools.DevToolsUtilsTest
    ✔ validDevToolsUrl (3.4s)
  io.flutter.run.bazelTest.BazelTestFieldsTest
    ✔ shouldUpgradeFieldsFromOldXml
    ✔ shouldReadFieldsFromXml
    ✔ roundTripShouldPreserveFields
  io.flutter.run.bazelTest.LaunchCommandsTest
    ✘ producesCorrectCommandLineForTestName (10.3s)
```
I bet we can stabilize the tests by not running FlutterWidgetPerfTest.

[...]

PR #4013, opened 2019-10-24
re-enable our disabled tests

devoncarew:
- re-enable our disabled tests

A trial run to see if our tests now pass; I'll cycle the bots a few times to make sure things aren't flakey.


devoncarew:
address some portion of https://github.com/flutter/flutter-intellij/issues/3583

Issue #3892, opened 2019-08-09
Android Studio device list shows "loading" and no devices connected


sempre76:
0

I've installed Android Studio with Flutter/Dart plugin on Win10. Android Studio 3.4.2 Flutter Plugin 38.2 Dart Plugin v183.6270

If i open a Flutter project in android studio, the device list always displays "loading", in the device list i see my three devices:
![image](https://user-images.githubusercontent.com/5462033/62770627-b15efb80-ba9b-11e9-9a20-a8ef22f9c0f5.png)

I selected one device and it opens on the emulator, but when i start the project this error message is displayed:
![image](https://user-images.githubusercontent.com/5462033/62770589-9ee4c200-ba9b-11e9-9047-72e1ebae1f28.png)

When i run "flutter devices", the device is listed as connected:
![image](https://user-images.githubusercontent.com/5462033/62770556-9096a600-ba9b-11e9-927f-21773e0c6494.png)

What is going wrong?

stevemessick:
Do you still see this problem after updating everything to latest (AS, plugin, SDK)?

somoy99:
having the same problem

devoncarew:
We'll investigate and see if we can repro/

IncognitoJam:
I'm also having the same issue.

Output from `flutter doctor -v`:
```
$ flutter doctor -v
[√] Flutter (Channel beta, v1.9.1+hotfix.3, on Microsoft Windows [Version 10.0.18990.1], locale en-GB)
• Flutter version 1.9.1+hotfix.3 at C:\src\flutter
• Framework revision a72edc2706 (23 hours ago), 2019-09-25 15:41:01 -0700
• Engine revision b863200c37
• Dart version 2.5.0

[√] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
• Android SDK at C:\Users\camer\Android
• Android NDK location not configured (optional; useful for native profiling support)
• Platform android-29, build-tools 29.0.2
• ANDROID_HOME = C:\Users\camer\Android
• Java binary at: C:\Users\camer\AppData\Local\JetBrains\Toolbox\apps\AndroidStudio\ch-0\191.5791312\jre\bin\java
• Java version OpenJDK Runtime Environment (build 1.8.0_202-release-1483-b03)
• All Android licenses accepted.

[√] Android Studio (version 3.5)
[...]

PR #4014, opened 2019-10-25
Wait Device Daemon connected to prevent race condition in devices list

Volune:
## Issue description

The devices list was keeping showing "Loading..." in Android Studio.
Windows 10, Android Studio `3.5.1`, Flutter `1.9.1+hotfix.5`, flutter-intellij `4.2.2`

May be related to #3892

## Fix

Flutter-doctor is able to detect the devices. Android Studio logs showed that AS didn't receive the "device.enable" command response. Just putting some breakpoints were solving the issue, identifying a race condition (sending the command too early).

The fix waits for the "daemon.connected" before sending any commands.

## Test

I was not able to correctly build the plugin. The version I built showed the fix is working, but other stuff were broken. It may be caused by me using the "master" version of the plugin in AS 3.5.1, or by bad project/dependencies configuration.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,3 +11,4 @@ Alexander Zolotov <goldifit@gmail.com>
 QuangSon <quangson91@gmail.com>
 ChangJoo Park <pcjpcj2@gmail.com>
 Amol Grover <frextrite@gmail.com>
+Jeremy Judeaux <jeremy.judeaux@volune.net>
--- a/src/io/flutter/run/daemon/DaemonApi.java
+++ b/src/io/flutter/run/daemon/DaemonApi.java
@@ -101,7 +101,7 @@ public class DaemonApi {
     return send("app.callServiceExtension", new AppServiceExtension(appId, methodName, params));
   }
 
-  CompletableFuture enableDeviceEvents() {
+  CompletableFuture<Void> enableDeviceEvents() {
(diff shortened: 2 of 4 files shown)


stevemessick:
@Volune Thanks for the PR! I have never been able to reproduce the original problem, so your sleuthing is appreciated.

@devoncarew I patched this in and verified that it didn't change anything on my Mac.

Volune:
@stevemessick AUTHORS updated

Travis build failed with *Failed to retrieve the Dart SDK*, can you restart it?

devoncarew:
restarted...

stevemessick:
Thanks for doing this! Unfortunately, it didn't make it into the latest release, which is due to be published in an hour or two. The next release is likely to be early, perhaps in less than 3 weeks, and it will be in that one.

AshikAzeez:
I am facing same issue after flutter upgrade.
and resolved by update android studio flutter plugin.

jims57:
I fixed this by killing the process, like this, run in **Terminal**.
Step 1: Get Android Studio process ID it is using now, command as follows
**launchctl list | egrep studio**

Step: Kill the process by the process ID, **{PID}** is the ID your **Android Studio** is using.
**sudo kill -15 {PID}**

Step: Open Android Studio again, and it should be OK.

PR #4016, opened 2019-10-25
detect whether the vm service connection supports the getMemoryUsage API

devoncarew:
- detect whether the vm service connection supports the `getMemoryUsage` API
- cleanup code around memory polling to move more of the domain knowledge from `VMServiceManager` into `HeapMonitor`

Instead of polling the vm connection once a second for the device's memory usage, we use the error return code to detect whether that connection supports the `getMemoryUsage()` API (package:dwds doesn't currently support it). If it doesn't support the API we stop polling.

### DIFF
--- a/src/io/flutter/vmService/HeapMonitor.java
+++ b/src/io/flutter/vmService/HeapMonitor.java
@@ -13,15 +13,12 @@ import org.dartlang.vm.service.element.*;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.*;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 
 public class HeapMonitor {
   private static final Logger LOG = Logger.getInstance(HeapMonitor.class);
 
-  private static final ScheduledExecutorService executor = Executors.newScheduledThreadPool(1);
+  private static final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
 
   private static final int POLL_PERIOD_IN_MS = 1000;
 
@@ -89,8 +86,6 @@ public class HeapMonitor {
 
   @NotNull private final VmServiceWrapper vmServiceWrapper;
 
-  private boolean isPolling;
-
   public HeapMonitor(@NotNull VmServiceWrapper vmServiceWrapper) {
     this.vmServiceWrapper = vmServiceWrapper;
   }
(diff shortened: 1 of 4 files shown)


src/io/flutter/vmService/HeapMonitor.java
```diff
@@ -134,40 +128,48 @@ private void collectMemoryUsage() {
...
         @Override
         public void received(Sentinel sentinel) {
-          errorCount++;
+          latch.countDown();
         }
 
         @Override
         public void onError(RPCError error) {
-          errorCount++;
+          // {"jsonrpc":"2.0","error":{"code":-32603,"message":"UnimplementedError..."}}
```
devoncarew: @grouma @jakemac53 @alan-knight - I'm using the return error code from the call to the vm service's `getMemoryUsage()` API to detect whether the current connection supports that call. Is `-32603` a good code to use for that? Is it a 'I don't support that API' code? I note that the message I get is essentially a stack trace, so I don't know if the code is 'unsupported' or 'the server threw an exception'.

```
==> {"jsonrpc":"2.0","id":"41","method":"getMemoryUsage","params":{"isolateId":"1"}}
```

```
<== {"jsonrpc":"2.0","error":{"code":-32603,"message":"UnimplementedError
	#0      ChromeProxyService.getMemoryUsage (package:dwds/src/services/chrome_proxy_service.dart:309:5)\n<asynchronous suspension>\n
	#1      VmServerConnection._delegateRequest (package:vm_service/vm_service.dart:1017:51)\n<asynchronous suspension>\n
	#2      _rootRunUnary (dart:async/zone.dart:1132:38)\n
	#3      _CustomZone.runUnary (dart:async/zone.dart:1029:19)\n
	#4      _CustomZone.runUnaryGuarded (dart:async/zone.dart:931:7)\n
```

jakemac53: It looks like we should really be returning [-32601](https://www.jsonrpc.org/specification#error_object) - I would check for that code and we should change our handling to return that for unimplemented errors

src/io/flutter/vmService/HeapMonitor.java
```diff
@@ -134,40 +128,48 @@ private void collectMemoryUsage() {
...
+          if (error.getCode() == -32603) {
+            handleMemoryApiNotSupported();
```
kenzieschmoll: Can you just check the vm service version since we should know which version started supporting this API? This is what we do for backwards compatibility in DevTools for recently added vm service APIs.

jakemac53: The problem is that dwds will return unimplemented codes for a lot of apis even though its vm service version has been updated. That really just means we will accept a request for that method but we might return an unimplemented response still.

kenzieschmoll: Gotcha that makes sense. Nevermind then 👍

src/io/flutter/vmService/HeapMonitor.java
jakemac53: https://github.com/dart-lang/webdev/issues/781

devoncarew: Makes sense. I'll handle both for now, and file an issue on dwds to move to returning -32601:Method not found.

devoncarew: ah, filed already as https://github.com/dart-lang/webdev/issues/781

devoncarew:
Updated to check for both:

- rpcInternalError / -32603;
- rpcMethodNotFound / -32601;

PR #4019, opened 2019-10-25
remove support for file path breakpoints; handle multiple vm breakpoints better

devoncarew:
- remove support for file path breakpoints (continue to send in `file:`, `dart:`, and `package:` breakpoint uris); file paths are no longer used with the switch to the CFE
- handle multiple vm breakpoints better; register all successful vm breakpoints created from a user breakpoint; only convert the user breakpoint to a 'breakpointFailed' type if no vm breakpoints succeeded

Additional cleanup to the codebase while working in this area:
- renamed the `PositionMapper` class to `FlutterPositionMapper` to avoid a name conflict with an other `PositionMapper`

### DIFF
--- a/src/io/flutter/run/FlutterPositionMapper.java
+++ b/src/io/flutter/run/FlutterPositionMapper.java
@@ -43,11 +43,12 @@ import java.util.Set;
  * <p>
  * Used when setting breakpoints, stepping through code, and so on while debugging.
  */
-public class PositionMapper implements DartVmServiceDebugProcess.PositionMapper {
+public class FlutterPositionMapper implements DartVmServiceDebugProcess.PositionMapper {
+  private static final Logger LOG = Logger.getInstance(FlutterPositionMapper.class);
+
   @NotNull
   private final Project project;
 
-  // TODO(skybrian) for Bazel this should be a list of source roots.
   /**
    * The directory containing the Flutter application's source code.
    * <p>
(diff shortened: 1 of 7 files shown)


devoncarew:
@jwren, ping -

src/io/flutter/run/FlutterPositionMapper.java
```diff
@@ -196,9 +197,6 @@ private String findRemoteSourceRoot(String remotePath) {
       results.add(threeSlashize(new File(file.getPath()).toURI().toString()));
     }
 
-    // straight path - used by some VM embedders
-    results.add(file.getPath());
```
alexander-doroshko: @devoncarew Should I remove similar line from the Dart plugin as well?
It was added by this [commit](https://github.com/JetBrains/intellij-plugins/pull/423/commits/60df87d36f2d85aed2bab720e4ca3ac34fb6d93b) but I'm not sure how useful it was/is.

alexander-doroshko: Thanks, I'll remove.
If there are any other improvements/fixes that are not mirrored in the Dart plugin yet, PRs are welcome!

devoncarew: For sure; I do try and move over relevant items or file issues (don't always capture all the Flutter plugin changes).

PR #4032, opened 2019-10-29
Fix ant script for #4011

stevemessick:
Modify the ant compile script to pull down the new Java-Websocket library and use it. This should fix the build errors in #4011.

No promises! This is untested beyond checking that `plugin build` succeeds.


stevemessick:
Closing since this is now part of #4011.

PR #4043, opened 2019-11-01
Build for beta not canary

stevemessick:
Drop support for canary; to be added later in a dev channel.


product-matrix.json
```diff
@@ -23,23 +23,23 @@
       "untilBuild": "191.*"
     },
     {
-      "comments": "Android Studio 3.6 Beta 1",
+      "comments": "Android Studio 3.6 Beta 2",
```
devoncarew: Does this now support `2019.2.3`?

stevemessick: Sorry, but I don't know if beta 2 is based on 2019.2.3 or not. I think it is.

Issue #4047, opened 2019-11-04
IndexOutOfBoundsException from WidgetIndentsHighlightingPass


devoncarew:
IntelliJ plugin 41.0.2
Android Studio/3.5.1

```
java.lang.IndexOutOfBoundsException: Wrong offset: -1. Should be in range: [0, 283]
 	at c.i.o.editor.impl.LineSet.findLineIndex(LineSet.java:176)
 	at c.i.o.editor.impl.DocumentImpl.getLineNumber(DocumentImpl.java:991)
 	at io.flutter.editor.WidgetIndentsHighlightingPass$WidgetCustomHighlighterRenderer.updateSelected(WidgetIndentsHighlightingPass.java:145)
 	at io.flutter.editor.WidgetIndentsHighlightingPass.onCaretPositionChanged(WidgetIndentsHighlightingPass.java:496)
 	at io.flutter.editor.WidgetIndentsHighl
```

cc @jacob314

PR #4050, opened 2019-11-05
fix a IndexOutOfBoundsException from WidgetIndentsHighlightingPass

devoncarew:
- fix a IndexOutOfBoundsException from WidgetIndentsHighlightingPass
- fix https://github.com/flutter/flutter-intellij/issues/4047 (`IndexOutOfBoundsException: Wrong offset: -1. Should be in range: [0, 283]`)

### DIFF
--- a/src/io/flutter/editor/WidgetIndentsHighlightingPass.java
+++ b/src/io/flutter/editor/WidgetIndentsHighlightingPass.java
@@ -135,13 +135,13 @@ public class WidgetIndentsHighlightingPass {
       final Document doc = highlighter.getDocument();
       final int caretOffset = carat.getOffset();
 
-      if (startOffset >= doc.getTextLength()) {
+      if (startOffset < 0 || startOffset >= doc.getTextLength()) {
         return setSelection(false);
       }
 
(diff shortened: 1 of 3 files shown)


src/io/flutter/editor/WidgetIndentsHighlightingPass.java
```diff
@@ -135,13 +135,13 @@ boolean updateSelected(@NotNull Editor editor, @NotNull RangeHighlighter highlig
...
 
-      if (startOffset >= doc.getTextLength()) {
+      if (startOffset < 0 || startOffset >= doc.getTextLength()) {
         return setSelection(false);
       }
 
       final int endOffset = highlighter.getEndOffset();
 
-      int off = startOffset;
+      int off;
```
jacob314: was this change intentional? Seems accidental.

devoncarew: It was - IntelliJ helpfully told me that `off` was assigned to again before it was first read, so this initial value was never used.

jacob314:
lgtm

Issue #4057, opened 2019-11-06
IncorrectOperationException: Sorry but parent: main.dart:emulator-5554 has already been disposed


devoncarew:
```
c.i.util.IncorrectOperationException: Sorry but parent: main.dart:emulator-5554 has already been disposed (see the cause for stacktrace) so the child: io.flutter.utils.StreamSubscription@6c7a3df0 will never be disposed
	at c.i.o.util.objectTree.ObjectTree.register(ObjectTree.java:79)
	at c.i.o.util.Disposer.register(Disposer.java:97)
	at c.i.o.util.Disposer.register(Disposer.java:93)
	at io.flutter.vmService.VMServiceManager.hasServiceExtension(VMServiceManager.java:533)
	at io.flutter.run.daemon.Flutt...
```

PR #4058, opened 2019-11-06
fix an issue with an Disposable being already disposed

devoncarew:
- fix an issue with an Disposable being already disposed
- fix https://github.com/flutter/flutter-intellij/issues/4057

### DIFF
--- a/src/io/flutter/vmService/VMServiceManager.java
+++ b/src/io/flutter/vmService/VMServiceManager.java
@@ -358,6 +358,7 @@ public class VMServiceManager implements FlutterApp.FlutterAppListener, Disposab
 
       // Query for display refresh rate and add the value to the stream.
       // This needs to happen on the UI thread.
+      //noinspection CodeBlock2Expr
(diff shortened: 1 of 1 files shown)


src/io/flutter/vmService/VMServiceManager.java
```diff
@@ -530,7 +531,9 @@ public boolean hasServiceExtensionNow(String name) {
   }
 
   public void hasServiceExtension(String name, Consumer<Boolean> onData, Disposable parentDisposable) {
-    Disposer.register(parentDisposable, hasServiceExtension(name, onData));
+    if (!Disposer.isDisposed(parentDisposable)) {
```
jacob314: Do you have any details on what is further up the stack? It seems like something else has gone badly wrong if we are calling this method when parentDisposable has already been disposed.

devoncarew: From looking through the source, I think this can only happen if we're still setting up parts of the UI when the app terminates.

PR #4062, opened 2019-11-07
Re-enable attach button for add-to-app projects

stevemessick:
IntelliJ seems to have tightened the expectations on UI performance, and we may be lagging.

The event used here changes status from running to canceled by the time the code at line 107 executes. Once canceled, we can't get the project anymore.

Also, Android apps (without Flutter) won't display the attach button. After a module is added the button needs to be made visible again (without restarting the IDE).

### DIFF
--- a/src/io/flutter/actions/AttachDebuggerAction.java
+++ b/src/io/flutter/actions/AttachDebuggerAction.java
@@ -104,7 +104,7 @@ public class AttachDebuggerAction extends FlutterSdkAction {
       return;
     }
     boolean enabled;
-    RunnerAndConfigurationSettings settings = RunFlutterAction.getRunConfigSettings(e);
+    RunnerAndConfigurationSettings settings = RunManagerEx.getInstanceEx(project).getSelectedConfiguration();
     if (settings == null) {
       enabled = false;
     }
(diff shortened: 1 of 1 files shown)


src/io/flutter/actions/AttachDebuggerAction.java
```diff
@@ -104,14 +104,15 @@ public void update(AnActionEvent e) {
...
-    RunnerAndConfigurationSettings settings = RunFlutterAction.getRunConfigSettings(e);
+    RunnerAndConfigurationSettings settings = RunManagerEx.getInstanceEx(project).getSelectedConfiguration();
     if (settings == null) {
       enabled = false;
     }
     else {
       RunConfiguration configuration = settings.getConfiguration();
       enabled = configuration instanceof SdkRunConfig;
     }
+    e.getPresentation().setVisible(true);
```
devoncarew: I assume that we do want this visible even if enabled == false? We'll want it visible for Flutter enabled projects, just not enabled if if doesn't apply to the currently running app?

stevemessick: Yes, it would be disabled if the run config is for Android, and enabled when a Flutter run config is selected.

Issue #4069, opened 2019-11-08
java.nio.charset.MalformedInputException: Input length = 1 from FlutterModuleImporter.editFileContent(FlutterModuleImporter.java:114)


devoncarew:
`41.0.2`, `Android Studio/3.5.2`

```
java.io.UncheckedIOException: java.nio.charset.MalformedInputException: Input length = 1
 	at java.io.BufferedReader$1.hasNext(BufferedReader.java:574)
 	at java.util.Iterator.forEachRemaining(Iterator.java:115)
 	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
 	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:580)
 	at io.flutter.module.FlutterModuleImporter.editFileContent(FlutterModuleImporter.java:114)
 	at io.flutter.module.FlutterModuleI...
```

@stevemessick, from a brief look at the code, it looks like one of the two files we're reading can sometimes be of length 1, and the char encoding we're passing into `Files.newBufferedReader` doesn't like that. I don't know if that means we should change the encoding we're using to read the file (the other Files.newBufferedReader method uses UTF-8), or find some other way to read the contents.

stevemessick:
I'm changing it to use StandardCharsets.UTF_8.

devoncarew:
You may want to test how that reacts to a 1-char long file. And - it looks like we read two files w/ the `editFileContent()` method - are they both expected to be in UTF8?

devoncarew:
Fixed in https://github.com/flutter/flutter-intellij/pull/4097

Issue #4064, opened 2019-11-08
Add-to-app release features


xster:
Extracting some discussion items out of #3554 for release tracking.

- [x] Let the AndroidX setting follow the parent project's AndroidX setting **P1**
- [x] Just do include_flutter.groovy for now **P1**
- [x] The auto integration probably needs to add compileOptions {
        sourceCompatibility 1.8
        targetCompatibility 1.8
    } to the host app's build.gradle too **P1** #4134
- [x] Resurface the flutter attach button in the toolbar **P1**
- [x] Let the user choose the directory when creating new modules **P2** #4134
- [x] Choose the debug/profile/release mode of the AAR built when building an AAR from the menu ** P2 ** Done via https://github.com/flutter/flutter/pull/44797
- [x] If only one Flutter fun config exists then the Flutter attach button is active even if it is not selected **P2** #4134

And when https://github.com/flutter/flutter/issues/32999 makes it in:

- [x] Let flutter attach monitoring start as soon as the Android host app is launched to allow hot reload on Dart file save when the user navigates into the Flutter screen. **P2**

devoncarew:
@stevemessick, should we break these out into individual issues? That may be helpful if some items have different priority, and or may land in different releases.

xster:
Thanks. Commented inline.

stevemessick:
Thanks @xster, that helps.

xster:
One more note that @matthew-carroll just noticed. It would be nice (for Flutter modules in general) to be able to choose the debug/profile/release mode of the AAR built when building an AAR from the menu (might already be possible?). P2

stevemessick:
I added that to #4064

xster:
Thanks for the demo plugin @stevemessick 

I tried it a bit. It's much smoother than the last one. A few more observations

- In the outer app's settings.gradle, `include ':flutter_module'` appeared multiple times, though it's probably inconsequential

![Screen Shot 2019-11-18 at 6 19 58 PM](https://user-images.githubusercontent.com/156888/69111711-3c547800-0a33-11ea-8ca4-7743c9ad4466.png)

- The Java 8 compileOptions wasn't added in the outer app's build.gradle. The build fails to compile until added manually

- The Flutter attach is waiting for a service connection even though the Flutter screen already launched 

[...]

Issue #4065, opened 2019-11-08
IllegalArgumentException from FlutterBuildActionGroup.build()


devoncarew:
`41.0.2`, `Android Studio/3.5.2`

```
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'root' of io/flutter/sdk/FlutterSdk.flutterBuild must not be null
 	at io.flutter.sdk.FlutterSdk.$$$reportNull$$$0(FlutterSdk.java)
 	at io.flutter.sdk.FlutterSdk.flutterBuild(FlutterSdk.java)
 	at io.flutter.actions.FlutterBuildActionGroup.build(FlutterBuildActionGroup.java:33)
 	at io.flutter.module.FlutterModuleModel$FlutterGradleLinker.linkNewModule(FlutterModuleModel.java:115)
 	at io.flutter.module.FlutterModuleModel$FlutterGradleLink...
```

devoncarew:
Hmm, @stevemessick, I have a fix for this, but I'm not sure why we can't find the PubRoot in the method `FlutterGradleLinker.linkNewModule(FlutterModuleModel.java:115)`, or what an early exit will mean in terms of the work that the linkNewModule() method is trying to do.

devoncarew:
@stevemessick, do you know if this was address by https://github.com/flutter/flutter-intellij/pull/4097?

stevemessick:
I don't believe it was.

stevemessick:
That was incorrect; #4097 deleted the code that caused this exception.

PR #4097, opened 2019-11-12
Support for debugging add-to-app modules in Android Studio

stevemessick:
Address all P1 items in #4064.
cc: @xster 

### Create an Android app

Start Android Studio and use the File > New > New Project menu to fire 
up the new project wizard. Choose the template that includes a Basic
Activity. One the next page, use Kotlin and minimum API level 16.
Let the system stabilize (Gradle sync).

Close the editor for content_main.xml. Switch the Project view to
the Project Files mode. Collapse the Gradle window if it is open.

### Create a Flutter module

Use the File > New > New Module menu to start the new module wizard.
Choose the Flutter Module template. Fill out the wizard pages and
click Finish, then wait a bit. Android Studio needs to do two
successive Gradle sync's. The first one generates an error message,
which is corrected by the second one. Ignore the error. Let the
system stabilize again.

### Link the Flutter module to the Android app

Go to the editor for MainActivity.kt. Change the onCreate method:
```kotlin
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_main)
        setSupportActionBar(toolbar)
        val engine = FlutterEngine(this.applicationContext)
        engine.dartExecutor.executeDartEntrypoint(DartExecutor.DartEntrypoint.createDefault())
        fab.setOnClickListener { view ->
            FlutterEngineCache.getInstance().put("1", engine)
            startActivity(FlutterActivity.withCachedEngine("1").build(this))
        }
    }
```
You need to add some imports. Click on each red name then type
`alt-return`. Accept the suggestion. "FlutterActivity" has two choices;
use the one from io.flutter.embedGradle.

Open an editor on the AndroidManifest.xml. Add this line to the
<application> tag:
```xml
        <activity android:name="io.flutter.embedding.android.FlutterActivity" />
```
### Debug the app

The run config should show the Flutter run config, which makes the Flutter Attach
button active. Click the Flutter Attach button, or use the menu item Run > Flutter
Attach. (Note: with the kotlin code above you need to start the attach process
before starting the Android app.)

Change the run config to "app" for the Android app. Click the Debug button.
Wait for the app to launch.

When the app is visible, click the mail icon at bottom right. The Flutter
screen should become visible. At this point you can set breakpoints and
do hot reload as for a stand-alone Flutter app.

### DIFF
--- a/flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
+++ b/flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
@@ -15,11 +15,14 @@ import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VfsUtilCore;
 import com.intellij.openapi.vfs.VirtualFile;
 import io.flutter.project.FlutterProjectModel;
+import io.flutter.pub.PubRoot;
+import io.flutter.utils.FlutterModuleUtils;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.function.Consumer;
@@ -49,7 +52,7 @@ public class FlutterModuleImporter {
   }
 
   public void importModule() {
-    String location = myModel.projectLocation().get()+ "/" + myModel.projectName().get();
+    String location = myModel.projectLocation().get() + "/" + myModel.projectName().get();
     VirtualFile moduleRoot = VfsUtil.findFileByIoFile(new File(location), true);
     assert (moduleRoot != null);
     assert (myModel.project().get().isPresent());
@@ -62,7 +65,7 @@ public class FlutterModuleImporter {
     }
 
     String newPath = Paths.get(myRelativePath, ".android", "include_flutter.groovy").normalize().toString();
-    if (!new File(moduleRoot.getParent().getPath(), newPath).exists()) {
+    if (!new File((newPath.startsWith("..") ? moduleRoot : moduleRoot.getParent()).getPath(), newPath).exists()) {
       showHowToEditDialog();
       return;
     }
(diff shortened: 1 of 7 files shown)


xster:
Thanks @stevemessick! This is super awesome.

2 questions on the readme. 

@matthew-carroll, I think we don't need System.loadLibrary("flutter"); FlutterMain.startInitialization(this.applicationContext); FlutterMain.ensureInitializationComplete(this, null) anymore. If you're on master of the framework (what we'll most likely ship by December), it should work with just the remaining lines. Were you getting an error @stevemessick?

@blasten, it seems like flutter attach here "just works" on Android. Is the IDE doing something special to catch the debug-uri for the observatory?

matthew-carroll:
@xster that's correct. I don't think we should need any loading/initialization code from the app developer anymore.

stevemessick:
@xster TBH I just wrote the test code based on a test app from the Flutter repo. The simplification works so I changed the testing doc. I'll edit the comment, too.

And yes, the IDE does quite a bit to handle attach. It's been over a year since I looked at that code and am fuzzy on details. It sets up a debug listener that runs when the VM service protocol says a debug launch is happening.

xster:
Understood

wrt the service protocol thing @devoncarew, @zanderso, @blasten when we're cleaning up the attach logic at some point later on in Q4, it would be nice to deduplicate these IDE and/or tool logic.

xster:
Can't review the plugin code but the usage instructions LGTM!

Thanks Steve

devoncarew:
fixes https://github.com/flutter/flutter-intellij/issues/4069

flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
```diff
@@ -98,18 +101,23 @@ public void importModule() {
...
+        buildFile.setBinaryContent(buildWriter.toString().getBytes(StandardCharsets.UTF_8));
       }
       catch (IOException e) {
         // Should not happen
       }
     });
+
+    // Setup a default run configuration for 'main.dart' (if it's not there already and the file exists).
+    PubRoot pubRoot = PubRoot.forDirectory(moduleRoot);
+    assert(pubRoot != null);
```
devoncarew: How similar is this code path to the one in `FlutterBuildActionGroup.build()`? I ask because we (very infrequently) see an NPE from there when the pubRoot is null. https://github.com/flutter/flutter-intellij/issues/4065

devoncarew: If it's possible here as well, we'd want to handle null instead of asserting.

xster:
Is there a way I can try this plugin locally?

stevemessick:
@xster I'll contact you via email.

flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
stevemessick: I don't see any similarity. Looking at the caller of `FlutterBuildActionGroup.build()` I see it does `PubRoot.forEventWithRefresh(event)`. My guess is the null arises because the event changes status before that expression runs, and the Project cannot be retrieved from the event. All I have is anecdotal evidence for this, but I think JetBrains is tightening controls on how much can be done on the EDT, probably to improve responsiveness. We should look at callers of `forEventWithRefresh` and rewrite them to ensure the Project is fetched early.

PR #4102, opened 2019-11-13
Update FlutterDartAnalysisServer to support Widget manipulation methods.

jacob314:


devoncarew:
Is this largely copied from generated code (I'm asking in terms of reviewing the PR)?

jacob314:
Not quite ready for review yet as I'm fixing the travis failure. This is almost all generated code. I'll call out the couple files that aren't.

src/io/flutter/dart/FlutterDartAnalysisServer.java
```diff
@@ -98,11 +108,60 @@ private void sendSubscriptions() {
...
+    Uninterruptibles.awaitUninterruptibly(latch, 100, TimeUnit.MILLISECONDS);
+    return result.get();
+  }
+
   private void processString(String jsonString) {
-    // Perform a quick check to see if we should parse this into a json object.
-    if (jsonString.contains(FLUTTER_NOTIFICATION_OUTLINE_KEY)) {
-      processResponse(new Gson().fromJson(jsonString, JsonObject.class));
-    }
+    processResponse(new Gson().fromJson(jsonString, JsonObject.class));
```
jacob314: @devoncarew
I had to remove this optimization to skip all json that didn't contain a specific key. This seems like a bit of a hack of an optimization to be performing. As long as we make sure we are calling processResponse on a background thread it shouldn't be needed.

testSrc/unit/io/flutter/editor/ActiveEditorsOutlineServiceTest.java
```diff
@@ -39,11 +39,11 @@
   // The outline service doesn't care too much what the outline's contents are, only that the outlines update.
   // However, the request method uses the outline's length to determine if the outline is up-to-date.
   private static final FlutterOutline firstFlutterOutline =
-    new FlutterOutline("First", 0, 0, 0, 0, "", null, null, null, null, null, null, 1, false, null, null, null, null);
+    new FlutterOutline("First", 0, 0, 0, 0, "", null, null, null, null, null, null);
```
jacob314: update the test to match the latest version of the FlutterOutline class. The existing version was about a year out of date.

src/io/flutter/dart/FlutterDartAnalysisServer.java
```diff
@@ -7,24 +7,27 @@
...
 import com.jetbrains.lang.dart.analyzer.DartAnalysisServerService;
-import org.dartlang.analysis.server.protocol.FlutterOutline;
-import org.dartlang.analysis.server.protocol.FlutterService;
-import org.dartlang.analysis.server.protocol.SourceChange;
+import org.dartlang.analysis.server.protocol.*;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
 import java.util.HashMap;
```
jacob314: @scheglov this code should look pretty familiar. This is similar code to the code you used to have to access  the design time constructor apis from here.

devoncarew:
From the 'dart check' bot:

```
Found proscribed imports:
src/org/dartlang/analysis/server/protocol/ExtractWidgetFeedback.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
src/org/dartlang/analysis/server/protocol/ExtractWidgetOptions.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
src/org/dartlang/analysis/server/protocol/FlutterOutline.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
src/org/dartlang/analysis/server/protocol/FlutterOutlineAttribute.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
src/org/dartlang/analysis/server/protocol/FlutterWidgetProperty.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
src/org/dartlang/analysis/server/protocol/FlutterWidgetPropertyEditor.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
src/org/dartlang/analysis/server/protocol/FlutterWidgetPropertyValue.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
src/org/dartlang/analysis/server/protocol/FlutterWidgetPropertyValueEnumItem.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
```

src/io/flutter/dart/FlutterDartAnalysisServer.java
devoncarew: We may want to keep this. With the available imports work, json parsing (from this location) was showing up on performance profiles. For large repos we're sending dozen's of megabytes of json to clients to parse. (I agree that the work shouldn't be done on the EDT; I don't remember if that was the case when we were looking at the performance profiles a few months ago).

What was it you're trying avoid? Are we going to need a more flexible filter mechanism?

jacob314:
> From the 'dart check' bot:
> 
> ```
> Found proscribed imports:
> src/org/dartlang/analysis/server/protocol/ExtractWidgetFeedback.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
> src/org/dartlang/analysis/server/protocol/ExtractWidgetOptions.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
> src/org/dartlang/analysis/server/protocol/FlutterOutline.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
> src/org/dartlang/analysis/server/protocol/FlutterOutlineAttribute.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
> src/org/dartlang/analysis/server/protocol/FlutterWidgetProperty.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
> src/org/dartlang/analysis/server/protocol/FlutterWidgetPropertyEditor.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
> src/org/dartlang/analysis/server/protocol/FlutterWidgetPropertyValue.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
> src/org/dartlang/analysis/server/protocol/FlutterWidgetPropertyValueEnumItem.java:import org.apache.commons.lang3.builder.HashCodeBuilder;
> ```

I've removed this check from the dart check bot. It was added a year ago as there was an issue running the plugin on all platforms if it used those imports but that issue appears to have gone away based on checking Android Studio and Intellij.
https://github.com/flutter/flutter-intellij/commit/527ba3b0fd5a498fd08cc92450952ed6a22c4cc4#diff-27d6443fab617e0acff17b548af3a6d0

src/io/flutter/dart/FlutterDartAnalysisServer.java
jacob314: We would need a more flexible approach to filtering as otherwise we filter out all the analysis server messages about widget refactoring. Let me check if we are calling this on the EDT thread. If it isn't actually done on the EDT thread it shouldn't really be impacting user performance even it shows up in profiles as a few % of cpu usage.

src/io/flutter/dart/FlutterDartAnalysisServer.java
```diff
@@ -7,24 +7,27 @@
 
 import com.google.common.annotations.VisibleForTesting;
```
jacob314: this is the only file with non-generated changes.

devoncarew:
> I've removed this check from the dart check bot. It was added a year ago as there was an issue running the plugin on all platforms if it used those imports but that issue appears to have gone away based on checking Android Studio and Intellij.

OK; I'm a bit superstitious about usage of those classes, but we can always switch the code to other implementations if we find that there are still issues.

jacob314:
> > I've removed this check from the dart check bot. It was added a year ago as there was an issue running the plugin on all platforms if it used those imports but that issue appears to have gone away based on checking Android Studio and Intellij.
> 
> OK; I'm a bit superstitious about usage of those classes, but we can always switch the code to other implementations if we find that there are still issues.

I'm happy to port over or fix the generator in the analyzer if we find there is still an issue.

src/io/flutter/dart/FlutterDartAnalysisServer.java
devoncarew: For posterity, here's the original perf issue: https://youtrack.jetbrains.com/issue/WEB-40646 (w/ some data about the perf profiles).

testData/sample_tests/test/simple_outline.txt
```diff
@@ -2,11 +2,13 @@
...
   "offset": 0,
   "length": 878,
+  "codeOffset": 0,
+  "codeLength": 878,
   "dartElement": {
     "kind": "COMPILATION_UNIT",
     "name": "<unit>",
     "location": {
-      "file": "/usr/local/google/home/djshuckerow/Code/github.com/DaveShuckerow/flutter-intellij/testData/sample_tests/test/simple_test.dart",
+      "file": "/usr/local/google/home/jacobr/git/flutter-intellij/1/flutter-intellij/testData/sample_tests/test/simple_test.dart",
```
devoncarew: Not necessarily for this PR, but I wonder if we shouldn't normalize these files somehow (to sanitize paths).

src/io/flutter/dart/FlutterDartAnalysisServer.java
jacob314: Ok how about this. Currently this code does execute on the ServerResponseReaderThread which is the same thread that autocomplete will be using so I can understand how it is causing problems. What I can do is tweak FlutterDartAnalysisService to do its json parsing on a separate thread and return immediately. That way there should be less of an impact on other FlutterDartAnalysisServer consumers if I am understanding the threading model correctly.

tool/plugin/lib/lint.dart
```diff
@@ -80,8 +80,6 @@ class LintCommand extends Command {
       'com.android.annotations.NonNull',
       'io.netty.',
       'javax.annotation.Nullable',
-      'org.apache.commons.lang3.StringUtils',
```
devoncarew: Let's keep this first exclusion - we're already using `org.apache.commons.lang.StringUtils` in the codebase.

src/io/flutter/dart/FlutterDartAnalysisServer.java
devoncarew: OK, sgtm. Longer term we may want to investigate a faster json parser (Alex was going to look at `com.fasterxml.jackson.core.JsonParser`), and plan to try and use the already parsed json from the Dart plugin instead of re-parsing it a 2nd time in the Flutter plugin. That would be less work overall and simplify any threading hijinks we'll have to do in the short term.

jacob314: switched to a slightly faster json parsing api (less reflection I think) and moved work to a background thread so we don't block the dart plugin.

testData/sample_tests/test/simple_outline.txt
jacob314: that would be nice but lets leave it for a cleanup pr

tool/plugin/lib/lint.dart
jacob314: done

jacob314:
PTAL.
FlutterDartAnalysisServer has some new changes to cleanup the threading. I think the existing code had some bad threading assumptions anyway given callbacks were made on one thread and listeners were registered on a different thread.

Issue #3503, opened 2019-05-20
Flutter IntelliJ opens devtools in the default browser instead of chrome


jacob314:
The Dart DevTools server now supports launching DevTools in chrome and VSCode users it to launch web browsers. We should use this functionality from IntelliJ as well or use existing IntelliJ functionality to launch Chrome.

A Mac user I was helping debug had Safari as their default value so had to copy and paste the URL from Safari to Chrome each time they launched DevTools fro IntelliJ.

devoncarew:
Note - docs for using the devtools server to launch a browser are here: https://github.com/flutter/devtools/pull/1324.

devoncarew:
There's a slim chance I'll take this on in the M42 timeframe.

hBurt:
Hey, so I had this same issue pop up and saw someone asking this same question on stack overflow as well. Turns out it's quite a simple fix. 

In my case, my computers default browser is set to Safari but Safari doesn't work well with Flutters Dev Tools at the moment. So.. how can you keep your computers default browser and specify a browser you want to use when starting dev tools?

In Android Studio open up your Preferences and in the top left there is a search bar.

Type in "Web Browsers".

You will see a list of web browsers and you can click and drag. Place the one you want to use by default on top. For example I have chrome on the top of my list. 
<img width="996" alt="Screen Shot 2021-04-08 at 9 39 18 AM" src="https://user-images.githubusercontent.com/4024074/114038450-e19e0500-984f-11eb-8c85-89c8313fcbe5.png">

Now just set this from "System default" to "First listed" and apply your changes

<img width="730" alt="Screen Shot 2021-04-08 at 9 41 20 AM" src="https://user-images.githubusercontent.com/4024074/114038546-f7132f00-984f-11eb-83f1-ed62b5905a7b.png">

When you press this button to open dev tools it will now, in my case, open chrome instead of safari even though Safari is my computers default browser. 

<img width="364" alt="Screen Shot 2021-04-08 at 9 45 19 AM" src="https://user-images.githubusercontent.com/4024074/114038602-01cdc400-9850-11eb-8f22-2c7cce9ca1d2.png">

PR #4105, opened 2019-11-13
Improve devtools launching

devoncarew:
Improve devtools launching:

- no longer route devtools starting up output to the message console (fixes an issue where we steal focus from the run/debug console)
- address an issue where some hyperlinks in the performance tool window would launch devtools, but would not perform the devtools app install flow (would fail if devtools was not already installed)
- remove the deprecated `WebDevManager.java` class

This is in prep for additional devtools launching improvements (https://github.com/flutter/flutter-intellij/issues/3503).


src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -194,21 +195,28 @@ private boolean isBazel(Project project) {
...
+          devToolsInstance = instance;
+
+          devToolsInstance.openBrowserAndConnect(uri, page);
+        }, () -> {
+          // Listen for closing; null out the devToolsInstance.
+          devToolsInstance = null;
+        });
+      }
+      else {
+        devToolsInstance = null;
```
kenzieschmoll: If we hit this case, is there any feedback we need to show to the user?

src/io/flutter/view/PerfFPSTab.java
```diff
@@ -95,11 +96,16 @@ private void buildUI() {
...
   private void openInDevTools() {
+    final DevToolsManager devToolsManager = DevToolsManager.getInstance(app.getProject());
+
     // open the timeline view
-    DevToolsManager.getInstance(app.getProject()).openBrowserAndConnect(
-      app.getConnector().getBrowserUrl(),
-      "timeline"
-    );
+    if (devToolsManager.hasInstalledDevTools()) {
+      devToolsManager.openBrowserAndConnect(app.getConnector().getBrowserUrl(), "timeline");
```
kenzieschmoll: perhaps put this in a method inside `DevToolsManager` to consolidate the two uses here in the memory tab below. Something like:
```
public void openToScreen(String screen) {
  openBrowserAndConnect(app.getConnector().getBrowserUrl(), screen);
}
```

or even better consolidate all the conditional logic in DevToolsManager, and just call `devToolsManager.openToScreen("timeline")` here (and below for memory)
```
public void openDevTools(String screen) {
  if (hasInstalledDevTools()) {
     openToScreen(screen);
  } 
  else {
      final CompletableFuture<Boolean> result = installDevTools();
      result.thenAccept(o -> openToScreen(screen));
  }
}

private void openToScreen(String screen) {
  openBrowserAndConnect(app.getConnector().getBrowserUrl(), screen);
}
```

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -133,8 +132,26 @@ public void openBrowserAndConnect(String uri) {
...
+    openBrowserImpl(uri, screen);
+  }
+
+  /**
+   * Open a browser connected to DevTools with the given (optional) screen.
+   * <p>
+   * Calling this method may cause DevTools to be installed (in which case, this API will provide
+   * appropriate progress feedback to the user).
+   */
+  public void openToScreen(@NotNull FlutterApp app, @Nullable String screen) {
```
kenzieschmoll: As long as this behaves properly when screen is null, lgtm

devoncarew: Yup, it should (it's existing logic in DevToolsUtils.generateDevToolsUrl()).

PR #4107, opened 2019-11-13
reduce the severity of a warning

devoncarew:
- reduce the severity of a warning (issues starting the device daemon up are already reported to the user)
- remove an obsolete todo:


jwren:
LGTM

PR #4109, opened 2019-11-14
Build for Android Studio beta 4 and later

stevemessick:
@jacob314 I added the library to both module definitions. I only tested the one I'm using. I think this is safe, but it it causes problems we can revert the change to fultter-intellij.iml.


stevemessick:
Thanks, @jacob314 for the quick review!

PR #4111, opened 2019-11-15
refactor the UI and location of the Flutter Performance tool window

devoncarew:
Refactor the UI and location of the Flutter Performance tool window:

- move the `Flutter Performance` tool window from the bottom right position to the right hand side; this locates it next to the Outline and Inspector views, and will hopefully improve its discoverability
- change it from a series of three sub-tabs into one, vertically oriented, UI. This simplifies the UI (which I think is good, as this tool window will likely be used mostly for issue awareness, and devtools used for diagnosing the problem). It will also help with discovery of features that were not in the first tool window tab (memory usage and the widgets rebuild count feature)
- simplified the look of the views in order to make the new combined UI less busy
- have the top two areas be fixed height (FPS and memory); that will allow the 3rd area - widget rebuild counts - to get the remaining vertical space as the user re-sizes the height of the tool window

This tool window now no longer looks ideal in a horizontal form factor. I believe that current workspaces will still try and open the view in the old location. We may want to rename the view - I believe a new class name will mean that IntelliJ will position the view in the default location (instead of the last location it was in).

Screen Shot 2019-11-15 at 10 47 58 AM

### DIFF
--- a/src/io/flutter/inspector/WidgetPerfTipsPanel.java
+++ b/src/io/flutter/inspector/WidgetPerfTipsPanel.java
@@ -156,8 +156,6 @@ public class WidgetPerfTipsPanel extends JPanel {
   private void showPerfTips(@NotNull ArrayList<PerfTip> tips) {
     perfTips.removeAll();
 
-    final PerfTip lastMainTip = !currentTips.isEmpty() ? currentTips.get(0) : null;
-
     currentTips.clear();
     currentTips.addAll(tips);
 
--- a/src/io/flutter/inspector/WidgetPerfTipsPanel.java
+++ b/src/io/flutter/inspector/WidgetPerfTipsPanel.java
@@ -35,16 +35,11 @@ import java.util.*;
  */
 public class WidgetPerfTipsPanel extends JPanel {
   static final int PERF_TIP_COMPUTE_DELAY = 1000;
-  private final FlutterApp app;
+
   private final FlutterWidgetPerfManager perfManager;
 
-  // Computing performance tips is somewhat expensive so we don't want to
-  // compute them too frequently. Performance tips are only computed when
-  // new performance stats are available but performance stats are updated
-  // at 60fps so to be conservative we delay computing perf tips.
-  private final Timer perfTipComputeDelayTimer;
-  long lastUpdateTime;
-  final private JPanel perfTips;
+  private long lastUpdateTime;
+  private final JPanel perfTips;
 
   /**
    * Currently active file editor if it is a TextEditor.
@@ -53,13 +48,12 @@ public class WidgetPerfTipsPanel extends JPanel {
   private ArrayList<TextEditor> currentTextEditors;
 
   final LinkListener<PerfTip> linkListener;
-  private boolean visible = true;
-  private List<PerfTip> currentTips;
+  final private List<PerfTip> currentTips = new ArrayList<>();
 
   public WidgetPerfTipsPanel(Disposable parentDisposable, @NotNull FlutterApp app) {
-    this.app = app;
-
     setLayout(new VerticalLayout(5));
+
+    add(new JSeparator());
 
     perfManager = FlutterWidgetPerfManager.getInstance(app.getProject());
     perfTips = new JPanel();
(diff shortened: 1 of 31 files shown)


devoncarew:
Updated:

- moved `io.flutter.view.FlutterPerfView` into `io.flutter.performance.FlutterPerformanceView` (to let the tool window take on the new default location)
- renames the enable checkbox for widget reload tracking to `'Track widget rebuilds'`

src/io/flutter/inspector/WidgetPerfTipsPanel.java
```diff
@@ -155,10 +153,14 @@ private void updateTip() {
 
   ArrayListMultimap<TextEditor, PerfTip> tipsPerFile;
 
-  private void showPerfTips(ArrayList<PerfTip> tips) {
+  private void showPerfTips(@NotNull ArrayList<PerfTip> tips) {
     perfTips.removeAll();
-    final PerfTip lastMainTip = currentTips != null && currentTips.size() > 0 ? currentTips.get(0) : null;
-    currentTips = tips;
+
+    final PerfTip lastMainTip = !currentTips.isEmpty() ? currentTips.get(0) : null;
```
kenzieschmoll: is lastMainTip used?

src/io/flutter/performance/FlutterPerformanceView.java
```diff
@@ -250,57 +260,19 @@ FlutterViewAction registerAction(FlutterViewAction action) {
...
-                         ToolWindow toolWindow,
-                         boolean selectedTab) {
-    final PerfFPSTab perfTab = new PerfFPSTab(runnerTabs, app, toolWindow);
-    final TabInfo tabInfo = new TabInfo(perfTab)
-      .append(PERFORMANCE_TAB_LABEL, SimpleTextAttributes.REGULAR_ATTRIBUTES);
-    runnerTabs.addTab(tabInfo);
-    if (selectedTab) {
-      runnerTabs.select(tabInfo, false);
-    }
+  private void addFPSTab(JComponent parentPanel, @NotNull FlutterApp app, @NotNull Disposable parentDisposable) {
```
kenzieschmoll: should these methods be renamed now that they are no longer "tabs"?

src/io/flutter/inspector/WidgetPerfTipsPanel.java
devoncarew: No, removed (apparently it hadn't been used before the refactor either).

src/io/flutter/performance/FlutterPerformanceView.java
devoncarew: I removed them - they're now in-lined into their calling locations.

devoncarew:
For posterity, here is the previous UI:

Screen Shot 2019-11-17 at 12 49 29 PM

Screen Shot 2019-11-17 at 12 50 15 PM

Screen Shot 2019-11-17 at 12 49 56 PM

src/io/flutter/performance/FlutterPerformanceView.java
```diff
@@ -212,9 +224,15 @@ private void addPerformanceViewContent(FlutterApp app, ToolWindow toolWindow) {
...
-      addWidgetRebuildsTab(runnerTabs, app, state);
+      // Create the three FPS, memory, and widget recount areas.
+      final PerfFPSTab perfTab = new PerfFPSTab(app, this);
+      perfViewsPanel.add(perfTab);
+
+      final PerfMemoryTab memoryTab = new PerfMemoryTab(app, this);
+      perfViewsPanel.add(memoryTab);
+
+      final PerfWidgetRebuildsTab rebuildsTab = new PerfWidgetRebuildsTab(app, this);
+      perfViewsPanel.add(rebuildsTab);
```
kenzieschmoll: comment about class naming as "tab" still stands. Since these are no longer tabs, maybe they should be replaces with `PerfFPSSection`, `PerfMemorySection`, `PerfWidgetRebuildsSection`

devoncarew: Ah, yeah - I'll rename away from 'tab'.

PR #4126, opened 2019-11-19
Force VFS refresh to prevent adding multiple include statements

stevemessick:
Force synchronous VFS refresh to prevent adding multiple, redundant include statements to settings.gradle.

### DIFF
--- a/src/io/flutter/utils/AndroidUtils.java
+++ b/src/io/flutter/utils/AndroidUtils.java
@@ -500,6 +500,7 @@ public class AndroidUtils {
         addIncludeStatement();
         if (errorDuringOperation.get()) return;
         addCoeditTransformedProject(project);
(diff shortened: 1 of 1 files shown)


src/io/flutter/utils/AndroidUtils.java
```diff
@@ -500,7 +500,7 @@ private void enable() {
         addIncludeStatement();
         if (errorDuringOperation.get()) return;
         addCoeditTransformedProject(project);
-        projectRoot.refresh(true, true);
+        projectRoot.refresh(false, true);
```
devoncarew: note: it looks like this change forces the refresh to happen synchronously; perhaps add a comment in-line, so people will know that the use of that param is important?

stevemessick: Done!

PR #4129, opened 2019-11-21
Smooth out the rough spots in add-to-app support

stevemessick:
Finish up some in-progress changes and clean up a few things for add-to-app:
- Add Java code to the testing instructions, and note that both Java and Kotlin need to be tested.
(The compileOptions section is added when using Kotlin but not Java.)
- Eliminate redundant Gradle sync listeners (cleans up duplicated lines in settings.gradle).
- Change the build.gradle editing to use the proper framework. Add compileOptions.
- Use the module location setting that was added to the UI in a previous PR.
- Set the project type to "io.flutter" when a project is created so it doesn't use the default.
- If there is only one Flutter run config, use it for attach whether it is selected or not.
- Remove TODOs that only apply to IJ support for add-to-app. That isn't happening.

### DIFF
--- a/flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
+++ b/flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
@@ -5,8 +5,17 @@
  */
 package io.flutter.module;
 
+import static com.intellij.openapi.command.WriteCommandAction.runWriteCommandAction;
+
+import com.android.tools.idea.gradle.dsl.api.GradleBuildModel;
+import com.android.tools.idea.gradle.dsl.api.ProjectBuildModel;
+import com.android.tools.idea.gradle.dsl.api.android.AndroidModel;
+import com.android.tools.idea.gradle.dsl.api.android.CompileOptionsModel;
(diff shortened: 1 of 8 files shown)


xster:
Can't review the code but functionally, it works superbly!

flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
```diff
@@ -87,32 +97,49 @@ public void importModule() {
       return;
     }
 
+    editSettingsFile(settingsFile);
+    editBuildFile(buildFile);
+
+    // Setup a default run configuration for 'main.dart' (if it's not there already and the file exists).
+    PubRoot pubRoot = PubRoot.forDirectory(moduleRoot);
+    assert (pubRoot != null);
```
devoncarew: Do we know whether this will fire? If so, we may instead want to check for null and handle it.

stevemessick: The assert is just to keep IntelliJ happy. I don't think it can fire.

PR #4134, opened 2019-11-22
Use Build Variants view to control Flutter build mode

stevemessick:
After creating an add-to-app module in an Android app, use Android Studio's Build Variants tool window to set the build variant for the "flutter" module. That value (debug/profile/release) will be passed as the corresponding parameter to the build command when a build is initiated from the menu, Build > Flutter > Build {AAR,APK,App Bundle,iOS}.

The add-to-app (Flutter) module must be selected to see the Build > Flutter  menu.

@devoncarew @xster @matthew-carroll for comments on using the tool window this way.

Screen Shot 2019-11-22 at 2 13 09 PM

### DIFF
--- a/flutter-studio/src/io/flutter/android/AndroidBuildVariantDetector.java
+++ b/flutter-studio/src/io/flutter/android/AndroidBuildVariantDetector.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2019 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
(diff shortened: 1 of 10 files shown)


matthew-carroll:
@stevemessick does this change impact how the Flutter module is listed under dependencies in build.gradle?

stevemessick:
@matthew-carroll No. However, if the buildTypes are deleted then the option won't show up in the view.

stevemessick:
https://github.com/flutter/flutter/pull/44797 means this is not needed anymore.

Issue #3562, opened 2019-06-10
simplify integration with dartdoc code samples


devoncarew:
Our current integration with the dartdoc code samples are:

- we show an option to create them in our new project wizards
- we show an editor banner to create them when on a page that contains them

We've learned a bit more about how user's typically interact with these samples, and will have more options in the future (i.e. FlutterPad). I have some suggestions for an update of our UI:

- remove the ability to create the samples in new project wizards (I believe the samples are intended to be more 'in the moment' samples than canonical examples)
- change the editor banner code to regex the current file to locate dartdoc samples
- possibly, have one banner per sample in the file (the current pulldown in the editor banner is easy to miss)

With the above changes, we won't need to download a samples index per flutter sdk (and won't see the issues with the samples pulldown in project wizards not being populated - reported on this issue tracker, as well as on flutter/flutter; https://github.com/flutter/flutter/issues/31916).

@pq @InMatrix for thoughts

InMatrix:
Sounds reasonable. In addition, it would be really useful to cache a preview image of a sample and show it somewhere in the IDE as the user encounters the sample code via Go to Definition, in a spirit similar to showing color and icon previews in the gutter. We can remove the friction of "I'm not sure if this sample will be useful unless I run it to find out."

devoncarew:
What's left on this issue:

- change the editor banner code to regex the current file to locate dartdoc samples

After that, we should be able to remove the code that uses's flutter to enumerate the list of samples.

devoncarew:
Ah, this has landed.

PR #4139, opened 2019-11-25
remove FlutterSampleManager; for embedded flutter samples, open the hosted docs

devoncarew:
Refactor how we handle flutter samples embedded in dartdoc:
- when we see a sample embedded in dartdoc in package:flutter, show an editor banner offering to open the sample in a browser; for the next flutter stable release, this will show in an embedded DartPad instance. For the current stable version, this will show the source for the sample
- remove the (now unused) FlutterSampleManager
- add new tests

I may refactor our tests a bit; I had difficulty testing the code that uses the IntelliJ `Document` class. We may need to resurrect the older `FlutterCodeInsightFixtureTestCase` class (a subclass of `LightPlatformCodeInsightFixtureTestCase`)?

Screen Shot 2019-11-21 at 5 31 07 PM

Screen Shot 2019-11-21 at 5 31 14 PM

https://master-api.flutter.dev/flutter/material/AppBar-class.html:

Screen Shot 2019-11-25 at 12 00 10 PM

(@pq for fyi)


devoncarew:
fix https://github.com/flutter/flutter-intellij/issues/3562

src/io/flutter/samples/FlutterSampleNotificationProvider.java
```diff
@@ -5,65 +5,155 @@
...
-      if (filePath.equals(samplePath)) {
-        samples.add(sample);
+
+    // Find all candidate class definitions.
+    final PsiFile psiFile = PsiDocumentManager.getInstance(project).getPsiFile(document);
+    assert (psiFile != null);
+
+    final List<DartClass> classDeclartions = new ArrayList<>();
+
+    psiFile.acceptChildren(new PsiElementVisitor() {
```
jwren: Consider using, `PsiTreeUtil`, perhaps `getChildrenOfType`

src/io/flutter/samples/FlutterSampleNotificationProvider.java
```diff
@@ -5,65 +5,155 @@
...
-      return Collections.emptyList();
+      return null;
     }
 
+    final String flutterPackagePath = FileUtil.normalize(sdk.getHomePath()) + "/packages/flutter/lib/src/";
     final String filePath = FileUtil.normalize(file.getPath());
-    final String pathSuffix = FileUtil.normalize(sdk.getHomePath()) + "/packages/flutter/";
 
+    // Only show for files in the flutter sdk.
+    if (!filePath.startsWith(flutterPackagePath)) {
```
jwren: Given some `VirtualFile` (even before referencing the `FlutterSdk`), it should be able to bail if the file isn't in the IntelliJ `Library`.  Exiting faster is important when an editor notification is called on every file.

src/io/flutter/samples/FlutterSampleNotificationProvider.java
```diff
@@ -5,65 +5,155 @@
...
+          samples.add(sample);
+        }
       }
     }
 
     return samples;
   }
+
+  // "/// {@tool dartpad ...}"
+  private static final Pattern toolDartPadPattern = Pattern.compile("\\{@tool.*\\sdartpad.*}");
```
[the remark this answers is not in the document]
devoncarew: I think you're right - the convention here is all caps; updated.

src/io/flutter/samples/FlutterSampleNotificationProvider.java
devoncarew: Cool, that's the util I was looking for; switched to `PsiTreeUtil`.

src/io/flutter/samples/FlutterSampleNotificationProvider.java
devoncarew: I definitely agree with the performance considerations. Do you have a pointer to the additional check here - should the File be contained by the Library? Should the Project have a dart sdk Library in the LibraryTable?

devoncarew:
I'm going to land this as is, but will follow up if we think there are performance improvements to FlutterSampleNotificationProvider.createNotificationPanel().

pq:
Thanks for the FYI; looks like a solid approach to me! 👍

src/io/flutter/samples/FlutterSampleNotificationProvider.java
jwren: Change the first if statement to something like

`if (!(fileEditor instanceof TextEditor) && FileTypeRegistry.getInstance().isFileOfType(file, DartFileType.INSTANCE) && !file.isWritable()) {`

This should exit quickly only for Dart files that are not writable by the user (with very few exceptions this will only be Dart/Flutter library files.)

The mentioned file index checks would be checks like this:

`(!)ProjectFileIndex.getInstance(project).isInContent(file) or ProjectFileIndex.getInstance(project).isInLibrary/[Classes/Source](file)`

These would be the APIs if you needed some more generic check (such as for all languages).

I suspect though assuming that `isWritable()` returns `false` above as it should for files in a library, I doubt that the calls out to `ProjectFileIndex` would be faster than directly checking the specific library paths like you are doing here.

src/io/flutter/sdk/FlutterCreateAdditionalSettings.java
```diff
@@ -219,8 +211,7 @@ public Builder setAndroidX(boolean selected) {
     }
 
     public FlutterCreateAdditionalSettings build() {
-      return new FlutterCreateAdditionalSettings(includeDriverTest, type, description, org, swift, kotlin, offlineMode, sampleContent,
```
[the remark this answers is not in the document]
devoncarew: ah, good catch

Issue #4136, opened 2019-11-24
Can not create New flutter project in Android Studio 4.0 Canary 4


ytyng:
## What happened

Android Studio > File (menu) > New > New flutter project...

No dialog opened. Alert "IDE Error Occurred" in right bottom of Android Studio window.

## Version information

Android Studio `4.0 Canary 4` • Flutter plugin `io.flutter 41.2.3` • Dart plugin `192.7761`

Flutter 1.9.1+hotfix.6 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 68587a0916 (2 months ago) • 2019-09-13 19:46:58 -0700
Engine • revision b863200c37
Tools • Dart 2.5.0

## Exception

null

```
java.lang.NoSuchMethodError: com.android.tools.adtui.validation.ValidatorPanel.registerValidator(Lcom/android/tools/idea/observable/ObservableValue;Lcom/android/tools/adtui/validation/Validator;)V
	at io.flutter.project.FlutterProjectStep.onWizardStarting(FlutterProjectStep.java:148)
	at com.android.tools.idea.wizard.model.ModelWizard.<init>(ModelWizard.java:124)
	at com.android.tools.idea.wizard.model.ModelWizard.<init>(ModelWizard.java:47)
	at com.android.tools.idea.wizard.model.ModelWizard$Builder.build(ModelWizard.java:560)
	at io.flutter.actions.FlutterNewProjectAction.actionPerformed(FlutterNewProjectAction.java:52)
	at com.intellij.openapi.actionSystem.ex.ActionUtil$1.run(ActionUtil.java:265)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAware(ActionUtil.java:282)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.lambda$actionPerformed$0(ActionMenuItem.java:292)
	at com.intellij.openapi.wm.impl.FocusManagerImpl.runOnOwnContext(FocusManagerImpl.java:265)
	at com.intellij.openapi.wm.impl.IdeFocusManagerImpl.runOnOwnContext(IdeFocusManagerImpl.java:107)
	at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.actionPerformed(ActionMenuItem.java:282)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
[...]

PR #4140, opened 2019-11-25
Show error dialog when creating a project with canary builds

stevemessick:
We can't currently restrict installation of the beta plugin into Android Studio canary. When trying to create a project you get an error. This adds a catch-block for that error which pops up a message, rather than just reporting a random error.

Screen Shot 2019-11-25 at 11 53 09 AM

Hopefully, this will reduce the number of issues like #4136


stevemessick:
Thanks @jacob314 !

PR #4141, opened 2019-11-26
Refactor of Color parsing code to share more logic with HotUI.

jacob314:


jacob314:
Refactor so that it is easier for the HotUI code to understand Color values written as expressions.
The only new functionality is the getColorName method to return a pretty Colors.someField name for a Color object.

src/io/flutter/editor/ExpressionParsingUtils.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  public static Color parseColor(String text) {
+    final Color color = parseColor(text, "const Color(");
+    if (color != null) return color;
+    return parseColor(text, "Color(");
+  }
+
+  public static Color parseColor(String text, String colorText) {
+    final Integer val = ExpressionParsingUtils.parseNumberFromCallParam(text, colorText);
```
stevemessick: You don't need the class name ExpressionParsingUtils here.

src/io/flutter/editor/ExpressionParsingUtils.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright 2019 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.editor;
+
+import java.awt.*;
+
+public class ExpressionParsingUtils {
```
stevemessick: Could add some tests for this class.

src/io/flutter/editor/ExpressionParsingUtils.java
jacob314: wow. Surprising we don't have Java lint catching that. Only reason it is there is because of my mechanical steps refactoring these methods.

src/io/flutter/editor/ExpressionParsingUtils.java
jacob314: Filed bug https://github.com/flutter/flutter-intellij/issues/4145 to track this.

Issue #4146, opened 2019-11-28
IllegalArgumentException: null file at io.flutter.pub.PubRoot.forDescendant(PubRoot.java:129)


alexander-doroshko:
I see this in user logs. IntelliJ IDEA 2019.2.4  Build #IU-192.7142.36. OS: Windows 10. Flutter plugin version: 41.1.4.
```
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'file' of com/intellij/openapi/roots/impl/ProjectFileIndexImpl.getContentRootForFile must not be null
	at com.intellij.openapi.roots.impl.ProjectFileIndexImpl.$$$reportNull$$$0(ProjectFileIndexImpl.java)
	at com.intellij.openapi.roots.impl.ProjectFileIndexImpl.getContentRootForFile(ProjectFileIndexImpl.java)
	at io.flutter.pub.PubRoot.forDescendant(PubRoot.java:129)
	at io.flutter.run.test.TestFields.getPubRoot(TestFields.java:132)
	at io.flutter.run.test.TestFields.getRelativePath(TestFields.java:140)
	at io.flutter.run.test.TestFields.getSuggestedName(TestFields.java:165)
	at io.flutter.run.test.TestConfig.suggestedName(TestConfig.java:70)
	at com.intellij.execution.impl.RunConfigurable$installUpdateListeners$1.stateChanged(RunConfigurable.kt:363)
	at com.intellij.openapi.options.SettingsEditor.fireEditorStateChanged(SettingsEditor.java:136)
	at com.intellij.openapi.options.SettingsEditor$2.stateChanged(SettingsEditor.java:114)
	at jdk.internal.reflect.GeneratedMethodAccessor2137.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.intellij.util.EventDispatcher.dispatchVoidMethod(EventDispatcher.java:130)
	at com.intellij.util.EventDispatcher.access$000(EventDispatcher.java:24)
	at com.intellij.util.EventDispatcher$1.invoke(EventDispatcher.java:88)
	at com.sun.proxy.$Proxy198.stateChanged(Unknown Source)
	at com.intellij.ui.UserActivityWatcher.fireUIChanged(UserActivityWatcher.java:92)
[...]

PR #4147, opened 2019-11-28
fix an NPE in PubRoot.forDescendant

devoncarew:
- fix an NPE in PubRoot.forDescendant
- fix https://github.com/flutter/flutter-intellij/issues/4146


devoncarew:
I'm going to TBR this, but will update w/ any review comments after the fact.

PR #4164, opened 2019-12-03
Slowness

jacob314:


devoncarew:
Thanks for taking this over and landing it!

src/io/flutter/bazel/Workspace.java
```diff
@@ -193,6 +194,7 @@ public int hashCode() {
    */
   @Nullable
   public static Workspace load(@NotNull Project project) {
+    if (!FlutterSettings.getInstance().shouldUseBazel()) return null;
```
devoncarew: This will address the performance issue for people not in google3, but users in google3 will still see a significant performance degradation. We're performing an expensive operation for every PSI element in open editor files. This is something we should address; @DaveShuckerow. The ability to have test editor annotations is valuable, but we want to follow up and make sure this feature lands well.

PR #4173, opened 2019-12-04
Stop showing the screen mirror when the version of Flutter a user has won't support it.

jacob314:
image


jacob314:
I'm fine with just calling it Properties. Only argument for "Widget Properties" is we only show it for widgets.

PR #4180, opened 2019-12-05
Fix module path bug on windows

stevemessick:
@kenzieschmoll I think this fixes the problem you found when creating an add-to-app module independently of its Android app.

### DIFF
--- a/flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
+++ b/flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
@@ -79,7 +79,7 @@ public class FlutterModuleImporter {
       showHowToEditDialog();
       return;
     }
-    myRelativePath = newPath;
+    myRelativePath = newPath.replace('\\', '/');
     VirtualFile settingsFile = projectRoot.findChild("settings.gradle");
     if (settingsFile == null) {
       showHowToEditDialog();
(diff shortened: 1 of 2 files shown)


flutter-studio/src/io/flutter/module/FlutterModuleImporter.java
```diff
@@ -79,7 +79,7 @@ public void importModule() {
       showHowToEditDialog();
       return;
     }
-    myRelativePath = StringUtil.escapeBackSlashes(newPath);
+    myRelativePath = newPath.replace('\\', '/');
```
jacob314: nit: maybe add a helper utils method that normalizes paths to use '/' instead of '\'

devoncarew: I think you want `FileUtil.normalize()` here (https://github.com/flutter/flutter-intellij/pull/4172/files#diff-fcaf5e6bc7e277244e7374343c3b29e3L63).

stevemessick: FileUtil.normalize() does it.

alexander-doroshko: Generally we use `com.intellij.openapi.util.io.FileUtil#toSystemDependentName` and `com.intellij.openapi.util.io.FileUtil#toSystemIndependentName`.

Issue #4187, opened 2019-12-07
NPE from ActiveEditorsOutlineService.updateActiveEditors(), ActiveEditorsOutlineService.java:139


devoncarew:
`42.0.1`

```
java.lang.NullPointerException
com.intellij.openapi.util.io.FileUtilRt.toSystemDependentName(), FileUtilRt.java:391
com.intellij.openapi.util.io.FileUtilRt.toSystemDependentName(), FileUtilRt.java:386
com.intellij.openapi.util.io.FileUtil.toSystemDependentName(), FileUtil.java:597
io.flutter.editor.ActiveEditorsOutlineService.updateActiveEditors(), ActiveEditorsOutlineService.java:139
io.flutter.editor.ActiveEditorsOutlineService.access$000(), ActiveEditorsOutlineService.java:44
io.flutter.editor.ActiveEditorsOutlineService$1.selectionChanged(), ActiveEditorsOutlineService.java:78
sun.reflect.GeneratedMethodAccessor60.invoke(), undefined:-1
sun.reflect.DelegatingMethodAccessorImpl.invoke(), DelegatingMethodAccessorImpl.java:43
java.lang.reflect.Method.invoke(), Method.java:498
com.intellij.util.messages.impl.MessageBusConnectionImpl.deliverMessage(), MessageBusConnectionImpl.java:114
com.intellij.util.messages.impl.MessageBusImpl.doPumpMessages(), MessageBusImpl.java:393
com.intellij.util.messages.impl.MessageBusImpl.pumpWaitingBuses(), MessageBusImpl.java:354
com.intellij.util.messages.impl.MessageBusImpl.pumpMessages(), MessageBusImpl.java:343
com.intellij.util.messages.impl.MessageBusImpl.pumpMessages(), MessageBusImpl.java:327
com.intellij.util.messages.impl.MessageBusImpl.sendMessage(), MessageBusImpl.java:320
com.intellij.util.messages.impl.MessageBusImpl.access$100(), MessageBusImpl.java:29
com.intellij.util.messages.impl.MessageBusImpl$1.invoke(), MessageBusImpl.java:192
com.sun.proxy.$Proxy119.selectionChanged(), undefined:-1
com.intellij.openapi.fileEditor.impl.FileEditorManagerImpl.lambda$fireSelectionChanged$16(), FileEditorManagerImpl.java:1598
[...]

PR #4188, opened 2019-12-07
fix two NPEs from the outline view

devoncarew:
- fix two NPEs from the outline view
- fix https://github.com/flutter/flutter-intellij/issues/4187

The VirtualFile.getCanonicalPath() call can return null for some file (editor?) types.


src/io/flutter/editor/ActiveEditorsOutlineService.java
```diff
@@ -110,19 +110,21 @@ private void updateActiveEditors() {
 
     final Set<String> newPaths = new HashSet<>();
     for (VirtualFile file : files) {
-      if (FlutterUtils.isDartFile(file)) {
+      if (FlutterUtils.isDartFile(file) && file.getCanonicalPath() != null) {
```
jacob314: wow. didn't realize we could have Dart files without an actual path. Guess it makes sense.

Issue #4189, opened 2019-12-07
java.util.concurrent.TimeoutException from io.flutter.inspector.EvalOnDartLibrary


devoncarew:
`42.0.2`

```
java.util.concurrent.TimeoutException
	java.util.concurrent.CompletableFuture.encodeThrowable(), CompletableFuture.java:331
	java.util.concurrent.CompletableFuture.completeThrowable(), CompletableFuture.java:346
	java.util.concurrent.CompletableFuture$UniApply.tryFire(), CompletableFuture.java:632
	java.util.concurrent.CompletableFuture.postComplete(), CompletableFuture.java:506
	java.util.concurrent.CompletableFuture.completeExceptionally(), CompletableFuture.java:2088
	io.flutter.inspector.EvalOnDartLibrary.lambda$timeoutAfter$4(), EvalOnDartLibrary.java:182
	java.util.concurrent.FutureTask.run(), FutureTask.java:264
	java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(), ScheduledThreadPoolExecutor.java:304
	java.util.concurrent.ThreadPoolExecutor.runWorker(), ThreadPoolExecutor.java:1128
	java.util.concurrent.ThreadPoolExecutor$Worker.run(), ThreadPoolExecutor.java:628
	java.lang.Thread.run(), Thread.java:834
```

devoncarew:
Bumping this to the next milestone - we'll have better info from the next failures then.

pq:
Related to the legacy inspector; closing as stale.

PR #4193, opened 2019-12-10
provide more stack trace details for inspector timeouts

devoncarew:
- provide more stack trace details for inspector timeouts

This will help us track down issues like https://github.com/flutter/flutter-intellij/issues/4189 - to know whether they represent real problems, or things that we should ignore (not log as errors) at the call sites.


devoncarew:
This works by creating an exception object in the `timeout()` method, and - if we do timeout - to provide that exception as the cause to the timeout.

devoncarew:
note: updated to not use nested exceptions - just create the timeout exception at method entry, but don't throw it until and if we actually time out

PR #4200, opened 2019-12-11
Launch flutter attach when the Android app starts

stevemessick:
When an Android project containing an add-to-app module is opened, add a listener that detects app launches. The listener then launches `fluter attach` for it. The Attach button on the toolbar will seldom need to be used; only when there are multiple Flutter run configs is it needed. In conjunction, change the Attach button to be disabled when the attach process is running, and re-enable it when the process terminates.

@xster I'm not sure if this gets everything in the video you added to #4064, but I'll get you a new build to check out.
@devoncarew @DaveShuckerow for review.

### DIFF
--- a/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
+++ b/flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
@@ -38,6 +38,7 @@ import io.flutter.run.FlutterReloadManager;
 import io.flutter.run.SdkRunConfig;
 import io.flutter.sdk.FlutterSdk;
 import io.flutter.settings.FlutterSettings;
+import io.flutter.utils.AddToAppUtils;
 import io.flutter.utils.AndroidUtils;
 import io.flutter.utils.FlutterModuleUtils;
 import java.lang.reflect.Field;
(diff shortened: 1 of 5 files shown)


xster:
Oh this is a great idea that doesn't need to disrupt a lot of existing code. 

cc @blasten for his thoughts too. 

Does attaching have 'side effects' like opening panels in the IDE etc? If the user is only trying to develop Android but end up opening an activity which creates the VM and triggers the attach.

flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
```diff
@@ -109,4 +132,62 @@ public void sourceGenerationFinished(@NotNull Project project) {
...
+      };
+
+      @Override
+      public void sessionCreated(DebuggerSession session) {
+        session.getProcess().addDebugProcessListener(dpl);
+      }
+
+      @Override
+      public void sessionRemoved(DebuggerSession session) {
+        session.getProcess().removeDebugProcessListener(dpl);
```
blasten: Does this kill the attach process? Could it ever leak now that attach stays alive after the observatory is closed?

stevemessick: @blasten This runs when the Android app exits. At that point, my experiments all showed the attach process had already terminated. I don't think the observatory is involved here, but I honestly don't know for sure.

stevemessick:
@xster 
> Does attaching have 'side effects' like opening panels in the IDE etc? If the user is only trying to develop Android but end up opening an activity which creates the VM and triggers the attach.

When the Debug button is clicked the Debugger will open up. You'll see both Android and Flutter debuggers sharing the same screen real estate. Other than that this is pretty much invisible. The Attach button gets disabled, since you don't want it launching another attach process. And `ps xa | grep flutter` will show an additional process. It only works for the Debug button; the Run button is not affected, keeping in line with Jetbrains design for the debugger.

xster:
SGTM! Thanks

flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
```diff
@@ -56,6 +73,12 @@ public void moduleAdded(@NotNull Project proj, @NotNull Module mod) {
       });
       return;
     }
+    else {
+      if (ANDROID_PROJECT_TYPE.equals(ProjectTypeService.getProjectType(project))) {
```
devoncarew: Move project checking logic into a utility method in an add to app utils class?

flutter-studio/src/io/flutter/FlutterStudioStartupActivity.java
```diff
@@ -37,17 +54,17 @@
   public void runActivity(@NotNull Project project) {
     MessageBusConnection connection = project.getMessageBus().connect(project);
     // GRADLE_SYNC_TOPIC is not public in Android Studio 3.5. It is in 3.6. It isn't defined in 3.4.
-    Topic topic = getStaticFieldValue(GradleSyncState.class, Topic.class, "GRADLE_SYNC_TOPIC");
-    assert topic != null;
     //noinspection unchecked
-    connection.subscribe((Topic<GradleSyncListener>)topic, makeSyncListener(project));
+    Topic<GradleSyncListener> topic = getStaticFieldValue(GradleSyncState.class, Topic.class, "GRADLE_SYNC_TOPIC");
```
devoncarew: There's a lot of add to app code in `FlutterStudioStartupActivity`; perhaps split the code out into a separate class and reference it from here?

stevemessick:
> Are there any downsides to running flutter attach for every app launch? We may want to test this out as a user preference first - shipping initially as as opt-in preference. Thoughts?

That's a good question. `flutter attach` is a daemon process, similar to the one the plugin uses already. It only runs if the Android app has an add-to-app module, so it seems relatively low-risk. I can add a preference anyway -- wouldn't hurt, except we need to document it. I'll do the refactoring you suggested, too. 

cc @xster for work flow change.

xster:
I think always launching flutter attach for add-to-app co-editing users seems reasonable. I don't see any particular downsides since then it behaves the same way as Android (i.e. they can disconnect or just run without debug if they don't want their breakpoints to hit)

PR #4213, opened 2019-12-13
Generate dev log

stevemessick:
Add a channel option, to specify stable or dev build.
Automatically create the change log for the dev channel.
Misc clean up, like deleting `new` and changing `Future(() => value)` to `value`.

Requires #4200 to build the dev channel successfully.


.travis.yml
```diff
@@ -38,3 +38,4 @@ env:
   - IDEA_VERSION=3.6
   - IDEA_VERSION=2019.2.4
   - IDEA_VERSION=2019.3
+  - IDEA_VERSION=4.0
```
devoncarew: Just confirming that you do intend to compile against 4.0 as part of this PR.

tool/plugin/lib/plugin.dart
```diff
@@ -896,11 +887,11 @@ class DeployCommand extends ProductCommand {
   Future<int> doit() async {
     if (isReleaseMode) {
       if (!await performReleaseChecks(this)) {
-        return new Future(() => 1);
+        return 1;
```
devoncarew: Not necessarily for this PR, but we should probably change this to returning a constant (like `cliExitError`).

tool/plugin/lib/plugin.dart
```diff
@@ -1148,3 +1147,35 @@ class TestCommand extends ProductCommand {
     throw 'integration test execution not yet implemented';
   }
 }
+
+Future<String> makeDevLog(BuildSpec spec) async {
+  _checkGitDir();
+  var gitDir = await GitDir.fromExisting(rootPath);
+  var since = await lastRelease();
+  var processResult =
+      await gitDir.runCommand(['log', '$since..HEAD', '--oneline']);
```
devoncarew: Can you include some sample output from this? There are some options you can pass to `log` to tweak the output.

tool/plugin/lib/plugin.dart
```diff
@@ -1148,3 +1147,35 @@ class TestCommand extends ProductCommand {
...
+    devLog.writeln(m.replaceFirst(RegExp(r'^[0-9A-Fa-f]+ '), '- '));
+  });
+  devLog.writeln();
+  return devLog.toString();
+}
+
+Future<String> lastRelease() async {
+  _checkGitDir();
+  var gitDir = await GitDir.fromExisting(rootPath);
+  var processResult = await gitDir.runCommand(['branch', '--list', 'release*']);
```
devoncarew: Is this output sorted in some way (chronologically)? It looks like below we do depend on a specific ordering.

devoncarew: Also, will developers have these branches locally? These will be branches on the upstream repo.

.travis.yml
stevemessick: Not really. That's the dev channel build, which isn't (currently) built on travis. It does go through setup but the plugin tool exits without doing any work. Perhaps the `plugin generate` function should be changed to only generate the builds for the given channel; I haven't gotten so far into this that I know how that will work yet.

tool/plugin/lib/plugin.dart
stevemessick: I could make it check for capitalization of the first word, but haven't gotten that far.
```
## Changes since release_42
- if track widget creation is not enabled, pass in --no-track-widget-creation (#4196)
- Surface reload exceptions (#4198)
- provide more stack trace details for inspector timeouts (#4193)
- Update change log for M42.1 (#4192)
```

tool/plugin/lib/plugin.dart
stevemessick: According to the git docs it is sorted by byte value. If we're still doing this in 5 years we will have a problem when we get to release_100.

I have the branches locally, as will anyone who does a build. But to run this as a service it will need to make sure the branch is checked out locally. If that's travis we'll have to modify the .travis.yml to include the release_* branch for a dev channel build -- if we use travis. Sounds like another change to `plugin generate` once we know where this will run.

tool/plugin/lib/plugin.dart
devoncarew: Thanks; sounds like we can tweak the output here as necessary.

tool/plugin/lib/plugin.dart
devoncarew: The issue as I see it is that people likely won't have all the remote branches available locally (this would happen if one person does M40 and another does M41, or, if someone hasn't driven a release for a while).

We may want to look at some additional git-fu here. If there's a remote repo named upstream, we should likely query the branches for the upstream fork. If there's no upstream, we can fall-back on querying the local branches.

devoncarew: It looks like `-r` will show remote tracking branches. Those remote branches might not exist locally however if the users hasn't run `git fetch upstream`.

stevemessick: I added a TODO here, and fixed plugin.xml generation. It was setting the untilVersion too low.
PTAL

devoncarew: 👍 lgtm

Issue #4215, opened 2019-12-13
AssertionError from PropertyEditorPanel.getOffset(PropertyEditorPanel.java:334)


devoncarew:
`42.1.3` / `IntelliJ IDEA/2019.2.4`

```
java.lang.AssertionError
	at io.flutter.editor.PropertyEditorPanel.getOffset(PropertyEditorPanel.java:334)
	at io.flutter.editor.PropertyEditorPanel.updateWidgetDescription(PropertyEditorPanel.java:351)
	at io.flutter.editor.PropertyEditorPanel.maybeLoadProperties(PropertyEditorPanel.java:432)
	at io.flutter.editor.PropertyEditorPanel.outlinesChanged(PropertyEditorPanel.java:407)
	at io.flutter.utils.StreamSubscription.notify(StreamSubscription.java:43)
	at io.flutter.utils.EventStream.lambda$setValue...
```

cc @jacob314

devoncarew:
fixed by #4215

PR #4220, opened 2019-12-16
Fail gracefully if there is not an active file just as we fail gracefully if there is no outline.

jacob314:
…

### DIFF
--- a/src/io/flutter/editor/PropertyEditorPanel.java
+++ b/src/io/flutter/editor/PropertyEditorPanel.java
@@ -330,8 +330,8 @@ public class PropertyEditorPanel extends SimpleToolWindowPanel {
   int getOffset() {
     if (outline == null) return -1;
     final VirtualFile file = activeFile.getValue();
-    if (activeFile == null) return -1;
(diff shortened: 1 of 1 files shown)


jacob314:
Fixes https://github.com/flutter/flutter-intellij/issues/4215

src/io/flutter/editor/PropertyEditorPanel.java
```diff
@@ -330,8 +330,7 @@ InspectorObjectGroupManager getGroupManager() {
   int getOffset() {
     if (outline == null) return -1;
     final VirtualFile file = activeFile.getValue();
-
-    assert (activeFile.getValue() != null);
+    if (activeFile == null) return -1;
```
devoncarew: We've already dereferenced `activeFile` above; perhaps you meant `file` here? We should also replace the `activeFile.getValue()` reference below with `file`.

jacob314: doh. meant to wrote file instead of activeFile. fixed

devoncarew:
(merging)

Issue #4221, opened 2019-12-17
Exception when the color of Icon changes through the Flutter Outline


cjamcu:
## What happened

I was changing the color to an `Icon` widget across` Flutter Outline`

## Version information

Android Studio `3.5.3` • Flutter plugin `io.flutter 42.1.1` • Dart plugin `191.8593`

Flutter 1.9.1+hotfix.6 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 68587a0916 (3 months ago) • 2019-09-13 19:46:58 -0700
Engine • revision b863200c37
Tools • Dart 2.5.0

## Exception

null

```
java.lang.NullPointerException
	at io.flutter.editor.ColorField.buildColorExpression(ColorField.java:115)
	at io.flutter.editor.ColorField.applyColor(ColorField.java:219)
	at io.flutter.editor.AndroidStudioColorPickerProvider.lambda$show$0(AndroidStudioColorPickerProvider.java:46)
	at com.android.tools.idea.ui.resourcechooser.colorpicker2.OperationPanel$2.actionPerformed(OperationPanel.kt:58)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
	at java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:289)
	at java.awt.Component.processMouseEvent(Component.java:6550)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3325)
	at java.awt.Component.processEvent(Component.java:6315)
	at java.awt.Container.processEvent(Container.java:2239)
	at java.awt.Component.dispatchEventImpl(Component.java:4899)
	at java.awt.Container.dispatchEventImpl(Container.java:2297)
	at java.awt.Component.dispatchEvent(Component.java:4721)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4904)
	at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4535)
	at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4476)
[...]

PR #4249, opened 2020-01-03
fix an npe related to hot ui

devoncarew:
- fix an npe related to hot ui (`buildColorExpression()` could see a null `color` parameter)
- fix https://github.com/flutter/flutter-intellij/issues/4221

### DIFF
--- a/src/io/flutter/editor/FlutterColors.java
+++ b/src/io/flutter/editor/FlutterColors.java
@@ -97,8 +97,8 @@ public class FlutterColors {
   /**
    * Returns the the shortest material color name matching a color if one exists.
    */
-  public static @Nullable
-  String getColorName(Color color) {
+  @Nullable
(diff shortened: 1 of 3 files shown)


src/io/flutter/editor/FlutterColors.java
```diff
@@ -97,8 +97,8 @@ else if (colors.containsKey(key + primarySuffix)) {
   /**
    * Returns the the shortest material color name matching a color if one exists.
    */
-  public static @Nullable
-  String getColorName(Color color) {
+  @Nullable
+  public static String getColorName(@Nullable Color color) {
     String name = colorToName.get(color);
```
kenzieschmoll: will `colorToName.get(color)` throw an exception if `color` is null?

devoncarew: I don't believe so - it'll just return `null`.

PR #4253, opened 2020-01-05
draw a separator line between the main inspector tree and the details tree

devoncarew:
- draw a separator line between the main inspector tree and the details tree
- remove an override that will be removed from the platform in the 2020.1 version


src/io/flutter/inspector/InspectorTree.java
```diff
@@ -48,11 +48,6 @@ public void setHighlightedRoot(DefaultMutableTreeNode value) {
     repaint();
   }
 
-  @Override
```
devoncarew: This is deprecated and will be removed for 2020.1.

PR #4275, opened 2020-01-10
fix an issue with sending in non-absolute paths to flutter.subscriptions

devoncarew:
- fix an issue with sending in non-absolute paths to flutter.subscriptions
- fix https://github.com/flutter/flutter-intellij/issues/4273

I'm not sure how non-absolute paths are getting into our manager tracking `flutter.setSubscriptions` for files; this change may address the issue.

### DIFF
--- a/src/io/flutter/editor/ActiveEditorsOutlineService.java
+++ b/src/io/flutter/editor/ActiveEditorsOutlineService.java
@@ -180,7 +180,7 @@ public class ActiveEditorsOutlineService implements Disposable {
    * To get an outline that is guaranteed in-sync with the file it outlines, see {@link #getIfUpdated}.
    */
   @Nullable
-  public FlutterOutline getOutline(String path) {
+  public FlutterOutline getOutline(@Nullable String path) {
     return pathToOutline.get(path);
   }
 
@@ -192,7 +192,7 @@ public class ActiveEditorsOutlineService implements Disposable {
    */
   @Nullable
   public FlutterOutline getIfUpdated(@NotNull PsiFile file) {
-    final FlutterOutline outline = getOutline(file.getVirtualFile().getCanonicalPath());
+    final FlutterOutline outline = getOutline(file.getOriginalFile().getVirtualFile().getPath());
     if (isOutdated(outline, file)) {
       return null;
     }
--- a/src/io/flutter/editor/ActiveEditorsOutlineService.java
+++ b/src/io/flutter/editor/ActiveEditorsOutlineService.java
@@ -110,7 +110,7 @@ public class ActiveEditorsOutlineService implements Disposable {
 
     final Set<String> newPaths = new HashSet<>();
     for (VirtualFile file : files) {
-      if (file.isInLocalFileSystem() && FlutterUtils.isDartFile(file)) {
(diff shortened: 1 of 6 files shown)


src/io/flutter/editor/ActiveEditorsOutlineService.java
```diff
@@ -192,7 +192,7 @@ public FlutterOutline getOutline(String path) {
    */
   @Nullable
   public FlutterOutline getIfUpdated(@NotNull PsiFile file) {
-    final FlutterOutline outline = getOutline(file.getVirtualFile().getCanonicalPath());
+    final FlutterOutline outline = getOutline(file.getOriginalFile().getVirtualFile().getPath());
```
stevemessick: `getCanonicalPath()` is used in many other places. Why change to `getPath()` here? It looks like the initialization is (line 114 in the original):
`newPaths.add(file.getCanonicalPath());`.

alexander-doroshko: I haven't reviewed neither this nor other usages of getCanonicalPath(). Just noticed this word and remembered that normally we don't need to use this method. Its javadoc says that in fact in most cases we should work with usual paths. As I understand, we might need to work with canonical paths only for some known corner cases with symlinks. But I don't think there're a lot. For example, in Dart plugin there are zero usages of getCanonicalPath().

devoncarew: Thanks @stevemessick, @alexander-doroshko; I'll review our uses here. We should definitely use the same technique (either getPath or getCanonicalPath) when storing clearing from the path list.

The main part of this specific fix is that for PsiFiles, you can sometimes get non-absolute paths from getPath() - I believe if the PsiFile file is backed by a LightVirtualFile? The getOriginalFile() should avoid that, and then getFile() should return an absolute path (which we need later for the analysis server).

alexander-doroshko: - `PsiFile.getOriginalFile()` normally returns `this`. There are just a few cases when original file is different. The main use case is code completion. Platform creates a fake copy of `PsiFile` and performs completion there, so `CompletionContributors` might need to use original files. In all other cases we usually don't care and use `PsiFile` as is without calling `getOriginalFile()`. So as a cleanup, we may get rid of unneeded `getOriginalFile()` calls in our plugins. OTOH, the only harm from such calls is 17 useless characters in the source code.
- In those cases when `PsiFile` is not equal to `PsiFile.getOriginalFile()`, `PsiFile.getVirtualFile()` returns `null`. So `getOriginalFile()` can't fix the exception from #4273
- There are 116 subclasses of `VirtualFile` and most of them may be opened in an editor. So, it's not safe to send `fileInEditor.getPath()` to the server, and #4273 proves that. In the Dart plugin we check `VirtualFile.isInLocalFileSystem()` and `isFileNameRespectedByAnalysisServer()` (see `DartAnalysisServerService.isLocalAnalyzableFile()`). You may want to do the same in the Flutter plugin.

devoncarew: Great, thanks for the review. I'll update or re-create this PR w/ the new info.

Issue #4269, opened 2020-01-08
Only display Flutter UI elements on projects that have a Flutter module


stevemessick:
This was initially reported as #267 and it mostly works. But my Flutter dev project has the Flutter device selector showing for some reason. The project that is open is intellij.idea.community.main and does not have any Flutter modules, but it does have Java modules named 'flutter-intellij-community' and 'flutter-studio'.

Also, 'Flutter Inspector', 'Flutter Outline', and 'Flutter Performance' should not be visible on non-Flutter projects. They do show up on Android projects and Java projects today.

devoncarew:
@stevemessick, https://github.com/flutter/flutter-intellij/pull/4281 address part of this issue.

Can you add a bit more details about why the device selector shows up in some non-Flutter projects? From reading the code, that should only happen if the `flutter deamon` process has been started for that project. This may be a degenerate case? `flutter-intellij-community` is likely a valid Dart project, and the Dart SDK you used for it might be the dart vended into the flutter sdk.

stevemessick:
I think you are right, it only shows up in that one project. Not worth the trouble to track down why.

devoncarew:
See the closed, partial PR for this: https://github.com/flutter/flutter-intellij/pull/4281.

PR #4281, opened 2020-01-15
make displaying the Flutter tool windows conditional on having a Flutter module

devoncarew:
- make displaying the Flutter tool windows conditional on having a Flutter module
- address part of https://github.com/flutter/flutter-intellij/issues/4269

cc @stevemessick, @alexander-doroshko

### DIFF
--- a/src/io/flutter/sdk/IsFlutterProjectCondition.java
+++ b/src/io/flutter/sdk/IsFlutterProjectCondition.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2020 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.sdk;
+
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.Condition;
+import io.flutter.utils.FlutterModuleUtils;
+
+import java.util.List;
+
+/**
+ * A condition that returns true if any modules include a content root with a Dart package
+ * with a 'flutter:' dependency.
+ */
+public class IsFlutterProjectCondition implements Condition<Project> {
+  @Override
+  public boolean value(Project project) {
+    final List<Module> flutterModules = FlutterModuleUtils.findModulesWithFlutterContents(project);
+
+    return !flutterModules.isEmpty();
+  }
+}
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -1150,15 +1150,18 @@
     <errorHandler implementation="io.flutter.FlutterErrorReportSubmitter"/>
 
     <toolWindow id="Flutter Outline" anchor="right" icon="FlutterIcons.Flutter_13"
-                factoryClass="io.flutter.preview.PreviewViewFactory"/>
+                factoryClass="io.flutter.preview.PreviewViewFactory"
+                conditionClass="io.flutter.sdk.IsFlutterProjectCondition"/>
     <projectService serviceImplementation="io.flutter.preview.PreviewView" overrides="false"/>
 
     <toolWindow id="Flutter Inspector" anchor="right" icon="FlutterIcons.Flutter_13"
-                factoryClass="io.flutter.view.FlutterViewFactory"/>
+                factoryClass="io.flutter.view.FlutterViewFactory"
+                conditionClass="io.flutter.sdk.IsFlutterProjectCondition"/>
     <projectService serviceImplementation="io.flutter.view.FlutterView" overrides="false"/>
 
     <toolWindow id="Flutter Performance" anchor="right" icon="FlutterIcons.Flutter_13"
-                factoryClass="io.flutter.performance.FlutterPerformanceViewFactory"/>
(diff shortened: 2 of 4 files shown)


alexander-doroshko:
I see that `conditionClass` is now deprecated in master, docs suggest to use `ToolWindowFactory#isApplicable(Project)` instead. But this `isApplicable(Project)` is available since 2020.1 only.

But I'm not sure about `conditionClass` / `isApplicable(Project)` approach anyway. Will this work if you open an empty project (`IsFlutterProjectCondition == false`) and then copy Flutter project content to this project?

You might want to look at Dart Analysis tool window management. It has `DartAnalysisToolWindowFactory.shouldBeAvailable == false`. It appears only when `DartProblemsView` explicitly calls `toolWindow.getContentManager().getContent()`. See https://github.com/JetBrains/intellij-plugins/commit/3101f19ba3033ba9e19d34469e0a3a751bc3d50d.

devoncarew:
Thanks for the comments. I did try `isApplicable` first, but didn't see that it took effect.

> Will this work if you open an empty project (IsFlutterProjectCondition == false) and then copy Flutter project content to this project?

This wouldn't handle that case. I did test with a newly created Java project, and a newly created Flutter one, and saw the expected results, but suspect that a project changing types after it was opened would not be handled.

I'll keep this PR open and look at the Dart plugin's inpl.

devoncarew:
Similar to that referenced PR, we could do something like show the views the first time a Flutter SDK was retrieved or set for a project. They wouldn't show for Flutter project's w/o SDKs set, but that's probably OK.

Or we could introduce a listener and callback for Flutter-ness for a project - trigger the tool windows becoming visible once we first detect that it's a Flutter project.

alexander-doroshko:
Empty tool windows don't make much sense. At which moment does some meaningful content appear there? I think tool windows should appear exactly at this moment. If you follow the idea behind Dart Analysis tool window, you'll probably get this behavior automatically. The first call of `toolWindow.getContentManager().getContent()` (in a method that adds something meaningful to the tool window) will effectively call your `createToolWindowContent()` and the tool window will appear. AFAIU, API that you need has been stable stable for a long time and you'll need neither deprecated not not-yet-available-in-2019.x API.

devoncarew:
> At which moment does some meaningful content appear there? I think tool windows should appear exactly at this moment. 

For the `Flutter Outline` tool window, that's as soon as we know that the module is a Flutter (or Dart) module. For the `Flutter Inspector` and `Flutter Performance` tool windows, that's when at app has started.

The reason we want to show these two other tool windows all the time is because:
- generally, we think the `Flutter Outline` view surfaces information that's useful to users; we want them to have it open unless they've explicitly closed it
- when an app runs, user often inspect the app with the inspector or performance tool window
- if these tool window close when the app exits, it'll collapse the right-hand side tool window area
- this also closes the Flutter Outline view; there are then discoverability issues with the view - users often don't remember to re-open it
- so, we show 'no app running' content in those tool windows after an app exits, but keep the tools windows open

If they don't show the tool windows on app open, we could run into the same issue; if the flutter inspector is the active tool window when they close intellij, when intellij re-opens, if we don't show the flutter inspector tool window, the right hand side tool window area will be collapsed, taking the flutter outline tool window with it.

alexander-doroshko:
> For the Flutter Outline tool window, that's as soon as we know that the module is a Flutter (or Dart) module.

Maybe as soon as a Dart file is opened in the editor?

> we want them to have it open unless they've explicitly closed it

We do this for the Dart Analysis tool window. Unfortunately, Platform can't help us to restore tool window state if a tool window is not visible right from the start. But we remember the tool window state ourselves and call `toolWindow.activate()` once created if needed.  See https://github.com/JetBrains/intellij-plugins/commit/ec02e4d03aa27846a5fb7e8e7e53449287b20c78. The logic is not trivial (additional listener used) but I haven't found a better way.

> we show 'no app running' content in those tool windows after an app exits, but keep the tool windows open

I think it's fine. I think it's important that tool windows don't appear earlier than they are useful. But if they stay open afterwards - that's less a problem (or not at all).

src/io/flutter/sdk/IsFlutterProjectCondition.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.List;
+
+/**
+ * A condition that returns true if any modules include a content root with a Dart package
+ * with a 'flutter:' dependency.
+ */
+public class IsFlutterProjectCondition implements Condition<Project> {
+  @Override
+  public boolean value(Project project) {
+    final List<Module> flutterModules = FlutterModuleUtils.findModulesWithFlutterContents(project);
```
jacob314: You may want to check this with a Bazel project. I just saw a case where a bazel project didn't show up as having a Flutter module but otherwise worked fine.
WorkspaceCache.getInstance(project).isBazel()
will cheaply indicate whether a project uses bazel. The downside is it might report some bazel projects that don't actually use Dart.

jkoenig134:
When will this be addressed?

jacob314:
Should this pull request be closed?

devoncarew:
> Should this pull request be closed?

Yes, I keep meaning to come back to it. I'll close it, and reference it in an issue, so the work can be found if we need it.

PR #4290, opened 2020-01-16
Restore ability to use Java 11

stevemessick:
If you have the option set to convert multiple imports from a package into a single implicit import, and you change the explicit import of IntelliJ's `Module` into `*`, and you are using Java 8, you will break compilation using Java 11.


devoncarew:
Hmm, I wonder if we should have an additional bot on our CI that tries to compile with Java 11?

devoncarew:
Or change the settings for import organization for IntelliJ?

stevemessick:
I added an issue to compile with 11, https://github.com/flutter/flutter-intellij/issues/4291

PR #4295, opened 2020-01-16
Switch to using WorkspaceCache

jacob314:
Fix race condition in workspace cache and simplify WorkspaceCache to stop
trying to do work in the background.

### DIFF
--- a/src/io/flutter/bazel/WorkspaceCache.java
+++ b/src/io/flutter/bazel/WorkspaceCache.java
@@ -5,15 +5,21 @@
  */
 package io.flutter.bazel;
 
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableSet;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.registry.Registry;
+import io.flutter.FlutterUtils;
 import io.flutter.project.ProjectWatch;
 import io.flutter.utils.FileWatch;
-import io.flutter.utils.Refreshable;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import javax.swing.*;
+import java.util.LinkedHashSet;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 
 /**
@@ -23,24 +29,25 @@ import java.util.concurrent.atomic.AtomicReference;
  */
 public class WorkspaceCache {
   @NotNull private final Project project;
-  @NotNull private final Refreshable<Workspace> cache;
+  @Nullable private Workspace cache;
+
+  private boolean refreshScheduled = false;
+
+  private final Set<Runnable> subscribers = new LinkedHashSet<>();
 
   private WorkspaceCache(@NotNull final Project project) {
     this.project = project;
 
-    cache = new Refreshable<>();
-    cache.setDisposeParent(project);
-
     // Trigger a reload when file dependencies change.
     final AtomicReference<FileWatch> fileWatch = new AtomicReference<>();
-    cache.subscribe(() -> {
+    subscribe(() -> {
       if (project.isDisposed()) return;
 
-      final Workspace next = cache.getNow();
+      final Workspace next = cache;
 
       FileWatch nextWatch = null;
       if (next != null) {
-        nextWatch = FileWatch.subscribe(next.getRoot(), next.getDependencies(), this::refreshAsync);
+        nextWatch = FileWatch.subscribe(next.getRoot(), next.getDependencies(), this::scheduleRefresh);
         nextWatch.setDisposeParent(project);
       }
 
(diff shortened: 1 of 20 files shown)


jacob314:
This fixes performance issues loading bazel workspaces and race conditions where WorkspaceCache.getNow() would sometimes return there wasn't a bazel workspace when there was causing race conditions where projects would incorrectly use the non-bazel mode with surprising results.

WorkspaceCache used the `Refreshable` api which was overkill given that loading the workspace file can be performed on the UI thread and the workspace needs to be accessed reliably from the UI thread.

The documentation for the .getNow() getter on WorkspaceCache claimed to always return a valid bazel Workspace but that doesn't match the implementation using `Refreshable`. Changing the implementation of `Refreshable` to block on startup was too complex and risky given that complexity isn't needed.

Fix UI showing the Flutter SDK to reflect the fact that a fixed SDK is used for a bazel project and the only time any other SDK was used was when there was a race condition and we failed to enter bazel mode.
Removed the setting to assume bazel mode from FlutterSettings as it isn't needed and was somewhat confusing as the setting it changed was a DartSDK setting not a Flutter setting. As a followup we should eliminate the setting from the DartSDK as well.
Updated settings dialog:
image

devoncarew:
> WorkspaceCache used the `Refreshable` api which was overkill given that loading the workspace file can be performed on the UI thread and the workspace needs to be accessed reliably from the UI thread.
> 
> The documentation for the .getNow() getter on WorkspaceCache claimed to always return a valid bazel Workspace but that doesn't match the implementation using `Refreshable`. Changing the implementation of `Refreshable` to block on startup was too complex and risky given that complexity isn't needed.

As a drive by comment, the DeviceDaemon service also uses `Refreshable`. We could simplify that code by removing the use of refreshable.

src/io/flutter/bazel/WorkspaceCache.java
```diff
@@ -64,46 +82,82 @@ public static WorkspaceCache getInstance(@NotNull final Project project) {
...
+    final Workspace workspace = Workspace.loadUncached(project);
+    if (workspace == cache) return;
+    cache = workspace;
+
+    // If the current workspace is a bazel workspace, update the Dart plugin
+    // registry key to indicate that there are dart projects without pubspec
+    // registry keys. TODO(jacobr): it would be nice if the Dart plugin was
+    // instead smarter about handling Bazel projects.
+    if (cache != null) {
+      if (!Registry.is(dartProjectsWithoutPubspecRegistryKey, false)) {
```
devoncarew: Does this indicate to the Dart plugin that all projects are bazel projects? Or that the current one might be?

We should probably push some version of bazel `Workspace` knowledge down to the Dart plugin; it'll then have unify behavior with the Flutter plugin and the analysis server. cc @jwren

src/io/flutter/run/bazelTest/FlutterBazelTestLineMarkerContributor.java
```diff
@@ -23,7 +23,6 @@ public FlutterBazelTestLineMarkerContributor() {
   @Nullable
   @Override
   public Info getInfo(@NotNull PsiElement element) {
```
devoncarew: I haven't looked at the parent impl, but is this method override necessary now?

jacob314: @jwren I believe it is only supposed to indicate that the current one might be given that I can open a non-bazel project after setting this option and everything still seems fine.
@jwren  any background on the intended behavior of this option in the Dart plugin?

jwren: I don't have context with `FlutterBazelTestLineMarkerContributor`, did I add this, or someone else, @daveshuckerow?

src/io/flutter/bazel/WorkspaceCache.java
jwren: Changing this registry changes it for all projects. This registry disables features such as notifications around pubspec files themselves and disabling notifications around new available Dart SDK versions.

From my perspective having Flutter set this registry, is unifying the behavior between the Dart and Flutter plugin.  We considered the idea of moving a bunch of the `Workspace` data down to the Dart plugin but  Jacob made the point that the in the Flutter plugin, `Workspace` manages Flutter specific bazel data, mainly changes in executing flutter applications, i.e. the json file knowledge

devoncarew: I wasn't suggesting moving the specific `Workspace` or `WorkspaceCache` class down to the Dart plugin, but instead removing the flag, and having a similar class in the dart plugin which detects when its in a bazel workspace, and configures the UI to operate slightly differently in those situations.

jacob314:
Great! Disabling notifications around available Dart SDK versions and pubspec files sounds like an ok compromise for bazel users. Bazel users who occasionally edit a non-bazel workspace won't be negatively impacted. Typical users will never open a bazel workspace so will never be impacted at all.

src/io/flutter/run/bazelTest/FlutterBazelTestLineMarkerContributor.java
jacob314: Dave did this. My take on this code is it would be nice if we short circuited earlier and didn't check whether the project was bazel every time in this traversal but that this code isn't too much of a performance problem as written given the caching.

PR #4320, opened 2020-01-22
Refactor build script

stevemessick:
@devoncarew This is ready for review. It isn't finished, but a number of technical issues have been solved.

### DIFF
--- a/tool/kokoro/setup.sh
+++ b/tool/kokoro/setup.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# Initialize everything required by the plugin tool.
+setup() {
+  # Fail on any error.
+  set -e
+
+  # Display commands being run. Only do this while debugging and be careful
+  # that no confidential information is displayed.
+  # set -x
+
+  git clone --depth 1 https://github.com/flutter/flutter.git ../flutter
+  export PATH="$PATH":`pwd`/../flutter/bin:`pwd`/../flutter/bin/cache/dart-sdk/bin
+  flutter config --no-analytics
+  flutter doctor
+  export FLUTTER_SDK=`pwd`/../flutter
+
+  java -version
+  echo "JAVA_HOME=$JAVA_HOME"
+
+  echo "install ant"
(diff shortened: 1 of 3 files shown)


stevemessick:
TBR to see if I can trigger the webhook. I'd still like a review, please.

tool/kokoro/test.sh
```diff
@@ -1,36 +1,15 @@
...
+source ./setup.sh
+setup
 
 echo "kokoro test start"
 ./bin/plugin test
 
 echo "kokoro test finished"
 
-
+# This is temporary, duplicating build.sh
```
devoncarew: Add a TODO: here is this is temporary?

tool/kokoro/setup.sh
```diff
@@ -0,0 +1,29 @@
...
+# Initialize everything required by the plugin tool.
+setup() {
+  # Fail on any error.
+  set -e
+
+  # Display commands being run. Only do this while debugging and be careful
+  # that no confidential information is displayed.
+  # set -x
+
+  git clone --depth 1 https://github.com/flutter/flutter.git ../flutter
```
devoncarew: This shallow clone is faster, but this checkout of flutter won't have accurate access to things like its version info (which is stored in git metadata). I assume this won't be a problem for our use of Flutter here, but FYI if we run into issues down the road.

tool/kokoro/setup.sh
```diff
@@ -0,0 +1,29 @@
...
+  export PATH="$PATH":`pwd`/../flutter/bin:`pwd`/../flutter/bin/cache/dart-sdk/bin
+  flutter config --no-analytics
+  flutter doctor
+  export FLUTTER_SDK=`pwd`/../flutter
+
+  java -version
+  echo "JAVA_HOME=$JAVA_HOME"
+
+  echo "install ant"
+  curl https://www-us.apache.org/dist//ant/binaries/apache-ant-1.10.7-bin.tar.gz > ../ant.tar.gz
```
devoncarew: `www-us.apache.org/dist//ant/binaries...` ==> `www-us.apache.org/dist/ant/binaries` (remove a double `//`)

stevemessick:
Thanks!

PR #4324, opened 2020-01-23
update the dart plugin to IntelliJ 2019.2

devoncarew:
- update the dart plugin to IntelliJ 2019.2

2019.1 is our oldest supported IntelliJ platform, but we expect to move off it soon, so instead updating to 2019.2.


devoncarew:
cc @alexander-doroshko

PR #4327, opened 2020-01-23
simplify the logic we use to prevent reload on save

devoncarew:
simplify the logic we use to prevent reload on save:

- remove the logic that would block a reload on save if there are analysis errors in the project; this was overly conservative - errors in tests shouldn't block reload, as well as errors in code that's not part of the app's execution path
- revert back to some older logic which does block reload on save if there are analysis errors in the file currently being edited (and that file is a Dart file)

cc @InMatrix for fyi about the UX change (to remove blocking reload on save on analysis errors - which had been discussed - but to continue to block on a much stricter sub-set; analysis errors in the current file)

### DIFF
--- a/src/io/flutter/run/FlutterReloadManager.java
+++ b/src/io/flutter/run/FlutterReloadManager.java
@@ -26,44 +26,38 @@ import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.fileEditor.FileDocumentManager;
-import com.intellij.openapi.module.Module;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.popup.Balloon;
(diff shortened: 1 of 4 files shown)


devoncarew:
The latest settings page:

Screen Shot 2020-01-23 at 11 09 05 AM

devoncarew:
cc @DanTup - I'm not sure what logic we use in VS Code to prevent reload-on-save (if we do at all)

DanTup:
This sounds similar - in VS Code we'll currently only hot-reload-on-save if:

- File saved is in the current workspace
- File saved has a dart, htm, html, css extension (I think the web ones were added so hot-restart-on-save worked better for web)
- File saved has no errors

So we won't skip it if there are errors in files other than the one you saved. This prevents more false positives, though might mean trying to reload invalid code if you're editing multiple files.

If you save multiple files in one go (eg. Save All), only one of them has to pass the above checks to trigger the reload.

PR #4328, opened 2020-01-23
Define dev channel artifacts

stevemessick:
Put dev channel artifacts in releases/release_dev and identify that to Kokoro. Also, fix .gitignore to stop trying to add lib/dart-plugin/deps/commons-lang3-3.2.1.jar.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ flutter-gui-tests/.gradle/
 flutter-gui-tests/guitest.log
 flutter-intellij.jar
 flutter-intellij.zip
-lib/
+lib/*.jar
 material-design-icons/
 out/
 packages
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ flutter-gui-tests/.gradle/
 flutter-gui-tests/guitest.log
 flutter-intellij.jar
 flutter-intellij.zip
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -12,7 +12,7 @@ flutter-gui-tests/.gradle/
 flutter-gui-tests/guitest.log
 flutter-intellij.jar
 flutter-intellij.zip
-lib/*.jar
+lib/
```
devoncarew: Does `lib/dart-plugin/deps/commons-lang3-3.2.1.jar` exist for you locally? It doesn't for me. Perhaps just delete it?

I make this exclusion more specific in a recent PR in order to be able to easily check in new dart plugin versions (`lib/dart-plugin/192.7761`).

stevemessick: `lib/*.jar` doesn't match anything for me. Actually, why do we want to exclude `lib` anyway? Seems like it should be included.

devoncarew: I have ~a dozen jar files in lib - I think I used it as my maven location.

If other people aren't (if this issue is specific to me), I can change where my maven jars get provisioned to.

stevemessick:
Kokoro is still under development; that failure does not matter here.

PR #4360, opened 2020-02-05
Change the plugin tool to properly build the dev channel

stevemessick:
The initial dev channel support only built the specs that identified as dev channel specs. This makes a number of changes in how things are built:

- A dev channel build produces all versions
- A master branch build produces all versions, plus the version for integration tests
- A release build produces only the versions intended to be published (no change)

The dev channel version is now derived from the previous release and the number of weeks since its branch was committed. This is how it looks in the plugin management page:

Screen Shot 2020-02-05 at 10 46 46 AM

(There may be unresolved issues with point releases. I couldn't test that.)

I did a little refactoring, too. Initialization of globals was really confusing, and now that it has to be done async it needed to be cleaned up.

@devoncarew

### DIFF
--- a/tool/plugin/lib/plugin.dart
+++ b/tool/plugin/lib/plugin.dart
@@ -982,16 +982,7 @@ class DeployCommand extends ProductCommand {
       return 1;
     }
 
-    var token;
-    if (isTesting) {
-      stdout.writeln('Please enter the permanent token '
-          'for the JetBrains Marketplace');
-      stdout.write('Token: ');
-      token = stdin.readLineSync();
-    } else {
-      token = readTokenFile();
-    }
-
+    var token = readTokenFile();
     var value = 0;
     var originalDir = Directory.current;
     for (var spec in specs) {
@@ -999,15 +990,19 @@ class DeployCommand extends ProductCommand {
       var filePath = releasesFilePath(spec);
       log("uploading $filePath");
       var file = File(filePath);
-      Directory.current = file.parent;
+      changeDirectory(file.parent);
       var pluginNumber = plugins[spec.pluginId];
       value = await upload(p.basename(file.path), pluginNumber, token, spec.channel);
       if (value != 0) {
         return value;
       }
     }
-    Directory.current = originalDir;
+    changeDirectory(originalDir);
     return value;
+  }
+
+  void changeDirectory(Directory dir) {
+    Directory.current = dir.path;
   }
 
   String readTokenFile() {
--- a/tool/plugin/lib/plugin.dart
+++ b/tool/plugin/lib/plugin.dart
@@ -228,7 +228,7 @@ Future<bool> performReleaseChecks(ProductCommand cmd) async {
       return true;
     }
     if (cmd.isDevChannel) {
-      log('release node is incompatible with the dev channel');
+      log('release mode is incompatible with the dev channel');
       return false;
     }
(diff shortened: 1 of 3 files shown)


stevemessick:
Looks like this needs more git magic.

devoncarew:
taking a look at this now

tool/plugin/lib/plugin.dart
```diff
@@ -241,7 +253,7 @@ Future<bool> performReleaseChecks(ProductCommand cmd) async {
           log('the .travis.yml file needs updating: plugin generate');
         }
       } else {
-        log('the current git branch must be named "release_${cmd.releaseMajor}"');
+        log('the current git branch must be named "$expectedName"');
```
devoncarew: Do we need to fail the build here?

tool/plugin/lib/plugin.dart
```diff
@@ -271,6 +283,15 @@ void separator(String name) {
   log('${ansi.red}${ansi.bold}$name${ansi.none}', indent: false);
 }
 
+int get devBuildNumber {
```
devoncarew: Do we have the ability to publish a rev to a weekly dev build? What if we find a P0 in the dev channel and what to patch it in order to help validate a fix? a dev channel `.1`

tool/plugin/lib/plugin.dart
```diff
@@ -316,6 +348,12 @@ Future<int> zip(String directory, String outFile) async {
   return await exec('zip', args, cwd: p.dirname(directory));
 }
 
+String _nextRelease() {
+  var num = RegExp(r'release_(\d+)').matchAsPrefix(lastReleaseName).group(1);
```
devoncarew: `num` is a type; I assume this is valid Dart (as it analyzes correctly), but perhaps name the variable something else?

tool/plugin/lib/plugin.dart
```diff
@@ -632,7 +672,7 @@ class BuildCommand extends ProductCommand {
         processedFile.writeAsStringSync(source);
       }
 
-      if (spec.channel == 'stable') {
+      if (!spec.version.startsWith('4.')) {
```
devoncarew: Why check against a specific number? If we need to check against `4.` perhaps document in-line for other readers.

tool/plugin/lib/plugin.dart
stevemessick: It does fail. The function returns false, which causes failure.

tool/plugin/lib/plugin.dart
stevemessick: We don't have that currently. We can add it if needed. We'd probably have to add a --patch option or similar, and it should work for stable, master, and dev builds.

tool/plugin/lib/plugin.dart
stevemessick: Good point. Fixed.

tool/plugin/lib/plugin.dart
stevemessick: That's standard practice; checking stable was a mistake. The edits need to be made regardless of which channel we are building. 

There is a TODO upstream. Someday we should refactor the editing code into a separate object.

stevemessick:
Thanks for the review! I captured some work items in https://github.com/flutter/flutter-intellij/issues/4367.

Issue #4369, opened 2020-02-10
Flutter plugin caused IDE freeze


markathomas:
## What happened

(please describe what you were doing when this exception occurred)

## Version information

IntelliJ IDEA `2020.1 EAP` • Flutter plugin `io.flutter 43.0.4` • Dart plugin `201.4865.19`

No Flutter sdk configured.

## Exception

Freeze in EDT for 50 seconds
IDE KILLED! Sampled time: 42300ms, sampling rate: 100ms, GC time: 0ms (0%), Class loading: 0%, load average: 5.16

```
com.intellij.diagnostic.Freeze
	at java.base@11.0.6/jdk.internal.misc.Unsafe.park(Native Method)
	at java.base@11.0.6/java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
	at java.base@11.0.6/java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1796)
	at java.base@11.0.6/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3128)
	at java.base@11.0.6/java.util.concurrent.CompletableFuture.waitingGet(CompletableFuture.java:1823)
	at java.base@11.0.6/java.util.concurrent.CompletableFuture.join(CompletableFuture.java:2043)
	at io.flutter.vmService.DisplayRefreshRateManager.getDisplayRefreshRate(DisplayRefreshRateManager.java:87)
	at io.flutter.vmService.DisplayRefreshRateManager.lambda$queryRefreshRate$0(DisplayRefreshRateManager.java:41)
	at io.flutter.vmService.DisplayRefreshRateManager$$Lambda$6421/0x00000001035f3040.run(Unknown Source)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:192)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.doRun(LaterInvocator.java:467)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.runNextEvent(LaterInvocator.java:443)
	at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:420)
	at java.desktop@11.0.6/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:313)
	at java.desktop@11.0.6/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:776)
[...]

PR #4373, opened 2020-02-11
Fix crash caused by getting flutter view id with .join()

kenzieschmoll:
I was unable to reproduce the original issue, but this CL is intended to fix https://github.com/flutter/flutter-intellij/issues/4369

@devoncarew


src/io/flutter/vmService/DisplayRefreshRateManager.java
```diff
@@ -80,18 +80,18 @@ public Double getCurrentDisplayRefreshRateRaw() {
...
-      return null;
-    }).join();
-
-    if (flutterViewId == null) {
+    try {
+      final String flutterViewId = vmServiceManager.getFlutterViewId().exceptionally(exception -> {
+        // We often see "java.lang.RuntimeException: Method not found" from here; perhaps a race condition?
+        LOG.warn(exception.getMessage());
+        return null;
+      }).get();
```
devoncarew: It looks like `get()` may also block the thread until the value is ready?

I believe the underlying issue here is that we're using clocking calls and this code path can be called from the event thread. I suspect what we need to do is to move to a non-blocking call (which may be something like `.whenComplete()`?), and then use that to determine the display refresh rate, and return the value as a future.

(Java futures definitely have a much larger API surface area than the Dart variant)

kenzieschmoll: The issue with `whenComplete` is that it returns a `CompletableFuture<T> `. In this case T is a String because that is the value we expect from `getFlutterViewId`. However, we need to return CompletableFuture<double> for displayRefreshRate.

devoncarew: You may try creating a `CompletableFuture<Double> foo` at the start of the method, starting the future work to get the flutterViewId, and return the `foo` completable future. When you have the flutter view ID, you can then request the display rate. Once you have a value for that you should be able to call `foo.complete(displayRefreshRate)`.

kenzieschmoll: I think I may have found a way around this by returning outside of whenComplete. PTAL.

src/io/flutter/vmService/DisplayRefreshRateManager.java
```diff
@@ -78,20 +78,22 @@ public Double getCurrentDisplayRefreshRateRaw() {
...
-    }
-
-    return invokeGetDisplayRefreshRate(flutterViewId);
+    }).whenComplete((String id, Throwable throwable) -> {
+      if (id == null) {
+        // Fail gracefully by returning the default.
+        displayRefreshRate.complete(defaultRefreshRate);
+      } else {
+        invokeGetDisplayRefreshRate(id).whenComplete((Double refreshRate, Throwable t) -> {
+          displayRefreshRate.complete(refreshRate);
```
devoncarew: If `t` is not null here, should we instead complete `displayRefreshRate` with the `defaultRefreshRate`?

kenzieschmoll: Added exception handling via `exceptionally()`

devoncarew: I think you either want to use `thenApplyAsync()` and `exceptionally()`, or use `whenComplete()`. It looks like whenComplete gives you both the value, if it completes nornally or the exception if it fails.

kenzieschmoll: done.

src/io/flutter/vmService/DisplayRefreshRateManager.java
```diff
@@ -78,20 +78,25 @@ public Double getCurrentDisplayRefreshRateRaw() {
...
+        LOG.warn(throwable.getMessage());
+        // Fail gracefully by returning the default.
+        displayRefreshRate.complete(defaultRefreshRate);
+      } else {
+        invokeGetDisplayRefreshRate(id).whenComplete((Double refreshRate, Throwable t) -> {
+          if (t != null) {
+            LOG.warn(t.getMessage());
+            // Fail gracefully by returning the default.
+            displayRefreshRate.complete(defaultRefreshRate);
+          }
```
devoncarew: This needs an else statement, otherwise when `t != null` we'll call `displayRefreshRate.complete()` twice.

kenzieschmoll: whoops. thanks for catching. fixed.

PR #4379, opened 2020-02-14
Make deploy script executable

stevemessick:
Also update some comments.


stevemessick:
TBR no code changes @devoncarew

PR #4385, opened 2020-02-18
Switch common test config utils to use a per-file cache

DaveShuckerow:


src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -84,21 +84,22 @@ private void clearCachedInfo() {
...
       outlineService.addListener(getListenerForFile(file));
       return new HashMap<>();
     }
 
     synchronized (this) {
       // Visit the fields on the outline to get which calls are actual named tests.
-      if (cachedCallToTestType == null) {
-        cachedCallToTestType = new HashMap<>();
-        populateTestTypeMap(outline, cachedCallToTestType, file);
+      if (!cachedCallToTestType.containsKey(path)) {
```
jacob314: what if the outline has changed? we need to clear out the cache in that case.

src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -64,11 +64,11 @@ public TestType asTestCall(@NotNull PsiElement element) {
     return null;
   }
 
-  Map<DartCallExpression, TestType> cachedCallToTestType;
+  private final Map<String, Map<DartCallExpression, TestType>> cachedCallToTestType = new HashMap<>();
```
jacob314: caching DartCallExpressions will cause us to hang on to old IntelliJ ASTs and leave us vulnerable to if IntelliJ recomputes AST nodes. It is safer to use file offsets as keys.

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: The outline given will be null and the cache will be cleared.

CF: the comments above and the documentation for outlineService.getIfUpdated:

https://github.com/flutter/flutter-intellij/blob/f69eec4dfa31dcd3964fa79d925474376d24e2ad/src/io/flutter/editor/ActiveEditorsOutlineService.java#L187-L194

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: Done.

src/io/flutter/run/common/CommonTestConfigUtils.java
jacob314: that is not a safe assumption.
The only assumption safe to depend on is the outline will be null if it is out of date.
For example, imagine if another service such as the the widget indent guides or HotUI is also requesting outlines using ActiveEditorsOutlineService, you could easily get a non-null outline even though it is different than the outline your cache is based on.

src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -160,7 +161,7 @@ protected ActiveEditorsOutlineService getActiveEditorsOutlineService(@NotNull Pr
    * Traverses the {@param outline} tree and adds to {@param callToTestType } the {@link DartCallExpression}s that are tests or test groups.
    */
   private void populateTestTypeMap(@NotNull FlutterOutline outline,
-                                   @NotNull Map<DartCallExpression, TestType> callToTestType,
+                                   @NotNull Map<Integer, TestType> callToTestType,
                                    @NotNull PsiFile file) {
     final PsiElement element = file.findElementAt(outline.getOffset());
```
jacob314: optimization: move this findElementAt call so you only call it when unit_test_group or unit_test_test is found. Currently we are doing this fairly expensive operation a lot more times than we need to.
I do like that you are still using
file.findElementAt to avoid location calculation bugs.

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: This code checks if the outline is outdated and returns null if so.

If you have suggestions for how to improve the heuristic, they are welcome.

jacob314: the getIfUpdated method is fine as written. What you need to do is not depend on getting a null outline for the case that there was an update. To do that you should store what the outline was in CommonTestConfigUtils so that you can tell if you have gotten a new outline so need to update your cache.

DaveShuckerow:
> It would be good to add a unittest as this logic has some pretty subtle edge cases.
> lets discuss how to avoid leading memory and write the code a bit more defensively by avoiding assumptions about lifetimes of objects like DartCallExpression.

Done.

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: Done.

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: Done.

DaveShuckerow:
PTAL.

src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -64,49 +64,57 @@ public TestType asTestCall(@NotNull PsiElement element) {
     return null;
   }
 
-  Map<DartCallExpression, TestType> cachedCallToTestType;
+  private final Map<String, Map<Integer, TestType>> cachedCallToTestType = new HashMap<>();
+  private final Map<String, FlutterOutline> pathToPreviouslyUsedOutline = new HashMap<>();
```
jacob314: nit: rather than two maps with the same key, define a class  with both the 
FltuterOutline and the Map<Integer, TestType> as fields.

testSrc/unit/io/flutter/testing/FakeActiveEditorsOutlineService.java
```diff
@@ -38,16 +40,20 @@ public void loadOutline(@NotNull String flutterOutlinePath) {
...
+    pathToFlutterOutline.put(filePath, flutterOutline);
+
   }
 
   @Nullable
   @Override
   public FlutterOutline getOutline(String path) {
-    return flutterOutline;
+    // The path string that we get will be prepended with a '/' character, compared to how the cache was initialized.
+    final FlutterOutline result = pathToFlutterOutline.get(path);
```
jacob314: nit: remove the `result` local variable.

DaveShuckerow: Done.

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: Done.

src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -89,7 +87,7 @@ private void clearCachedInfo(String path) {
     final String path = file.getVirtualFile().getPath();
     final boolean outlineOutdated;
     synchronized (this) {
-      outlineOutdated = pathToPreviouslyUsedOutline.containsKey(path) && outline != pathToPreviouslyUsedOutline.get(path);
+      outlineOutdated = cache.containsKey(path) && outline != cache.get(path).outline;
```
jacob314: nit: the cleaner pattern for this in java is
final entry = cache.get(path);

src/io/flutter/run/common/CommonTestConfigUtils.java
```diff
@@ -99,14 +97,8 @@ private void clearCachedInfo(String path) {
...
     synchronized (this) {
-      pathToPreviouslyUsedOutline.put(path, outline);
-      // Visit the fields on the outline to get which calls are actual named tests.
-      if (!cachedCallToTestType.containsKey(path)) {
-        cachedCallToTestType.put(path, new HashMap<>());
-        populateTestTypeMap(outline, cachedCallToTestType.get(path), file);
-      }
-
-      return cachedCallToTestType.get(path);
+      cache.put(path, new OutlineCache(outline, file));
```
jacob314: nit: cache the OutlineCache entry as a local variable so you don't have to read it out of a cache.

DaveShuckerow: Done.

src/io/flutter/run/common/CommonTestConfigUtils.java
DaveShuckerow: Done.

PR #4386, opened 2020-02-18
Update ant URL

stevemessick:
Apache changed the location. Our Kokoro runs are failing but this should fix it.


stevemessick:
TBR @devoncarew The presubmit succeeds with this change.

PR #4409, opened 2020-02-27
Ensure indexing is finished before updating library model

stevemessick:
Remove usage of @internal API, and ensure write access and smart mode before updating the library model.


stevemessick:
@devoncarew TBR

Issue #4416, opened 2020-03-03
Incompatibility problems


stevemessick:
See the Jetbrains plugin page for a formatted version of this report.

```
14 compatibility problems
Flutter 44.0.3 is binary incompatible with IntelliJ IDEA Ultimate IU-193.6494.35 (2019.3.3) due to the following problems
Abstract method is not implemented (6 problems)
Abstract method ModuleGalleryEntry.createStep(NewModuleModel) is not implemented (6 problems)
Concrete class FlutterDescriptionProvider.ImportFlutterModuleGalleryEntry inherits from ModuleGalleryEntry but doesn't implement the abstract method createStep(NewModuleModel). This can lead to AbstractMethodError exception at runtime.
Concrete class FlutterDescriptionProvider.FlutterPluginGalleryEntry inherits from ModuleGalleryEntry but doesn't implement the abstract method createStep(NewModuleModel). This can lead to AbstractMethodError exception at runtime.
Concrete class FlutterDescriptionProvider.AddToAppModuleGalleryEntry inherits from ModuleGalleryEntry but doesn't implement the abstract method createStep(NewModuleModel). This can lead to AbstractMethodError exception at runtime.
Concrete class FlutterDescriptionProvider.FlutterApplicationGalleryEntry inherits from ModuleGalleryEntry but doesn't implement the abstract method createStep(NewModuleModel). This can lead to AbstractMethodError exception at runtime.
Concrete class FlutterDescriptionProvider.FlutterPackageGalleryEntry inherits from ModuleGalleryEntry but doesn't implement the abstract method createStep(NewModuleModel). This can lead to AbstractMethodError exception at runtime.
Concrete class FlutterDescriptionProvider.FlutterModuleGalleryEntry inherits from ModuleGalleryEntry but doesn't implement the abstract method createStep(NewModuleModel). This can lead to AbstractMethodError exception at runtime.
Method not found (6 problems)
Invocation of unresolved method ValidatorPanel.registerValidator(...) (2 problems)
Constructor ImportFlutterModuleStep.<init>(...) contains an invokevirtual instruction referencing an unresolved method ValidatorPanel.registerValidator(...). This can lead to NoSuchMethodError exception at runtime.
The method might have been declared in the super classes or in the super interfaces:
JComponent
JPanel
Disposable
Accessible
TransferHandler.HasGetTransferHandler
[...]

PR #4417, opened 2020-03-03
Stop using the Android Studio new module wizard for add-to-app

stevemessick:
We were adding Flutter modules to the New Module Wizard of Android Studio when users wanted to use add-to-app in an Android app. We can't do that (at least for now) so this adds a "New Flutter Module" parallel to the "New Flutter Project" on the New menu.

This is derived from the New Project Wizard from 3/8/19. The Kotlin conversion had not started at that time.

See https://github.com/flutter/flutter-intellij/issues/4416 and https://youtrack.jetbrains.com/issue/IDEA-234445

The current version of the NPW is mostly Kotlin and we can't re-use it, due to overuse of the `final` keyword (or lack of `open` in Kotlin). And we can't compile Kotlin in our build process, so I couldn't copy the current version.

After coverage analysis proved nearly all of it was unused I remove about 95% of the NPW classes. Those remaining are in the package with suffix `npwOld` and can be ignored.


stevemessick:
This builds for AS 3.5 and 3.6 but not 4.0.

I added a rename command to the plugin tool.

stevemessick:
Closing, obsolete.

PR #4423, opened 2020-03-10
Fix API incompatibilities

stevemessick:
This replaces #4417.

This rebases the sources on 3.6 (stable) Android Studio. We have to do some editing in the plugin tool to build for 4.0 now.

### DIFF
--- a/flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
+++ b/flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
@@ -9,7 +9,7 @@ import com.android.tools.idea.projectsystem.AndroidModuleSystem;
 import com.android.tools.idea.projectsystem.AndroidProjectSystem;
 import com.android.tools.idea.projectsystem.LightResourceClassService;
 import com.android.tools.idea.projectsystem.ProjectSystemSyncManager;
-import com.android.tools.idea.projectsystem.SourceProvidersFactory;
+//import com.android.tools.idea.projectsystem.SourceProvidersFactory;
 import com.android.tools.idea.projectsystem.gradle.GradleProjectSystem;
 import com.android.tools.idea.projectsystem.gradle.GradleProjectSystemProvider;
 import com.intellij.openapi.diagnostic.Logger;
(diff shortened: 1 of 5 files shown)


stevemessick:
The presubmit failure has nothing to do with this PR and can be ignored, I think.

flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
```diff
@@ -102,27 +102,19 @@ public ProjectSystemSyncManager getSyncManager() {
...
 
-  @NotNull
-  @SuppressWarnings("override")
-  public Collection<Module> getSubmodules() {
-    return gradleProjectSystem.getSubmodules();
-  }
+  //@NotNull
+  //public Collection<Module> getSubmodules() {
+  //  return gradleProjectSystem.getSubmodules();
+  //}
```
kenzieschmoll: does this need "// for 4.0"

tool/plugin/lib/plugin.dart
```diff
@@ -715,26 +715,60 @@ class BuildCommand extends ProductCommand {
...
-          'FlutterProjectModel model(@NotNull Project project,',
-          'FlutterProjectModel model(@NotNull NewModuleModel project,',
+          '//  return gradleProjectSystem.getSourceProvidersFactory();',
+          '  return gradleProjectSystem.getSourceProvidersFactory();',
         );
         source = source.replaceAll(
-          'mySharedModel.getValue().project().setValue(project);',
-          'mySharedModel.getValue().project().setValue(project.getProject().getValue());',
+          '//} // for 4.0',
+          '} // for 4.0',
```
kenzieschmoll: do we need to uncomment the @NotNull  annotations too?

flutter-studio/src/io/flutter/project/FlutterProjectSystem.java
stevemessick: It is not used. I didn't delete it because I think it will be needed in a later version.

tool/plugin/lib/plugin.dart
stevemessick: Those are only used by IntelliJ, and the editing is only temporary during a build, so I don't think there's a need.

stevemessick:
Thanks for the review!

PR #4432, opened 2020-03-21
Revive the GUI test NewModuleTest

stevemessick:
This fixes the GUI test framework and updates NewModuleTest. Had to rebuild the sample projects, too.

Classes in com.android...fixture are "test fixtures". They provide an API that allows programmatic control of a UI element. The actual tests are in io.flutter.

Changes to integration tests are still a work in progress.


.gitignore
```diff
@@ -26,3 +26,7 @@ releases/
 .idea/modules/
 .idea/libraries/Gradle_*.xml
 .idea/modules.xml
+flutter-studio/testSrc/config
```
devoncarew: Just to confirm, we don't want this source checked in?

stevemessick: TBH I have not looked at it. It is a cache created by the GUI test framework. I just assumed that others would prefer to not have it taking up space.

Thanks for the review!

PR #4435, opened 2020-03-25
Fix link to Flutter docs

stevemessick:
The Flutter docs moved. Update the link for add-to-app.


stevemessick:
I was wondering if we had enough to justify a release this month. Now we do!

Thanks for the review.

Issue #4429, opened 2020-03-18
hyperlinking issues with the run console for tests


devoncarew:
When running flutter tests, we don't properly link to source lines in the console in a few situations:

<img width="944" alt="Screen Shot 2020-03-18 at 8 35 07 AM" src="https://user-images.githubusercontent.com/1269969/76978100-7fe0c500-68f3-11ea-9535-658762e1ff3b.png">

For above:
- the first link is I believe provided by the core platform; when clicking on, it displays an error message for file not found; this might be due to the trailing line numbers
- none of the package: links are detected, including the self-references for the package containing the tests

Additionally, on a mac, I've seen errors for links that tried to open dart file; they report that the disk format was not understood (I believe these links are provided by the platform, try to open a dart file using the OS, and give an error as the dart file type is an older disk format on the mac).

Issue #3851, opened 2019-09-05
When running framework tests, stack traces are not linked to source code.


gspencergoog:
When I run a test in the flutter framework, I should be able to click on the error locations in a stack trace and open it in the editor.

For instance, here's a screenshot of what I see when I run a test.  I'm expecting all of the `package:` bits in parens to be links, but they're not.
![image](https://user-images.githubusercontent.com/8867023/64369731-d2026e80-cfd1-11e9-8274-273676d6ca65.png)

```
[✓] Flutter (Channel route_focus2, v1.9.8-pre.99, on Linux, locale en_US.UTF-8)
    • Flutter version 1.9.8-pre.99 at /usr/local/google/home/gspencer/code/flutter
    • Framework revision 108035623b (18 hours ago), 2019-09-04 17:33:52 -0700
    • Engine revision e7f9ef6aa0
    • Dart version 2.5.0 (build 2.5.0-dev.4.0 36985859e4)

[✓] IntelliJ IDEA Community Edition (version 2019.2)
    • IntelliJ at /opt/my-intellij-ce-current
    • Flutter plugin version 39.0.5
    • Dart plugin version 192.6459
```

gspencergoog:
FYI, I have "Show structured errors for Flutter framework issues" checked in the preferences, and I have the experiment "Replace the Run and Debug console output with a custom Flutter Logging view" unchecked.

kenzieschmoll:
@devoncarew @jacob314

gspencergoog:
I realized that maybe I was remembering incorrectly: I'm probably remembering the `file:` links, but in any case it would be super useful to be able to link to `package:` links, if that's possible.

jacob314:
we should support linking to package: urls in all our tools.

gspencergoog:
@jacob314 was that "should" meaning that it should be a feature request, or that it should already work?

jacob314:
updated my comment to be clearer. we don't support this yet.

PR #4443, opened 2020-03-28
fix an issue with discovering hyperlinks in test consoles

devoncarew:
Fix an issue with discovering hyperlinks in test consoles:

- add a `DartConsoleFilter` console filter to the test console
- some related cleanup and normalization to our filters for various consoles
- fix https://github.com/flutter/flutter-intellij/issues/4429
- fix https://github.com/flutter/flutter-intellij/issues/3851

We'd never contributed a `DartConsoleFilter` to the test consoles before (which we should have), and recent changes in the `FlutterConsoleFilter` had caused the test console hyperlink support to regress further.

@gspencergoog, this will likely address the issue you're seeing. We have a release scheduled for next Friday, so you'll see this change then. If you want to try out a version before that, we now have a beta channel that gets automatically pushed to every Sunday AM. I can send you docs on how to switch to the beta channel if you like.

### DIFF
--- a/src/io/flutter/run/bazelTest/BazelTestLaunchState.java
+++ b/src/io/flutter/run/bazelTest/BazelTestLaunchState.java
@@ -17,7 +17,7 @@ import com.intellij.execution.testframework.ui.BaseTestsOutputConsoleView;
 import com.intellij.execution.ui.ConsoleView;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
-import com.jetbrains.lang.dart.ide.runner.DartRelativePathsConsoleFilter;
+import com.jetbrains.lang.dart.ide.runner.DartConsoleFilter;
 import com.jetbrains.lang.dart.util.DartUrlResolver;
 import io.flutter.bazel.Workspace;
 import io.flutter.bazel.WorkspaceCache;
@@ -47,8 +47,8 @@ public class BazelTestLaunchState extends CommandLineState {
     this.config = config;
     this.fields = config.getFields();
     if (testFile == null) {
-      Workspace workspace = WorkspaceCache.getInstance(env.getProject()).get();
-      assert(workspace != null);
+      final Workspace workspace = WorkspaceCache.getInstance(env.getProject()).get();
+      assert (workspace != null);
       testFile = workspace.getRoot();
     }
     this.testFile = testFile;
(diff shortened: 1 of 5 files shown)


src/io/flutter/run/bazelTest/BazelTestLaunchState.java
```diff
@@ -106,8 +106,8 @@ protected ConsoleView createConsole(@NotNull Executor executor) throws Execution
     final ConsoleProps props = ConsoleProps.forBazel(config, executor, resolver);
     final BaseTestsOutputConsoleView console = SMTestRunnerConnectionUtil.createConsole(ConsoleProps.bazelFrameworkName, props);
 
-    final String baseDir = workspace.getRoot().getPath();
-    console.addMessageFilter(new DartRelativePathsConsoleFilter(project, baseDir));
```
stevemessick: DartRelativePathsConsoleFilter is intended to be used in test environs. Are you sure deleting it is the right thing to do?

src/io/flutter/run/daemon/DaemonConsoleView.java
```diff
@@ -46,6 +46,7 @@ protected ConsoleView createConsole() {
     };
 
     // Set up basic console filters. (More may be added later.)
+    // TODO(devoncarew): Do we need this filter? What about DartConsoleFilter (for package: uris)?
```
stevemessick: I think you want DartConsoleFilter here, too, but it may be getting added at a later stage. I forgot just how convoluted this is.

src/io/flutter/run/test/TestLaunchState.java
```diff
@@ -113,15 +113,15 @@ protected ConsoleView createConsole(@NotNull Executor executor) throws Execution
     final DartUrlResolver resolver = DartUrlResolver.getInstance(project, testFileOrDir);
     final ConsoleProps props = ConsoleProps.forPub(config, executor, resolver);
     final BaseTestsOutputConsoleView console = SMTestRunnerConnectionUtil.createConsole(ConsoleProps.pubFrameworkName, props);
+
     final Module module = ModuleUtil.findModuleForFile(testFileOrDir, project);
-    if (module != null) {
-      console.addMessageFilter(new FlutterConsoleFilter(module));
```
stevemessick: Are you sure you want to remove this? It has some specialized logic for handling Flutter stuff, like plugins.

src/io/flutter/run/daemon/DaemonConsoleView.java
devoncarew: We may just need the DartConsoleFilter in the SdkRunConfig.java (the run/debug console)? If you're certain I can add it, but would otherwise hold off.

src/io/flutter/run/test/TestLaunchState.java
devoncarew: I'll figure out what all the FlutterConsoleFilter is responsible for. I _think_ it's more important for the run console (to link to lib.main.dart, and some flutter doctor output?). It may have several different concerns in that class.

devoncarew: I added docs to FlutterConsoleFilter. I _think_ we only want to use it for the run and debug consoles.

Issue #4440, opened 2020-03-27
Rename packages commands to pub commands


mit-mit:
A few quarters (aligned with the launch of pub.dev) we changed the default Flutter packages commands to use 'pub' terminology, not 'packages':

```
$ flutter help pub
Commands for managing Flutter packages.

Usage: flutter pub <subcommand> [arguments]
```

The action button UI on a `pubspec.yaml` file still has 'Packages get' and 'Packages upgrade'. Can we change those to 'pub get' and 'pub upgrade'?

Issue #4439, opened 2020-03-27
Expose `flutter pub outdated` as an action on `pubspec.yaml` files


mit-mit:
Pub has a new `outdated` feature which allows developers to view stale dependencies. We'd like to expose this with a new `pub outdated` feature.

```
$ flutter help pub
Commands for managing Flutter packages.

Usage: flutter pub <subcommand> [arguments]
-h, --help    Print this usage information.

Available subcommands:
...
  outdated    Analyze dependencies to find which ones can be upgraded.
```

PR #4444, opened 2020-03-28
add support for 'flutter pub outdated'

devoncarew:
- add support for 'flutter pub outdated' (expose it in the pubspec file's header action bar); this is contingent on having a recent enough Flutter SDK
- remove `flutter upgrade` from the pubspec file's header action bar
- rename user facing references to `flutter packages` to `flutter pub`
- fix https://github.com/flutter/flutter-intellij/issues/4439
- fix https://github.com/flutter/flutter-intellij/issues/4440

Screen Shot 2020-03-27 at 4 19 04 PM

Screen Shot 2020-03-27 at 4 19 14 PM

### DIFF
--- a/src/io/flutter/ProjectOpenActivity.java
+++ b/src/io/flutter/ProjectOpenActivity.java
@@ -81,15 +81,15 @@ public class ProjectOpenActivity implements StartupActivity, DumbAware {
     @NotNull private final PubRoot myRoot;
 
     public PackagesOutOfDateNotification(@NotNull Project project, @NotNull PubRoot root) {
-      super("Flutter Packages", FlutterIcons.Flutter, "Flutter packages get.",
+      super("Flutter Packages", FlutterIcons.Flutter, "Flutter pub get.",
             null, "The pubspec.yaml file has been modified since " +
-                  "the last time 'flutter packages get' was run.",
+                  "the last time 'flutter pub get' was run.",
             NotificationType.INFORMATION, null);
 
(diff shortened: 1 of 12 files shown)


src/io/flutter/ProjectOpenActivity.java
```diff
@@ -81,15 +81,15 @@ public void runActivity(@NotNull Project project) {
     @NotNull private final PubRoot myRoot;
 
     public PackagesOutOfDateNotification(@NotNull Project project, @NotNull PubRoot root) {
-      super("Flutter Packages", FlutterIcons.Flutter, "Flutter packages get.",
+      super("Flutter Packages", FlutterIcons.Flutter, "Flutter pub get.",
             null, "The pubspec.yaml file has been modified since " +
-                  "the last time 'flutter packages get' was run.",
+                  "the last time 'flutter pub get' was run.",
```
stevemessick: I wonder if we're getting to the point of needing to have all user-visible strings in the bundle.

devoncarew: That's likely a useful change to make.

I wonder what tools JetBrains has for this? Are there lints we can enable? Is there a way to know how many non-localized strings we currently have?

stevemessick:
In Preferences, search for "Hard coded strings". It has some options, but the defaults look good to me.

devoncarew:
cc @alexander-doroshko for the changes we made to the Flutter plugin

Issue #4398, opened 2020-02-26
pull survey dates and url from the hosted json info


devoncarew:
Move to pulling our survey dates and url from https://github.com/flutter/website/blob/master/src/f/flutter-survey-metadata.json (or, the flutter.dev version of that url).

@pq

devoncarew:
For posterity, here's the flutter.dev version of the url: https://flutter.dev/f/flutter-survey-metadata.json .

devoncarew:
@pq, can we get this done by this milestone? Once this is in we won't have to worry about manually changing date, and will be able to get good counts of older clients.

PR #4449, opened 2020-03-30
data drive survey

pq:
Still a little testing to do but ready for preliminary feedback.

/cc @devoncarew @DanTup 

/fyi @jayoung-lee 

Fixes: #4398.

### DIFF
--- a/src/io/flutter/survey/FlutterSurvey.java
+++ b/src/io/flutter/survey/FlutterSurvey.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2020 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.survey;
+
+import com.google.gson.JsonObject;
+import org.jetbrains.annotations.NotNull;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class FlutterSurvey {
+  final String uniqueId;
+  final LocalDate startDate;
+  final LocalDate endDate;
+
+  final String title;
+  final String urlPrefix;
+
+  FlutterSurvey(String uniqueId, String startDate, String endDate, String title, String urlPrefix) {
+    this.uniqueId = uniqueId;
+    this.startDate = parseDate(startDate);
+    this.endDate = parseDate(endDate);
+    this.title = title;
+    this.urlPrefix = urlPrefix;
+  }
+
+  private static LocalDate parseDate(String date) {
+    return LocalDate.from(DateTimeFormatter.ISO_DATE_TIME.parse(date));
+  }
+
+  public static FlutterSurvey fromJson(@NotNull JsonObject json) {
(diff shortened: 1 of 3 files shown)


src/io/flutter/survey/FlutterSurvey.java
```diff
@@ -0,0 +1,47 @@
...
+                             json.getAsJsonPrimitive("startDate").getAsString(),
+                             json.getAsJsonPrimitive("endDate").getAsString(),
+                             json.getAsJsonPrimitive("title").getAsString(),
+                             json.getAsJsonPrimitive("url").getAsString()
+    );
+  }
+
+  boolean isSurveyOpen() {
+    final LocalDate now = LocalDate.now();
+    return now.isAfter(startDate) && now.isBefore(endDate);
```
devoncarew: I assume that the start date is inclusive, and the end date is exclusive? Or are they both inclusive?

This should likely read something like `now.isEqualToOrAfter(startDate)`.

jayoung-lee: I guess inclusive/exclusive won't matter that much, because both start date and end date actually specify the time, in addition to the date. So something consistent would work..?

```
"startDate": "2020-03-03T09:00:00-08:00",
"endDate": "2020-03-10T09:00:00-07:00"
```

Or perhaps I should change the names of the objects to `startDateAndTime` and 'endDateAndTime` to avoid confusion?

devoncarew: Ah, I'd written that thinking that it was for a day, not a date+time; agree that it doesn't matter if we're just talking about a second or millisecond.

pq: >  if we're just talking about a second or millisecond

Exactly my thinking.

> Or perhaps I should change the names of the objects to startDateAndTime and 'endDateAndTime` to avoid confusion?

I don't  think it's necessary but  happy to go with the flow.

pq: Updated to `now.isEqual(startDate) || now.isAfter(startDate)` ...

src/io/flutter/survey/FlutterSurveyNotifications.java
```diff
@@ -94,26 +72,23 @@ private void check(@NotNull VirtualFile file) {
     });
   }
 
-  @NotNull final Project myProject;
-
-  FlutterSurveyNotifications(@NotNull Project project) {
-    this.myProject = project;
-  }
-
   private void checkForDisplaySurvey() {
```
devoncarew, line 75: @pq - it looks like the following code happens synchronously on whatever thread we were called on (potentially the EDT)?

You may change this to working on a background thread (and if necessary, switching back the EDT at the very end).

We use a few different ways to execute in background threads - one is `ApplicationManager.getApplication().executeOnPooledThread()` (not sure if other's may be more appropriate).

pq: Yup. I'm looking at this very thing! Sorry I missed it first time through!

PR #4454, opened 2020-04-01
Update the build for AS canary 4.1

stevemessick:
We had a lot of catching up to do.

@DaveShuckerow @devoncarew


devoncarew:
Looking now (looks like compilation errors for 2020.1?: https://travis-ci.org/github/flutter/flutter-intellij/jobs/669923128?utm_medium=notification&utm_source=github_status)

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+# 45.0
+- Add support for 'flutter pub outdated' (#4444)
+- Improve how we calculate when to summarize flutter errors (#4447)
+- Fix an issue with discovering hyperlinks in test consoles (#4443)
```
devoncarew, line 4: These first three items (and the new supported platforms) would be good fodder for the blog.

src/io/flutter/utils/AndroidLocationProvider.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2020 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.utils;
+
+import com.android.tools.idea.gradle.dsl.model.BuildModelContext;
```
devoncarew, line 8: We may have issues w/ local development, for people developing against something like IntelliJ CE?

tool/plugin/lib/plugin.dart
```diff
@@ -638,121 +638,161 @@ class BuildCommand extends ProductCommand {
...
+          source = processedFile.readAsStringSync();
+          files[processedFile] = source;
+          source = source.replaceAll(
+            'findGradleTarget', // 3 matches
+            'findImportTarget',
+          );
+          processedFile.writeAsStringSync(source);
+        }
+      } catch (ex) {
+        // Restore sources.
```
devoncarew, line 780: Do we want this restore code to happen on an exception, or in a `finally`?

stevemessick: It happens in a finally after the compiler is finished.

src/io/flutter/utils/AndroidLocationProvider.java
stevemessick: Yes, Until we have time to devise a proper solution people can replace AndroidLocationProvider with its stub file content.

stevemessick:
PTAL

Needed to do several things.
- Update test driver
- Ignore AndroidxTests
- Swap base and stub files for new class

See https://github.com/flutter/flutter-intellij/issues/4459 and #3583

PR #4520, opened 2020-04-28
Update setup instructions

helin24:
This updates some of the instruction details, fixes a few capitalizations, and adds a few contextual comments.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,17 +14,14 @@ file.
 
 * Download and install the latest stable version of IntelliJ
   - [IntelliJ Downloads](https://www.jetbrains.com/idea/download/)
-  - either the community edition (free) or Ultimate will work.
+  - Either the community edition (free) or Ultimate will work.
 * Start IntelliJ
 * In the Project Structure dialog (`File | Project Structure`), select "Platform Settings > SDKs" click the "+" sign at the top "Add New SDK (Alt+Insert)" to configure an IntelliJ Platform Plugin SDK
-  - point it to the directory of your downloaded IntelliJ Community Edition installation (e.g, `IntelliJ IDEA CE.app/Contents` or `~/idea-IC-183.4886.37`)
-  - change the name to `IntelliJ IDEA Community Edition`
-  - extend it with additional plugin libraries (only until Android Q is released)
-    - plugins/android/lib/android.jar
-    - plugins/gradle/lib/gradle-common.jar
-* One-time Dart plugin install - first-time a new IDE is installed and run you will need to install the Dart plugin.  `Configure | Plugins` and install the Dart plugin, then restart the IDE
-* Open flutter-intellij project in IntelliJ (select and open the directory of the flutter-intellij repository). Build it using `Build` | `Make Project`
-* Try running the plugin; there is an existing launch config for "Flutter IntelliJ".
+  - Point it to the directory of your downloaded IntelliJ Community Edition installation (e.g, `IntelliJ IDEA CE.app/Contents` or `~/idea-IC-183.4886.37`)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -14,17 +14,14 @@ file.
 
 * Download and install the latest stable version of IntelliJ
   - [IntelliJ Downloads](https://www.jetbrains.com/idea/download/)
-  - either the community edition (free) or Ultimate will work.
+  - Either the community edition (free) or Ultimate will work.
 * Start IntelliJ
 * In the Project Structure dialog (`File | Project Structure`), select "Platform Settings > SDKs" click the "+" sign at the top "Add New SDK (Alt+Insert)" to configure an IntelliJ Platform Plugin SDK
-  - point it to the directory of your downloaded IntelliJ Community Edition installation (e.g, `IntelliJ IDEA CE.app/Contents` or `~/idea-IC-183.4886.37`)
-  - change the name to `IntelliJ IDEA Community Edition`
-  - extend it with additional plugin libraries (only until Android Q is released)
```
helin24, line 22: I think we decided that this isn't needed any more because it was for a particular change that eventually was not merged.

CONTRIBUTING.md
```diff
@@ -14,17 +14,14 @@ file.
...
-    - plugins/android/lib/android.jar
-    - plugins/gradle/lib/gradle-common.jar
-* One-time Dart plugin install - first-time a new IDE is installed and run you will need to install the Dart plugin.  `Configure | Plugins` and install the Dart plugin, then restart the IDE
-* Open flutter-intellij project in IntelliJ (select and open the directory of the flutter-intellij repository). Build it using `Build` | `Make Project`
-* Try running the plugin; there is an existing launch config for "Flutter IntelliJ".
+  - Point it to the directory of your downloaded IntelliJ Community Edition installation (e.g, `IntelliJ IDEA CE.app/Contents` or `~/idea-IC-183.4886.37`)
+  - Change the name to `IntelliJ IDEA Community Edition`
+* One-time Dart plugin install - first-time a new IDE is installed and run you will need to install the Dart plugin. Find `Plugins` (in Settings/Preferences) and install the Dart plugin, then restart the IDE
+* Open flutter-intellij project in IntelliJ (select and open the directory of the flutter-intellij repository). Build it using `Build` | `Build Project`
```
helin24, line 23: I'm not sure if some of these naming differences (e.g. 'Make project' vs. 'Build project') are just differences between IntelliJ versions. These were the names I found while setting up so I'm assuming they're latest, but let me know if there's some other reason for the discrepancy.

CONTRIBUTING.md
helin24: Actually, added this back because then later it did look like I needed this - So I'm unclear on what's going on here.

CONTRIBUTING.md
stevemessick: JetBrains changes menu times sometimes. It used to be Make. I don't know about Configure but your change is better.

CONTRIBUTING.md
stevemessick: Yes, you need the plugin SDK to resolve references to IntelliJ classes. Unless you set up a full source configuration, and include all the IntelliJ source files in your project.

CONTRIBUTING.md
```diff
@@ -56,7 +57,10 @@ In order to be able to run a single test class or test method you need to do the
   VM Options to the corresponding field
 * Repeat the same with Working directory field - it must point to intellij-community/bin
 
+If running the full unit tests suite fails, check the run configuration and verify that 'Use classpath of module' is set to `flutter-intellij.test`.
```
stevemessick, line 60: "unit test" suite -- no "s" after test.

PR #4522, opened 2020-04-29
Prevent drawing child lines through other chars (https://github.com/flutter/flutter-intellij/issues/4297)

### LINKED ISSUE #4297: UI Guides' lines not auto-updating
Sometimes I start composing my widgets and the UI Guides' lines don't update and I get some weird behavior like this:

image

I have to go to the specific line, hit `Enter` then `Backspace` again in order to get it right:

image

helin24:
Fixes #4297 

Before (note line through `child:` briefly appears):
no-guidelines_

After:
guidelines

### DIFF
--- a/src/io/flutter/editor/WidgetIndentsHighlightingPass.java
+++ b/src/io/flutter/editor/WidgetIndentsHighlightingPass.java
@@ -365,6 +365,10 @@ public class WidgetIndentsHighlightingPass {
               );
             }
             else {
+              // If there are other characters on the same line as the widget, avoid drawing a backwards line
+              if (widgetLineStartX != widgetPoint.x) {
+                return;
+              }
               // Edge case where we draw a backwards line to clarify
               // that the node is still a child even though the line is in
               // the wrong direction. This is mainly for debugging but could help
--- a/src/io/flutter/editor/WidgetIndentsHighlightingPass.java
+++ b/src/io/flutter/editor/WidgetIndentsHighlightingPass.java
@@ -334,22 +334,11 @@ public class WidgetIndentsHighlightingPass {
           if (childLine != null) {
             final int childIndent = childLine.getIndent();
             // Draw horizontal line to the child.
-            final VisualPosition widgetVisualPosition = editor.offsetToVisualPosition(childLine.getGuideOffset());
+            final GuidelineOffsetDetail guidelineOffsetDetail = getGuidelineOffsetDetail(childLine, editor, doc, chars);
+            final VisualPosition widgetVisualPosition = editor.offsetToVisualPosition(guidelineOffsetDetail.textStartOffset);
             final Point widgetPoint = editor.visualPositionToXY(widgetVisualPosition);
 
-            // This additional point is computed because sometimes there are other characters on a line before the widget, e.g. if a
-            // widget has been moved to a line with other code but the new outline has not been received yet (see issue #4297)
-            // We want to compute the earliest position to avoid drawing a line through any characters before the widget if they're present.
-            final int startIndex = doc.getLineStartOffset(childLine.getLine());
-            final int endIndex = doc.getLineEndOffset(childLine.getLine());
-            int firstCharPosition = 0;
-            while (startIndex + firstCharPosition < endIndex && Character.isWhitespace(chars.charAt(startIndex + firstCharPosition))) {
-              firstCharPosition += 1;
-            }
-            final Point firstCharPoint = editor.visualPositionToXY(new VisualPosition(childLine.getLine(), firstCharPosition));
-
-            final int widgetLineStartX = min(widgetPoint.x, firstCharPoint.x);
-            final int deltaX = widgetLineStartX - start.x;
+            final int deltaX = widgetPoint.x - start.x;
             // We add a larger amount of panding at the end of the line if the indent is larger up until a max of 6 pixels which is the max
             // amount that looks reasonable. We could remove this and always used a fixed padding.
             final int padding = max(min(abs(deltaX) / 3, 6), 2);
@@ -360,13 +349,13 @@ public class WidgetIndentsHighlightingPass {
                 start.x + 2,
                 newY + lineHeight * 0.5,
                 //start.x + charWidth  * childIndent - padding,
-                widgetLineStartX - padding,
+                widgetPoint.x - padding,
                 newY + lineHeight * 0.5
               );
             }
             else {
               // If there are other characters on the same line as the widget, avoid drawing a backwards line
-              if (widgetLineStartX != widgetPoint.x) {
+              if (guidelineOffsetDetail.textStartOffset != guidelineOffsetDetail.offset) {
                 return;
               }
               // Edge case where we draw a backwards line to clarify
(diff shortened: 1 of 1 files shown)


helin24:
Do we want to change the loop back condition as well? It looks like it intersects with the first letter a little but I'm not sure if it would look stranger without a horizontal bit.
Screen Shot 2020-04-29 at 10 27 51 AM


Screen Shot 2020-04-29 at 10 28 07 AM

devoncarew:
@helin24 - does this address https://github.com/flutter/flutter-intellij/issues/4297? If so, you might want to update the description in the first comment to say something like `fix #xxx` - that will auto-close the issue when this PR lands.

devoncarew:
https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

src/io/flutter/editor/WidgetIndentsHighlightingPass.java
```diff
@@ -336,7 +336,17 @@ public void paint(@NotNull Editor editor, @NotNull RangeHighlighter highlighter,
             // Draw horizontal line to the child.
             final VisualPosition widgetVisualPosition = editor.offsetToVisualPosition(childLine.getGuideOffset());
             final Point widgetPoint = editor.visualPositionToXY(widgetVisualPosition);
-            final int deltaX = widgetPoint.x - start.x;
+
+            final int startIdx = doc.getLineStartOffset(childLine.getLine());
```
jacob314, line 340: nit. Here and elsewhere, abbreviations so write
startIndex rather than startIdx.

src/io/flutter/editor/WidgetIndentsHighlightingPass.java
```diff
@@ -347,7 +357,7 @@ public void paint(@NotNull Editor editor, @NotNull RangeHighlighter highlighter,
                 start.x + 2,
                 newY + lineHeight * 0.5,
                 //start.x + charWidth  * childIndent - padding,
-                widgetPoint.x - padding,
+                widgetLineStartX - padding,
```
jacob314, line 360: Also update the edge case where we draw a line backward (starts at line 364). A reasonable option would be to just omit drawing the line backward if widgetLineStartX != widgetPoint.x as the backwards lines are distracting so drawing them for a case where we aren't confident that something strange has happened would be a bad thing.

src/io/flutter/editor/WidgetIndentsHighlightingPass.java
jacob314: add a comment explaining why this code is needed as well as tracking widgetPoint. You could reference the bug and talk about when the code has changed but we don't yet have a new outline.
there is an additional issues that we are just working around that there seems to be a race condition that @devoncarew is able to reproduce where the invalid offsets sometimes stay around even after a new outline is sent.

helin24: Should I remove referring to #4297 for this change? Because this change is related but it doesn't address getting invalid offsets in the new outline.

src/io/flutter/editor/WidgetIndentsHighlightingPass.java
helin24: I'm unclear on the backwards lines - the original situation would cause a loop to be drawn from the parent element to the beginning of the widget (before any outline update):
Screen Shot 2020-05-01 at 9 38 23 AM
This seems okay? But is it confusing because maybe the loop should be drawn to before `child:`?

With this change, then there would be no backwards line in this case:
Screen Shot 2020-05-01 at 9 42 27 AM

jacob314: some background: the backwards lines don't really provide much value as they look pretty bad so that is why we should stop showing them if there is a risk we are showing a misleading line. The main reasons we show them at all is because it is easy to show them and because that explains why we don't draw a line in cases where the user's code isn't indented nicely. 
in the case you are dealing with where the outline doesn't match the current source code, it is safer to just not show the line.

jacob314: the blue lines from the row to its children also look bad in this case so it is a good thing if we don't show them when given bad data like this. In this case they are on the wrong side of the lines for their parents so the guides are hard to read so distract more than add value.

helin24:
@jacob314 hopefullly I did this the way you expected from our discussion. I was initially thinking I'd make changes to the `getGuideOffset` function in `OutlineLocation`, but after playing around in there I realized that `OutlineLocation` gets information from the widget tree that takes a while to update. 

For the other goal of not drawing lines if a parent is indented further than its child, I'm not sure what we want exactly - do we want there to be no vertical lines at all, or do we want the darker line from parent to child start to be removed while keeping the lighter line spanning the parent's entire definition?

I suppose for a parent with more than one child, if any of them are indented further to the left than the parent, then we maybe don't want to draw a line? Or we just make sure not to draw a line through any children that are too far left, sort of like soft wrap?

And in any case, do we want to remove the backwards loop logic completely? If we're removing some vertical lines I guess it doesn't make sense to have any backwards loops.

src/io/flutter/editor/WidgetIndentsHighlightingPass.java
```diff
@@ -426,6 +432,38 @@ public void paint(@NotNull Editor editor, @NotNull RangeHighlighter highlighter,
...
+        this.offset = offset;
+        this.textStartOffset = textStartOffset;
+      }
+    }
+
+    private GuidelineOffsetDetail getGuidelineOffsetDetail(OutlineLocation childLine,
+                                                           Editor editor,
+                                                           Document doc,
+                                                           CharSequence chars) {
+      // This additional point is computed because sometimes there are other characters on a line before the widget, e.g. if a
```
jacob314, line 452: nit: prefer to wrap comment lines at 80 chars.

src/io/flutter/editor/WidgetIndentsHighlightingPass.java
```diff
@@ -352,6 +354,10 @@ public void paint(@NotNull Editor editor, @NotNull RangeHighlighter highlighter,
               );
             }
             else {
+              // If there are other characters on the same line as the widget, avoid drawing a backwards line
```
jacob314, line 357: end all comments with a period.

jacob314:
> For the other goal of not drawing lines if a parent is indented further than its child, I'm not sure what we want exactly - do we want there to be no vertical lines at all, or do we want the darker line from parent to child start to be removed while keeping the lighter line spanning the parent's entire definition?
Lets keep it simple. We can still draw the vertical lines.
> 
> I suppose for a parent with more than one child, if any of them are indented further to the left than the parent, then we maybe don't want to draw a line? Or we just make sure not to draw a line through any children that are too far left, sort of like soft wrap?
As you describe that gets complicated quickly. I'd suggest we don't try to handle that case at all or only try to handle the case of moving the x coordinate of the vertical line so it matches where we decided to actually end the matching horizontal line. Anything else will be a lot of work for a case that should just be an edge case when data has briefly gotten out of sync.
> 
> And in any case, do we want to remove the backwards loop logic completely? If we're removing some vertical lines I guess it doesn't make sense to have any backwards loops.
What you have is fine. The backwards lines are ok to show if they are from an outline view we have confidence in. My suggestion is just to not show the backwards lines when we know something has gone wrong and our data is out of sync with the outlines.

jacob314:
lgtm I would suggest we land as is after fixing the tiny nit comments. I'm not sure the other edge cases to fix will be worth the effort required to fix them.

PR #4530, opened 2020-05-05
Update tests for 2020.1

stevemessick:
IdeaTestApplication has been removed. This update causes a new test failure. It is being ignored for now.


stevemessick:
@jacob314 this fixes the problem @helin24 discovered with 2020.1.

PR #4537, opened 2020-05-08
Build with gradle

stevemessick:
See flutter-idea/README.md for a description of the goal of the PR.

It will be easier to make sense of the top-level build.gradle if it is reviewed without comparing to the old one. The old one had some integration test support that is no longer relevant, so it was only useful for running unit tests. I tried to remove all non-essential declarations.

Mods to the plugin tool are coming later (if this doesn't cause CI to go red).


flutter-idea/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+The original flutter-intellij structure has been preserved as a courtesy to the engineers that
+already have it checked out. The original metadata files are checked in to the repository and will
+be used when the project is opened. To use the Gradle structure, import
+flutter-intellij/build.gradle instead of opening the project. That will rebuild the metadata
+files according to the Gradle structure.
+
+## Building the plugin
+
+Make sure the proper IDE is unpacked in the artifacts directory. Open a terminal and do
```
devoncarew, line 16: 'Make sure the proper IDE is unpacked in the artifacts directory' - is this a manual step that engineers have to perform, or one handled by a script?

stevemessick:
I use the plug-in tool with the -o and -u options.

PR #4547, opened 2020-05-14
Refactor plugin

stevemessick:
There is no new code here. I just split the large Dart file into several less-large files. 

Had to upgrade to Gradle 6.4 due to: https://github.com/gradle/gradle/issues/9202

The error message I got was:
>
  Execution failed for task ':flutter-idea:compileJava'.
  > Unable to find source java class: '/Users/messick/src/flutter/flutter-intellij/flutter-idea/src/main/java/io/flutter/sdk/FlutterSdk.java' because it does not belong to any of the source dirs: '[/Users/messick/src/flutter/flutter-intellij/flutter-idea/src/main/java, /Users/messick/src/flutter/flutter-intellij/flutter-idea/gen, /Users/messick/src/flutter/flutter-intellij/flutter-idea/third_party/vmServiceDrivers, /Users/messick/src/flutter/flutter-intellij/flutter-idea/src/main/resources]'

Getting from there to the Gradle issue took some digging. As you can see, the error message is incorrect.

Next up: Refactor pre-compilation editing, then add a new command to build using Gradle.


tool/plugin/lib/globals.dart
```diff
@@ -0,0 +1,15 @@
...
+
+const Map<String, String> plugins = const {
+  'io.flutter': '9212',
+  'io.flutter.as': '10139', // Currently unused.
+};
+
+const int cloudErrorFileMaxSize = 1000; // In bytes.
+
+// Globals are initialized early in ProductCommand.
+String rootPath;
```
devoncarew, line 12: Could these also work as statics off of `ProductCommand`?

tool/plugin/lib/globals.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright 2017 The Chromium Authors. All rights reserved. Use of this source
+// code is governed by a BSD-style license that can be found in the LICENSE file.
+
+const Map<String, String> plugins = const {
```
devoncarew, line 4: This could use a more descriptive name; `plugins` ==> `supportedPlatformVersions'? `currentIntelliJVersions`?

tool/plugin/lib/runner.dart
```diff
@@ -0,0 +1,56 @@
+// Copyright 2017 The Chromium Authors. All rights reserved. Use of this source
+// code is governed by a BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+import 'package:args/command_runner.dart';
```
devoncarew, line 6: nit: I'd organize imports here

tool/plugin/lib/globals.dart
stevemessick: I'll look into it for the next PR.

tool/plugin/lib/globals.dart
stevemessick: True, I'm not sure why we used that name. I'll do something in the next PR;  wanted this one to be as straight-forward as possible.

tool/plugin/lib/runner.dart
stevemessick: Done; in next PR.

stevemessick:
Thanks for the review!

PR #4557, opened 2020-05-19
Add null check to device refresh

stevemessick:
Need to check for null project. I should have caught this in the code review.


helin24:
Ah of course! I think you mentioned that to me earlier but I forgot when I was putting together the change. LGTM

stevemessick:
The kokoro run passed but github doesn't know that.

PR #4587, opened 2020-05-29
Display structured errors immediately

helin24:
https://github.com/flutter/flutter/pull/58118 added a flag to set structured errors on startup. This change sets the flag for the run command without requiring changes to an individual project's config.

Running with an early error should cause:
```
Launching lib/main.dart on iPhone SE (2nd generation) in debug mode...
Running pod install...                                              1.7s
Running Xcode build...
Xcode build done.                                           33.6s
Waiting for iPhone SE (2nd generation) to report its views...
Debug service listening on ws://127.0.0.1:50854/5Ztl1ytHvmY=/ws
Syncing files to device iPhone SE (2nd generation)...

════════ Exception caught by widgets library ═══════════════════════════════════════════════════════
The following FormatException was thrown building GalleryApp(dirty):
FormatException

The relevant error-causing widget was: 
  GalleryApp file:///Users/helinx/Documents/gallery/lib/main.dart:22:16
When the exception was thrown, this was the stack: 
#0      GalleryApp.build (package:gallery/main.dart:37:5)
#1      StatelessElement.build (package:flutter/src/widgets/framework.dart:4585:28)
#2      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:4511:15)
#3      Element.rebuild (package:flutter/src/widgets/framework.dart:4227:5)
#4      BuildOwner.buildScope (package:flutter/src/widgets/framework.dart:2632:33)
...
════════════════════════════════════════════════════════════════════════════════════════════════════
```

### DIFF
--- a/src/io/flutter/run/SdkFields.java
+++ b/src/io/flutter/run/SdkFields.java
@@ -130,7 +130,7 @@ public class SdkFields {
     if (buildFlavor != null) {
       args = ArrayUtil.append(args, "--flavor=" + buildFlavor);
     }
-    if (FlutterSettings.getInstance().isShowStructuredErrors()) {
+    if (FlutterSettings.getInstance().isShowStructuredErrors() && flutterSdk.getVersion().isDartDefineSupported()) {
       args = ArrayUtil.append(args, "--dart-define=flutter.inspector.structuredErrors=true");
     }
     command = flutterSdk.flutterRun(root, main.getFile(), device, runMode, flutterLaunchMode, project, args);
--- a/src/io/flutter/run/SdkFields.java
+++ b/src/io/flutter/run/SdkFields.java
@@ -20,6 +20,7 @@ import io.flutter.pub.PubRootCache;
 import io.flutter.run.common.RunMode;
 import io.flutter.sdk.FlutterCommand;
 import io.flutter.sdk.FlutterSdk;
+import io.flutter.settings.FlutterSettings;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/SdkFields.java
```diff
@@ -129,6 +130,9 @@ public GeneralCommandLine createFlutterSdkRunCommand(@NotNull Project project,
     if (buildFlavor != null) {
       args = ArrayUtil.append(args, "--flavor=" + buildFlavor);
     }
+    if (FlutterSettings.getInstance().isShowStructuredErrors()) {
+      args = ArrayUtil.append(args, "--dart-define=flutter.inspector.structuredErrors=true");
```
devoncarew, line 134: @DanTup for FYI

DanTup: Great, this should fix https://github.com/Dart-Code/Dart-Code/issues/1850.

I presume it's safe to pass `--dart-define=` to older SDKs and this won't need a version check?

helin24: I'm not sure about this - is the concern that an older SDK will break when using `--dart-define` or this particular flag? If the SDK doesn't consider the flag yet there should be no problem, but I don't know if anything happens if `--dart-define` doesn't exist.

DanTup: > is the concern that an older SDK will break when using `--dart-define` or this particular flag?

Either, I guess. I wondered if it's safe for me to just always send this (if structured errors is enabled), or if that might result in errors for users with older Flutter SDKs. As long as it works in the latest stable and later it's probably not a real concern, but since I already have version checks on some flags, it's easy to add one if required to have compatibility with more SDKs.

(I can test it if nobody knows the answer off-hand, I just thought I'd ask and save the multiple sets of downloads changing between versions 😁)

helin24: Yes can you test and update me too? If there is an issue I think I would need a version check as well.

DanTup: Looks like `--dart-define` exists in both the current stable, and the previous stable (v1.12), added in https://github.com/flutter/flutter/pull/44083.

I think I'll add a version in VS Code check since it's easy and I already have it for some other things, but it's probably not entirely necessary.

PR #4595, opened 2020-06-01
remove a delay on hot reload

devoncarew:
- remove a delay on hot reload

This delay was an artifact from when we used the results of the analysis server to inform whether we should issue a hot reload on save or not. We no longer use the analysis server results - we always reload, unless there are structural issues with the file (that info comes from IntelliJ's PSI model, and we don't need a delay to get that info).

Related to https://github.com/flutter/flutter/pull/55376.

cc @DanTup @jonahwilliams 

@stevemessick and @helin24 for review.

### DIFF
--- a/src/io/flutter/run/FlutterReloadManager.java
+++ b/src/io/flutter/run/FlutterReloadManager.java
@@ -158,11 +158,6 @@ public class FlutterReloadManager {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


stevemessick:
Ignore the kokoro failure. I have a PR out to fix it.

PR #4603, opened 2020-06-05
Avoid updating file watch if bazel project is disconnected

helin24:
When prodaccess is lost, a bazel project's root directory disappears and file watch is disconnected in existing code. I think if we maintain the file watch settings when a disconnect occurs, the plugin will continue listening for changes to the root file and refresh appropriately if access is regained.

I tested this by keeping a project open during prodaccess loss and regain and verifying with breakpoints that refresh occurs in `WorkspaceCache`.

### DIFF
--- a/src/io/flutter/bazel/WorkspaceCache.java
+++ b/src/io/flutter/bazel/WorkspaceCache.java
@@ -30,6 +30,7 @@ import java.util.concurrent.atomic.AtomicReference;
 public class WorkspaceCache {
   @NotNull private final Project project;
   @Nullable private Workspace cache;
+  @NotNull private boolean disconnected = false;
 
   private boolean refreshScheduled = false;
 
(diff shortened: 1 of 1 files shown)


src/io/flutter/bazel/WorkspaceCache.java
```diff
@@ -129,7 +132,12 @@ public void unsubscribe(Runnable callback) {
    */
   private void refresh() {
     final Workspace workspace = Workspace.loadUncached(project);
-    if (workspace == cache) return;
+    if (workspace == cache && !disconnected) return;
+    if (cache != null && workspace == null) {
+      disconnected = true;
```
jacob314, line 137: should we instead return early here after setting 
`disconnected = true;`
? 
setting cache=null in this case seems a bit odd.

helin24: I thought it would be better for cache to be null when going on to notify listeners, because the only listener we want to behave differently is the workspace itself so that it keeps watching the previous file. For devices I think we still want the menu to disappear and for developers to not be able to run a project if they're disconnected.

PR #4604, opened 2020-06-05
Update build to support canary

stevemessick:


.travis.yml
```diff
@@ -36,5 +36,5 @@ env:
   - UNIT_TEST_BOT=true
   - IDEA_VERSION=3.6
   - IDEA_VERSION=4.0
-  - IDEA_VERSION=2020.1
   - IDEA_VERSION=2020.2
+  - IDEA_VERSION=4.1
```
devoncarew, line 40: move this line up? (I assume that 4.1 == 2020.1)

product-matrix.json
```diff
@@ -44,6 +32,19 @@
       "dartPluginVersion": "202.4357.23",
       "sinceBuild": "202.4357.23",
       "untilBuild": "202.*"
+    },
+    {
```
devoncarew, line 36: Similarly, I'd move this block of json up so that the platform versions are roughly in order.

stevemessick: Done

PR #4622, opened 2020-06-12
Start paused during run mode

helin24:
App should be paused on run mode as well as on debug mode so we won't miss any events from the beginning of an app's running.

Verified that a project can be run without hangs in run mode (no breakpoints), profile mode, and release mode.

### DIFF
--- a/src/io/flutter/sdk/FlutterSdk.java
+++ b/src/io/flutter/sdk/FlutterSdk.java
@@ -244,7 +244,7 @@ public class FlutterSdk {
       args.add("--device-id=" + device.deviceId());
     }
 
-    if (mode == RunMode.DEBUG) {
+    if (mode == RunMode.DEBUG || mode == RunMode.RUN) {
       args.add("--start-paused");
     }
 
@@ -320,6 +320,8 @@ public class FlutterSdk {
       if (!myVersion.flutterTestSupportsMachineMode()) {
         throw new IllegalStateException("Flutter SDK is too old to debug tests");
       }
+    }
+    if (mode == RunMode.DEBUG || mode == RunMode.RUN) {
       args.add("--start-paused");
     }
     if (FlutterSettings.getInstance().isVerboseLogging()) {


src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -244,7 +244,7 @@ public FlutterCommand flutterRun(@NotNull PubRoot root,
       args.add("--device-id=" + device.deviceId());
     }
 
-    if (mode == RunMode.DEBUG) {
+    if (mode == RunMode.DEBUG || mode == RunMode.RUN) {
```
jacob314, line 247: is there any other mode? also make sure we are not breaking release/profile mode.

helin24: There is also `RunMode.PROFILE` - I guess it would make sense to enable --start-paused for all of the IntelliJ modes (so in that case just remove this conditional)?

I'm having trouble testing flutter profile mode and release mode. Do I need to go through some additional setup steps for this? I se this doc for iOS at least https://flutter.dev/docs/deployment/ios

jacob314:
lgtm

gmcdowell:
Guys it looks like you have **caused considerable chaos** for us Flutter Developers with this change.

In short when I now run a Flutter Test via IntelliJ, **tests no longer run at all** as you have paused them with NO way of changing this behaviour.

Please revert

Screen Shot 2020-06-29 at 1 05 07 PM

blasko:
This is a _**full stop**_ breaking change that needs to be addressed asap.

franferri:
Same problem here.

benoitskipr:
running them in debug mode shows the error in console still, but tests run (as a temp fix)

franferri:
I have open a TT to flutter themselves: https://github.com/flutter/flutter/issues/60486#issuecomment-651050511
Maybe I had to post it here

franferri:
I have installed today in Android Studio the plugins Flutter and Dart. Last versions.

I was working for hours doing tests in my project and after the update suddenly I can't run the Flutter tests anymore, it keeps saying:

```
Warning: You should set `SILENT_OBSERVATORY` to true when debugging the VM as it will output the observatory URL by default.
This breaks the various reporter contracts.
To set the value define `DART_VM_OPTIONS=-DSILENT_OBSERVATORY=true`.
```

I'm using macOS last version with Flutter stable channel:

```
Flutter 1.17.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 1ad9baa8b9 (11 days ago) • 2020-06-17 14:41:16 -0700
Engine • revision ee76268252
Tools • Dart 2.8.4
I have tried to add the DART_VM_OPTIONS=-DSILENT_OBSERVATORY=true to the environment, export in a console and then start Android Studio with open -a... nothing works.
```

Either Flutter is rewriting the var or the build-in dart-sdk not picking it.

Since setting the variable as global don't work, nor in the terminal session before running Android Studio, nor Android Studio let you change anything in the dart plugin panel... I'm out of options

if anybody knows how to fix this please help.

Posted in stackoverflow too in a desperate attempt to get help: https://stackoverflow.com/questions/62629506/dart-vm-options-dsilent-observatory-true

**Steps to reproduce:**

macos, download android studio, install flutter plugin, new flutter project, run the test with right click, run test (you may have edit the runtest configuration to mark it as flutter test instead of dart test).

As simple as that, the sample project don't run the tests, nor any other project you may have.

franferri:
Tested solution till this get solved for Android Studio.

1. Go to plugins and remove dart and flutter ones, uninstall completelly
2. From the following URL's download the previous versions (click load more at the end of the page till the versions show):
- https://plugins.jetbrains.com/plugin/9212-flutter/versions
- https://plugins.jetbrains.com/plugin/6351-dart/versions
3. Use the following versions (before the Jun25 aberration versions
* Dart Plugin (193.6911.31)
* Flutter Plugin (46.0.2)
4. Live the dream

To manually install a plugin from disk go to "Preferences" -> "Plugins" -> Click the little cog icon near the "Installed" tab -> Choose "Install Plugin from Disk..."

helin24:
@franferri thanks for the detailed report and suggestion - I've reverted this change and this will be released soon. In the meantime, using v46 of the flutter plugin (without changing the dart plugin) should work, and you should also be able to run your tests in debug mode.

Arkangel12:
> **Tested solution till this get solved for Android Studio.**
> 
> 1. Go to plugins and remove dart and flutter ones, uninstall completelly
> 2. From the following URL's download the previous versions (click load more at the end of the page till the versions show):
> 
> * https://plugins.jetbrains.com/plugin/9212-flutter/versions
> * https://plugins.jetbrains.com/plugin/6351-dart/versions
> 
> 1. Use the following versions (before the Jun25 aberration versions
> 
> * Dart Plugin (193.6911.31)
> * Flutter Plugin (46.0.2)
> 
> 1. Live the dream
> 
> To manually install a plugin from disk go to "Preferences" -> "Plugins" -> Click the little cog icon near the "Installed" tab -> Choose "Install Plugin from Disk..."

This steps fail for me in IntelliJ, when i remove Dart and Flutter plugins and try to manually added it sent a message of no compatibility.

when we can have this fix  at the IntelliJ marketplace?

franferri:
In the lower part of the screens for the plugins download, there is a
button to load more. Load till you find the indicated versions. 3 - 4 times
till you find them. The exact versions. Must work as it does in Android
studio.

Arkangel12:
i did download the exact files it not a file issue, looks like i another problem trying to load a plugin manually  in IntelliJ.

 [This i the message i receive](https://stackoverflow.com/questions/61294519/plugin-flutter-is-incompatible-with-this-installation-android-studio-3-6-3) when i try to do it

lukechi1219:
after install latest Flutter plugin with this fix, Android Studio (IntelliJ) runs flutter test with wrong flag "--start-paused" which would wait for a debugger to connect. Yet in run test mode there won't be a debugger, so it waits until timeout.

PLEASE rollback asap...

unknown (1)

safarmer:
Also, there is no option to run tests in debug mode if you are running all tests in a folder. Everything works when I run an individual file.

helin24:
@safarmer thanks for noting this - going to be tracking this in https://github.com/flutter/flutter-intellij/issues/4737 if you want to follow updates.

Elias8:
Any updates, please?

helin24:
@Elias8 we made a temporary fix for directory tests in run mode, but debug mode for directory tests will require more work and is not one of our priorities currently. You can upvote or comment on https://github.com/flutter/flutter-intellij/issues/4742 if this is important to you. (alternatively, if you've seen a regression for a separate problem mentioned here, feel free to post details in a new issue)

PR #4628, opened 2020-06-16
Update tool/colors to support generating mapping for CupertinoColor

DanTup:
This adds another script to tool/color to create a mapping for CupertinoColors to allow them to be shown in editor gutters:

Screenshot 2020-06-16 at 11 39 01

The `CupertinoColor` class references quite a few other classes, so reusing its file as-is required creating quite a few empty stub classes (an alternative could be to make it a real Flutter app and use the SDK directly - which I think Icons already does - but I wasn't sure if that was overkill). 

@devoncarew


tool/colors/colors_main_cupertino.dart
```diff
@@ -0,0 +1,87 @@
...
+  // https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/cupertino/colors.dart
+
+  StringBuffer buf = new StringBuffer();
+  buf.writeln('# Generated file - do not edit.');
+  buf.writeln();
+  buf.writeln('# suppress inspection "UnusedProperty" for whole file');
+  buf.writeln();
+
+  // colors
+  final Map<String, Color> colors = <String, Color>{
```
devoncarew, line 18: This is a lot of code that could potential get out of date wrt CupertinoColors and or will need maintenance.

To speak to your comment in the PR, could we reduce the amount of code here if we generate the properties files from a Flutter app, ala the current icons code?

If I remember correctly, the code to match colors was written before the code to generate icons, so just may not have been aware of the other way to generate these files.

devoncarew:
@DanTup - do you have similar code to generate the metadata files in the Dart-Code repo?

If so, we may want to move this to a more common location - in flutter/devtools - packages/framework_metadata? in `flutter/metadata`? I think the color and icons files would be generally useful to development tools.

DanTup:
@devoncarew I don't - I'm just running this script and then tweaking the output to match the format I needed. I think moving them out of this repo makes sense though I don't have an opinion on where is best (though putting it in DevTools might be equally strange - if DevTools already has some stuff like this, maybe it would all fit in a tool-agnostic repo?).

I'll take a look at the icons project and try to clean this up anyway, and I can send a PR wherever you think would make most sense.

DanTup:
@devoncarew I re-did this based on `update_icons`. However, it's a little different - the Icons one can get all the info it needs from a regex, but colors are more tricky - for example:

```dart
  static const CupertinoDynamicColor systemGrey3 = CupertinoDynamicColor.withBrightnessAndContrast(
    debugLabel: 'systemGrey3',
    color: Color.fromARGB(255, 199, 199, 204),
    darkColor: Color.fromARGB(255, 72, 72, 74),
    highContrastColor: Color.fromARGB(255, 174, 174, 178),
    darkHighContrastColor: Color.fromARGB(255, 84, 84, 86),
  );
```

Rather than making regexs to parse all of this, I made it in two passes:

- First pass extracts just the colour names (these are easy to get) and builds a `Map<String, Color>`
- Second pass uses the file written by the first pass so it can just read the fields from the (Material|Cupertino)Color instance directly

It seems to work great, though it feels a little weird to write Dart files and then execute a Dart script that needs them (and this means committing the generated intermediate list for it to compile in the repo). It might be slightly cleaner to split `update_colors.dart` into two scripts to do this.

Let me know what you think (and if I should put this somewhere else).

DanTup:
Also - if this is a pain to review because there's so much codegen and copied Flutter files, I can make a PR/commit with just the "real" changes to make it easier - just lmk!

devoncarew:
> It seems to work great, though it feels a little weird to write Dart files and then execute a Dart script that needs them (and this means committing the generated intermediate list for it to compile in the repo). It might be slightly cleaner to split `update_colors.dart` into two scripts to do this.

Can you add todo: for known cleanup items?

I think we should move this code to an IDE neutral repo, but don't need to gate landing this PR on that change.

DanTup:
I've added a TODO about splitting the script into two (I'm happy to do that now, if you think it makes sense). I don't know of a way to eliminate having two steps/script besides something like mirrors, but not sure if that's worthwhile.

> I think we should move this code to an IDE neutral repo, but don't need to gate landing this PR on that change.

Ok, if you think of somewhere suitable, let me know and I'm happy to move them over.

devoncarew:
@DanTup - I'm having trouble re-generating the colors locally now. Can you confirm that you don't have any local changes that aren't checking in (or, that w/ a clean checkout of this repo, you're able to run `grind colors`)? Thanks!

devoncarew:
(I have a few minor updates to the Cupertino colors support that I'd like to send as s PR)

devoncarew:
I'm making some progress on re-generating the files -

DanTup:
@devoncarew what's the error you're having? I've no idea what `grind` is, I was just running the script manually - but I'll take a look. It does look like we might now have two copies of the Color files (one downloaded temporarily to regex, and the other used to import when running the other step) so I'll fix that up.

DanTup:
Oh, I see there's more code related to the colours in `grind.dart` I hadn't seen - I thought the other script was standalone 😬 It's actually doing the bit I need to fix the above (downloading the file and replacing the imports), so fixing them both will probably be the same thing.

devoncarew:
Gotcha; I'll have an update to capture a few more CupertioColors. If you wouldn't mind de-duping any work between the grind script and `update_colors.dart` after that PR lands, that would be much appreciated.

DanTup:
@devoncarew sure - I've got a change ready at https://github.com/DanTup/flutter-intellij/commit/f4c0ed5693d62d47818a205013339458f9bd0676 which:

- fixes up the downloading (and consolidates it with what was in grind.dart - though now it's all in update_colors)
- splits the script into two (addressing the TODO) so if you run it and generate bad code, you're not left unable to re-run the script to generate it because it imports the broken code

Just let me know when you're done and I'll rebase and fix up.

PR #4631, opened 2020-06-17
Delay flutter daemon start until indexing finished

helin24:
Verify that opening workbench, opening project(s), quitting, then re-opening workbench consistently results in re-opened projects having device options.
Fixes https://github.com/flutter/flutter-intellij/issues/3892

### DIFF
--- a/src/io/flutter/run/daemon/DeviceService.java
+++ b/src/io/flutter/run/daemon/DeviceService.java
@@ -12,6 +12,7 @@ import com.intellij.execution.ExecutionException;
 import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/daemon/DeviceService.java
```diff
@@ -225,6 +226,8 @@ private DeviceDaemon chooseNextDaemon(Refreshable.Request<DeviceDaemon> request)
     emulatorManager.refresh();
 
     try {
+      // Wait for indexing to finish before starting daemon process
```
jacob314, line 229: nit: end comment with a period.
Do we have an idea why starting the daemon service in dumb mode causes problems?
If is worth commenting on it here as it is a little surprising that we have to wait for smart mode.

helin24:
Closing this since https://github.com/flutter/flutter-intellij/pull/4633 is a better fix

PR #4633, opened 2020-06-17
Recognize \r as newlines from flutter

helin24:
Verify that opening workbench, opening project(s), quitting, then re-opening workbench consistently results in re-opened projects having device options.

Fixes https://github.com/flutter/flutter-intellij/issues/3892

### DIFF
--- a/src/io/flutter/utils/StdoutJsonParser.java
+++ b/src/io/flutter/utils/StdoutJsonParser.java
@@ -21,6 +21,7 @@ public class StdoutJsonParser {
   private boolean bufferIsJson = false;
   private final List<String> lines = new ArrayList<>();
   private boolean eatNextEol = false;
+  private boolean isPotentialWindowsReturn = false;
 
   /**
    * Write new output to this [StdoutJsonParser].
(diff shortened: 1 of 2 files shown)


src/io/flutter/utils/StdoutJsonParser.java
```diff
@@ -32,11 +33,25 @@ public void appendOutput(String string) {
...
+          eatNextEol = true;
+          isPotentialWindowsReturn = true;
           continue;
         }
       }
 
+      if (isPotentialWindowsReturn) {
+        isPotentialWindowsReturn = false;
+
+        if (c != '\n') {
```
jacob314, line 50: a little odd that we don't flush when the `c` is a `'\n'`
is this really
`assert(c!= '\n');`
?

testSrc/unit/io/flutter/utils/StdoutJsonParserTest.java
```diff
@@ -49,6 +51,21 @@ public void appendsWithoutLineBreaks() {
     );
   }
 
+  @Test
+  public void jsonWithLineBreaks() {
+    final StdoutJsonParser parser = new StdoutJsonParser();
+    parser.appendOutput("[{'foo':'bar'}]\ntest\n");
```
jacob314, line 57: somewhat minor: do we have to worry about `\r` or `\n` in the middle of a json block? I don't know off the top of my head how the dart json encoder deals with them.

helin24: It seems this is dealt with fine, I tried putting \r, \n and \r\n in the middle of flutter tools output and it's recognized correctly.

DanTup:
@helin24 thanks for the ping! Dart-Code did have a similar bug (if the Flutter startup lock was held by another process, it'd fail to read the `daemon.connected` event, and therefore never enable device polling).

Issue #4635, opened 2020-06-12
Android studio flutter plugin snippet auto-complete for InheritedWidget still uses deprecated inheritFromWidgetOfExactType method


CareF:
Currently the default snippet for `InheritedWidget` (inh+`<tab>`) is still using `inheritFromWidgetOfExactType`. 

According to [API documents](https://api.flutter.dev/flutter/widgets/BuildContext/inheritFromWidgetOfExactType.html) `inheritFromWidgetOfExactType` is deprecated and should be replaced by `dependOnInheritedWidgetOfExactType`.

PR #4636, opened 2020-06-18
update the InheritedWidget template

devoncarew:
- update the `InheritedWidget` template
- fix https://github.com/flutter/flutter-intellij/issues/4635

Screen Shot 2020-06-17 at 5 58 44 PM

### DIFF
--- a/resources/liveTemplates/flutter_miscellaneous.xml
+++ b/resources/liveTemplates/flutter_miscellaneous.xml
@@ -11,7 +11,7 @@
       <option name="DART_TOPLEVEL" value="true" />
     </context>
   </template>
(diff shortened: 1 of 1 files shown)


resources/liveTemplates/flutter_miscellaneous.xml
```diff
@@ -11,7 +11,7 @@
       <option name="DART_TOPLEVEL" value="true" />
     </context>
   </template>
-  <template name="inh" value="class $NAME$ extends InheritedWidget {&#10;  const $NAME$({&#10;    Key key,&#10;    @required Widget child,&#10;  }) : assert(child != null),&#10;       super(key: key, child: child);&#10;&#10;  static $NAME$ of(BuildContext context) {&#10;    return context.inheritFromWidgetOfExactType($NAME$) as $NAME$;&#10;  }&#10;&#10;  @override&#10;  bool updateShouldNotify($NAME$ old) {&#10;    return $SHOULD_NOTIFY$;&#10;  }&#10;}" description="New Inherited widget" toReformat="true" toShortenFQNames="true">
```
jacob314, line 14: Is there a way we can write these templates with line breaks? Without line breaks this is slow to review. Maybe it would make sense to even include a section in a comment with the code with #10 replaced with line breaks.

jacob314:
lgtm

PR #4655, opened 2020-06-25
Add changes for M47

stevemessick:


stevemessick:
Merging before kokoro finishes since this doesn't affect the kokoro run.

PR #4657, opened 2020-06-25
Fix build issues for M47

stevemessick:


stevemessick:
TBR

stevemessick:
Our build process is complex. We have to make separate zip files for each IDE: Android Studio 3.6, 4.0, 4.1, 4.2 and IntelliJ 2020.2 EAP. (The list keeps changing; the oldest version is generally required internally.) The ones for Android Studio also work in their corresponding IntelliJ version. The APIs keep changing, especially in Android Studio. The change that caused this (and the one before) came in the latest canary build.

Anyway, to make all this happen we have a Dart script that drives the Gradle build. It edits source files as needed for the version being built. This latest API change I thought wouldn't need any changes to that, but 3.6 is so old (based on IntelliJ 2019.2) it doesn't have one of the classes in the method parameter types. So that has to be removed when building 3.6.

Issue #2708, opened 2018-10-19
Use the "What's new in Android Studio" panel for release notes


InMatrix:
When I upgraded Android Studio to 3.2, it showed me this nice graphical release notes in a side panel (screenshot below). Can we use this to provide better release notes after a new version of the Flutter plugin is installed? The panel can be re-opened via `Help > What's New in Android Studio`.

![image](https://user-images.githubusercontent.com/348942/47233928-0274f600-d389-11e8-939a-74cbe8609880.png)

pq:
That would be great.  I'm not sure if that's something we can contribute to...

@stevemessick : any idea?

stevemessick:
Not off hand; I'll have to see if they defined a contribution mechanism. We wouldn't want to overwrite it but if we can add to it that might be nice. Or maybe we want a `Help > What's New in Flutter` as a separate panel.

@devoncarew Seems like it would really help discovery of new stuff.

InMatrix:
> Or maybe we want a Help > What's New in Flutter as a separate panel.
That works, too. It's probably cleaner than adding to "What's New in Android Studio".

devoncarew:
Yup, I agree that this would makes sense as a new `What's New in Flutter` panel.

stevemessick:
The Assistant framework can be used for more than a "What's New" panel. Go to Tools > Firebase for a much more elaborate example. It basically implements a built-in tutorial that does some amount of code editing for you, assuming you have an Android (not Flutter) project open.

As far as making a "What's New" for Flutter, it looks straight-forward. We'd need to implement some new classes and define the content to display in an XML file.

InMatrix:
Let's give it a try for the next plugin release?

Cc: @mjohnsullivan We can use some help from DevRel to make this space more useful.

mjohnsullivan:
Certainly; we can help.

InMatrix:
Cc: @csells

PR #4658, opened 2020-06-25
[WIP] Flutter news

stevemessick:
This will be a good way to view Flutter SDK release notes. We will have to work with them to get the content properly formatted as XML and possibly design a download protocol for retrieving channel-specific content. The latter would be important to have so the plugin does not have to be updated to see SDK release notes.

To see an example, launch Android Studio and select Help > What's New in Android Studio. Another example is Tools > Firebase, which is much more complex. Here's a full screen shot showing the current content in this PR.

Screen Shot 2020-06-25 at 3 04 12 PM

(Those buttons don't actually work yet.)

TODO:
- remove action "What's New in Flutter"
- remove sample content
- remove as-hacks (mods to classes named "Android*")
- fix copyright in OpenFlutterNewsSidePanelAction

cc @devoncarew @csells @InMatrix 
See #2708


stevemessick:
Closing to create a cleaned-up PR for review.

PR #4664, opened 2020-06-26
Add JxBrowser dependencies and show in panel

helin24:
This change allows the plugin to download jxbrowser files and load them onto the class path dynamically. If installed correctly, running a project will open the devtools inspector page in the 'Fluter Inspector' tab:
Jul-15-2020 12-14-22

This change is gated so nothing related to JxBrowser will run unless we change `ENABLE_JXBROWSER` to `true` in `JxBrowserManager` in development. Gating can be opened once all of the pieces are merged.


stevemessick:
You'll need to put the JxBrowser API jar file(s) into lib/jxbrowser in order to compile. It looks like there's a problem with the travis run that shows passing when lots of compilation errors were reported. The API isn't too large, so that won't be a problem.

src/io/flutter/actions/TestJxBrowserAction.java
```diff
@@ -0,0 +1,57 @@
+package io.flutter.actions;
```
stevemessick, line 1: Delete this class.

src/io/flutter/view/FlutterView.java
```diff
@@ -182,11 +177,11 @@ private PerAppState getOrCreateStateForApp(FlutterApp app) {
   }
 
   private void addInspectorViewContent(FlutterApp app, @Nullable InspectorService inspectorService, ToolWindow toolWindow) {
+    String browserUrl = app.getConnector().getBrowserUrl();
+
+    System.setProperty("jxbrowser.license.key", "6P830J66YAO1XQRI1FNEXP8ZTHUOQTY2YUJRCN9RTARQ58QFAP63GRRH20B3P5PB7PC8");
```
stevemessick, line 182: Get this key from an environment variable. And document it :)

resources/META-INF/plugin.xml
```diff
@@ -1075,6 +1075,9 @@
       <action id="flutter.submitFeedback" class="io.flutter.actions.FlutterSubmitFeedback"
               text="Submit Feedback..."
               description="Provide feedback for the Flutter plugin"/>
+      <action id="flutter.testJxBrowser" class="io.flutter.actions.TestJxBrowserAction"
```
stevemessick, line 1078: Delete this.

resources/META-INF/plugin_template.xml
```diff
@@ -101,6 +101,9 @@
       <action id="flutter.submitFeedback" class="io.flutter.actions.FlutterSubmitFeedback"
               text="Submit Feedback..."
               description="Provide feedback for the Flutter plugin"/>
+      <action id="flutter.testJxBrowser" class="io.flutter.actions.TestJxBrowserAction"
```
stevemessick, line 104: Delete this.

src/io/flutter/view/FlutterView.java
helin24: Ah yeah, I have tried it as an environment variable, but I'm unclear on whether we eventually want to expose the key  or if this is something that's part of the build but not in the source code. Is it okay to have the key in our .iml file?

helin24:
I think I added the right files to version control, but I'm not quite sure - one of the travis builds is still broken. There's no need to review this carefully right now, since I still need to add some things before this would be landable (but I'm happy to make changes to fix the setup). This was more an option for @jacob314 to check out and see whether the performance looks adequate.

src/io/flutter/view/FlutterView.java
stevemessick: In general, auth keys are not included in source code, whether java or iml.

I'm not sure what the commercial auth is like. I saw some issues referring to files used for authentication, but that may have been an earlier version. In any case, this will need to be part of the build process.

stevemessick:
There are still several failures that are reporting success. I need to fix the false positive, but I'm not sure what's causing the compilation problems.

src/io/flutter/run/FlutterDebugProcess.java
```diff
@@ -17,10 +17,7 @@
 import com.intellij.xdebugger.XDebugSession;
 import com.jetbrains.lang.dart.util.DartUrlResolver;
 import io.flutter.FlutterUtils;
-import io.flutter.actions.ReloadAllFlutterApps;
-import io.flutter.actions.ReloadFlutterApp;
-import io.flutter.actions.RestartAllFlutterApps;
-import io.flutter.actions.RestartFlutterApp;
+import io.flutter.actions.*;
```
stevemessick, line 20: This should be rolled back. I'm pretty sure we settled on enumerating all imports.

stevemessick:
You'll need to cherry-pick or rebase to get the change to the Gradle script that solves your test bot failure.

helin24:
> You'll need to cherry-pick or rebase to get the change to the Gradle script that solves your test bot failure.

Hmm I did rebase (build.gradle files in flutter-idea and flutter-studio are the same as in master) but maybe another file is off. tool/plugin/compile.xml and .idea/runConfigurations/Flutter_Plugin.xml are a little different, will try making those the same.

src/io/flutter/ProjectOpenActivity.java
```diff
@@ -35,12 +54,73 @@
...
+          System.out.println("Unable to download file");
+        }
+      }
+    };
+    BackgroundableProcessIndicator processIndicator = new BackgroundableProcessIndicator(task);
+    processIndicator.setIndeterminate(false);
+    ProgressManager.getInstance().runProcessWithProgressAsynchronously(task, processIndicator);
+  }
+
+  private void loadClasses(File file) {
```
stevemessick, line 104: I think we should do a little refactoring. See AddToAppUtils.getStaticFieldValue(). If you move that method to, say, SystemUtils then you can just use it here. Unfortunately, I defined it in a class that's part of flutter-studio so you can't use it directly.

src/io/flutter/utils/FileUtils.java
```diff
@@ -0,0 +1,54 @@
...
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.utils;
+
+import com.intellij.openapi.util.SystemInfoRt;
+import com.intellij.util.SystemProperties;
+
+public class FileUtils {
+  private static final String IDE_VENDOR_NAME = System.getProperty("idea.vendor.name", "JetBrains");
```
stevemessick, line 12: I'm not sure JetBrains approves of outsiders (like us) modifying their internal data. If we make this "Google" then it can be shared with all IDEs, and the file will be only downloaded once.

src/io/flutter/ProjectOpenActivity.java
helin24: Should this function return the `addUrl` method? Something like `Method method = SystemUtils.getStaticFieldValue(URLClassLoader.class, Function<String, void>.class, "addUrl");` to replace everything before `method.invoke(classLoader, url);` ?

stevemessick: Hmm, I don't know what I was thinking. See various methods in ReflectionUtils for helpers that do some of what you're doing here. Please ignore AddToAppUtils.

devoncarew:
Hey - some drive by comments:

- is there UI we show the user during the download (an 'in-progress' UI)?
- is there provision for downloading new versions of the files when we update them?
- what are the tradeoffs associated w/ including the files in the plugin?

When I downloaded the above linked files, I see:

```
             jxbrowser-7.8.jar  9,155kb
      jxbrowser-javafx-7.8.jar    161kb
     jxbrowser-linux64-7.8.jar 73,060kb
 jxbrowser-linux64-arm-7.8.jar 70,422kb
         jxbrowser-mac-7.8.jar 81,130kb
       jxbrowser-swing-7.8.jar    131kb
         jxbrowser-swt-7.8.jar    121kb
       jxbrowser-win32-7.8.jar 63,175kb
       jxbrowser-win64-7.8.jar 65,886kb
```

I think we only need to include the 64 OS versions, and likely don't need linux arm. At a rough estimate, that's 229MB of jar files.

helin24:
> * is there UI we show the user during the download (an 'in-progress' UI)?
Currently, no, but it should be easy to add something for this. There's an in-progress status while downloading files and we would also need to check the progress of installing devtools.

> * is there provision for downloading new versions of the files when we update them?
If we started requiring a new version of the jxbrowser files in a new version of our plugin, once a user downloads our plugin and restarts, our plugin will see that the new version of the jxbrowser files aren't present locally and a download of the new files will start.

A related (lesser) problem that is not yet handled is how to get rid of old jxbrowser files that are outdated, because there may be multiple IDE applications that use a common set of files, so e.g. intellij might be using v2 jxbrowser files while android studio is still using v1. So maybe we need another local file that tracks what version of files each application needs.

> * what are the tradeoffs associated w/ including the files in the plugin?
Including all of the ones that we need (the 230 MB you mentioned) would be simpler for sure. But it would be problematic for users on slow or metered connections, especially since the files would be downloaded for every plugin update instead of for every jxbrowser version update. It would also be hard for Steve to upload about five copies of the plugin for each version of intellij we support.

src/io/flutter/ProjectOpenActivity.java
```diff
@@ -41,6 +42,8 @@ public ProjectOpenActivity() {
 
   @Override
   public void runActivity(@NotNull Project project) {
+    JxBrowserManager.get().setUp(project);
```
devoncarew, line 45: I think we only want to start this if we know this is a flutter project; I'd move this down somewhere below the `declaresFlutter` check.

src/io/flutter/view/FlutterView.java
devoncarew: Consider doing something like reading the key from a properties file, having a demo / mock key checked in for that file, and overwriting the file with the real key as part of the build process.

src/io/flutter/view/FlutterView.java
```diff
@@ -201,58 +196,12 @@ private void addInspectorViewContent(FlutterApp app, @Nullable InspectorService
       tabName = device.getUniqueName(existingDevices);
     }
 
-    final Content content = contentManager.getFactory().createContent(null, tabName, false);
-    tabContainer.add(runnerTabs.getComponent(), BorderLayout.CENTER);
-    content.setComponent(tabContainer);
-    content.putUserData(ToolWindow.SHOW_CONTENT_ICON, Boolean.TRUE);
-    content.setIcon(FlutterIcons.Phone);
-    contentManager.addContent(content);
+    DevToolsManager.getInstance(app.getProject()).openBrowserIntoPanel(browserUrl, contentManager, tabName);
```
devoncarew, line 199: How confident are we with the process here (downloading, jsbrowser, ...)? We should probably do something like have an experimental setting in our preferences, and let people opt into trying this out for release or two. That would allow us to iron out any issues.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -217,6 +229,36 @@ private boolean isBazel(Project project) {
...
+
+    if (devToolsInstance != null) {
+      devToolsInstance.openPanel(uri, contentManager, tabName);
+    }
+    else {
+      @Nullable final OSProcessHandler handler =
+        isBazel(project) ? getProcessHandlerForBazel() : getProcessHandlerForPub();
+
+      if (handler != null) {
+        // TODO(devoncarew) Add a Task.Backgroundable here; "Starting devtools..."
```
jacob314, line 243: nit: todos should use the name of the cl author.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -311,6 +353,31 @@ public void openBrowserAndConnect(String serviceProtocolUri, String page) {
...
+    String address = DevToolsUtils.generateDevToolsUrl(devtoolsHost, devtoolsPort, serviceProtocolUri, null) + "&embed=true&page=inspector";
+
+    EngineOptions options =
+      EngineOptions.newBuilder(HARDWARE_ACCELERATED).build();
+    Engine engine = Engine.newInstance(options);
+    Browser browser = engine.newBrowser();
+
+    final Content content = contentManager.getFactory().createContent(null, tabName, false);
+
+    SwingUtilities.invokeLater(() -> {
```
jacob314, line 367: why the invoke later? I would have thought this code was already running on the ui thread.

src/io/flutter/view/FlutterView.java
helin24: We're going to start with these features disabled, since there are some other pieces that would be too much to add in this change. I think it would make sense to add a setting to opt in/out, especially since we will need a permanent way to opt out.

src/io/flutter/devtools/DevToolsManager.java
helin24: I'm not sure how to tell, actually. This is called from a process listening for devtools to start, is that on a separate thread?

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+    return manager;
+  }
+
+  public boolean isInstalled() {
+    return status.get().equals(JxBrowserStatus.INSTALLED);
+  }
+
+  public void setUp(Project project) {
+    if (!ENABLE_JXBROWSER) {
```
jacob314, line 63: Should be ENABLE_JX_BROWSER not ENABLE_JXBROWSER. Also move this check to outside this class to where setup is called. That way it is obvious by looking at that code that we aren't doing anything unless the flag is set.

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,209 @@
...
+      // If already in progress, let calling point wait until success or failure (it may make sense to call setUp but proceed).
+      // If already succeeded or failed, no need to continue.
+      return;
+    }
+
+    LOG.info(project.getName() + ": Installing JxBrowser");
+
+    final File directory = new File(DOWNLOAD_PATH);
+    if (!directory.exists()) {
+      //noinspection ResultOfMethodCallIgnored
```
jacob314, line 78: log something if making the directory fails and abort.

src/io/flutter/devtools/DevToolsManager.java
jacob314: an easy way to tell is to set a breakpoint and see what thread you are on or add an assert that you are on the UI thread.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -311,6 +353,31 @@ public void openBrowserAndConnect(String serviceProtocolUri, String page) {
       null
     );
   }
+
+  public void openPanel(String serviceProtocolUri, ContentManager contentManager, String tabName) {
+    String address = DevToolsUtils.generateDevToolsUrl(devtoolsHost, devtoolsPort, serviceProtocolUri, null) + "&embed=true&page=inspector";
+
+    EngineOptions options =
+      EngineOptions.newBuilder(HARDWARE_ACCELERATED).build();
+    Engine engine = Engine.newInstance(options);
```
jacob314, line 360: Are these blocking calls that take over 10ms to initialize the Engine + browser? If so this work should be done on a background thread.

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,198 @@
...
+
+    if (!created) {
+      // This means another IDE already created this file and has started downloading.
+      // Wait for download to finish and then try loading again.
+      LOG.info(project.getName() + ": Waiting for JxBrowser file to download");
+      int attempts = 0;
+      while (attempts < 100) {
+        if (tempLoadingFile.exists()) {
+          try {
+            Thread.sleep(1000);
```
jacob314, line 118: you should use exponential backoff and we should display a user visible message if the download fails.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -311,6 +351,31 @@ public void openBrowserAndConnect(String serviceProtocolUri, String page) {
       null
     );
   }
+
+  public void openPanel(String serviceProtocolUri, ContentManager contentManager, String tabName) {
+    String address = DevToolsUtils.generateDevToolsUrl(devtoolsHost, devtoolsPort, serviceProtocolUri, null) + "&embed=true&page=inspector";
```
jacob314, line 356: nit: address -> url

jacob314, line 356: update DevToolsUtils to accept passing in embed=true and pass in the tabName instead of hardcoding to the inspector.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -311,6 +351,31 @@ public void openBrowserAndConnect(String serviceProtocolUri, String page) {
...
+
+    SwingUtilities.invokeLater(() -> {
+      // Creating Swing component for rendering web content
+      // loaded in the given Browser instance.
+      BrowserView view = BrowserView.newInstance(browser);
+      view.setPreferredSize(new Dimension(contentManager.getComponent().getWidth(), contentManager.getComponent().getHeight()));
+
+      content.setComponent(view);
+      content.putUserData(ToolWindow.SHOW_CONTENT_ICON, Boolean.TRUE);
+      content.setIcon(FlutterIcons.Phone);
```
jacob314, line 373: add a todo: we should use a different icon based on the tab and we should copy the icons from the inspector toolbar rather than using the Phone icon.

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,209 @@
...
+  private static final String DOWNLOAD_PATH = FileUtils.platformPath();
+  private static final AtomicReference<JxBrowserStatus> status = new AtomicReference<>(JxBrowserStatus.NOT_INSTALLED);
+  private static final Logger LOG = Logger.getInstance(JxBrowserManager.class);
+  // We will be gating JxBrowser features until all of the features are landed.
+  // To test JxBrowser, set this to true and also add license key to VM options (-Djxbrowser.license.key=<key>).
+  public static final boolean ENABLE_JX_BROWSER = false;
+
+  private JxBrowserManager() {}
+
+  public static JxBrowserManager get() {
```
jacob314, line 52: Nit: for consistency with singletons in the rest of the code base call this getInstance() instead of get().

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,209 @@
...
+    catch (IOException e) {
+      e.printStackTrace();
+    }
+
+    if (!created) {
+      // This means another IDE already created this file and has started downloading.
+      // Wait for download to finish and then try loading again.
+      LOG.info(project.getName() + ": Waiting for JxBrowser file to download");
+      int attempts = 0;
+      while (attempts < 100) {
```
jacob314, line 115: make 100 a constant and consider removing if using exponential backoff.

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,209 @@
...
+    // Check for file or loading file
+    final String[] fileNames = {JxBrowserUtils.getPlatformFileName(), JxBrowserUtils.getApiFileName(), JxBrowserUtils.getSwingFileName()};
+    boolean allDownloaded = true;
+    final List<File> files = new ArrayList<>();
+    for (String fileName : fileNames) {
+      final File file = new File(DOWNLOAD_PATH + File.separator + fileName);
+      files.add(file);
+      if (!file.exists()) {
+        allDownloaded = false;
+        break;
```
jacob314, line 91: this break seems wrong. The list of `files` will sometimes be missing a file.

src/io/flutter/devtools/DevToolsManager.java
helin24: `generateDevToolsUrl` has a `page` param but is this now inactive? e.g. the method can produce this URL (from a test): `"http://127.0.0.1:9100/?ide=Android-Studio&uri=http%3A%2F%2F127.0.0.1%3A50224%2FWTFTYus3IPU%3D%2F#timeline"` but adding `#timeline` doesn't seem to do anything when I try in devtools.

src/io/flutter/devtools/DevToolsManager.java
helin24: Actually, I'm still confused about this - should I put these lines into `invokeLater` as well? Or is it fine to remove `invokeLater` because all of this is on a non-main thread (verified that this is the case).

src/io/flutter/jxbrowser/JxBrowserManager.java
helin24: Wow good catch. Could be helpful to add unit tests in a next change.

src/io/flutter/jxbrowser/JxBrowserManager.java
helin24: Do you mean keep increasing the sleep time with more iterations? This would still require some stopping condition, you're thinking something other than number of attempts?

jacob314: yes. what you would do is double the time to sleep each time.
rather than having a stopping condition, what would be ideal would be to have the code that displays the browser check whether the browser is available yet and if it is not available display the following:
1. "Installing JX Browser" (for the common case where JX Browser has not failed to download)
2. JX Browser Download failed (retrying in X seconds) for the case where the download has failed at least once.
This is similar to what you see if you open gmail when offline. Bonus points: add a "retry download now" button.
That way users only see the message about a failed download when they care when they are about to try to use the inspector and it isn't available.

src/io/flutter/devtools/DevToolsManager.java
jacob314: Which thread is it on specifically? Even if it isn't the main thread it could be bad to block the thread for a couple seconds waiting for the engine to load.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -307,10 +347,36 @@ public void processTerminated(@NotNull ProcessEvent event) {
...
+      EngineOptions.newBuilder(HARDWARE_ACCELERATED).build();
+    Engine engine = Engine.newInstance(options);
+    Browser browser = engine.newBrowser();
+
+    final Content content = contentManager.getFactory().createContent(null, tabName, false);
+
+    SwingUtilities.invokeLater(() -> {
+      // Creating Swing component for rendering web content
+      // loaded in the given Browser instance.
+      BrowserView view = BrowserView.newInstance(browser);
```
jacob314, line 368: because it could have been some time from when the action was requested until we are ready to launch, we need to make sure the contentManager has not been disposed or perhaps that the project is still valid.

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,227 @@
...
+      created = tempLoadingFile.createNewFile();
+    }
+    catch (IOException e) {
+      e.printStackTrace();
+    }
+
+    if (!created) {
+      // This means another IDE already created this file and has started downloading.
+      // Wait for download to finish and then try loading again.
+      LOG.info(project.getName() + ": Waiting for JxBrowser file to download");
```
jacob314, line 117: what if the other ide was killed while the file was being downloaded? Now JXBrowser will never work. We need a generally more robust way to do this or we will get a lot of hard to track down bugs. I would suggest we err on the side of a simpler solution than a more efficient one. It is preferable if it wasn't possible for multiple IDEs to be trying to download files to the same locations.

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,227 @@
...
+          }
+
+          tempLoadingFile.delete();
+          loadClasses(files);
+        }
+        catch (IOException e) {
+          LOG.info(project.getName() + ": JxBrowser file downloaded failed: " + currentFileName);
+          e.printStackTrace();
+          status.set(JxBrowserStatus.INSTALLATION_FAILED);
+          tempLoadingFile.delete();
```
jacob314, line 201: keep in mind if the IDE is killed you may not get a chance to delete this file. Likely a file lock that is released if the IDE is killed or similar would be more robust.

src/io/flutter/jxbrowser/JxBrowserManager.java
helin24: If we don't have to manage downloads among different IDEs then we don't need this check.

src/io/flutter/jxbrowser/JxBrowserManager.java
helin24: Removing this since we're going to do a download per IDE

src/io/flutter/jxbrowser/JxBrowserManager.java
helin24: Removing the `tempLoadingFile` since we will only try once on initial startup of an IDE. We can let users manually try another download if they open up a devtools panel.

src/io/flutter/jxbrowser/JxBrowserManager.java
helin24: Will be adding panel messages in a next change.

src/io/flutter/view/FlutterView.java
```diff
@@ -181,6 +182,33 @@ private PerAppState getOrCreateStateForApp(FlutterApp app) {
...
+      }
+      tabName = device.getUniqueName(existingDevices);
+    }
+
+    if (emptyContent != null) {
+      contentManager.removeContent(emptyContent, true);
+      emptyContent = null;
+    }
+
+    // TODO(helin24): Add message to panel to communicate that opening devtools is in progress.
```
jacob314, line 206: nit:
Add message to panel to communicate --> add a message to communicate

src/io/flutter/view/FlutterView.java
```diff
@@ -373,6 +401,37 @@ private void debugActiveHelper(FlutterApp app, @Nullable InspectorService inspec
...
 
+    if (JxBrowserManager.getInstance().isInstalled()) {
+      final DevToolsManager devToolsManager = DevToolsManager.getInstance(app.getProject());
+
+      if (devToolsManager.hasInstalledDevTools()) {
+        addBrowserInspectorViewContent(app, inspectorService, toolWindow);
+      }
+      else {
+        final CompletableFuture<Boolean> result = devToolsManager.installDevTools();
+        result.thenAccept(o -> addBrowserInspectorViewContent(app, inspectorService, toolWindow));
```
jacob314, line 412: do we need to handle if devtools failed to install?

src/io/flutter/view/FlutterView.java
```diff
@@ -373,6 +401,37 @@ private void debugActiveHelper(FlutterApp app, @Nullable InspectorService inspec
...
+      }
+
+      onAppChanged(app);
+
+      app.addStateListener(new FlutterApp.FlutterAppListener() {
```
jacob314, line 417: why is this code needed here? this seems copied from somewhere and not clearly relevant.

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    if (allDownloaded) {
+      LOG.info(project.getName() + ": JxBrowser platform file already exists, skipping download");
+      loadClasses(files);
+      return;
+    }
+
+    // Delete any already existing files.
+    // TODO: Handle if files cannot be deleted.
```
jacob314, line 101: add your username to the TODO

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -0,0 +1,173 @@
...
+      for (File file : files) {
+        final URL url = file.toURI().toURL();
+        classLoader.addURL(url);
+        LOG.info("Loaded JxBrowser file successfully: " + url.toString());
+      }
+
+      status.set(JxBrowserStatus.INSTALLED);
+    }
+    catch (Exception e) {
+      LOG.info("Failed to load JxBrowser files");
```
jacob314, line 168: what exception are you expecting? should we delete the cached browser files in this case?

src/io/flutter/utils/JxBrowserUtils.java
```diff
@@ -0,0 +1,51 @@
...
+      if (SystemInfo.is32Bit) {
+        name = "win32";
+      } else if (SystemInfo.is64Bit) {
+        name = "win64";
+      }
+    } else if (SystemInfo.isLinux && SystemInfo.is64Bit) {
+      name = "linux64";
+    }
+
+    if (name.isEmpty()) {
```
jacob314, line 29: are we catching this exception? It would be nice to just fail gracefully if a platform we don't support is used.

src/io/flutter/utils/JxBrowserUtils.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public static String getApiFileName() {
+    return String.format("%s-%s.%s", JXBROWSER_FILE_PREFIX, JXBROWSER_FILE_VERSION, JXBROWSER_FILE_SUFFIX);
+  }
+
+  public static String getSwingFileName() {
+    return String.format("%s-swing-%s.%s", JXBROWSER_FILE_PREFIX, JXBROWSER_FILE_VERSION, JXBROWSER_FILE_SUFFIX);
+  }
+
+  public static String getLoadingFileName() {
```
jacob314, line 44: should this be removed?

src/io/flutter/jxbrowser/JxBrowserManager.java
helin24: `toURL` can throw a malformed URL exception. I'm not sure how to handle this case. We could delete the files, but if this occurred because of a bug in our file naming, then it would be annoying for the user to have to keep downloading the files to a bad spot each time they opened their IDE.

src/io/flutter/utils/JxBrowserUtils.java
helin24: Ah true. Changed this to a checked exception.

src/io/flutter/view/FlutterView.java
helin24: Yes, what would make sense here? I could pull the code below that starts the old inspector into a separate function and run that instead. Or maybe just display a message?

PR #4671, opened 2020-06-28
Split icon generation into two scripts + use single set of downloaded files

DanTup:
This splits the `update_colors.dart` script into two scripts so it no longer calls itself (which simplifies debugging, and avoids getting into the situation where a broken generated file prevents running the script to regenerate it) and also consolidates the downloaded files so there's just one set.

Running `grind colors` works properly this time, and generates the same files that were in the repo - with the exception of the downloaded files being unformatted (I didn't commit the format changes because it would make this PR harder to review - I can add them once this is reviewed, or just send another PR with just the (un)formatting changes).

@devoncarew the rebase was messy, so I manually applied your changes to my existing change. If you think I've missed anything, let me know!

### DIFF
--- a/tool/colors/flutter/colors_cupertino.dart
+++ b/tool/colors/flutter/colors_cupertino.dart
@@ -1,14 +1,16 @@
+/// This file was downloaded by update_colors.dart.
+
 // Copyright 2014 The Flutter Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
 // @dart = 2.8
 
+import '../stubs.dart';
+
 // Examples can assume:
 // Widget child;
 // BuildContext context;
-
-import '../stubs.dart';
 
 /// A palette of [Color] constants that describe colors commonly used when
 /// matching the iOS platform aesthetics.
--- a/tool/colors/flutter/colors_cupertino.dart
+++ b/tool/colors/flutter/colors_cupertino.dart
@@ -79,8 +79,7 @@ class CupertinoColors {
   /// Not the same grey as disabled buttons etc.
   ///
   /// This is the disabled color in the iOS palette.
-  static const CupertinoDynamicColor inactiveGray =
-      CupertinoDynamicColor.withBrightness(
+  static const CupertinoDynamicColor inactiveGray = CupertinoDynamicColor.withBrightness(
     debugLabel: 'inactiveGray',
     color: Color(0xFF999999),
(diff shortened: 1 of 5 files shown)


tool/colors/flutter/colors_cupertino.dart
```diff
@@ -1,15 +1,17 @@
+/// This file was downloaded by update_colors.dart.
```
devoncarew, line 1: nit: `//`, not `///`

tool/colors/flutter/colors_material.dart
```diff
@@ -1,3 +1,5 @@
+/// This file was downloaded by update_colors.dart.
```
devoncarew, line 1: should probably use a double slash comment

tool/colors/update_colors.dart
```diff
@@ -20,62 +23,40 @@ void main(List<String> args) async {
...
-    stderr.write(results.stderr);
-
-    exit(0);
-  }
+  print('Generating dart files:');
+  await generateDartFiles();
 }
 
-Future generateDartFiles() async {
-  // TODO: Use the files from the local flutter checkout instead of a download
```
devoncarew, line 46: This todo is likely worth retaining - I think we're better off avoiding downloads when re-generating these files if we can avoid them.

tool/colors/flutter/colors_cupertino.dart
DanTup: Done!

tool/colors/flutter/colors_material.dart
DanTup: Done!

tool/colors/update_colors.dart
DanTup: Ah yes, I slightly misread it thinking it was about having two downloaded copies. I've added this back.

DanTup:
> Re-formatting before commit sgtm.

It's actually un-formatting... The downloaded files are from Flutter which isn't formatted, but I'd edited them manually at some point which formatted them (this meant anytime you ran the script it would cause a big diff as the format was undone).

So I've added those unformatted (but as they are in Flutter) files. Another option would be to run `dartfmt` on them after downloading. If you're prefer that, let me know and I'll fix in another PR.

I have a few other tweaks I'd like to land here - including sorting the colours for simpler diffs, and outputting a TS-friendly format - if you think it's worth moving to another repo now, if you can create a new repo somewhere appropriate, I can copy the files over and then send the PRs there instead.

PR #4674, opened 2020-06-29
Update build script

stevemessick:
Fix a few problems:
- JxBrowser files were not available to unit tests
- The script wasn't getting updated with the proper IDE and plugin versions
- The result of the external build was not being returned, causing broken bots to be green

### DIFF
--- a/flutter-idea/build.gradle
+++ b/flutter-idea/build.gradle
@@ -33,6 +33,8 @@ intellij {
 dependencies {
   testCompile "org.powermock:powermock-api-mockito2:2.0.0"
   testCompile "org.powermock:powermock-module-junit4:2.0.0"
+  implementation fileTree(dir: "$project.rootDir/lib/jxbrowser",
+                          includes: ["*.jar"])
 }
 
 /* Need to swizzle source and resource paths to match standard or buildPlugin does nothing. */
--- a/tool/plugin/lib/plugin.dart
+++ b/tool/plugin/lib/plugin.dart
@@ -366,10 +366,34 @@ class AntBuildCommand extends BuildCommand {
 }
 
 class GradleBuildCommand extends BuildCommand {
+  static const String propertiesTemplate = """
+org.gradle.parallel=true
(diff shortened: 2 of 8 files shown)


tool/plugin/lib/plugin.dart
```diff
@@ -387,9 +411,9 @@ class GradleBuildCommand extends BuildCommand {
       return await super.doit();
     } finally {
       if (Platform.isWindows) {
-        return await exec('.\\gradlew.bat', ['--stop']);
+        await exec('.\\gradlew.bat', ['--stop']);
```
jacob314, line 414: wow. should we have a lint about return statements in finally clauses? @pq @munificent 
Determining which return statement should actually apply in this case is a bit confusing.

stevemessick: I like that idea. I can't imagine why I typed "return" in the first place but I certainly didn't want it. Note the final statement in the protected block is a return, but you might not be able to determine that all the time.

Thanks for the review! Looks like I have a problem in the build script, which I'll fix.

stevemessick:
I had to make so many changes that another review is probably warranted. However, I want to get the patch out.

TBR

flutter-idea/build.gradle
```diff
@@ -33,6 +33,8 @@ intellij {
 dependencies {
   testCompile "org.powermock:powermock-api-mockito2:2.0.0"
   testCompile "org.powermock:powermock-module-junit4:2.0.0"
+  implementation fileTree(dir: "$project.rootDir/lib/jxbrowser",
```
helin24, line 36: Is this change leftover from working on jxbrowser stuff? Or is it necessary for dealing with the jxbrowser files that are in your directory but not committed?

stevemessick:
That will make the unit test bot in your PR turn green. Just restart it.

helin24:
> That will make the unit test bot in your PR turn green. Just restart it.

Oh ha! Magic! I'll try it.

PR #4678, opened 2020-07-01
Start paused for run mode and use FlutterTestRunner for run tests

helin24:
We reverted a change (https://github.com/flutter/flutter-intellij/pull/4673) that was similar to this one because it was breaking tests.

However, we still want to start with the app paused to catch early errors since we're now expecting structured errors from flutter_tools. If we do not start paused, there's a chance that the app will start running before flutter_tools has started listening for errors and we will miss early ones.

While looking into why run mode during tests doesn't resume app execution, I found that running tests (NOT in debug mode) uses the runner `DefaultRunProgramRunner` from the IntelliJ platform - it seems like we should be using a flutter-specific test runner for run and debug modes? But I'm not sure whether there's a reason we have been using the default runner.

I tested this change by running the gallery app and running the sample tests of the gallery app.

### DIFF
--- a/src/io/flutter/run/test/FlutterTestRunner.java
+++ b/src/io/flutter/run/test/FlutterTestRunner.java
@@ -17,6 +17,7 @@ import com.intellij.execution.runners.GenericProgramRunner;
 import com.intellij.execution.ui.RunContentDescriptor;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.util.Key;
+import com.intellij.openapi.wm.ToolWindowId;
 import com.intellij.xdebugger.XDebugProcess;
 import com.intellij.xdebugger.XDebugProcessStarter;
 import com.intellij.xdebugger.XDebugSession;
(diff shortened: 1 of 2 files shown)


helin24:
> What about RunMode.PROFILE? Does it already start paused?

I'm not sure, actually. We talked about only needing to add this for run mode, but I don't recall why and it's not clear to me from looking at the plugin and flutter_tools. Can wait for Jacob to return before merging this change.

helin24:
Verified that we don't need to make this change for `RunMode.PROFILE` - flutter_tools ignores the structured errors flag for profile mode anyway, so there's no need to start paused.

panudetjt:
Hello @helin24 
This merged make me cannot only run specify test without pause.
It's very inconvenience while working with test that throw Exception.
Thank you.

panudetjt:
It's have debug button already why do we need duplicate functionality of that button into run button ?

🤔

helin24:
@panudetjt I tested that this change allows for running tests without debugging pauses, but it sounds like you have a case that doesn't work. Can you describe the features of your test and what happens when you run?

panudetjt:
```dart
import 'package:flutter_test/flutter_test.dart';
void main() {
 thisWillThrow() {
    throw UnimplementedError();
  }
  test('should pause', () {
    expect(thisWillThrow(), throwsA(isA<UnimplementedError>()));
  });
}
```
image
when pressed ``Run token_controller_test.dart`` 
it have functionality as ``Debug token_controller_test.dart``
image
when I tested large group it's very inconvenience.

thank you 👍

PR #4680, opened 2020-07-01
Flutter news

stevemessick:
See #4658 for full description.

The action that would display the news panel is currently disabled. Once we get real content we can re-enable it.


src/io/flutter/actions/OpenFlutterNewsSidePanelAction.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    @Override
+    public void stateChanged(@NotNull ToolWindowManager toolWindowManager) {
+      ApplicationManager.getApplication().invokeLater(() -> {
+        if (myProject.isDisposed()) {
+          myProjectToListenerMap.remove(myProject);
+          return;
+        }
+
+        ToolWindow window = ToolWindowManager.getInstance(myProject).getToolWindow(OpenAssistSidePanelAction.TOOL_WINDOW_TITLE);
```
helin24, line 136: Is the argument `toolWindowManager` the same as what's retrieved here?

stevemessick: Not necessarily. The tool window manager is project-specific. This could have been triggered by a different project. It might make sense to check that it is same and do nothing if not.

stevemessick:
Thanks for the review. I'm going to have to move most of these files into the flutter-studio module before committing.

PR #4687, opened 2020-07-06
Simplify regex and sort colors

DanTup:
This slightly simplifies the regex (to avoid needing three for the different color types) since we only need the names (we don't care about what's after the `=` if the type on the left is some `Color` type, and also sorts the colors so that it's easier to se what's added removed (for ex. if the order changes in Flutter or the regexes change).

The real changes are in `update_colors.dart` and the others are generated as a result of that change.


DanTup:
I also have a change that allows writing the colours in JS format (used by Dart-Code):

https://github.com/DanTup/flutter-intellij/commit/19da4c0b946bdb047f84017f2023bf4feb85bc3c

If you think it's worth committing here (or we move this to its own repo and then commit there), let me know (otherwise it's easy for me to just run from that branch if I need to regenerate again).

tool/colors/update_colors.dart
```diff
@@ -66,27 +66,20 @@ Future<void> downloadFile(String url, File file) async {
...
-    ...regexpColor3.allMatches(data).map((Match match) => match.group(1)),
-  ];
+  List<String> names = regexpColor
+      .allMatches(data)
+      .map((Match match) => match.group(1))
+      .toList();
 
   // Remove any duplicates.
-  return Set<String>.from(names).toList();
+  return Set<String>.from(names).toList()..sort();
```
devoncarew, line 82: 👍 for sorting

DanTup:
> If you need additional metadata files to be written, I'd contribute the generation code here (and hopefully try to make the new metadata files ~reasonably general).

It's fairly generic, though it outputs a TypeScript file and includes type definitions, so I'll tweak it to just write simple JS objects (with `.json` extension), and I can just paste the contents into the TS for Dart-Code. 

> I think that's either a new repo - something like flutter/framework-metadata - or using an existing repo, like https://github.com/flutter/assets-for-api-docs

Yeah, the latter sounds like it's quite specific for the API docs online, so it might be better to keep this separate? (unless you mean rename that one, though I'm not sure if that will affect/break anyone/anything). I'll go with whatever you think is best (if a new repo, you might need to create it) and I can move the code over to it and remove from here in a future PR.

Thanks!

PR #4688, opened 2020-07-06
Build issues

stevemessick:
There were some issues hidden by the broken build script.
- Bad package name
- Fix `@Override` that causes error in old versions
- Change from instance to static methods in PluginManager


tool/plugin/lib/edit.dart
```diff
@@ -69,6 +69,18 @@ List<EditCommand> editCommands = [
...
     version: '4.1',
   ),
+  Subst(
+    path:
+        'flutter-studio/src/io/flutter/assistant/whatsnew/FlutterNewsBundleCreator.java',
+    initial: """
+    PluginManager pluginManager = PluginManager.getInstance();
+    IdeaPluginDescriptor descriptor = pluginManager.findEnabledPlugin(PluginId.getId("io.flutter"));
+""",
+    replacement: """
```
helin24, line 79: Confused - how does this work when initial and replacement are the same?

stevemessick: There are two hard-to-read differences. The first line is removed and "p" becomes "P" in the second.

helin24: I completely misread, not sure how I missed the difference!

PR #4712, opened 2020-07-22
Add handling and messaging for when JxBrowser is not installed

helin24:
We want to provide some status messages in the inspector panel while downloading JxBrowser files or while starting DevTools.

The paths once a project is run:
- FlutterView::debugActiveHelper
  - JxBrowser is not enabled -> skip to normal inspector
  - JxBrowser is installed -> `handleJxBrowserInstalled` (show that devtools is starting)
  - JxBrowser installation failed -> `handleJxBrowserInstallationFailed` (offer user option to retry installation)
  - JxBrowser installation is in progress -> `handleJxBrowserInstallationInProgress` (wait for installation)
    - Success -> `handleJxBrowserInstalled`
    - Failure -> `handleJxBrowserInstallationFailed`
    - Time out -> Show error message

Messages will be shown in the panel, e.g.:
Screen Shot 2020-07-22 at 11 10 15 AM

### DIFF
--- a/src/io/flutter/jxbrowser/JxBrowserManager.java
+++ b/src/io/flutter/jxbrowser/JxBrowserManager.java
@@ -81,6 +81,11 @@ public class JxBrowserManager {
     setUp(project);
   }
 
+  private void setStatusFailed() {
+    status.set(JxBrowserStatus.INSTALLATION_FAILED);
+    installation.complete(JxBrowserStatus.INSTALLATION_FAILED);
+  }
+
   public void setUp(Project project) {
     if (!status.compareAndSet(JxBrowserStatus.NOT_INSTALLED, JxBrowserStatus.INSTALLATION_IN_PROGRESS)) {
       // This check ensures that an IDE only downloads and installs JxBrowser once, even if multiple projects are open.
@@ -96,8 +101,7 @@ public class JxBrowserManager {
     if (!directory.exists()) {
       if (!directory.mkdirs()) {
         LOG.info(project.getName() + ": Unable to create directory for JxBrowser files");
-        status.set(JxBrowserStatus.INSTALLATION_FAILED);
-        installation.complete(JxBrowserStatus.INSTALLATION_FAILED);
+        setStatusFailed();
         return;
       }
     }
@@ -109,8 +113,7 @@ public class JxBrowserManager {
     catch (FileNotFoundException e) {
       LOG.info(project.getName() + ": Unable to find JxBrowser platform file");
       e.printStackTrace();
-      status.set(JxBrowserStatus.INSTALLATION_FAILED);
-      installation.complete(JxBrowserStatus.INSTALLATION_FAILED);
+      setStatusFailed();
       return;
     }
 
@@ -179,8 +182,7 @@ public class JxBrowserManager {
         catch (IOException e) {
           LOG.info(project.getName() + ": JxBrowser file downloaded failed: " + currentFileName);
           e.printStackTrace();
-          status.set(JxBrowserStatus.INSTALLATION_FAILED);
-          installation.complete(JxBrowserStatus.INSTALLATION_FAILED);
+          setStatusFailed();
         }
       }
     };
@@ -203,8 +205,7 @@ public class JxBrowserManager {
     }
     catch (MalformedURLException e) {
       LOG.info("Failed to load JxBrowser files");
-      status.set(JxBrowserStatus.INSTALLATION_FAILED);
-      installation.complete(JxBrowserStatus.INSTALLATION_FAILED);
+      setStatusFailed();
     }
   }
 }
--- a/src/io/flutter/jxbrowser/JxBrowserManager.java
+++ b/src/io/flutter/jxbrowser/JxBrowserManager.java
@@ -60,11 +60,10 @@ public class JxBrowserManager {
     return status.get();
   }
 
+  /**
+   * Call {@link #setUp} before this function to ensure that an installation has started.
+   */
   public JxBrowserStatus waitForInstallation(int seconds) throws TimeoutException {
-    if (installation == null) {
-      return null;
-    }
-
     try {
       return installation.get(seconds, TimeUnit.SECONDS);
     }
--- a/src/io/flutter/jxbrowser/JxBrowserManager.java
+++ b/src/io/flutter/jxbrowser/JxBrowserManager.java
@@ -55,6 +55,13 @@ public class JxBrowserManager {
     return status.get();
   }
 
+  public void retryFromFailed(Project project) {
+    if (!status.compareAndSet(JxBrowserStatus.INSTALLATION_FAILED, JxBrowserStatus.NOT_INSTALLED)) {
+      return;
+    }
+    setUp(project);
+  }
+
   public void setUp(Project project) {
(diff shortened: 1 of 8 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -373,6 +383,109 @@ public void debugActive(@NotNull FlutterViewMessages.FlutterDebugEvent event) {
...
+      waitForJxBrowserInstallation(app, inspectorService, toolWindow);
+    });
+  }
+
+  protected void waitForJxBrowserInstallation(FlutterApp app, InspectorService inspectorService, ToolWindow toolWindow) {
+    final JxBrowserManager jxBrowserManager = JxBrowserManager.getInstance();
+    // Periodically check for download to finish.
+    // TODO(helin24): Maybe set up JxBrowserManager as listener instead.
+    int waitSeconds = 1;
+    while (waitSeconds <= 128) {
```
jacob314, line 434: you should listen for an event from JxBrowserManager indicating that installation succeeded or failed rather than sleeping.

src/io/flutter/view/FlutterView.java
```diff
@@ -373,6 +383,109 @@ public void debugActive(@NotNull FlutterViewMessages.FlutterDebugEvent event) {
...
+  }
+
+  protected void presentLabel(ToolWindow toolWindow, String text) {
+    final JBLabel label = new JBLabel(text, SwingConstants.CENTER);
+    label.setForeground(UIUtil.getLabelDisabledForeground());
+    replacePanelLabel(toolWindow, label);
+  }
+
+  protected void presentClickableLabel(ToolWindow toolWindow, String text, MouseAdapter mouseAdapter) {
+    final JBLabel label = new JBLabel(text, SwingConstants.CENTER);
```
jacob314, line 470: I believe there is a JBLink or similar class that is already styled to look like a link. I think the performance tab sohws an example of using it.

src/io/flutter/view/FlutterView.java
```diff
@@ -373,6 +383,109 @@ public void debugActive(@NotNull FlutterViewMessages.FlutterDebugEvent event) {
...
+  }
+
+  protected void handleJxBrowserInstallationFailed(FlutterApp app, InspectorService inspectorService, ToolWindow toolWindow) {
+    // Allow user to manually restart.
+    presentClickableLabel(toolWindow, "JxBrowser installation failed. Click to retry.", new MouseAdapter() {
+      @Override
+      public void mouseClicked(MouseEvent e) {
+        JxBrowserManager.getInstance().retryFromFailed(app.getProject());
+        handleJxBrowserInstallationInProgress(app, inspectorService, toolWindow);
+
```
jacob314, line 458: nit: remove whitespace line.

src/io/flutter/view/FlutterView.java
jacob314: you could also use CompletableFuture completing the future when JxBrowserManager is ready.

src/io/flutter/utils/ThreadUtil.java
```diff
@@ -61,4 +61,14 @@ public static ThreadGroup getRootThreadGroup() {
 
     return result;
   }
+
+  public static boolean sleep(int seconds) {
```
jacob314, line 65: Would be good to avoid using sleep outside of tests. Remove this method.

src/io/flutter/view/FlutterView.java
```diff
@@ -89,6 +89,7 @@ public void dispose() {
   protected static final String INSTALLATION_TIMED_OUT_LABEL = "Waiting for JxBrowser installation timed out. Restart your IDE to try again.";
   protected static final String INSTALLING_DEVTOOLS_LABEL = "Installing DevTools...";
   protected static final String DEVTOOLS_FAILED_LABEL = "Setting up DevTools failed.";
+  protected static final int INSTALLATION_WAIT_LIMIT_SECONDS = 200;
```
jacob314, line 92: is there a reason to not make this limit 10X higher? Users may have limited bandwidth.

helin24: I have no opinion on this and chose a random number. This doesn't end the download attempt, only the completable future, but since we display a message suggesting a restart we should probably just make this bigger.

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -55,28 +60,48 @@ public JxBrowserStatus getStatus() {
...
+  }
+
   public void setUp(Project project) {
     if (!status.compareAndSet(JxBrowserStatus.NOT_INSTALLED, JxBrowserStatus.INSTALLATION_IN_PROGRESS)) {
       // This check ensures that an IDE only downloads and installs JxBrowser once, even if multiple projects are open.
       // If already in progress, let calling point wait until success or failure (it may make sense to call setUp but proceed).
       // If already succeeded or failed, no need to continue.
       return;
     }
+    installation = new CompletableFuture<>();
```
jacob314, line 96: This logic is only safe if waitForInstallation is called after setUp is called.
Here's a way to tweak it to make it a bit more robust.
Only create a new installation CompletableFuture if the previous future is already completed (indicates a previous failed install). Document he assumptions about the completed future.

src/io/flutter/view/FlutterView.java
```diff
@@ -414,36 +415,23 @@ protected void handleJxBrowserInstallationInProgress(FlutterApp app, InspectorSe
     if (jxBrowserManager.getStatus().equals(JxBrowserStatus.INSTALLED)) {
       handleJxBrowserInstalled(app, inspectorService, toolWindow);
     } else {
-      startJxBrowserInstallationWaitingThread(app, inspectorService, toolWindow);
-    }
-  }
+      try {
+        final JxBrowserStatus newStatus = jxBrowserManager.waitForInstallation(INSTALLATION_WAIT_LIMIT_SECONDS);
```
jacob314, line 419: why has the executeOnPolledThread call gone away?
The CompletableFuture code as written is still blocking but you could switch to the async version if you don't want to block.

src/io/flutter/jxbrowser/JxBrowserManager.java
helin24: Is my change and comment here accurate?

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -58,8 +56,35 @@ public static JxBrowserManager getInstance() {
...
   }
 
-  public boolean isInstalled() {
-    return status.get().equals(JxBrowserStatus.INSTALLED);
+  public JxBrowserStatus getStatus() {
+    return status.get();
+  }
+
+  public JxBrowserStatus waitForInstallation(int seconds) throws TimeoutException {
+    if (installation == null) {
```
jacob314, line 64: installation is never null. you could annotate it as non nullable.

src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -58,8 +56,35 @@ public static JxBrowserManager getInstance() {
     return manager;
   }
 
-  public boolean isInstalled() {
-    return status.get().equals(JxBrowserStatus.INSTALLED);
+  public JxBrowserStatus getStatus() {
+    return status.get();
+  }
+
+  public JxBrowserStatus waitForInstallation(int seconds) throws TimeoutException {
```
jacob314, line 63: nit: comment that you must call setup as well as calling this message or it will time out.

PR #4719, opened 2020-07-27
Use off screen rendering for jxbrowser on mac

helin24:
There's currently a bug that causes a zoom effect on resize using hardware accelerated render mode on mac. Off screen rendering performance is slower (fewer frames per second, see metrics [here](https://jxbrowser-support.teamdev.com/docs/guides/browser-view.html#rendering)) but fixes this problem.

Before:
Jul-27-2020 10-06-03

After:
Jul-27-2020 09-28-29

### DIFF
--- a/src/io/flutter/jxbrowser/EmbeddedBrowser.java
+++ b/src/io/flutter/jxbrowser/EmbeddedBrowser.java
@@ -5,25 +5,29 @@
  */
 package io.flutter.jxbrowser;
 
+import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.wm.ToolWindow;
 import com.intellij.ui.content.Content;
 import com.intellij.ui.content.ContentManager;
 import com.teamdev.jxbrowser.browser.Browser;
 import com.teamdev.jxbrowser.engine.Engine;
 import com.teamdev.jxbrowser.engine.EngineOptions;
+import com.teamdev.jxbrowser.engine.RenderingMode;
 import com.teamdev.jxbrowser.view.swing.BrowserView;
 import icons.FlutterIcons;
 
 import java.awt.*;
 
 import static com.teamdev.jxbrowser.engine.RenderingMode.HARDWARE_ACCELERATED;
+import static com.teamdev.jxbrowser.engine.RenderingMode.OFF_SCREEN;
 
 public class EmbeddedBrowser {
   public void openPanel(ContentManager contentManager, String tabName, String url) {
-    EngineOptions options =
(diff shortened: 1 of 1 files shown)


helin24:
Bots seem unable to build flutter tools for tests... seems unrelated but I'm also not sure how to fix.

devoncarew:
I'm not sure why it's failing:

> Error: Unable to 'pub upgrade' flutter tool. Retrying in five seconds... (9 tries left)
> Resolving dependencies... 
> The current Dart SDK version is 2.10.0-0.0.dev.flutter-a3815b6590.

Perhaps trying to repro locally? Normally I'd say trying to move the bots to using the dev channel of flutter (if the regression is on master), but from the above we might already be running the bots against dev.

src/io/flutter/jxbrowser/EmbeddedBrowser.java
```diff
@@ -5,25 +5,29 @@
...
 import static com.teamdev.jxbrowser.engine.RenderingMode.HARDWARE_ACCELERATED;
+import static com.teamdev.jxbrowser.engine.RenderingMode.OFF_SCREEN;
 
 public class EmbeddedBrowser {
   public void openPanel(ContentManager contentManager, String tabName, String url) {
-    EngineOptions options =
-      EngineOptions.newBuilder(HARDWARE_ACCELERATED).build();
-    Engine engine = Engine.newInstance(options);
-    Browser browser = engine.newBrowser();
+    final RenderingMode renderingMode = SystemInfo.isMac ? OFF_SCREEN : HARDWARE_ACCELERATED;
```
jacob314, line 26: Add a TODO with a bug # to switch back to HARDWARE_ACCELERATED rendering once the JxBrowser bug is fixed upstream.
Also, how is the frame rate with OFF_SCREEN rendering? DevTools uses WebGL heavily so there is a risk OFF_SCREEN rendering is too slow to enable as the default on Mac.

helin24: The OFF_SCREEN frame rate is estimated at 23 FPS, compared to 60 FPS for HARDWARE_ACCELERATED. Is there a frame rate that WebGL requires?

jacob314:
Lets not land that if it is only hitting 23fps instead of 60fps. Users can deal with a little bit more flicker when they resize.
What tool are you using to measure the frame rate?

helin24:
Fair enough - I'm not measuring the frame rate, just taking that number from JxBrowser's documentation of the two render modes. Is there a way to measure frame rate myself?

PR #4728, opened 2020-07-29
Show notification for when devtools build is slow

helin24:
Jul-29-2020 10-12-20

### DIFF
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -16,6 +16,9 @@ import com.intellij.execution.process.ProcessAdapter;
 import com.intellij.execution.process.ProcessEvent;
 import com.intellij.execution.process.ProcessOutput;
 import com.intellij.ide.browsers.BrowserLauncher;
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationType;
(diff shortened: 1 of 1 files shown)


src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -317,6 +335,7 @@ public void onTextAvailable(@NotNull ProcessEvent event, @NotNull Key outputType
           catch (JsonSyntaxException e) {
             processHandler.destroyProcess();
           }
+          notification.expire();
```
jacob314, line 338: glad to see the notification is being closed as soon as devtools is open!

jacob314: what happens if notification.expire() is called before the notification is shown?
Will that call throw or will the subsequent call to Notification.Bus.notify show the notification?

jacob314:
Discussed offline. 
lgtmlgtm

src/io/flutter/devtools/DevToolsManager.java
helin24: I tested by putting an extra `notification.expire()` statement before calling `Notification.Bus.notify` and verified that no notification is shown in that situation. Also verified that external projects where this step takes negligible time does not display the notification.

PR #4733, opened 2020-08-03
Use daemon API to get devtools host and port for internal use

helin24:
Instead of building and serving devtools independently, we want to make a request to flutter daemon to start serving devtools which should already be built.

This change requires https://github.com/flutter/flutter/pull/62702 to be merged internally.

### DIFF
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -25,6 +25,7 @@ import com.intellij.openapi.progress.ProgressManager;
 import com.intellij.openapi.progress.Task;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.Key;
+import com.intellij.openapi.util.Pair;
 import com.intellij.ui.content.ContentManager;
 import io.flutter.FlutterMessages;
 import io.flutter.bazel.Workspace;
@@ -38,6 +39,7 @@ import io.flutter.utils.JsonUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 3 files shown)


src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -259,26 +260,47 @@ private void openBrowserImpl(String uri, String screen) {
       devToolsInstance.openBrowserAndConnect(uri, screen);
     }
     else {
-      @Nullable final OSProcessHandler handler =
-        isBazel(project) ? getProcessHandlerForBazel() : getProcessHandlerForPub();
+      if (isBazel(project)) {
```
jacob314, line 263: comment why we do this for Bazel only for now (non-bazel projects currently have a flutter daemon with an older version of devtools).

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -259,26 +260,49 @@ private void openBrowserImpl(String uri, String screen) {
...
-        DevToolsInstance.startServer(handler, instance -> {
-          devToolsInstance = instance;
+      // For internal users, we can connect to the DevTools server started by flutter daemon. For external users, the flutter daemon has an
+      // older version of DevTools, so we launch the server using `pub global run` instead.
+      if (isBazel(project)) {
+        final Optional<FlutterApp> first =
+          FlutterApp.allFromProjectProcess(project).stream().filter((FlutterApp app) -> app.getProject() == project).findFirst();
+
+        if (!first.isPresent()) {
+          return;
```
jacob314, line 270: should we report an error here? We weren't able to open devtools as the app closed.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -259,26 +260,49 @@ private void openBrowserImpl(String uri, String screen) {
...
+        }
 
-          devToolsInstance.openBrowserAndConnect(uri, screen);
-        }, () -> {
-          // Listen for closing; null out the devToolsInstance.
-          devToolsInstance = null;
+        first.get().serveDevTools().thenAccept((Pair<String, Integer> hostAndPort) -> {
+          if (hostAndPort == null) {
```
jacob314, line 274: at each async boundary we probably need to make sure the project hasn't been closed in whcih case we should quietly cancel the request.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -259,26 +260,49 @@ private void openBrowserImpl(String uri, String screen) {
...
+        startDevToolsServerAndConnect(handler, uri, screen);
       }
-      else {
-        // TODO(devoncarew): We should provide feedback to callers that the open browser call failed.
+    }
+  }
+
+  private void startDevToolsServerAndConnect(OSProcessHandler handler, String uri, String screen) {
+    if (handler != null) {
+      // TODO(devoncarew) Add a Task.Backgroundable here; "Starting devtools..."
```
jacob314, line 291: use your username for new TODOs

src/io/flutter/devtools/DevToolsManager.java
jacob314: one more thing. we also don't do any flutter version checking so this wouldn't be safe externally as stable probably doesn't support this daemon api.

src/io/flutter/run/daemon/DaemonApi.java
```diff
@@ -80,6 +81,10 @@
     return send("app.stop", new AppStop(appId));
   }
 
+  CompletableFuture<Pair<String, Integer>> devToolsServe() {
```
jacob314, line 84: nit: create a little Value class for this instead of using Pair.
With Pair it isn't obvious what the two fields are but it would be really obvious with a simple data class with a `host` and `port` field.

src/io/flutter/devtools/DevToolsManager.java
helin24: This means a check of whether someone has closed down the IntelliJ project? And we probably should also check if the app has been disconnected?

src/io/flutter/devtools/DevToolsManager.java
helin24: I didn't want to put my name here since I just moved this block of code without changes, do you think this is enough of a change where I should take ownership of the TODO?

jacob314: you did the right thing. I saw the lower one was a move but I didn't scroll up high enough to spot this one.

Issue #4692, opened 2020-07-08
Can't run debug while `main.dart` has a `build` diretory sibling


Sunbreak:
If `main.dart` has a `build` directory silbing, while running debug, the plugin complains:
> Error running 'main.dart': 
> Entrypoint isn't within a Flutter pub root

Both IntelliJ Ultimate/Community and Android Studio have the problem. **But VS Code works fine.**

## Steps to Reproduce

https://github.com/Sunbreak/error_intellij_directory

## Version info

`flutter doctor -v`

```
[✓] Flutter (Channel unknown, v1.17.4, on Mac OS X 10.15.4 19E287, locale en-CN)
    • Flutter version 1.17.4 at /Users/sunbreak/flutter/flutter-1.17.4
    • Framework revision 1ad9baa8b9 (3 weeks ago), 2020-06-17 14:41:16 -0700
    • Engine revision ee76268252
    • Dart version 2.8.4

[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at /Users/sunbreak/Library/Android/sdk
    • Platform android-29, build-tools 29.0.3
    • Java binary at: /Users/sunbreak/Library/Application Support/JetBrains/Toolbox/apps/AndroidStudio/ch-0/192.6392135/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b4-5784211)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 11.5)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 11.5, Build version 11E608c
    • CocoaPods version 1.9.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 3.6)
    • Android Studio at /Users/sunbreak/Library/Application Support/JetBrains/Toolbox/apps/AndroidStudio/ch-0/192.6392135/Android Studio.app/Contents
    • Flutter plugin version 46.0.1
    • Dart plugin version 192.8052
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b4-5784211)

[✓] IntelliJ IDEA Ultimate Edition (version 2020.1.2)
[...]

PR #4735, opened 2020-08-04
Check for BUILD file for bazel projects only

helin24:
We removed a check for BUILD to fix https://github.com/flutter/flutter-intellij/issues/4692 but this caused problems for internal projects.

### DIFF
--- a/src/io/flutter/run/MainFile.java
+++ b/src/io/flutter/run/MainFile.java
@@ -20,6 +20,7 @@ import com.jetbrains.lang.dart.psi.DartFile;
 import com.jetbrains.lang.dart.psi.DartImportStatement;
 import com.jetbrains.lang.dart.util.DartResolveUtil;
 import io.flutter.FlutterBundle;
+import io.flutter.bazel.WorkspaceCache;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -120,14 +121,17 @@ public class MainFile {
   @Nullable
   private static VirtualFile findAppDir(@Nullable VirtualFile file, @NotNull Project project) {
     for (VirtualFile candidate = file; inProject(candidate, project); candidate = candidate.getParent()) {
-      if (isAppDir(candidate)) return candidate;
+      if (isAppDir(candidate, project)) return candidate;
     }
     return null;
   }
 
-  private static boolean isAppDir(@NotNull VirtualFile dir) {
+  private static boolean isAppDir(@NotNull VirtualFile dir, @NotNull Project project) {
     return dir.isDirectory() && (
-      dir.findChild("pubspec.yaml") != null || dir.findChild(".packages") != null);
+      dir.findChild("pubspec.yaml") != null ||
+      (WorkspaceCache.getInstance(project).isBazel() && dir.findChild("BUILD") != null) ||
+      dir.findChild(".packages") != null
+    );
   }
 
   private static boolean inProject(@Nullable VirtualFile file, @NotNull Project project) {
--- a/src/io/flutter/run/MainFile.java
+++ b/src/io/flutter/run/MainFile.java
@@ -122,19 +122,19 @@ public class MainFile {
   @Nullable
   private static VirtualFile findAppDir(@Nullable VirtualFile file, @NotNull Project project) {
     if (WorkspaceCache.getInstance(project).isBazel()) {
-      Workspace workspace = WorkspaceCache.getInstance(project).get();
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/MainFile.java
```diff
@@ -120,14 +121,17 @@ public boolean hasFlutterImports() {
...
     }
     return null;
   }
 
-  private static boolean isAppDir(@NotNull VirtualFile dir) {
+  private static boolean isAppDir(@NotNull VirtualFile dir, @NotNull Project project) {
     return dir.isDirectory() && (
-      dir.findChild("pubspec.yaml") != null || dir.findChild(".packages") != null);
+      dir.findChild("pubspec.yaml") != null ||
+      (WorkspaceCache.getInstance(project).isBazel() && dir.findChild("BUILD") != null) ||
```
jacob314, line 132: checking for a BUILD file could be fragile. Instead ,it would be better to check whether the directory is under the bazel root directory. I think that is the equivalent of what isAppDir is doing externally.

helin24: That sounds reasonable, but how do we define the bazel root directory? Is it related to what we're doing in Workspace, e.g. `findContainingWorkspaceFile` or `ProjectRootManager.getInstance(p).getContentRoots()`?

jacob314: findContainingWorkspaceFile will give you a breadcrumb to the answer. If you go back up the relative path from that directory to google3/ you will be in the right place.

mehmetf: Can this (i.e. asking for the "type" of workspace) be part of the flutter_tool interface? We are not able to customize the plugin easily in g3 but we are able to customize the tooling.

jacob314: To be clear. We are actually extremely robust on determining whether a project is bazel by checking for the existence of `dart/config/intellij-plugins/flutter.json`. This bug is a case of code being written that forgets that not all Dart projects have pubspec.yaml or .packages file. A solution would be to audit the code base for all checks for pubspec.yaml and replace them with a helper call that is specialized for bazel workspaces.

One advantage of the current system is we don't have to make users configure the flutter sdk directory manually for bazel projects as we are able to determine it automatically. If we relied on specifying a flutter sdk directory we would have a chicken and egg problem as a user would have to have configured flutter before we could ask it whether a particular directory is a bazel or vanilla project.

mehmetf: Understood.

> Audit the code base for all checks for pubspec.yaml and replace them with a helper call that is specialized for bazel workspaces.

SGTM. I implemented a similar strategy for dartdoc where all "package" related operations go through an overridable class called `PackageInfo` rather than random code doing package processing. This also has the added advantage that what the plugin needs from pubspec.yaml is well defined and testable rather than being scattered all over the code.

src/io/flutter/run/MainFile.java
```diff
@@ -120,16 +121,22 @@ public boolean hasFlutterImports() {
 
   @Nullable
   private static VirtualFile findAppDir(@Nullable VirtualFile file, @NotNull Project project) {
+    if (WorkspaceCache.getInstance(project).isBazel()) {
+      Workspace workspace = WorkspaceCache.getInstance(project).get();
+      if (workspace != null) {
```
jacob314, line 126: you should assert `workspace != null` instead of using an if.
`isBazel()` is equivalent to `workspace != null`.

src/io/flutter/run/MainFile.java
```diff
@@ -120,16 +121,22 @@ public boolean hasFlutterImports() {
...
+
     for (VirtualFile candidate = file; inProject(candidate, project); candidate = candidate.getParent()) {
-      if (isAppDir(candidate, project)) return candidate;
+      if (isAppDir(candidate)) return candidate;
     }
     return null;
   }
 
-  private static boolean isAppDir(@NotNull VirtualFile dir, @NotNull Project project) {
+  private static boolean isAppDir(@NotNull VirtualFile dir) {
```
jacob314, line 137: comment that this method is only appropriate for non-bazel projects or continue to require passing in the project and asserting. I'd hate for someone else to start using this method in a place where it is triggered for a bazel project.

src/io/flutter/run/MainFile.java
jacob314: Using a PackageInfo class sounds like a clean solution. Filed https://github.com/flutter/flutter-intellij/issues/4738 to track cleaning up this tech debt.

PR #4745, opened 2020-08-06
Fix Android Studio canary build

stevemessick:
You can mostly ignore `IdeaFrameFixture.java `. See its class comment for instructions to produce a new version of it.

The new method in `FlutterProjectStep` was copied from an old version of `StudioWizardStepPanel`, which has been deleted in Android Studio. I'm in the process of re-implementing these classes; this is part of the new project wizard.

Once this is merged the next dev channel build will work for canary again.

Tested by building a new plugin, installing it into canary 7, then creating a Flutter app.

### DIFF
--- a/flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/fixture/IdeaFrameFixture.java
+++ b/flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/fixture/IdeaFrameFixture.java
@@ -8,6 +8,7 @@ package com.android.tools.idea.tests.gui.framework.fixture;
 import static com.android.tools.idea.gradle.util.BuildMode.ASSEMBLE;
 import static com.android.tools.idea.gradle.util.BuildMode.SOURCE_GEN;
 import static com.android.tools.idea.gradle.util.GradleUtil.getGradleBuildFile;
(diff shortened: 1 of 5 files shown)


flutter-studio/testSrc/com/android/tools/idea/tests/gui/framework/fixture/IdeaFrameFixture.java
```diff
@@ -345,7 +334,7 @@ protected void selectProjectMakeAction() {
    * @param path the series of menu names, e.g. {@link invokeActionByMenuPath("Build", "Make Project")}
    */
   public IdeaFrameFixture invokeMenuPath(@NotNull String... path) {
-    getMenuFixture().invokeMenuPath(path);
+    getMenuFixture().invokeMenuPath(10, path);
```
helin24, line 337: is this the same as waitAndInvokeMenuPath below?

stevemessick: Sorry for not answering sooner; missed the question.

It looks like it is the same. I don't write (or even read) this. Copy/paste/edit per instructions at the top. :) I would prefer to re-use IdeFrameFixture but it isn't intended to be re-used, so I have to re-write it.

Issue #4737, opened 2020-08-04
Running app with plugin version 48 pauses isolates


bschiehl:
## Steps to Reproduce

1. Run the example app of [this plugin](https://github.com/michael-cheung-thebus/foreground_service) with "flutter run" on the command line. The app should execute the foregroundServiceFunction which is running in a background isolate and print the current time approximately every 5 seconds.
2. Run the same app through version 48 of the flutter plugin for IntelliJ/Android Studio. The foregroundServiceFunction doesn't get executed. On hot reload, the following is printed to the console: "An isolate is paused (probably due to --start-paused); interface might not update." After a hot restart the app resumes as expected. 

## Version info
    $ flutter doctor -v
    [√] Flutter (Channel stable, v1.17.5, on Microsoft Windows [Version 10.0.18363.959], locale de-DE)
        • Flutter version 1.17.5 at C:\src\flutter\flutter_windows_1.17.4-stable\flutter
        • Framework revision 8af6b2f038 (5 weeks ago), 2020-06-30 12:53:55 -0700
        • Engine revision ee76268252
        • Dart version 2.8.4

    [√] Android toolchain - develop for Android devices (Android SDK version 30.0.0)
        • Android SDK at C:\Users\bschiehl\AppData\Local\Android\sdk
        • Platform android-30, build-tools 30.0.0
        • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
        • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)
        • All Android licenses accepted.

    [√] Android Studio (version 4.0)
        • Android Studio at C:\Program Files\Android\Android Studio
        • Flutter plugin version 48.0.2
        • Dart plugin version 193.7361
        • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)

    [√] VS Code (version 1.46.1)
        • VS Code at C:\Users\bschiehl\AppData\Local\Programs\Microsoft VS Code
        • Flutter extension version 3.12.1

    [√] Connected device (1 available)
        • SM G950F • 9889db335a42484749 • android-arm64 • Android 9 (API 28)

devoncarew:
cc @helin24 - potentially related to https://github.com/flutter/flutter-intellij/pull/4678? It sounds like there can be other isolates, apart from the main one, that we need to resume on start.

helin24:
[...]

PR #4748, opened 2020-08-06
Avoid start-paused for directory scope tests

helin24:
This fixes https://github.com/flutter/flutter-intellij/issues/4737 temporarily.

We previously tried starting all tests with `--start-paused` for structured errors, but this causes directory scope tests to hang because we need to connect to multiple observatory URIs sequentially to run each contained test (and this isn't implemented yet).

### DIFF
--- a/src/io/flutter/sdk/FlutterSdk.java
+++ b/src/io/flutter/sdk/FlutterSdk.java
@@ -29,6 +29,7 @@ import io.flutter.pub.PubRoot;
 import io.flutter.run.FlutterDevice;
 import io.flutter.run.FlutterLaunchMode;
 import io.flutter.run.common.RunMode;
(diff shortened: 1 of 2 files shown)


src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -321,7 +322,7 @@ public FlutterCommand flutterTest(@NotNull PubRoot root, @NotNull VirtualFile fi
         throw new IllegalStateException("Flutter SDK is too old to debug tests");
       }
     }
-    if (mode == RunMode.DEBUG || mode == RunMode.RUN) {
```
jacob314, line 324: add a comment on why this check is here with a link to a bug. Otherwise the next person looking at this code will be confused.

PR #4749, opened 2020-08-07
Show run console and skip breakpoints during test run

helin24:
I had previously changed test runs to start paused and to use the FlutterTestRunner instead of the default runner so that we could resume once the IDE was listening for vm service messages. However, the FlutterTestRunner was set up for debugging and this caused test runs to show the debugging console and recognize breakpoints. This change skips the debugging setup for run mode tests.

Screen Shot 2020-08-07 at 9 49 20 AM

To check that process termination works correctly (so we don't keep waiting for the observatory URI if the user has stopped the test run), I ran with a debugging statement in the waiting loop and verified that the execution gives up if the run is stopped before the tests run:
```
      while (url == null) {
        if (launcher.isTerminated()) return;
        System.out.println("Waiting");
        TimeoutUtil.sleep(1000);
        url = connector.getWebSocketUrl();
      }
```

### DIFF
--- a/src/io/flutter/run/test/FlutterTestRunner.java
+++ b/src/io/flutter/run/test/FlutterTestRunner.java
@@ -14,10 +14,13 @@ import com.intellij.execution.executors.DefaultDebugExecutor;
 import com.intellij.execution.process.*;
 import com.intellij.execution.runners.ExecutionEnvironment;
 import com.intellij.execution.runners.GenericProgramRunner;
+import com.intellij.execution.runners.RunContentBuilder;
 import com.intellij.execution.ui.RunContentDescriptor;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.util.Key;
 import com.intellij.openapi.wm.ToolWindowId;
+import com.intellij.util.TimeoutUtil;
 import com.intellij.xdebugger.XDebugProcess;
 import com.intellij.xdebugger.XDebugProcessStarter;
 import com.intellij.xdebugger.XDebugSession;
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/test/FlutterTestRunner.java
```diff
@@ -62,9 +71,64 @@ public boolean canRun(@NotNull String executorId, @NotNull RunProfile profile) {
...
+    throws ExecutionException {
+    final ExecutionResult executionResult = launcher.execute(env.getExecutor(), this);
+    final ObservatoryConnector connector = new Connector(executionResult.getProcessHandler());
+
+    ApplicationManager.getApplication().executeOnPooledThread(() -> {
+      // Poll, waiting for "flutter run" to give us a websocket.
+      // This is adapted from DartVmServiceDebugProcess::scheduleConnect
+      String url = connector.getWebSocketUrl();
+
+      while (url == null) {
```
helin24, line 91: In `DartVmServiceDebugProcess::scheduleConnect` there's a check for whether the debug session has ended before the connection starts, but I'm not sure what the equivalent is to check for here.

devoncarew: Is there a process we can query here, to see if it's still alive? I image there's something in the run state that we can check.

src/io/flutter/run/test/FlutterTestRunner.java
```diff
@@ -62,9 +71,64 @@ public boolean canRun(@NotNull String executorId, @NotNull RunProfile profile) {
...
+        TimeoutUtil.sleep(100);
+        url = connector.getWebSocketUrl();
+      }
+
+      final VmService vmService;
```
devoncarew, line 96: Can this declaration move inside the `try` block?

src/io/flutter/run/test/FlutterTestRunner.java
```diff
@@ -62,9 +71,64 @@ public boolean canRun(@NotNull String executorId, @NotNull RunProfile profile) {
...
+      try {
+        vmService = VmService.connect(url);
```
devoncarew, line 98: Perhaps add a comment indicating that we're looking to resume any paused-at-start isolates here.

src/io/flutter/run/test/FlutterTestRunner.java
devoncarew: The other changes (in the new commit) look good.

I assume from the question here that you'd want to find a way to address this (checking for process termination) before landing this PR?

helin24: Yeah I didn't find an easy way to do this but figured I'd look around a little more.

helin24: Exposing the process handler status seems to work here.

PR #4769, opened 2020-08-21
Use JxBrowser version 7.10

helin24:
This fixes our zooming in problem.

Before:
Jul-27-2020 10-06-03

After:
Aug-21-2020 16-01-15


flutter-intellij.iml
```diff
@@ -119,5 +101,23 @@
     <orderEntry type="library" name="org.powermock:powermock-module-junit4:2.0.2" level="project" />
     <orderEntry type="library" name="org.powermock:powermock-api-mockito2:2.0.2" level="project" />
     <orderEntry type="library" name="com.google.protobuf:protobuf-java:3.5.1" level="project" />
+    <orderEntry type="module-library">
```
jacob314, line 104: nit: try to keep this XML in the same part of the file to minimize the diff.

PR #4778, opened 2020-08-25
Revert "Add tests for JxBrowserManager logic and add test comments (#…

helin24:
…4714)"

This reverts commit 36acbb6b7b83f69b7cdf04488cc72a5e8cb671f4.

My change refactoring JxBrowserManager broke the class loading process, but I'm not yet sure why.


helin24:
Seems to be an issue with not passing `File` objects directly. Making a separate fix.

PR #4780, opened 2020-08-26
FlutterDependencyInspection shouldn't work in a non-Flutter project

alexander-doroshko:
Dart SDK from `flutter/bin/cache/dart-sdk` may be used for pure Dart projects, without any Flutter-related code. Flutter-specific inspection shouldn't work in this case.

Once the Flutter plugin update with this fix is published, I'll enable 'Outdated Dependencies' inspection from the Dart plugin for such cases. I don't want to enable it right now to avoid having 2 similar warnings from 2 inspections at the same time.


alexander-doroshko:
I do. Merged, thanks for the quick review!

stevemessick:
Thanks Alex! This will be in the next dev channel build, which you can pick up Monday if you want. Our next general release will be Sept 3.

Issue #4224, opened 2019-12-17
Unnecessary "Pubspec has been edited" banner


stevemessick:
The "Pubspec has been edited" banner shows up after closing the pubspec.yaml editor.

Repro:
- open a dart file
- open pubspec.yaml
- change it
- close it
- when focus shifts to the dart editor the banner is displayed

`flutter doctor -v`
```
[✓] Flutter (Channel master, v1.13.1-pre.94, on Mac OS X 10.14.6 18G2022, locale
    en-US)
    • Flutter version 1.13.1-pre.94 at .../flutter/flutter
    • Framework revision 36e599eb5d (6 days ago), 2019-12-11 07:27:13 +0100
    • Engine revision 12bf95fd49
    • Dart version 2.7.0 (build 2.7.0-dev.2.1 8b8894648f)


[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
    • Android SDK at .../Library/Android/sdk
    • Android NDK location not configured (optional; useful for native profiling
      support)
    • Platform android-29, build-tools 29.0.2
    • Java binary at:
      /Library/Java/JavaVirtualMachines/jdk1.8.0_221.jdk/Contents/Home/bin/java
    • Java version Java(TM) SE Runtime Environment (build 1.8.0_221-b11)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 10.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 10.3, Build version 10G8
    • CocoaPods version 1.7.1

[✓] Android Studio (version 3.4)
    • Android Studio at /Applications/Android Studio 3.4.app/Contents
    • Flutter plugin version 39.0.1
    • Dart plugin version 183.6270
    • Java version OpenJDK Runtime Environment (build
      1.8.0_152-release-1343-b01)

[✓] Android Studio
    • Android Studio at /Applications/Android Studio 3.6 Preview.app/Contents
    • Flutter plugin version 42.1.2
    • Dart plugin version 192.7761
    • Java version OpenJDK Runtime Environment (build
      1.8.0_212-release-1586-b4-5784211)

[✓] Android Studio (version 3.5)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 42.1.1
    • Dart plugin version 191.8593
    • Java version OpenJDK Runtime Environment (build
      1.8.0_202-release-1483-b49-5587405)

[...]

PR #4781, opened 2020-08-26
Don't cache VirtualFiles for '.packages' file and 'lib' folder.

alexander-doroshko:
This cache is never updated although `.packages` file and `lib` folder may be created/deleted while user works with the open project.

This a better fix for #4224.

I've removed 2 `refresh()` calls
```
pubspec.refresh(true, false);
packagesFile.refresh(true, false);
```
because they effectively were noop in this case (asynchronous refresh can't help if you need to check timestamps immediately after calling `refresh`).

As a next step, you may revise the whole `PubRootCache` idea. (Pub root and pubspec files remain cached.) Caching `VirtualFiles` that are within project content usually doesn't help to improve performance. Everything is already cached in the VFS, and methods like `getParent()` or `findChild()` normally take no time. 

On the other hand, manual caching of `VirtualFiles` adds a huge headache with updating or clearing cache as needed. As far as I can see, cache updating and clearing is not implemented.

What definitely needs to be cached is information based on `pubspec` content. We don't want to parse pubspec again and again if its content hasn't changed. This cache is implemented in `validateUpdateCachedPubspecInfo()` and neighbor methods. This logic definitely needs to remain. The rest of caching (`PubRoot`, `PubRootCache`) probably doesn't add any performance boost but adds bugs and maintenance headache.

### DIFF
--- a/src/io/flutter/pub/PubRoot.java
+++ b/src/io/flutter/pub/PubRoot.java
@@ -37,18 +37,10 @@ public class PubRoot {
   @NotNull
   private final VirtualFile pubspec;
 
-  @Nullable
-  private final VirtualFile packagesFile;
-
-  @Nullable
-  private final VirtualFile lib;
-
-  private PubRoot(@NotNull VirtualFile root, @NotNull VirtualFile pubspec, @Nullable VirtualFile packagesFile, @Nullable VirtualFile lib) {
+  private PubRoot(@NotNull VirtualFile root, @NotNull VirtualFile pubspec) {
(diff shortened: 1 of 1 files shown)


alexander-doroshko:
Steps to reproduce the problem, which is fixed:
- Take some existing Flutter project and delete .packages file before opening it.
- Open the project and pen some Dart file. There will be a bar suggesting to get or upgrade dependencies.
- Clicking it will get / upgrade dependencies successfully but the yellow bar won't disappear until you close the project.

PR #4714, opened 2020-07-24
Add tests for JxBrowserManager logic and add test comments

helin24:

PR #4782, opened 2020-08-26
Load correct file path for JxBrowser files

helin24:
While refactoring the `JxBrowserManager` file (https://github.com/flutter/flutter-intellij/pull/4714), I left out the download path while loading files.

This can be tested by running `bin/plugin build` with JxBrowser enabled and the key included, then installing and running the plugin .jar file.

### DIFF
--- a/src/io/flutter/jxbrowser/JxBrowserManager.java
+++ b/src/io/flutter/jxbrowser/JxBrowserManager.java
@@ -210,7 +210,7 @@ public class JxBrowserManager {
     ProgressManager.getInstance().runProcessWithProgressAsynchronously(task, processIndicator);
   }
 
-  protected void loadClasses(String[] fileNames) {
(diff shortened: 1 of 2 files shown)


src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -161,8 +158,9 @@ public void setUp(Project project) {
     // Delete any already existing files.
     // TODO(helin24): Handle if files cannot be deleted.
     for (String fileName : fileNames) {
-      if (!FileUtils.deleteFile(fileName)) {
-        LOG.info(project.getName() + ": Existing file could not be deleted - " + fileName);
+      String filePath = getFilePath(fileName);
```
jacob314, line 161: nit: add `final`

PR #4788, opened 2020-08-31
Send IntelliJ component background color to devtools

helin24:
Found that the class `UIUtil` has static methods to get default colors that I'm assuming come from the theme. 

Light:
Screen Shot 2020-09-01 at 8 37 41 AM

Dark:
Screen Shot 2020-09-01 at 8 36 30 AM

### DIFF
--- a/testSrc/unit/io/flutter/devtools/DevToolsUtilsTest.java
+++ b/testSrc/unit/io/flutter/devtools/DevToolsUtilsTest.java
@@ -60,5 +60,10 @@ public class DevToolsUtilsTest {
       generateDevToolsUrl(devtoolsHost, devtoolsPort, serviceProtocolUri, page, false, null, "3c3f41"),
       "http://127.0.0.1:9100/?ide=Android-Studio&backgroundColor=3c3f41&uri=http%3A%2F%2F127.0.0.1%3A50224%2FWTFTYus3IPU%3D%2F#timeline"
     );
+
+    assertEquals(
+      generateDevToolsUrl(devtoolsHost, devtoolsPort, serviceProtocolUri, page, false, null, "ffffff"),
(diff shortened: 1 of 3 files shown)


testSrc/unit/io/flutter/devtools/DevToolsUtilsTest.java
```diff
@@ -55,5 +55,10 @@ public void validDevToolsUrl() {
       generateDevToolsUrl(devtoolsHost, devtoolsPort, serviceProtocolUri, page, false, null),
       "http://127.0.0.1:9100/?ide=Android-Studio&uri=http%3A%2F%2F127.0.0.1%3A50224%2FWTFTYus3IPU%3D%2F#timeline"
     );
+
+    assertEquals(
+      generateDevToolsUrl(devtoolsHost, devtoolsPort, serviceProtocolUri, page, false, null, "3c3f41"),
```
jacob314, line 60: This color doesn't seem like quite the one we want as it is a bit grey. Taking a screenshot, the background color in a window such as the code editor or the dart analysis window in the light theme is 0xffffff.
Screen Shot 2020-08-31 at 8 50 24 PM

helin24: This is actually the dark mode color, I've figured out how to get the color from IntelliJ's defaults and light mode is indeed 0xffffff. Do you care about the test value?

jacob314: would be nice to test light mode as well if that is easy.

helin24: It's not very meaningful since the hex code is passed into the function, but it's certainly easy!

PR #4790, opened 2020-09-01
Enable embedded browser by default in dev

helin24:
Users on the dev release of our plugin should automatically have the embedded browser setting enabled. I verified that this setting can also be turned off.

### DIFF
--- a/src/io/flutter/FlutterUtils.java
+++ b/src/io/flutter/FlutterUtils.java
@@ -11,6 +11,8 @@ import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.util.ExecUtil;
 import com.intellij.ide.actions.ShowSettingsUtilImpl;
 import com.intellij.ide.impl.ProjectUtil;
+import com.intellij.ide.plugins.IdeaPluginDescriptor;
+import com.intellij.ide.plugins.PluginManagerCore;
 import com.intellij.ide.util.PropertiesComponent;
 import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationManager;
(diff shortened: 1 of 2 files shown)


src/io/flutter/FlutterUtils.java
```diff
@@ -552,4 +555,10 @@ private static ModuleSourceOrderEntry findModuleSourceEntry(@NotNull Module modu
     }
     return null;
   }
+
+  public static boolean isPluginVersionDev() {
```
jacob314, line 559: I like it!
Should we also make sure that local builds behave like dev or provide an option to trigger dev/non-dev behavior?

helin24: I could make this return true for `SNAPSHOT` versions as well. I'm not sure what an option would look like and whether we foresee adding more differentiated handling for dev vs. non-dev releases; I was thinking this current use case is very temporary and it's sufficient to test by changing plugin.xml to have a version name like `SNAPSHOT-dev`. Do you want to describe a potential option more?

stevemessick:
I think it is fine as-is. The preference is sticky, right? Set it once and
keep this simple.

helin24:
> I think it is fine as-is. The preference is sticky, right? Set it once and
keep this simple.

It is sticky, so true, this is probably good enough.

src/io/flutter/FlutterUtils.java
jacob314: Making it return true for SNAPSHOT sounds good. As I'd like anyone building locally to also catch bugs before we enable features in STABLE.
The uses should be restricted to enabling certain experimental features by default instead of disabling them. Given that, maybe make this private and only in the FlutterSettings class to avoid misuse. It would be good to avoid uses of it that make it harder to test behavior in the release version by having dev/SNAPSHOT diverge.

helin24: Okay, that makes sense - updated here: https://github.com/flutter/flutter-intellij/pull/4794

PR #4791, opened 2020-09-01
Add jxbrowser installation events

helin24:
The table that's accessible in BigQuery seems to be delayed, so I think I have to check back tomorrow to see if my testing recorded events. 

Update - I still don't see this in BigQuery, but I do see it in analytics:
Screen Shot 2020-09-02 at 9 54 50 AM


jacob314 (approved):
lgtm

PR #4795, opened 2020-09-02
Update build for M49

stevemessick:


stevemessick:
I will fix the Kokoro problem later.

stevemessick:
The travis problem is the same as kokoro -- the gradle project expects to run tests with ideaIC but the product matrix is using android-studio everywhere.

PR #4802, opened 2020-09-03
Remove extra page param when launching DevTools

helin24:
There are currently two page params used while creating a DevTools URL, and the one in the form `#timeline` doesn't seem to do anything. I wanted to remove this since it's clearer and it will make adding a "simulated" embedded browser page (for developers without a JxBrowser license) a little easier.

I checked that opening DevTools from the run console and from the performance tab hasn't changed.


jacob314 (approved):
Thanks for cleaning this up!
lgtm

Issue #4803, opened 2020-09-03
DevTools is not working on Windows


stevemessick:
## What happened

Attempted to start embedded DevTools on Windows.

## Version information

Android Studio `4.0.1` • Flutter plugin `io.flutter 49.0.2` • Dart plugin `193.7547`

Error getting Flutter sdk information.

## Exception

null

```
com.teamdev.jxbrowser.browser.UnsupportedRenderingModeException: The current HARDWARE_ACCELERATED rendering mode is not supported
   at com.teamdev.jxbrowser.browser.internal.BrowserSettingsImpl.checkRenderingMode(BrowserSettingsImpl.java:370)
   at com.teamdev.jxbrowser.browser.internal.BrowserSettingsImpl.setTransparentBackgroundEnabled(BrowserSettingsImpl.java:358)
   at com.teamdev.jxbrowser.browser.internal.BrowserSettingsImpl.enableTransparentBackground(BrowserSettingsImpl.java:341)
   at io.flutter.jxbrowser.EmbeddedBrowser.openPanel(EmbeddedBrowser.java:30)
   at io.flutter.devtools.DevToolsInstance.lambda$openPanel$1(DevToolsManager.java:402)
   at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:309)
   at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.doRun(LaterInvocator.java:441)
   at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.runNextEvent(LaterInvocator.java:424)
   at com.intellij.openapi.application.impl.LaterInvocator$FlushQueue.run(LaterInvocator.java:407)
   at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
   at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:764)
   at java.awt.EventQueue.access$500(EventQueue.java:98)
   at java.awt.EventQueue$3.run(EventQueue.java:715)
   at java.awt.EventQueue$3.run(EventQueue.java:709)
   at java.security.AccessController.doPrivileged(Native Method)
   at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
   at java.awt.EventQueue.dispatchEvent(EventQueue.java:734)
   at com.intellij.ide.IdeEventQueue.defaultDispatchEvent(IdeEventQueue.java:908)
[...]

PR #4804, opened 2020-09-03
Skip transparent background if hardware accelerated causes error

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/4803

It seems that hardware accelerated mode can fail on Windows, but we aren't sure if this is due to a machine setting or a JxBrowser limitation. Adding this check so that we can try a backup mode.

~~Verified that this works, but it is comically zoomed in:
Annotation 2020-09-08 162604~~

The `OFFSCREEN` mode works fine; in the previous screenshot my windows device had a strange zoom setting.
Screenshot 2020-09-09 151929

The `HARDWARE_ACCELERATED` mode works too; in comments below the issue was also due to the local zoom setting.
Screenshot 2020-09-09 154410

### DIFF
--- a/src/io/flutter/jxbrowser/EmbeddedBrowser.java
+++ b/src/io/flutter/jxbrowser/EmbeddedBrowser.java
@@ -10,6 +10,7 @@ import com.intellij.openapi.wm.ToolWindow;
 import com.intellij.ui.content.Content;
 import com.intellij.ui.content.ContentManager;
 import com.teamdev.jxbrowser.browser.Browser;
+import com.teamdev.jxbrowser.browser.UnsupportedRenderingModeException;
 import com.teamdev.jxbrowser.engine.Engine;
 import com.teamdev.jxbrowser.engine.EngineOptions;
 import com.teamdev.jxbrowser.navigation.event.LoadFinished;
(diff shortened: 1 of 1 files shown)


jacob314:
Do we know why the zoom is broken? That is something it would be good to fix either in IntelliJ or DevTools. Perhaps the device pixel ratio isn't getting handled correctly? What is the device pixel ratio of your windows machine?

src/io/flutter/jxbrowser/EmbeddedBrowser.java
```diff
@@ -20,19 +21,30 @@
...
-    final Engine engine = Engine.newInstance(options);
-    final Browser browser = engine.newBrowser();
-    browser.settings().enableTransparentBackground();
+    Engine engine = Engine.newInstance(options);
+    Browser browser = engine.newBrowser();
+
+    try {
+      browser.settings().enableTransparentBackground();
+    } catch (UnsupportedRenderingModeException ex) {
+      options =
```
jacob314, line 36: does HARDWARE_ACCELERATED without the transparent background work on windows? Perhaps add a comment as the placement of the try-catch makes it look like that is the problem not the HARDWARE_ACCELERATED option. I'll put it this way, if I had to pick, HARDWARE_ACCELERATED over enableTransparentBackground as one impacts performance and the other just causes a UI flicker.

src/io/flutter/jxbrowser/EmbeddedBrowser.java
```diff
@@ -20,19 +21,30 @@
...
+      engine = Engine.newInstance(options);
+      browser = engine.newBrowser();
+      browser.settings().enableTransparentBackground();
+    }
 
     // Multiple LoadFinished events can occur, but we only need to add content the first time.
     final AtomicBoolean contentLoaded = new AtomicBoolean(false);
 
     browser.navigation().loadUrl(url);
+    final Browser finalBrowser = browser;
```
jacob314, line 47: nit: having a variable named `finalBrowser` is a bit ugly.  A slightly cleaner option would be to have a `createBrowser()`helper method.
`final Browser browser= createBrowser(..)`

src/io/flutter/jxbrowser/EmbeddedBrowser.java
helin24: I can skip enabling transparent background, but `HARDWARE_ACCELERATED` mode also causes the window to appear to the left of the panel. Looking into if I can fix this.. maybe a swing issue?

Annotation 2020-09-09 135226

jacob314: That seems like it is a jxbrowser bug on windows. Swing shouldn't have many windows specific issues.

helin24: Hmm.. I'll post an issue and see if they have any advice.

src/io/flutter/jxbrowser/EmbeddedBrowser.java
helin24: If I just skip the transparent browser step, then we don't really need a separate function here.

helin24:
Okay, that was silly. I initially had a setting on my Windows machine (Display -> Scale and layout -> Change the size of text, apps, and other items) that made both render modes problematic. After resetting this to default, both render modes run fine on Windows as long as `enableTransparentBackground` is skipped.

jacob314 (approved):
lgtm

PR #4818, opened 2020-09-10
Use Gradle to compile for Kokoro

stevemessick:
The Ant script has to manage the entire class path. Changing to Java 14 on Kokoro breaks it. This change switches to using Gradle instead. We won't know that it works until this is merged, but it should since the Gradle-driven tests do work.


helin24 (approved):

stevemessick:
It worked! There's some risk that this breaks the dev channel build. We'll keep an eye on that.

Issue #4812, opened 2020-09-09
NPE from JxBrowserUtils.getJxBrowserKey(JxBrowserUtils.java:56)


devoncarew:
I see this from a local dev environment:

```
java.lang.NullPointerException: inStream parameter is null
	at java.base/java.util.Objects.requireNonNull(Objects.java:246)
	at java.base/java.util.Properties.load(Properties.java:403)
	at io.flutter.utils.JxBrowserUtils.getJxBrowserKey(JxBrowserUtils.java:56)
	at io.flutter.jxbrowser.JxBrowserManager.setUp(JxBrowserManager.java:133)
	at io.flutter.ProjectOpenActivity.runActivity(ProjectOpenActivity.java:63)
	at com.intellij.ide.startup.impl.StartupManagerImpl.runActivity(StartupManagerImpl.java:300)
	at com.intellij.ide.startup.impl.StartupManagerImpl.lambda$runPostStartupActivities$4(StartupManagerImpl.java:226)
	at com.intellij.openapi.extensions.impl.ExtensionPointImpl.processWithPluginDescriptor
...
```

cc @helin24

PR #4821, opened 2020-09-11
Log exception when jxbrowser license key file is missing

helin24:
When there isn't a jxbrowser.properties file, the logs should show something like:
```
2020-09-10 17:26:22,153 [   6543]   INFO - ter.jxbrowser.JxBrowserManager - jxbrowser.properties file does not exist 
2020-09-10 17:26:22,153 [   6543]   INFO - ter.jxbrowser.JxBrowserManager - flutter_error_studies: Unable to find JxBrowser license key file 
```
instead of an NPE.

Since this should mostly occur in development, should I use `System.out.println` here instead?

Fixes https://github.com/flutter/flutter-intellij/issues/4812

### DIFF
--- a/src/io/flutter/jxbrowser/JxBrowserManager.java
+++ b/src/io/flutter/jxbrowser/JxBrowserManager.java
@@ -133,11 +133,11 @@ public class JxBrowserManager {
     // Retrieve key
     try {
       final String key = JxBrowserUtils.getJxBrowserKey();
-      System.setProperty(JxBrowserUtils.PROPERTY_NAME, key);
+      System.setProperty(JxBrowserUtils.LICENSE_PROPERTY_NAME, key);
     }
     catch (FileNotFoundException e) {
       LOG.info(e.getMessage());
-      LOG.info(project.getName() + ": Unable to find JxBrowser licence key file");
+      LOG.info(project.getName() + ": Unable to find JxBrowser license key file");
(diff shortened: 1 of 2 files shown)


src/io/flutter/jxbrowser/JxBrowserManager.java
```diff
@@ -133,11 +133,11 @@ public void setUp(Project project) {
     // Retrieve key
     try {
       final String key = JxBrowserUtils.getJxBrowserKey();
-      System.setProperty(JxBrowserUtils.PROPERTY_NAME, key);
+      System.setProperty(JxBrowserUtils.LICENSE_PROPERTY_NAME, key);
     }
     catch (FileNotFoundException e) {
       LOG.info(e.getMessage());
-      LOG.info(project.getName() + ": Unable to find JxBrowser licence key file");
+      LOG.info(project.getName() + ": Unable to find JxBrowser license key file");
```
devoncarew, line 140: You may be able to use `LOG.info(project.getName() + ..., e)` instead of two separate log calls.

devoncarew (approved):
👍 thanks!

devoncarew:
> Since this should mostly occur in development, should I use `System.out.println` here instead?

I don't have a real opinion here - like you say, we should only see this for local development (we wouldn't want to call print generally). Either works for this I think -

PR #4823, opened 2020-09-11
Open DevTools in standard browser if jxbrowser license is missing

helin24:
With this change, running the development workbench without a jxbrowser license shows a message in the inspector panel: "The JxBrowser license could not be found. Click to open Devtools in the browser." that can be clicked to open the same DevTools URL in the normal browser.

When we're ready to remove the old inspector view, we can also use this path for users who elect not to use JxBrowser.

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -48,6 +48,7 @@ import io.flutter.run.daemon.FlutterApp;
 import io.flutter.settings.FlutterSettings;
 import io.flutter.utils.AsyncUtils;
 import io.flutter.utils.EventStream;
+import io.flutter.utils.JxBrowserUtils;
 import io.flutter.utils.VmServiceListenerAdapter;
 import io.flutter.vmService.ServiceExtensions;
 import org.dartlang.vm.service.VmService;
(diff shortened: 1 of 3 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -447,11 +459,18 @@ protected void waitForJxBrowserInstallation(FlutterApp app, InspectorService ins
...
-      handleJxBrowserInstallationInProgress(app, inspectorService, toolWindow);
-    });
+    if (System.getProperty(JxBrowserUtils.LICENSE_PROPERTY_NAME) == null) {
+      // If the license isn't available, allow the user to open the equivalent page in a non-embedded browser window.
+      presentClickableLabel(toolWindow, "The JxBrowser license could not be found. Click to open Devtools in the browser.", (linkLabel, data) -> {
+        presentDevTools(app, inspectorService, toolWindow, false);
+      });
+    } else {
+      // Allow the user to manually restart.
+      presentClickableLabel(toolWindow, "JxBrowser installation failed. Click to retry.", (linkLabel, data) -> {
```
jacob314, line 469: nit: try to avoid using the work "Click"/ "Click here" in messages. This could just be "Retry?"

jacob314 (approved):
lgtm

PR #4813, opened 2020-09-09
upgrade the vm service protocol library

devoncarew:
- upgrade the vm service protocol library

This brings in the latest version of the (generated) vm service protocol library, and adapts our debugger code for the relatively minor changes.

Note that the new version of the vm service protocol library is now forward compatible with major version changes of the protocol; we'll try and work with any version of the protocol.

PR #4828, opened 2020-09-14
update to the latest VM service protocol library

devoncarew:
Follow up from https://github.com/flutter/flutter-intellij/pull/4813:
- update to the latest VM service protocol library (w/ the version number change)
- create an abstract `EmptyResumeConsumer` so we don't have so many empty implementations of a ResumeConsumer around


jacob314 (approved):
lgtm

PR #4842, opened 2020-09-16
remove a log error about no isolates

devoncarew:
- remove a log error about no isolates at initial app connection

I believe that DDS can change the timing of when isolates are reported to exist; @bkonyi to confirm.

cc @bkonyi, @jwren


bkonyi (approved):
Assuming this is just a timing change now that DDS landed. LGTM.

PR #4847, opened 2020-09-17
support pausing a running app

devoncarew:
- support pausing a running app
- make the `DartVmServiceDebugProcess` class abstract as we don't instantiate it but instead have 3 concrete subclasses

Screen Shot 2020-09-17 at 9 18 59 AM

W/ this change, the 'pause' button becomes enabled when the app is running (not paused).

cc @jwren, @alexander-doroshko  as we'll likely want a similar change in the Dart plugin


jacob314 (approved):
lgtm

PR #4857, opened 2020-09-18
remove FlutterLogView and associated classes

devoncarew:
- remove FlutterLogView and associated classes

This PR removes then `FlutterLogView` console contribution and associated classes. This has been disabled for ~a year now; we have some longer term goals for improvements to console output, but likely won't go the route of a full console replacement.

Removing this code drops our build time issue count down to 91 warnings (and deletes ~4.2k sloc).


pq (approved):
🔥

PR #4858, opened 2020-09-18
address lints on JxBrowserManager.java

devoncarew:
- address lints on JxBrowserManager.java

@helin24 - this isn't necessarily for landing; it looks like `files` is written to but never read from - I don't know what the intent of the variable was. In any case, FYI.


helin24:
`files` must have been left over from when I was doing class loading a different way, it's okay to remove.

devoncarew:
> `files` must have been left over from when I was doing class loading a different way, it's okay to remove.

OK, gotcha. This is ready for review then :)

helin24 (approved):

PR #4865, opened 2020-09-20
more general lint cleanups

devoncarew:
- more lint cleanups


src/io/flutter/view/FlutterView.java
```diff
@@ -229,7 +230,7 @@ private void addBrowserInspectorViewContent(FlutterApp app,
   private void addInspectorViewContent(FlutterApp app, @Nullable InspectorService inspectorService, ToolWindow toolWindow) {
     final ContentManager contentManager = toolWindow.getContentManager();
     final SimpleToolWindowPanel toolWindowPanel = new SimpleToolWindowPanel(true);
-    final JBRunnerTabs runnerTabs = new JBRunnerTabs(myProject, ActionManager.getInstance(), IdeFocusManager.getInstance(myProject), this);
+    final JBRunnerTabs runnerTabs = new JBRunnerTabs(myProject, this);
```
stevemessick, line 233: This constructor may not be available in AS 4.0. Have you done `plugin make`?

devoncarew: No; I can run that. Shouldn't that be run on the bots as well though - green bots should tell is this works across the range of supported IDEs?

stevemessick: > green bots should tell is this works across the range of supported IDEs?

That's a good point. It should, but it isn't working. There are compilation errors in your 4.0 build.

The `plugin buildPlugin` command returns success even when the compilation fails. That needs to be fixed.

devoncarew: Ah, ok. Sorry I've been committing breaking changes - I'll vet that things work locally (and roll back things that don't).

stevemessick: Not your fault! And you could just add a Subst() in edit.dart instead of rolling back. Or go ahead and commit; I can fix it up.

stevemessick (approved):
Since I broke `plugin make` I'll clean up any problems. Go ahead and merge this as-is.

src/io/flutter/view/FlutterView.java
devoncarew: I reverted the two api changes in there that depend on newer platform versions (and added todos indicating what versions they require).

stevemessick (approved):

Issue #4467, opened 2020-04-06
Clarify generic error names in analytics


InMatrix:
The Flutter framework error analytics include some rather generic and sometimes vague names such as:
* 'package
* rangeerror
* invalid-argument
* exception

Does each name refer to a unique type of error? Can you clarify what are those errors? Thanks!

devoncarew:
This is due to our extremely ad-hoc method for converting error messages into IDs. We could improve the heuristic, but I wonder if we should continue to iterate on that system, or just have real IDs for error messages.

devoncarew:
cc @jacob314

devoncarew:
Here's the current logic: https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/logging/FlutterErrorHelper.java#L30

devoncarew:
@InMatrix - what are the current generic error names we're seeing (and, any idea about how to narrow down the source errors we're seeing them from)? Our series of regexs unfortunately essentially obfuscates the original error text. 

This would be more reliable with actual error codes. cc @csells

InMatrix:
I think the only one that appears to be a parsing issue is the error logged as `'package` in our analytics. The other ones I mentioned in the original post seems to be just generic in its source, like `exception` and `invalid-argument`.

devoncarew:
From some investigation, we produce the `'package` ID when we encounter an error from an assert statement, used without an assert message:

```
assert(1 != 2);
```

`assert(1 != 2, 'my assert test`)` we would be able the create a good error id for.

InMatrix:
I'm still seeing "'package" logged as an error id in the framework errors dashboard, when I set the date range to the last seven days. Is there a way to verify if this fix worked? Cc: @jayoung-lee to check if the dashboard has been refreshed with the latest data.

PR #4870, opened 2020-09-21
improve the flutter error ids we generate for analytics

devoncarew:
- improve the flutter error ids we generate for analytics
- fix https://github.com/flutter/flutter-intellij/issues/4467

This fixes the `'package` error ID we'd seen in analytics.

### DIFF
--- a/testSrc/unit/io/flutter/logging/FlutterErrorHelperTest.java
+++ b/testSrc/unit/io/flutter/logging/FlutterErrorHelperTest.java
@@ -34,27 +34,84 @@ public class FlutterErrorHelperTest {
     assertEquals(
       "failed-assertion",
       FlutterErrorHelper.getAnalyticsId(
-        "package:flutter/src/widgets/framework.dart':\nfailed assertion: line 123 pos 123: 'owner._debugcurrentbuildtarget == this': is not true"));
+        "'package:flutter/src/widgets/framework.dart':\n" +
+        "failed assertion: line 123 pos 123: 'owner._debugcurrentbuildtarget == this': is not true"));
     assertEquals(
       "failed-assertion",
       FlutterErrorHelper.getAnalyticsId(
-        "'package:flutter/src/widgets/framework.dart':\nfailed assertion: line 123 pos 123: 'owner._debugcurrentbuildtarget == this': is not true"));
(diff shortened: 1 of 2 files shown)


testSrc/unit/io/flutter/logging/FlutterErrorHelperTest.java
```diff
@@ -34,27 +34,84 @@ public void testGetAnalyticsId_scrubbing() {
...
       FlutterErrorHelper.getAnalyticsId("could not find a generator for route routesettings(\"/foobar\", null) in the _widgetsappstate"));
     assertEquals(
-      "controller's-length-property-does-not-match-the",
-      FlutterErrorHelper.getAnalyticsId("controller's length property (123) does not match the\nnumber of tabs"));
+      "controller's-length-property-does-not-match-the-number-of-tabs",
+      FlutterErrorHelper.getAnalyticsId("controller's length property (123) does not match the number of tabs"));
+  }
+
+  @Test
+  public void testGetAnalyticsId_errorStudiesApp() {
```
devoncarew, line 66: @InMatrix - can you verify that the IDs here are roughly what we'd want to see for the given errors?

InMatrix: Yes, those IDs look good. I'm not sure if we need to include the package information though. Is it included to disambiguate similar assertions?

devoncarew: By the package info, do you mean the `package:flutter_error_studies/...` text in the source errors below? It's just the text that I copied in - it's nice to have real world examples (definitely an issue when I originally wrote this, to have to guess what the error might look like).

InMatrix: No, I was referring to "'package:flutter/src/material/reorderable_list.dart': failed assertion: line 123 pos 123: 'children.every((widget w) => w.key != null)'"));" under "all children of this widget must have a key.\n" at line 56. I'm not sure if the extra detail about the specific file in package flutter and assertion position in the file are necessary for analytics .

stevemessick (approved):

PR #4873, opened 2020-09-21
Fix legacy compilation issues

stevemessick:


devoncarew:
Thanks for running these issues down! We could also choose to remove / rewrite the issues in the source code directly - checked in - w/ comments to not re-add until 2020.1.

devoncarew (approved):

PR #4874, opened 2020-09-21
Fix plugin make

stevemessick:
Should return failure code now.


devoncarew (approved):
Ah, we had been returning the return code from the call to gradle stop, not the build command result.

PR #4876, opened 2020-09-22
revert some forward looking API changes

devoncarew:
- revert some forward looking API changes
- remove the associated source re-writing
- add todos in the affected places w/ the versions that the newer APIs require
- also, remove support for re-writing source to support Android Studio 3.6

@stevemessick

### DIFF
--- a/src/io/flutter/FlutterUtils.java
+++ b/src/io/flutter/FlutterUtils.java
@@ -50,7 +50,6 @@ import org.yaml.snakeyaml.resolver.Resolver;
 
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.nio.file.Paths;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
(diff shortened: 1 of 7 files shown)


stevemessick:
This appears to be the inverse of #4873. I'm unclear why we should do this. We'll have to make the changes eventually, so why not just keep the edits?

devoncarew:
> This appears to be the inverse of #4873. I'm unclear why we should do this.

Yeah, it mostly is. I'm assuming that we'd rather have source code on our disks locally that can work in all our supported versions, instead of relying on a re-write process. We don't have many of these re-writes, so this isn't a huge deal; happy to close this if you'd prefer the current state.

stevemessick:
> I'm assuming that we'd rather have source code on our disks locally that can work in all our supported versions

Now I see. That would be nice, but we haven't had it since Android Studio 3.0. The AS code has incompatible differences with each version. I just keep my local sources sync'd to the latest AS sources, which includes IntelliJ.

I'd prefer to keep what we have.

devoncarew:
Sounds good -

Issue #4825, opened 2020-09-14
Can't run Flutter project if it's within a Bazel directory


rohansingh:
I have a Flutter project that's within a Bazel directory. It's not built by Bazel in any way, but it happens to be in a monorepo that uses Bazel for most other things.

Running and debugging from within Android Studio used to work fine, but is now broken. Trying to launch `main.dart` results in an error:

> Entrypoint isn't within a Flutter pub root.

Looking at **Preferences > Languages & Frameworks > Flutter** shows that the Flutter SDK path has changed to `/Users/rohan/tidbyt/null <set by bazel project>`. That isn't a valid path on my machine. Resetting the setting here doesn't seem to do anything.

I have a hunch that this is related to #4735 (/cc @helin24 @jacob314). My guess is that that broke the IntelliJ plugin for Flutter projects in Bazel directories outside of Google.

# Steps to Reproduce

1. Have a Flutter project that's inside a Bazel workspace. (not at Google)
2. Try to run the project from within Android Studio

## Version info

```
[✓] Flutter (Channel unknown, 1.20.2, on Mac OS X 10.15.6 19G2021, locale en-US)
    • Flutter version 1.20.2 at /opt/flutter
    • Framework revision bbfbf1770c (5 weeks ago), 2020-08-13 08:33:09 -0700
    • Engine revision 9d5b21729f
    • Dart version 2.9.1

[!] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at /Users/rohan/Library/Android/sdk
    • Platform android-29, build-tools 29.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)
    ✗ Android license status unknown.
      Try re-installing or updating your Android SDK Manager.
      See https://developer.android.com/studio/#downloads or visit https://flutter.dev/docs/get-started/install/macos#android-setup for detailed instructions.

[✓] Xcode - develop for iOS and macOS (Xcode 11.7)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 11.7, Build version 11E801a
    • CocoaPods version 1.9.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 4.0)
[...]

PR #4886, opened 2020-09-24
Return null workspace if missing flutter.json

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/4825

To test:
- Typical flutter project like flutter gallery can be run 
- Flutter project included inside a directory with a bazel WORKSPACE file (but without flutter.json file) can be run as a flutter project - verify that in Flutter language settings, the flutter SDK is customizable and not set by bazel
- Flutter project that has a flutter.json file has SDK set by bazel

We discussed potentially simplifying the `Workspace` and `WorkspaceCache` setup process since it is confusing to follow; but for now this change should be sufficient.

### DIFF
--- a/src/io/flutter/bazel/Workspace.java
+++ b/src/io/flutter/bazel/Workspace.java
@@ -220,11 +220,11 @@ public class Workspace {
     final String readonlyPath = "../READONLY/" + root.getName();
     final VirtualFile readonlyRoot = root.findFileByRelativePath(readonlyPath);
     VirtualFile configFile = root.findFileByRelativePath(PLUGIN_CONFIG_PATH);
-    if (configFile == null && readonlyRoot == null) return null;
-
     if (configFile == null && readonlyRoot != null) {
       configFile = readonlyRoot.findFileByRelativePath(PLUGIN_CONFIG_PATH);
     }
+    if (configFile == null) return null;
+
     final PluginConfig config = configFile == null ? null : PluginConfig.load(configFile);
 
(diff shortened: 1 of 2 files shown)


src/io/flutter/bazel/Workspace.java
```diff
@@ -220,6 +220,8 @@ static Workspace loadUncached(@NotNull Project project) {
     final String readonlyPath = "../READONLY/" + root.getName();
     final VirtualFile readonlyRoot = root.findFileByRelativePath(readonlyPath);
     VirtualFile configFile = root.findFileByRelativePath(PLUGIN_CONFIG_PATH);
+    if (configFile == null && readonlyRoot == null) return null;
```
jacob314, line 223: why not short circuit if the configFile == null OR readonlyRoot == null?
That should indicate we are in an environment where you can't actually run a g3 app.

jacob314 (approved):
lgtm

src/io/flutter/bazel/Workspace.java
helin24: I'm not clear on how the readonlyRoot works. I thought that if `configFile` is null but `readOnlyRoot` exists, then we can still run a g3 app? Since the code below finds a config file within readonly.

Maybe I should instead check for only `configFile` later, e.g.:
```
    VirtualFile configFile = root.findFileByRelativePath(PLUGIN_CONFIG_PATH);
    if (configFile == null && readonlyRoot != null) {
      configFile = readonlyRoot.findFileByRelativePath(PLUGIN_CONFIG_PATH);
    }
    if (configFile == null) return null;
```

PR #4902, opened 2020-09-29
remove the explicit disable track widget creation option

devoncarew:
- remove the explicit disable track widget creation option (users can still pass it in via run command line options)
- pass in `--track-widget-creation` if the flutter sdk supports the option
- don't pass it in if the user has an explicit value for track-widget-creation in the additional args


jacob314 (approved):
lgtm

PR #4908, opened 2020-09-29
remove an older wizard for small IDEs

devoncarew:
- remove an older wizard for small IDEs

This has not been in use for a while.


stevemessick (approved):
It looks like Android Studio support does not rely on this.

devoncarew:
> It looks like Android Studio support does not rely on this.

Yes, from memory, this was to support WebStorm.

PR #4909, opened 2020-09-29
Change FileUtils to instance for easier mocking

helin24:
This follows https://github.com/flutter/flutter-intellij/pull/4893 since with using version 2020.3 there's a problem with verifying static methods in tests.

I checked that file actions like downloading continue to work.


stevemessick (approved):
Thanks!

PR #4910, opened 2020-09-29
collect analytics for the 'enable hot ui' setting

devoncarew:
- collect analytics for the 'enable hot ui' setting


jacob314 (approved):
lgtm

PR #4915, opened 2020-09-30
update the DartElementPresentationUtil class

devoncarew:
- update the DartElementPresentationUtil class

This class is copied from the Dart plugin - this updates to the latest version.

### DIFF
--- a/src/io/flutter/preview/DartElementPresentationUtil.java
+++ b/src/io/flutter/preview/DartElementPresentationUtil.java
@@ -44,6 +44,8 @@ public class DartElementPresentationUtil {
     switch (element.getKind()) {
       case ElementKind.CLASS:
         return element.isAbstract() ? AbstractClass : Class;
+      case ElementKind.EXTENSION:
+        return Include;
       case ElementKind.MIXIN:
         return AbstractClass;
       case ElementKind.CONSTRUCTOR:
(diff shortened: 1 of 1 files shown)


src/io/flutter/preview/DartElementPresentationUtil.java
```diff
@@ -60,7 +62,7 @@ public static Icon getIcon(@NotNull Element element) {
         if (element.isTopLevelOrStatic()) return STATIC_FIELD_ICON;
         return Field;
       case ElementKind.FUNCTION:
-        return element.isTopLevelOrStatic() ? TOP_LEVEL_FUNCTION_ICON : Function;
+        return element.isTopLevelOrStatic() ? TOP_LEVEL_FUNCTION_ICON : Lambda;
```
stevemessick, line 65: Shouldn't this also check if the function has a name before using `Lambda`?

devoncarew: I'm not sure here - the 'lamda' icon is the one that the Dart plugin is using for non-top level functions, whether the item is actually a lamda or not.

devoncarew: Note that this PR is just updating our code to the latest from the Dart plugin - if we think we should change something here we should adjust it upstream.

stevemessick (approved):
Sorry, I thought I'd already approved this one.

PR #4917, opened 2020-09-30
remove the widget catalog file and related code

devoncarew:
- remove the widget catalog file (`widgets.json`) and related code

This file is a copy of the one from the flutter website. We haven't maintained it wrt changes in the source file, and, the only thing we were using the info for was for icon previews for elements in the inspector view.

The DevTools inspector no longer uses the icons from the widget catalog info; this change matches what that version of the inspector is doing.


jacob314 (approved):
lgtm
lgtm

PR #4931, opened 2020-10-01
various lint fixes

devoncarew:
- various lint fixes

### DIFF
--- a/src/io/flutter/FlutterErrorReportSubmitter.java
+++ b/src/io/flutter/FlutterErrorReportSubmitter.java
@@ -56,7 +56,7 @@ public class FlutterErrorReportSubmitter extends ErrorReportSubmitter {
                         @NotNull Component parentComponent,
                         @NotNull Consumer<SubmittedReportInfo> consumer) {
     if (events.length == 0) {
-      fail(((Consumer<SubmittedReportInfo>)consumer));
+      fail(consumer);
       return false;
     }
 
@@ -96,7 +96,7 @@ public class FlutterErrorReportSubmitter extends ErrorReportSubmitter {
     final DataContext dataContext = DataManager.getInstance().getDataContext(parentComponent);
     final Project project = PROJECT.getData(dataContext);
     if (project == null) {
-      fail(((Consumer<SubmittedReportInfo>)consumer));
+      fail(consumer);
       return false;
     }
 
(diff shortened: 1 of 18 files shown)


src/io/flutter/FlutterErrorReportSubmitter.java
```diff
@@ -56,7 +56,7 @@ public boolean submit(@NotNull IdeaLoggingEvent[] events,
                         @NotNull Component parentComponent,
                         @NotNull Consumer<SubmittedReportInfo> consumer) {
     if (events.length == 0) {
-      fail(((Consumer<SubmittedReportInfo>)consumer));
```
stevemessick, line 59: This is required to compile with current AS sources.

src/io/flutter/FlutterErrorReportSubmitter.java
```diff
@@ -96,7 +96,7 @@ public boolean submit(@NotNull IdeaLoggingEvent[] events,
     final DataContext dataContext = DataManager.getInstance().getDataContext(parentComponent);
     final Project project = PROJECT.getData(dataContext);
     if (project == null) {
-      fail(((Consumer<SubmittedReportInfo>)consumer));
```
stevemessick, line 99: This is required to compile with current AS sources.

src/io/flutter/FlutterErrorReportSubmitter.java
```diff
@@ -189,7 +189,7 @@ public boolean submit(@NotNull IdeaLoggingEvent[] events,
     final VirtualFile file = scratchRoot.createScratchFile(project, "bug-report.md", PlainTextLanguage.INSTANCE, text);
 
     if (file == null) {
-      fail(((Consumer<SubmittedReportInfo>)consumer));
```
stevemessick, line 192: This is required to compile with current AS sources.

stevemessick (changes requested):
I found three changes that will cause compilation errors in the current Android Studio environment.

devoncarew:
> I found three changes that will cause compilation errors in the current Android Studio environment.

Thanks for checking! I'll revert and add comments so I don't unintentionally re-add.

How are these passing in the CI though?

stevemessick:
Thanks!

> How are these passing in the CI though?

They both use Java 11.

src/io/flutter/FlutterErrorReportSubmitter.java
devoncarew: Reverted (and the two places below as well).

devoncarew:
> They both use Java 11.

Ah, gotcha.

I see that we're installing jdk11 on the bots (https://github.com/flutter/flutter-intellij/blob/master/.travis.yml#L6). It looks like travis has some capability to change between jdk versions in the middle of a run:

https://docs.travis-ci.com/user/languages/java/#switching-jdks-java-8-and-below-within-one-job

stevemessick (approved):

PR #4935, opened 2020-10-02
update the FlutterSearchableOptionContributor class

devoncarew:
- update the FlutterSearchableOptionContributor class


stevemessick (approved):
Wow, thanks for doing this 🥇

Issue #4936, opened 2020-10-02
Argument for @NotNull parameter 'service' of io/flutter/logging/FlutterConsoleLogManager.getFullStringValue must not be null


devoncarew:
Seen from crash reporting:

```
Argument for @NotNull parameter 'service' of io/flutter/logging/FlutterConsoleLogManager.getFullStringValue must not be null
	io.flutter.logging.FlutterConsoleLogManager.$$$reportNull$$$0(), FlutterConsoleLogManager.java:-1
	io.flutter.logging.FlutterConsoleLogManager.getFullStringValue(), FlutterConsoleLogManager.java:-1
	io.flutter.logging.FlutterConsoleLogManager.processLoggingEvent(), FlutterConsoleLogManager.java:367
	io.flutter.logging.FlutterConsoleLogManager.lambda$handleLoggingEvent$2(), FlutterConsoleLogManager.java:335
	com.intellij.util.concurrency.QueueProcessor$RunnableConsumer.consume(), QueueProcessor.java:279
...
```

Note that the line numbers don't match up with our current source.

PR #4938, opened 2020-10-03
fix a null assertion in the inspector

devoncarew:
- fix a null assertion in the inspector
- fix https://github.com/flutter/flutter-intellij/issues/4936

This is being reported frequently to crash reporting; the code here hasn't changed, so perhaps we're seeing a timing difference in the vm service becoming available when starting an app.

The main issue here was that when we went to create an `InspectorService.ObjectGroup` in the FlutterConsoleLogManager class, the VM service connection wasn't always ready. We now delay construction until the ObjectGroup is first needed.


jacob314 (approved):
lgtm

PR #4944, opened 2020-10-04
address an issue with the JxBrowser singleton

devoncarew:
- address an issue with the JxBrowser singleton (we were creating it but not remembering it)
- some minor additions wrt nullability annotations

cc @helin24


helin24 (approved):
Thanks!

devoncarew:
Thanks for the review!

@stevemessick - the kokoro bot has failed twice on this PR (on the original PR, and then on a whitespace change). It's not immediately obvious to me why from looking at the log - any ideas?

stevemessick:
It looks like ant changed its current version. I'm not sure we even need ant anymore on Kokoro. Ant is only used to compile with Java 8, and I think Kokoro only uses Java 11.

devoncarew:
Thanks Steve! This is passing now.

PR #4945, opened 2020-10-05
adjust the layout for the jsbrowser error feedback ui

devoncarew:
- adjust the layout for the jsbrowser error feedback ui
- remove one or two null checks that are now no longer needed (since we've propagated our nullability annotations a bit further through the project)

This breaks the error feedback UI from a single long line into two shorter ones; this is to make it work better in (generally fairly narrow) tool windows.

Screen Shot 2020-10-04 at 3 55 08 PM

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -200,12 +200,18 @@ public class FlutterView implements PersistentStateComponent<FlutterViewState>, 
                                               boolean isEmbedded) {
     final ContentManager contentManager = toolWindow.getContentManager();
 
+    final String tabName;
     final FlutterDevice device = app.device();
-    final List<FlutterDevice> existingDevices = new ArrayList<>();
-    for (FlutterApp otherApp : perAppViewState.keySet()) {
-      existingDevices.add(otherApp.device());
+    if (device == null) {
+      tabName = app.getProject().getName();
     }
-    final String tabName = device.getUniqueName(existingDevices);
+    else {
+      final List<FlutterDevice> existingDevices = new ArrayList<>();
+      for (FlutterApp otherApp : perAppViewState.keySet()) {
+        existingDevices.add(otherApp.device());
+      }
+      tabName = device.getUniqueName(existingDevices);
+    }
 
     if (emptyContent != null) {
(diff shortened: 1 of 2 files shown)


devoncarew:
Helin - this PR isn't timely, just some useful UX changes. I fixed two test failures introduced by this PR but am having some trouble w/ the 3rd:

```
Wanted but not invoked:
devToolsManager.openBrowserIntoPanel(
    "http://www.testUrl.com",
    null,
    "Test Project Name",
    "inspector"
);
-> at io.flutter.view.FlutterViewTest.testHandleJxBrowserInstalledWithDevtoolsSucceeded(FlutterViewTest.java:101)

However, there were exactly 2 interactions with this mock:
devToolsManager.hasInstalledDevTools();
-> at io.flutter.view.FlutterView.presentDevTools(FlutterView.java:392)

devToolsManager.installDevTools();
-> at io.flutter.view.FlutterView.presentDevTools(FlutterView.java:398)
```

src/io/flutter/view/FlutterView.java
```diff
@@ -199,18 +200,12 @@ private void addBrowserInspectorViewContent(FlutterApp app,
...
-      tabName = app.getProject().getName();
-    }
-    else {
-      final List<FlutterDevice> existingDevices = new ArrayList<>();
-      for (FlutterApp otherApp : perAppViewState.keySet()) {
-        existingDevices.add(otherApp.device());
-      }
-      tabName = device.getUniqueName(existingDevices);
+    final List<FlutterDevice> existingDevices = new ArrayList<>();
+    for (FlutterApp otherApp : perAppViewState.keySet()) {
```
helin24, line 205: I'm seeing a null pointer exception here when debugging the test that failed, it looks like during the test `perAppViewState` is null. Is this likely to only happen in the test environment? Maybe we could put this retrieving of a unique name into a helper function.

devoncarew: I think that the `device` should be non-null for all normal invocations (normal app runs) but can be null for tests. I'll likely back out removing the null checks - it's not a critical back of this PR.

helin24 (approved):
Looks nice!

PR #4948, opened 2020-10-05
Add more detail for JxBrowser failed downloads

helin24:
This adds the exception message and elapsed time to analytics if one of the file downloads fails.

I was hoping to check bytes downloaded, but I couldn't find a way to do that using this API. I tried just checking the size of `new File(DOWNLOAD_PATH + "/" + currentFileName)` but it seems the file doesn't appear unless it's completely downloaded. Any other ideas on this?


jacob314 (approved):
lgtm

PR #4953, opened 2020-10-07
Fix dev channel build

stevemessick:
Specifying Java target compatibility apparently doesn't work they way I expect it to. Class files built by Java 11 that specify Java 8 compatibility still have class file version 55 so the plugin can't be used in Android Studio 4.0 and 4.1. That might be a bug in the IntelliJ plugin for Gradle.

There are a number of changes here:
- Add Java version to Gradle script, set it in the plugin tool based on build version
- Unroll the build loop to do a single build at a time, manually incrementing the minor version number
- Change the Java version halfway through
- Refactor two utility functions (devBuildNumber(), _nextRelease()) into util.dart
- Move build version constructor from a switch branch to buildVersionNumber() in util.dart
- Use buildVersionNumber() everywhere that needs to use that value
- Move ++pluginCount into the main loop

### DIFF
--- a/tool/kokoro/deploy.sh
+++ b/tool/kokoro/deploy.sh
@@ -4,7 +4,18 @@ source ./tool/kokoro/setup.sh
 setup
 
 echo "kokoro build start"
-./bin/plugin make --channel=dev
+
+./bin/plugin make --channel=dev -o4.0 -u -m1
+./bin/plugin make --channel=dev -o4.1 -u -m2
+
+curl https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_osx-x64_bin.tar.gz > ../java.tar.gz
+(cd ..; tar fx java.tar.gz)
+export JAVA_HOME=`pwd`/../jdk-11.0.2.jdk/Contents/Home
+export PATH=$PATH:$JAVA_HOME/bin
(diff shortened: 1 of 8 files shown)


tool/kokoro/deploy.sh
```diff
@@ -4,7 +4,18 @@ source ./tool/kokoro/setup.sh
 setup
 
 echo "kokoro build start"
-./bin/plugin make --channel=dev
+
+./bin/plugin make --channel=dev -o4.0 -u -m1
+./bin/plugin make --channel=dev -o4.1 -u -m2
+
+curl https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_osx-x64_bin.tar.gz > ../java.tar.gz
```
helin24, line 11: Is it worth moving this into a shared file? Or will it likely be unnecessary soon?

helin24 (approved):

tool/kokoro/deploy.sh
stevemessick: I considered doing that. I decided not to because it will make undoing this change easier. But it is going to be more difficult to maintain as we have to edit not just the product matrix but also these files. So I'll probably refactor it the next time the product matrix changes.

PR #4956, opened 2020-10-08
Use Java version

stevemessick:
Make the Gradle script use the correct version of Java.


stevemessick:
Travis failed because it needs to be updated to use Java 8 for two builds.

helin24 (approved):

PR #4963, opened 2020-10-13
Add analytics for download finished

helin24:
Last week I fixed downloading happening repeatedly on each open but didn't realize that "installed" is being recorded to analytics each time class loading happens (which happens even if a new download wasn't needed). This records to analytics only for each download so we can verify that downloading is only happening once per version.


devoncarew (approved):
lgtm as if, but you might also pull the "jxbrowser" string out into a constant - it looks like its referenced from ~3 places.

Issue #4952, opened 2020-10-07
Flutter plugin version 50.0 swallows errors that happen inside of widgets


marcglasberg:
For example, instead of this:

The ones who are not experiencing the problem are actually getting the error twice, once with the stacktrace and once without it:

```
════════ Exception caught by widgets library ════════ 
The following assertion was thrown building MyWidget(dirty):
Failed assertion: boolean expression must not be null
The relevant error-causing widget was: 
  MyWidget file:///C:/Users/Me/flutter_app/lib/my_widget/my_widget.dart:117:13
When the exception was thrown, this was the stack: 
#0      MyWidget.preambulo (package:xyz/participantes_widget.dart:44:27)
#1      MyWidget._varios (package:xyz/my_widget.dart:71:14)
#2      MyWidget.build (package:xyz/my_widget.dart:56:14)
#3      StatelessElement.build (package:flutter/src/widgets/framework.dart:4701:28)
#4      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:4627:15)
...
════════════════════════════════════

════════ Exception caught by widgets library ════════ 
Failed assertion: boolean expression must not be null
The relevant error-causing widget was: 
  MyWidget file:///C:/Users/Me/flutter_app/lib/my_widget/my_widget.dart:117:13
══════════════════════════════════════ 
```

You get only this:

```
════════ Exception caught by widgets library ════════ 
Failed assertion: boolean expression must not be null
The relevant error-causing widget was: 
  MyWidget file:///C:/Users/Me/flutter_app/lib/my_widget/my_widget.dart:117:13
══════════════════════════════════════ 
```

Note, it happens only in the IDE run tab. The terminal gives the complete information.

Also, if you downgrade the Flutter plugin from 50.0 to 49.04 the problem is gone. You can download the plugin in this page:
https://plugins.jetbrains.com/plugin/9212-flutter/versions/stable/96118

And the direct download link is this:
https://plugins.jetbrains.com/files/9212/96118/flutter-intellij.zip?updateId=96118&pluginId=9212&family=INTELLIJ

marcglasberg:
(ported this issue from https://github.com/flutter/flutter/issues/67418)

devoncarew:
Thanks for the report! We did do work in this area in the last month (https://github.com/flutter/flutter-intellij/pull/4870, https://github.com/flutter/flutter-intellij/pull/4867), but I don't see anything that would have changed behavior here.

[...]

PR #4966, opened 2020-10-14
fix an issue with displaying flutter errors in terse mode

devoncarew:
- fix an issue with displaying flutter errors in terse mode
- fix https://github.com/flutter/flutter-intellij/issues/4952

It looks like this had regressed - we'd been unintentionally printing out the 1st error in a frame in a terse mode.


jacob314 (approved):
Can we add a test for this?
lgtm

devoncarew:
> Can we add a test for this?

I'll dig around our existing tests and see if there's a good one to pattern after.

jacob314:
lgtm

PR #4976, opened 2020-10-19
minor tweaks to an error message

devoncarew:
- minor tweaks to the test seen when we have repeated issues starting the device daemon
- fix some grammar in another error message


jacob314 (approved):
lgtm

Issue #4967, opened 2020-10-14
Running tests causes an `RPCError` fairly often. 


gspencergoog:
## What happened

When running tests in the Flutter framework, but right clicking next to the `main()` declaration and clicking "Run tests in ...", this RPC error occurs fairly often now, probably every fourth time I try it or so.  It may have something to do with timing: it doesn't seem to happen if I'm not running tests one right after another. Maybe something is crashing and restarting when each test finishes, and it takes a moment to restart?

## Version information

IntelliJ IDEA `2020.2.2` • Flutter plugin `io.flutter 50.0` • Dart plugin `202.7319.5`

Flutter 1.23.0-19.0.pre.26 • channel nnbd_widgets_2 • git@github.com:gspencergoog/flutter.git
Framework • revision e46ec16b01 (4 hours ago) • 2020-10-14 10:47:44 -0700
Engine • revision adf5b59485
Tools • Dart 2.11.0 (build 2.11.0-217.0.dev)

## Exception

null

```
java.lang.Throwable: org.dartlang.vm.service.element.RPCError@178b3b7e
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:143)
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:147)
	at com.intellij.idea.IdeaLogger.error(IdeaLogger.java:71)
	at io.flutter.run.test.FlutterTestRunner$1$1.onError(FlutterTestRunner.java:119)
	at org.dartlang.vm.service.VmServiceBase.processResponse(VmServiceBase.java:629)
	at org.dartlang.vm.service.VmServiceBase.processMessage(VmServiceBase.java:422)
	at org.dartlang.vm.service.VmServiceBase$1.onMessage(VmServiceBase.java:86)
	at de.roderick.weberknecht.WebSocketReceiver.run(WebSocketReceiver.java:65)
```

devoncarew:
Thanks for the report!

Looks like you're on the latest Flutter plugin, and on an up-to-date version of the framework.

I'll see if I can repro locally. Is this with all tests? With tests in a certain project? Are there tests / projects where it doesn't repro?

I we can't repro I may put in more logging logic, and ask you to do something like switch to the plugin's dev channel (which auto-publishes from master every sunday morning).

gspencergoog:
It doesn't seem to depend on which test, and I don't see a pattern to which ones succeed and which don't. The same test will usually succeed if I try it again after it throws the exception.

[...]

Issue #4960, opened 2020-10-13
Flutter isolate test hangs


enchikiben:
## Steps to Reproduce

After plugin update> 45, flutter isolate test hangs:
```
import 'package:flutter/foundation.dart';
import 'package:flutter_test/flutter_test.dart';


dynamic testFunGlobal(var s) {
  return 111;
}

Future<dynamic> testFun() async {
  return await compute(testFunGlobal, "11");
}

void main() {
  test('test', () async {
    var a = await testFun();
    print(a);
  });
}
```

Run test icon:
![image](https://user-images.githubusercontent.com/1251978/95871653-56075800-0d87-11eb-9f4f-757e60c796ee.png)

Will be executed indefinitely and will not return any result.

## Version info

C:\projects\flutter\bin\flutter.bat doctor --verbose
[√] Flutter (Channel stable, 1.22.1, on Microsoft Windows [Version 10.0.19041.546], locale ru-RU)
    • Flutter version 1.22.1 at C:\projects\flutter
    • Framework revision f30b7f4db9 (5 days ago), 2020-10-08 10:06:30 -0700
    • Engine revision 75bef9f6c8
    • Dart version 2.10.1

[√] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at C:\Users\enchikiben\AppData\Local\Android\sdk
    • Platform android-29, build-tools 29.0.3
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)
    • All Android licenses accepted.

[√] Android Studio (version 3.6)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin version 49.0.1
    • Dart plugin version 192.8052
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)

[!] Connected device
    ! No devices available

! Doctor found issues in 1 category.
Process finished with exit code 0

devoncarew:
Thanks for the report! Can you try the latest plugin - version 50.0? That's only available in Android Studio 4.0 and later (I see that you're on AS 3.6 and Flutter plugin version 49.0).

enchikiben:
![image](https://user-images.githubusercontent.com/1251978/95881290-e2b71380-0d91-11eb-9aaa-008052e247c9.png)
I have the last one

devoncarew:
[...]

PR #4977, opened 2020-10-19
resume isolates at test start

devoncarew:
- fix an issue resuming isolates at the start of running tests
- fix https://github.com/flutter/flutter-intellij/issues/4967
- fix https://github.com/flutter/flutter-intellij/issues/4960

From the reported issue, and looking at the code, I believe the problem was that we were resuming all known isolates (reported by the VM object) at test start. However, there's a chance the test isolate might not be executable yet. I believe the state changes that isolates go through are (https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#streamlisten) `IsolateStart`, `IsolateRunnable`, and `PauseStart`. If you try and resume an isolate that's not yet runnable (or paused) the VM will return an error.

This changes to resuming all isolates that are `'PauseStart'` when we first connect, and all isolates for which we later received `'PauseStart'` events.

This alos changes from reporting issues via `LOG.error()` - which look like crashes to the user - to writing error messages to the run / test console.

@helin24 - this is for the `FlutterTestRunner` program runner. I haven't looked to see if we'll need similar changes for the bazel test runner as well.

### DIFF
--- a/src/io/flutter/run/test/FlutterTestRunner.java
+++ b/src/io/flutter/run/test/FlutterTestRunner.java
@@ -37,13 +37,12 @@ import io.flutter.sdk.FlutterSdk;
 import io.flutter.settings.FlutterSettings;
 import io.flutter.utils.JsonUtils;
 import io.flutter.utils.StdoutJsonParser;
+import io.flutter.utils.VmServiceListenerAdapter;
 import io.flutter.vmService.VmServiceConsumers;
+import io.flutter.vmService.VmServiceConsumers.EmptyResumeConsumer;
 import org.dartlang.vm.service.VmService;
 import org.dartlang.vm.service.consumer.VMConsumer;
-import org.dartlang.vm.service.element.ElementList;
-import org.dartlang.vm.service.element.IsolateRef;
-import org.dartlang.vm.service.element.RPCError;
-import org.dartlang.vm.service.element.VM;
+import org.dartlang.vm.service.element.*;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/test/FlutterTestRunner.java
```diff
@@ -37,13 +37,12 @@
...
+import io.flutter.utils.VmServiceListenerAdapter;
 import io.flutter.vmService.VmServiceConsumers;
+import io.flutter.vmService.VmServiceConsumers.EmptyResumeConsumer;
 import org.dartlang.vm.service.VmService;
 import org.dartlang.vm.service.consumer.VMConsumer;
-import org.dartlang.vm.service.element.ElementList;
-import org.dartlang.vm.service.element.IsolateRef;
-import org.dartlang.vm.service.element.RPCError;
-import org.dartlang.vm.service.element.VM;
+import org.dartlang.vm.service.element.*;
```
helin24, line 45: We may want to keep this expanded

helin24 (approved):

helin24:
Thanks for making this change! I was struggling to figure out why I didn't see events using a listener, but it turns out I was missing `vmService.streamListen`!

devoncarew:
> Thanks for making this change! I was struggling to figure out why I didn't see events using a listener, but it turns out I was missing `vmService.streamListen`!

Thanks!

Yeah, I think the docs at the vm service protocol (https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md) could be expanded - to cover common use cases, and things like isolate state changes when they start up, pause, and terminate.

src/io/flutter/run/test/FlutterTestRunner.java
devoncarew: sure, reverting this change

Issue #4771, opened 2020-08-22
Flutter Test shows <no name> instead of file name in Android Studio and IntelliJ


Peng-Qian:
### Problem
1. Run any test will show as following
![Screen Shot 2020-08-23 at 8 21 49 AM](https://user-images.githubusercontent.com/37643086/90965142-b6151700-e519-11ea-82ad-755ac349e669.png)

2. when running a single/group test by clicking the icon in the editor. 'SILENT_OBSERVATORY' will show up.
![Screen Shot 2020-08-23 at 8 27 10 AM](https://user-images.githubusercontent.com/37643086/90965233-76026400-e51a-11ea-8274-5145647cc25a.png)

jacob314:
Can you provide some details on what version of Flutter you are on and what action you took to run the tests?

Peng-Qian:
Hi @jacob314 ,

Thanks for your reply, please see the following details.

#### Flutter doctor
```
[✓] Flutter (Channel beta, 1.21.0-9.1.pre, on Mac OS X 10.15.6 19G2021, locale en-NZ)
 
[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
[✓] Xcode - develop for iOS and macOS (Xcode 11.6)
[✓] Android Studio (version 4.0)
[✓] IntelliJ IDEA Ultimate Edition (version 2020.2)
[✓] VS Code (version 1.46.0)
[✓] Connected device (1 available)
```
#### Plugin and IDE version
```
Flutter 48.1.2 for Android Studio 4.0.1 Build #AI-193.6911.18.40.6626763, built on June 25, 2020
Flutter 48.1.4 for IntelliJ IDEA 2020.2 (Ultimate Edition) Build #IU-202.6397.94, built on July 27, 2020
```

stevemessick:
@Peng-Qian does this problem occur if you use the stable channel of Flutter?

Peng-Qian:
Hi @stevemessick ,

In the stable channel, tests can display the filename, but the second problem still there.
![Screen Shot 2020-09-03 at 3 47 15 PM](https://user-images.githubusercontent.com/37643086/92069296-17cc5f80-edfd-11ea-9d55-68c016ac13dd.png)

stevemessick:
I'm using the master channel. It looks like the name is there but is hidden by "no name".

[...]

PR #4980, opened 2020-10-19
fix an issue with unnamed test groups

devoncarew:
- fix an issue with unnamed test groups (`<no name>`)
- fix https://github.com/flutter/flutter-intellij/issues/4771

@stevemessick - the only important change here is in `Group.from()` - updating for the change that unnamed groups changed to be reported w/ names of `""` instead of `null`. The other changes are just updating the `DartTestEventsConverterZ` class from the original version in the Dart plugin.


stevemessick (approved):
Thanks!

PR #4983, opened 2020-10-19
Show option to open DevTools in browser during JxBrowser installation

helin24:
If an app is started while JxBrowser is still installing, we present:
Screen Shot 2020-10-19 at 11 22 59 AM

If the installation times out:
Screen Shot 2020-10-19 at 11 22 16 AM

If something else goes wrong, like installation is interrupted:
Screen Shot 2020-10-19 at 11 15 46 AM

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -91,6 +91,8 @@ public class FlutterView implements PersistentStateComponent<FlutterViewState>, 
   protected static final String INSTALLATION_IN_PROGRESS_LABEL = "Installing JxBrowser and DevTools...";
   protected static final String INSTALLATION_TIMED_OUT_LABEL =
     "Waiting for JxBrowser installation timed out. Restart your IDE to try again.";
+  protected static final String INSTALLATION_STOPPED_LABEL =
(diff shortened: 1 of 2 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -91,6 +91,8 @@ public void dispose() {
   protected static final String INSTALLATION_IN_PROGRESS_LABEL = "Installing JxBrowser and DevTools...";
   protected static final String INSTALLATION_TIMED_OUT_LABEL =
     "Waiting for JxBrowser installation timed out. Restart your IDE to try again.";
+  protected static final String INSTALLATION_STOPPED_LABEL =
+    "The JxBrowser installation was stopped for unknown reasons. Restart your IDE to try again.";
```
jacob314, line 95: We should word smith this error message. 
How about
"The JxBrowser installation failed unexpectedly.  Restart your IDE to try again."
also change the label to 
JX_BROWSER_INSTALLATION_FAILED
to clarify this is a failure.

jacob314 (approved):
lgtm

src/io/flutter/view/FlutterView.java
helin24: This is slightly different because it's `installation.get()` that threw an exception, but the message makes sense.

PR #4999, opened 2020-10-26
Fork a shell to build old versions

### LINKED ISSUE #4921: Plugin tool needs to work with Java 8 and 11
M50 was difficult to produce. Each version of the plugin had to be built individually, instead of all at once, because half of them required Java 8 and the other half Java 11. I had to add a (long overdue) option to set the minor version number. It still needs a way to run with the proper JAVA_HOME setting for the two Javas. We will need this capability until Android Studio 4.2 is stable.

stevemessick:
Fixes #4921

### DIFF
--- a/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
+++ b/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
@@ -41,7 +41,7 @@ public class FlutterNewProjectAction extends AnAction implements DumbAware {
     if (NewWelcomeScreen.isNewWelcomeScreen(e)) {
       e.getPresentation().setIcon(getFlutterDecoratedIcon());
       e.getPresentation().setText(
-        Registry.is("use.tabbed.welcome.screen")
+        Registry.is("use.tabbed.welcome.screen", false)
         ? FlutterBundle.message("welcome.new.project.compact")
         : FlutterBundle.message("welcome.new.project.title"));
     }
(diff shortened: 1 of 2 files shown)


tool/plugin/lib/runner.dart
```diff
@@ -93,13 +93,39 @@ testing=$testing
...
+          result = await exec('./gradlew', command);
+        }
       }
     } finally {
       propertiesFile.writeAsStringSync(source);
     }
     return result;
   }
+
+  Future<int> runShellScript(List<String> command, BuildSpec spec) async {
```
helin24, line 114: Could this have a more descriptive name? Like `gradleWithOldJava`?

flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
```diff
@@ -41,7 +41,7 @@ public void update(@NotNull AnActionEvent e) {
     if (NewWelcomeScreen.isNewWelcomeScreen(e)) {
       e.getPresentation().setIcon(getFlutterDecoratedIcon());
       e.getPresentation().setText(
-        Registry.is("use.tabbed.welcome.screen")
+        Registry.is("use.tabbed.welcome.screen", false)
```
helin24, line 44: Is this a related change?

helin24 (approved):

flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
stevemessick: It fixes a problem when the item is not found.

tool/plugin/lib/runner.dart
stevemessick: Sure, I can change it. This won't be around very long though.

PR #5005, opened 2020-10-28
Build for M51

stevemessick:
I removed a few items in the change log dealing with CI stuff, otherwise just `git log --oneline` as always.

This also implements the bounded versioning for canary.


helin24 (approved):

product-matrix.json
```diff
@@ -32,7 +32,7 @@
       "ideaVersion": "202.6787931",
       "dartPluginVersion": "202.6397.47",
       "sinceBuild": "202.6397.94",
-      "untilBuild": "202.*",
+      "untilBuild": "202.7660.26",
```
devoncarew, line 35: Just to confirm, this does mean that we can only install into `2020.2.6397.94 - 2020.2.7660.26` of IntelliJ 2020.2 (it doesn't just limit us to AS 4.2 Canary versions)?

It looks like the current stable version of 2020.2 is:

```
IntelliJ IDEA 2020.2.3 (Community Edition)
Build #IC-202.7660.26, built on October 6, 2020
```

devoncarew (approved):

product-matrix.json
stevemessick: Correct. I tested that the plugin installs correctly everywhere. We'll be releasing a new build for canary when it comes out next month. If 2020.2.1 comes out we may have to do another.

stevemessick:
Kokoro has finished. Travis hasn't started. Not waiting for it.

PR #5006, opened 2020-10-29
Make embedded browser default enabled, not version-specific

helin24:
I changed the default value in `isEnableEmbeddedBrowsers` (the method above this change) but neglected to update the set method as well, so non-dev users would have been unable to disable the embedded browsers setting.

I verified with a non-dev version that the embedded browser option can be both enabled and disabled.


jacob314 (approved):
lgtm

PR #5007, opened 2020-10-29
Allow bazel projects to set up JxBrowser

helin24:
Previously we were exiting `ProjectOpenActivity` early for bazel projects so that JxBrowser files weren't installed. This change moves up JxBrowser installation in `ProjectOpenActivity` to after verifying that a project is either a Flutter project with a pubspec file or a bazel project.

### DIFF
--- a/src/io/flutter/ProjectOpenActivity.java
+++ b/src/io/flutter/ProjectOpenActivity.java
@@ -50,8 +50,15 @@ public class ProjectOpenActivity implements StartupActivity, DumbAware {
     if (FlutterUtils.isAndroidStudio() && AndroidUtils.isAndroidProject(project)) {
       AndroidUtils.addGradleListeners(project);
     }
-    if (!FlutterModuleUtils.declaresFlutter(project)) {
+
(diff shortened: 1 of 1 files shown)


src/io/flutter/ProjectOpenActivity.java
```diff
@@ -50,10 +50,17 @@ public void runActivity(@NotNull Project project) {
     if (FlutterUtils.isAndroidStudio() && AndroidUtils.isAndroidProject(project)) {
       AndroidUtils.addGradleListeners(project);
     }
-    if (!FlutterModuleUtils.declaresFlutter(project)) {
+
+    if (!FlutterModuleUtils.declaresFlutter(project) && !WorkspaceCache.getInstance(project).isBazel()) {
```
jacob314, line 54: The fact that 
FlutterModuleUtils.declaresFlutter(project) returns false for Bazel projects is probably a bug. We should fix that instead of working around it here. Audit all the other uses to make sure none will be problems for bazel.

jacob314 (approved):
lgtm

Issue #4982, opened 2020-10-19
ui guides have some subtle doubling of lines now


devoncarew:
This is for 2020.2 - I see some doubling of vertical lines when UI guides is enabled. Perhaps the default indent guides UI has shifted slightly where it points the lines?

<img width="344" alt="Screen Shot 2020-10-19 at 10 49 10 AM" src="https://user-images.githubusercontent.com/1269969/96492797-0696b100-11f9-11eb-8029-ee687609b2dd.png">

cc @jacob314 @helin24

PR #5009, opened 2020-10-29
Fix indent delta to match change to IntelliJ.

jacob314:
We could diverge from the IntelliJ default but the new default
seems fine and I like that the lines are now to the left of the cursor which is a bit clearer than when they were to the right.
Before:
Screen Shot 2020-10-29 at 10 00 20 AM
After:
Screen Shot 2020-10-29 at 3 33 00 PM

This is related to https://github.com/flutter/flutter-intellij/issues/4982
An existing CL already resolved the doubling aspect by ensuring our indent renderer wins but we still have the issue that our widget indent guides had diverged from the default indent guides in terms of where lines were drawn.


jacob314:
Fyi @InMatrix. This change modifies where indent guides are drawn to align with other indent guide lines in IntelliJ.

jacob314:
Delta is indent guides are now 2 pixels before the start of a line instead of 2 pixels after the start of a line.

stevemessick (approved):

PR #5016, opened 2020-11-02
Use github actions

stevemessick:
This runs unit tests on Windows! And does everything else that Travis does.

We can ignore the Kokoro failure. That problem is already fixed but I don't want to merge it until later. Travis hasn't even started yet so I cancelled it.

I still need to update the plugin tool to auto-generate the versions we build. I'll probably drop Travis in that PR.

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -1,6 +1,19 @@
 name: presubmit
 on: [push]
 jobs:
+  windows-test:
+    runs-on: windows-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: Setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+          java-package: jdk
+      - name: windows UNIT_TEST_BOT
+        run: ./tool/github.bat
+        shell: cmd
   presubmit:
     runs-on: ubuntu-latest
     strategy:
@@ -27,18 +40,3 @@ jobs:
         run: ./tool/github.sh
         env:
           BOT: ${{matrix.bot}}
-  windows-test:
-    runs-on: windows-latest
-    steps:
-      - name: checkout
-        uses: actions/checkout@v2
-      - name: Setup java
-        uses: actions/setup-java@v1
-        with:
-          java-version: '11'
-          java-package: jdk
-      - name: windows UNIT_TEST_BOT
-        run: ./tool/github.bat
-        shell: cmd
-        env:
-          BOT: UNIT_TEST_BOT
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -27,15 +27,6 @@ jobs:
         run: ./tool/github.sh
         env:
           BOT: ${{matrix.bot}}
-  mac-test:
-    runs-on: macos-latest
-    steps:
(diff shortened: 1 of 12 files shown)


stevemessick:
I see I need to do some clean up in the Windows script. I'll get that in the next PR with the plugin tool changes.

tool/github.sh
```diff
@@ -0,0 +1,78 @@
...
+
+export JAVA_HOME=$JAVA_HOME_11_X64
+
+git clone --depth 1 https://github.com/flutter/flutter.git ../flutter
+export PATH="$PATH":`pwd`/../flutter/bin:`pwd`/../flutter/bin/cache/dart-sdk/bin
+flutter config --no-analytics
+flutter doctor
+export FLUTTER_SDK=`pwd`/../flutter
+
+if [ "$IDEA_VERSION" = "4.0" -o "$IDEA_VERSION" = "4.1" ] ; then
```
devoncarew, line 18: What does this if statement do? Perhaps add a comment line (as bash if statements can be mysterious for people).

tool/github.sh
```diff
@@ -0,0 +1,78 @@
...
+# Get packages for the test data.
+(cd testData/sample_tests; echo "pub get `pwd`"; pub get)
+
+# Set up the plugin tool.
+(cd tool/plugin; echo "pub get `pwd`"; pub get)
+
+if [ "$DART_BOT" = true ] ; then
+
+  # analyze the Dart code in the repo
+  echo "pub global activate tuneup"
```
devoncarew, line 43: We should be able to convert this over to just `dart analyze` now (not necessarily for this PR).

tool/github.sh
```diff
@@ -0,0 +1,78 @@
...
+  pub global activate tuneup
+  pub global run tuneup
+
+  # ensure that the edits have been applied to template files (and their target
+  # files have been regenerated)
+  ./bin/plugin generate
+
+  # show any changed file
```
devoncarew, line 51: nit: files

.github/workflows/presubmit.yaml
```diff
@@ -0,0 +1,42 @@
+name: presubmit
+on: [push]
+jobs:
+  windows-test:
```
devoncarew, line 4: Should this be named `presubmit-windows`? Is this essentially the same as the presubmit bot below?

devoncarew (approved):
lgtm; some questions in-line

tool/github.sh
stevemessick: I updated the comment in the block.

tool/github.sh
stevemessick: That would be more readable.

tool/github.sh
stevemessick: lol, I copied that from travis.sh. It's been "file" for almost two years. I'll change it.

.github/workflows/presubmit.yaml
stevemessick: The presubmit jobs do version-specific builds. This one runs unit tests on Windows. It is more similar to the checker job but I want to have "windows" in the name. I think this name is descriptive.

stevemessick:
The Windows-based workflow is now disabled. Many unit tests don't run and I don't know enough about Windows to fix them.

Issue #5025, opened 2020-11-06
Plugin Issue


wahabshah009:
Thanks for the feedback! If your issue is related to the Flutter framework itself,
please open an issue at
[github.com/flutter/flutter](https://github.com/flutter/flutter/issues/new).

## What happened
To be honest i have no idea when and why this exception occurred still i want to contribute for bug fixation...

## Version information

Android Studio `4.1` • Flutter plugin `io.flutter 51.0.2` • Dart plugin `201.9245`

Error getting Flutter sdk information.

## Exception

null

```
com.teamdev.jxbrowser.engine.UserDataDirectoryCreationException: Failed to create the user data directory: C:\WINDOWS\TEMP\UserData\064f1223-c100-4184-84c4-17bdd219aa43
	at com.teamdev.jxbrowser.engine.internal.EngineImpl.newInstance(EngineImpl.java:203)
	at com.teamdev.jxbrowser.engine.Engine.newInstance(Engine.java:106)
	at io.flutter.jxbrowser.EmbeddedBrowser.openPanel(EmbeddedBrowser.java:29)
	at io.flutter.devtools.DevToolsInstance.lambda$openPanel$1(DevToolsManager.java:407)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:201)
	at com.intellij.openapi.application.impl.ApplicationImpl.runIntendedWriteActionOnCurrentThread(ApplicationImpl.java:831)
	at com.intellij.openapi.application.impl.ApplicationImpl.lambda$invokeLater$4(ApplicationImpl.java:310)
	at com.intellij.openapi.application.impl.FlushQueue.doRun(FlushQueue.java:80)
	at com.intellij.openapi.application.impl.FlushQueue.runNextEvent(FlushQueue.java:128)
	at com.intellij.openapi.application.impl.FlushQueue.flushNow(FlushQueue.java:46)
	at com.intellij.openapi.application.impl.FlushQueue$FlushNow.run(FlushQueue.java:184)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:764)
	at java.awt.EventQueue.access$500(EventQueue.java:98)
	at java.awt.EventQueue$3.run(EventQueue.java:715)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.security.AccessController.doPrivileged(Native Method)
[...]

PR #5035, opened 2020-11-10
Set user-data folder for embedded browser

helin24:
(maybe) fixes https://github.com/flutter/flutter-intellij/issues/5025

This changes `EmbeddedBrowser` to a singleton class so that we can maintain one JxBrowser `Engine` and `Browser` with a folder for the browser data.
Screen Shot 2020-11-10 at 10 09 58 AM

### DIFF
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -404,7 +404,7 @@ class DevToolsInstance {
 
     //noinspection CodeBlock2Expr
     ApplicationManager.getApplication().invokeLater(() -> {
-      new EmbeddedBrowser().openPanel(contentManager, tabName, url);
+      EmbeddedBrowser.getInstance().openPanel(contentManager, tabName, url);
     });
   }
 }
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -234,7 +234,7 @@ public class DevToolsManager {
     final String screen = null;
 
     if (devToolsInstance != null) {
-      devToolsInstance.openPanel(uri, contentManager, tabName, pageName);
+      devToolsInstance.openPanel(project, uri, contentManager, tabName, pageName);
     }
     else {
       @Nullable final OSProcessHandler handler =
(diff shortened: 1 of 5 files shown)


src/io/flutter/jxbrowser/EmbeddedBrowser.java
```diff
@@ -16,25 +16,46 @@
...
+    if (embeddedBrowser == null) {
+      embeddedBrowser = new EmbeddedBrowser();
+    }
+    return embeddedBrowser;
+  }
+
+  private Browser browser;
+
+  private EmbeddedBrowser() {
+    System.out.println(Paths.get(JxBrowserManager.DOWNLOAD_PATH + File.separatorChar + "user-data"));
```
devoncarew, line 42: log this instead?

src/io/flutter/jxbrowser/EmbeddedBrowser.java
```diff
@@ -16,25 +16,46 @@
...
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static com.teamdev.jxbrowser.engine.RenderingMode.HARDWARE_ACCELERATED;
 
 public class EmbeddedBrowser {
-  public void openPanel(ContentManager contentManager, String tabName, String url) {
+  private static EmbeddedBrowser embeddedBrowser;
+
+  @NotNull
+  public static EmbeddedBrowser getInstance() {
```
devoncarew, line 32: Do we want this to be an application singleton or a per project (per IntelliJ window) singleton? I'm guessing the 2nd one. There are several places in the code base where we have per-project singletons. You register them in the plugin's xml file and then have a slightly different singleton getter. See for example `FlutterReloadManager.getInstance(Project project)`.

devoncarew (approved):
I would experiment w/ a project singleton (I assume a project singleton is what we want here, but have less context on JxBrowser...).

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -404,7 +404,7 @@ public void openPanel(String serviceProtocolUri, ContentManager contentManager,
 
     //noinspection CodeBlock2Expr
     ApplicationManager.getApplication().invokeLater(() -> {
-      new EmbeddedBrowser().openPanel(contentManager, tabName, url);
+      EmbeddedBrowser.getInstance().openPanel(contentManager, tabName, url);
```
stevemessick, line 407: getInstance(project)
I think we'd want separate instances for each window.

src/io/flutter/jxbrowser/EmbeddedBrowser.java
helin24: Ack forgot about this! Changed to per-project.

src/io/flutter/devtools/DevToolsManager.java
helin24: Hmm.. I can't do it easily, so hopefully only one project per window.

stevemessick (approved):

PR #5044, opened 2020-11-12
Simplify the new project wizard for AS 4.2+

stevemessick:
For Android Studio 4.2 and later, change the new project wizard to this:

Screen Shot 2020-11-12 at 1 12 39 PM

This includes one goal for this quarter: adding a link on the first page of the wizard that points to the installation docs.

`FlutterNewProjectAction.replaceModuleBuilders()` (and `restoreModuleBuilders()`) does some tricky stuff with reflection. This could break, and our fallback is to simply stop using those methods. That would make the wizard look like it does in IntelliJ. There should be no changes to the IntelliJ wizard, although almost all the code is shared.

Android Studio has always highlighted the help message for the selected project type, so I found a way to do that here, too. 

FlutterNewProjectAction.java_stub is the previous version of `FlutterNewProjectAction`, which is required for AS 4.1 (see the change in product-matrix.json).

FYI @devoncarew @pq

### DIFF
--- a/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
+++ b/flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
@@ -11,9 +11,16 @@ import com.android.tools.idea.wizard.model.ModelWizardDialog;
 import com.intellij.icons.AllIcons;
 import com.intellij.ide.impl.NewProjectUtil;
 import com.intellij.ide.projectWizard.NewProjectWizard;
+import com.intellij.ide.projectWizard.ProjectCategory;
+import com.intellij.ide.util.projectWizard.ModuleBuilder;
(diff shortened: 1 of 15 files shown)


flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java
```diff
@@ -85,4 +112,129 @@ Icon getFlutterDecoratedIcon() {
...
+
+    originalModuleBuilders = null;
+    originalTypes = null;
+    originalModuleTypes = null;
+    originalModuleMap = null;
+  }
+
+  @NotNull
+  private ModuleBuilderFactory[] getFlutterModuleBuilders() {
+    // The order is not relevant. The builders are sorted by weight.
```
helin24, line 217: What do you mean by weight?

helin24 (approved):

PR #5046, opened 2020-11-13
Use runScript for bazel run

helin24:
This uses the updated `runScript`, which:
- Processes `bazel-options` as a string of all of the bazel arguments
- Handles the device-specific logic internally (so we can remove from the plugin)

### DIFF
--- a/src/io/flutter/run/bazel/BazelFields.java
+++ b/src/io/flutter/run/bazel/BazelFields.java
@@ -116,13 +116,13 @@ public class BazelFields {
   }
 
   @Nullable
-  private String getLaunchScriptFromWorkspace(@NotNull final Project project) {
+  private String getRunScriptFromWorkspace(@NotNull final Project project) {
     final Workspace workspace = getWorkspace(project);
-    String launchScript = workspace == null ? null : workspace.getLaunchScript();
-    if (launchScript != null) {
-      launchScript = workspace.getRoot().getPath() + "/" + launchScript;
+    String runScript = workspace == null ? null : workspace.getRunScript();
+    if (runScript != null) {
+      runScript = workspace.getRoot().getPath() + "/" + runScript;
     }
-    return launchScript;
+    return runScript;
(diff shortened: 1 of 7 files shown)


stevemessick (approved):

src/io/flutter/run/bazel/BazelFields.java
```diff
@@ -201,75 +201,38 @@ GeneralCommandLine getLaunchCommand(@NotNull Project project,
...
+    final StringBuilder fullBazelArgs = new StringBuilder(inputBazelArgs);
+
     // If the user hasn't overridden the flutter_build_mode, then
     if (!StringUtil.notNullize(bazelArgs).matches(".*--define[ =]flutter_build_mode.*")) {
+      if (!inputBazelArgs.isEmpty()) {
+        fullBazelArgs.append(" ");
+      }
 
       // Set the mode. This section needs to match the bazel versions of the flutter_build_mode parameters.
       if (enableReleaseMode) {
```
jacob314, line 215: Add a TODO to remove these defines and let the run script handle this part.

src/io/flutter/run/bazel/BazelFields.java
```diff
@@ -201,75 +201,38 @@ GeneralCommandLine getLaunchCommand(@NotNull Project project,
...
             // So far as flutter is concerned, there is no difference between debug mode and run mode;
             // the only difference is that a debug mode app will --start-paused.
-            commandLine.addParameters("--define", "flutter_build_mode=debug");
+            fullBazelArgs.append("--define=flutter_build_mode=debug");
             break;
         }
       }
     }
 
-    // User specified additional bazel arguments.
```
jacob314, line 228: great to see this go away!

jacob314 (approved):
lgtm
looks good. It will be nice to as a follow up switch to the standard run/debug/profile options instead of passing it in as a bazel specific flag.

PR #5048, opened 2020-11-14
Channel command: run 'flutter channel'

stevemessick:
The FlutterSdk.returnOutputOfQuery() is derived from queryFlutterConfigImpl(). The command to run becomes a parameter and post-processing is moved out of that method and into its caller.

### DIFF
--- a/src/io/flutter/sdk/FlutterSdk.java
+++ b/src/io/flutter/sdk/FlutterSdk.java
@@ -223,6 +223,10 @@ public class FlutterSdk {
     return new FlutterCommand(this, getHome(), FlutterCommand.Type.CONFIG, additionalArgs);
   }
 
+  public FlutterCommand flutterChannel() {
+    return new FlutterCommand(this, getHome(), FlutterCommand.Type.CHANNEL);
+  }
+
   public FlutterCommand flutterRun(@NotNull PubRoot root,
                                    @NotNull VirtualFile main,
                                    @NotNull FlutterDevice device,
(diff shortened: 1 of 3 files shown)


src/io/flutter/sdk/FlutterSdkChannel.java
```diff
@@ -0,0 +1,75 @@
...
+        case "master":
+          return MASTER;
+        case "dev":
+          return DEV;
+        case "beta":
+          return BETA;
+        case "stable":
+          return STABLE;
+        default:
+          throw new IllegalArgumentException("Unknown channel: " + name);
```
devoncarew, line 39: Do we want to throw here? Return null?

For a longer tail of users, the channel text will be just the git branch they're on for the flutter repo. We may want to handle unknown names, and map them to something like 'unknown' (instead of it being an error condition).

devoncarew (approved):
I think we should likely introduce an 'unknown' or similar channel - most users will be on the 4 official channels, but the channel result from flutter is an arbitrary git branch name.

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -478,72 +491,64 @@ public String queryFlutterConfig(String key, boolean useCachedValue) {
...
+      }
+      catch (JsonSyntaxException ignored) {
+      }
+    }
+    return null;
   }
 
-  private String queryFlutterConfigImpl(String key) {
-    final FlutterCommand command = flutterConfig("--machine");
+  // Do not run this on EDT.
```
helin24, line 516: what is EDT?

helin24 (approved):

stevemessick:
Event Dispatch Thread. Running long-running tasks without forking a new
thread reduces UI responsiveness.

And yes, Devon, I agree. I didn't know there were more than four channels
until I saw "unknown" in a recent issue.

stevemessick:
`AsyncRateLimiterTest` failed when run as a git hub action but ran as part of kokoro. Ignoring it.

I added an "unknown" channel.

PR #5064, opened 2020-11-18
Send SIGINT for bazel test processes

helin24:
SIGKILL is currently sent when a test process is stopped, but this can't be routed to remote processes spawned from the original one.

### DIFF
--- a/src/io/flutter/utils/MostlySilentOsProcessHandler.java
+++ b/src/io/flutter/utils/MostlySilentOsProcessHandler.java
@@ -12,6 +12,7 @@ import com.intellij.execution.process.OSProcessHandler;
 import com.intellij.execution.process.UnixProcessManager;
 import com.intellij.openapi.util.SystemInfo;
 import com.intellij.util.io.BaseOutputReader;
+import io.flutter.FlutterInitializer;
 import org.jetbrains.annotations.NotNull;
 
 /**
@@ -32,7 +33,12 @@ import org.jetbrains.annotations.NotNull;
  * for more information.
  */
 public class MostlySilentOsProcessHandler extends OSProcessHandler {
+  /*
+  This determines whether a soft kill (SIGINT) is sent to the process on destroy, instead of the default SIGKILL. SIGKILL can't be routed
+  to remote processes spawned from the original one.
+   */
   private boolean softKill;
+  private GeneralCommandLine commandLine;
 
   public MostlySilentOsProcessHandler(@NotNull GeneralCommandLine commandLine)
     throws ExecutionException {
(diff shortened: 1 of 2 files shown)


src/io/flutter/utils/MostlySilentOsProcessHandler.java
```diff
@@ -30,14 +32,33 @@
  * for more information.
  */
 public class MostlySilentOsProcessHandler extends OSProcessHandler {
+  private boolean softKill;
```
jacob314, line 35: Add a comment indicating why softKill should be preferred. 
Your comment for this CL is about the right level of granularity.
`SIGKILL is currently sent when a test process is stopped, but this can't be routed to remote processes spawned from the original one.`

src/io/flutter/utils/MostlySilentOsProcessHandler.java
```diff
@@ -30,14 +32,33 @@
...
   @Override
   protected BaseOutputReader.Options readerOptions() {
     return BaseOutputReader.Options.forMostlySilentProcess();
   }
+
+  @Override
+  protected void doDestroyProcess() {
+    final Process process = getProcess();
+    if (softKill && SystemInfo.isUnix && shouldDestroyProcessRecursively() && processCanBeKilledByOS(process)) {
+      UnixProcessManager.sendSigIntToProcessTree(process);
```
jacob314, line 58: Can we verify whether we were able to kill the process and fail if we weren't able to?

helin24: `UnixProcessManager.sendSigIntToProcessTree` will log a warning message and return false if the process isn't killed - what's a reasonable action to take if that's the case? Throw an exception or call `super.doDestroyProcess`?

jacob314: Lets kill the process completely but also log a message to analytics so we can diagnose what went wrong.

src/io/flutter/utils/MostlySilentOsProcessHandler.java
```diff
@@ -30,14 +33,43 @@
...
     return BaseOutputReader.Options.forMostlySilentProcess();
   }
+
+  @Override
+  protected void doDestroyProcess() {
+    final Process process = getProcess();
+    if (softKill && SystemInfo.isUnix && shouldDestroyProcessRecursively() && processCanBeKilledByOS(process)) {
+      final boolean result = UnixProcessManager.sendSigIntToProcessTree(process);
+      if (!result) {
+        FlutterInitializer.getAnalytics().sendEvent("process", "process kill failed: " + this.commandLine.getCommandLineString());
```
jacob314, line 67: don't need the full command line string which might identify the user, just the file name.

jacob314 (approved):
lgtm

helin24:
CC @jwren

Issue #5057, opened 2020-11-17
Exception


crystalztl:
## What happened

run flutter app

## Version information

IntelliJ IDEA `2020.2.3` • Flutter plugin `io.flutter 51.0.3` • Dart plugin `202.8070`

Flutter 1.22.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 1aafb3a8b9 (4 days ago) • 2020-11-13 09:59:28 -0800
Engine • revision 2c956a31c0
Tools • Dart 2.10.4

## Exception

null

```
com.teamdev.jxbrowser.engine.ChromiumProcessStartupFailureException: Process exit code: 139
	at com.teamdev.jxbrowser.engine.internal.EngineImpl.throwExceptionAndCleanup(EngineImpl.java:459)
	at com.teamdev.jxbrowser.engine.internal.EngineImpl.newInstance(EngineImpl.java:278)
	at com.teamdev.jxbrowser.engine.Engine.newInstance(Engine.java:106)
	at io.flutter.jxbrowser.EmbeddedBrowser.openPanel(EmbeddedBrowser.java:29)
	at io.flutter.devtools.DevToolsInstance.lambda$openPanel$1(DevToolsManager.java:407)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:201)
	at com.intellij.openapi.application.impl.ApplicationImpl.runIntendedWriteActionOnCurrentThread(ApplicationImpl.java:802)
	at com.intellij.openapi.application.impl.ApplicationImpl.lambda$invokeLater$4(ApplicationImpl.java:322)
	at com.intellij.openapi.application.impl.FlushQueue.doRun(FlushQueue.java:84)
	at com.intellij.openapi.application.impl.FlushQueue.runNextEvent(FlushQueue.java:132)
	at com.intellij.openapi.application.impl.FlushQueue.flushNow(FlushQueue.java:47)
	at com.intellij.openapi.application.impl.FlushQueue$FlushNow.run(FlushQueue.java:188)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:313)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:776)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:727)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:721)
[...]

PR #5065, opened 2020-11-19
Unset and disable embedded browser option if MacOS Big Sur

helin24:
Screen Shot 2020-11-19 at 10 59 34 AM

Temporarily fixes https://github.com/flutter/flutter-intellij/issues/5057

### DIFF
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.java
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.java
@@ -308,6 +308,9 @@ public class FlutterSettingsConfigurable implements SearchableConfigurable {
 
     // TODO(helinx): Remove this once JxBrowser is compatible with MacOS 11.
     myEnableEmbeddedBrowsersCheckBox.setEnabled(!SystemInfo.isMacOSBigSur);
+    if (SystemInfo.isMacOSBigSur) {
+      myEnableEmbeddedBrowsersCheckBox.setToolTipText("This option is not yet supported for MacOS Big Sur.");
+    }
 
     myOrganizeImportsOnSaveCheckBox.setEnabled(myFormatCodeOnSaveCheckBox.isSelected());
 
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.java
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.java
@@ -20,7 +20,6 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.ComboBox;
 import com.intellij.openapi.ui.FixedSizeButton;
 import com.intellij.openapi.ui.TextComponentAccessor;
-import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.io.FileUtilRt;
 import com.intellij.openapi.util.text.StringUtil;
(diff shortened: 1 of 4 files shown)


src/io/flutter/settings/FlutterSettings.java
```diff
@@ -269,11 +270,16 @@ public boolean isEnableHotUiInCodeEditor() {
   }
 
   public boolean isEnableEmbeddedBrowsers() {
-    return getPropertiesComponent().getBoolean(enableEmbeddedBrowsersKey, true);
+    // TODO(helinx): Change default value once JxBrowser is compatible with MacOS 11.
+    return getPropertiesComponent().getBoolean(enableEmbeddedBrowsersKey, !SystemInfo.isMacOSBigSur);
```
devoncarew, line 274: I think the 2nd param to this method is the default value if none is set. If they've previously set a value here, they'll get that value, which I assume we won't want for big sur.

I'd suggest just changing the start of this method to something like `if (SystemInfo.isMacOSBigSur) return false;`

src/io/flutter/settings/FlutterSettings.java
```diff
@@ -269,11 +270,16 @@ public boolean isEnableHotUiInCodeEditor() {
...
   }
 
   public void setEnableEmbeddedBrowsers(boolean value) {
-    getPropertiesComponent().setValue(enableEmbeddedBrowsersKey, value, true);
+    // TODO(helinx): Remove this and change default value below once JxBrowser is compatible with MacOS 11.
+    if (SystemInfo.isMacOSBigSur) {
```
devoncarew, line 279: I don't think we'll need to change the behavior of the setEnableEmbeddedBrowsers() w/ the change to isEnableEmbeddedBrowsers() I suggested above.

devoncarew (commented):
Having it disabled on big sure makes sense.

I have some comments on the FlutterSettings.java changes; also, it would be nice (but not critical), to change the tooltip for the settings checkbox on big sur to explain when the checkbox was disabled.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -305,6 +306,9 @@ public void reset() {
 
     myEnableEmbeddedBrowsersCheckBox.setSelected(settings.isEnableEmbeddedBrowsers());
 
+    // TODO(helinx): Remove this once JxBrowser is compatible with MacOS 11.
+    myEnableEmbeddedBrowsersCheckBox.setEnabled(!SystemInfo.isMacOSBigSur);
```
devoncarew, line 310: Perhaps also (conditionally on if we're on big sur):

```
myEnableEmbeddedBrowsersCheckBox.setToolTipText("This option not available...");
```

devoncarew (approved):

src/io/flutter/settings/FlutterSettings.java
helin24: I thought the setting would show up as checked but disabled, but it's always empty if disabled.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
helin24: Oh I didn't realize you could set the tool tip here! Thanks

stevemessick (approved):

PR #5066, opened 2020-11-19
Use existing devtools for bazel in panel and remove opening devtools for tests

helin24:
We're currently launching a new devtools server when opening the embedded browser, but for bazel projects it is faster to use the existing devtools server.

### DIFF
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -39,7 +39,6 @@ import io.flutter.utils.JsonUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -131,16 +130,16 @@ public class DevToolsManager {
     return result;
   }
 
-  public void openBrowser(Optional<FlutterApp> appOptional) {
-    openBrowserImpl(appOptional,null, null);
+  public void openBrowser(FlutterApp app) {
+    openBrowserImpl(app, null, null);
   }
 
-  public void openBrowserAndConnect(Optional<FlutterApp> appOptional, String uri) {
-    openBrowserAndConnect(appOptional, uri, null);
+  public void openBrowserAndConnect(FlutterApp app, String uri) {
+    openBrowserAndConnect(app, uri, null);
   }
 
-  public void openBrowserAndConnect(Optional<FlutterApp> appOptional, String uri, String screen) {
-    openBrowserImpl(appOptional, uri, screen);
+  public void openBrowserAndConnect(FlutterApp app, String uri, String screen) {
+    openBrowserImpl(app, uri, screen);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -237,23 +194,27 @@ public void openBrowserIntoPanel(String uri, ContentManager contentManager, Stri
       devToolsInstance.openPanel(project, uri, contentManager, tabName, pageName);
     }
     else {
-      @Nullable final OSProcessHandler handler =
-        isBazel(project) ? getProcessHandlerForBazel() : getProcessHandlerForPub();
+      if (isBazel(project)) {
+        try {
+          devToolsInstance = getDevToolsInstance().get(15, TimeUnit.SECONDS);
```
jacob314, line 199: why is there a 15 second timeout?

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -298,6 +242,36 @@ private void openBrowserImpl(String uri, String screen) {
     }
   }
 
+  private CompletableFuture<DevToolsInstance> getDevToolsInstance() {
+    final CompletableFuture<DevToolsInstance> instance = new CompletableFuture<>();
+    final Optional<FlutterApp> appsOptional =
+      FlutterApp.allFromProjectProcess(project).stream().filter((FlutterApp app) -> app.getProject() == project).findFirst();
```
jacob314, line 248: discussing offline if there is a more robust way to achieve the same result. I worry about all the extra edge cases this could introduce.

src/io/flutter/devtools/DevToolsManager.java
helin24: I can't remember what we decided on this - keep it? Otherwise it could wait indefinitely without raising an exception.

src/io/flutter/devtools/DevToolsManager.java
helin24: I tried passing an app through to this function, but it's kind of complicated. There are a couple issues:
- App has to be found this way when opening devtools from the run/debug tab, and this function is used from that workflow
- App doesn't exist when a test is running, but I think we still want to be able to open devtools for bazel projects during tests? I'm not sure about the best way to make requests to the daemon from a test.

src/io/flutter/run/OpenDevToolsAction.java
```diff
@@ -50,21 +51,29 @@ public void update(@NotNull final AnActionEvent e) {
...
+    if (project == null) {
       return;
     }
 
-    final DevToolsManager devToolsManager = DevToolsManager.getInstance(event.getProject());
+    final DevToolsManager devToolsManager = DevToolsManager.getInstance(project);
 
-    final Optional<FlutterApp> appOptional = FlutterUtils.findFlutterAppFromProject(event.getProject());
+    final Optional<FlutterApp> appOptional =
+      FlutterApp.allFromProjectProcess(project).stream().filter((FlutterApp app) -> app.getProject() == project).findFirst();
```
jacob314, line 62: This code still has the problem that we are finding all FlutterApps for the project and picking one rather than usinig the app for the app being debugged. Can we pick the right app?

helin24: Ah yeah, I can pass the app from where `OpenDevToolsAction` is created.

jacob314 (approved):
lgtm

PR #5068, opened 2020-11-20
Add dispatch events to run actions

stevemessick:
The `pull_request_target` should allow actions to run as expected on repo forks. The `workflow_dispatch` is supposed to enable manual triggering from the github UI.

### DIFF
--- a/src/io/flutter/module/settings/PlatformsForm.kt
+++ b/src/io/flutter/module/settings/PlatformsForm.kt
@@ -1,0 +1,2 @@
+package io.flutter.module.settings
+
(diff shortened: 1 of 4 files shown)


src/io/flutter/module/settings/PlatformsForm.kt
```diff
@@ -0,0 +1,2 @@
+package io.flutter.module.settings
```
helin24, line 1: is this related?

helin24 (approved):

src/io/flutter/module/settings/PlatformsForm.kt
stevemessick: I don't know how that got included.

PR #5085, opened 2020-12-01
Run devtools panel actions on UI thread

helin24:
May fix https://github.com/flutter/flutter-intellij/issues/5032

`addBrowserInspectorViewContent` should be run on the UI thread since it removes content from the panel and `app.getConnector().getBrowserUrl()` may be inaccurate when not run on the UI thread.

However, `DevToolsManager::openBrowserIntoPanel` has some actions that should be run on a non-UI thread since we have to wait for the devtools server to start.

### DIFF
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -251,13 +251,8 @@ public class DevToolsManager {
         }
       }
       else {
-        if (devToolsInstance != null) {
-          devToolsInstance.openBrowserAndConnect(uri, screen);
-        }
-        else {
-          @Nullable final OSProcessHandler handler = getProcessHandlerForPub();
-          startDevToolsServerAndConnect(handler, uri, screen);
-        }
+        @Nullable final OSProcessHandler handler = getProcessHandlerForPub();
+        startDevToolsServerAndConnect(handler, uri, screen);
       }
     });
   }
--- a/src/io/flutter/devtools/DevToolsManager.java
+++ b/src/io/flutter/devtools/DevToolsManager.java
@@ -190,6 +190,10 @@ public class DevToolsManager {
 
   public void openBrowserIntoPanel(FlutterApp app, String uri, ContentManager contentManager, String tabName, String pageName) {
     ApplicationManager.getApplication().executeOnPooledThread(() -> {
+      if (!project.isOpen()) {
+        return;
+      }
+
       final String screen = null;
 
       if (devToolsInstance != null) {
@@ -231,6 +235,10 @@ public class DevToolsManager {
     }
 
     ApplicationManager.getApplication().executeOnPooledThread(() -> {
+      if (!project.isOpen()) {
+        return;
+      }
+
       // For internal users, we can connect to the DevTools server started by flutter daemon. For external users, the flutter daemon has an
       // older version of DevTools, so we launch the server using `pub global run` instead.
       if (isBazel(project)) {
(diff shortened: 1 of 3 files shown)


src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -187,58 +187,62 @@ private boolean isBazel(Project project) {
   }
 
   public void openBrowserIntoPanel(FlutterApp app, String uri, ContentManager contentManager, String tabName, String pageName) {
-    final String screen = null;
+    ApplicationManager.getApplication().executeOnPooledThread(() -> {
+      final String screen = null;
```
jacob314, line 191: make sure project is still valid as soon as you get here.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -187,58 +187,62 @@ private boolean isBazel(Project project) {
...
-      catch (Exception e) {
-        LOG.info("Failed to get existing devToolsInstance");
-      }
-    }
-    else {
-      if (devToolsInstance != null) {
-        devToolsInstance.openPanel(project, uri, contentManager, tabName, pageName);
+      if (isBazel(project)) {
+        try {
+          getDevToolsInstance(app).get(15, TimeUnit.SECONDS).openPanel(project, uri, contentManager, tabName, pageName);
```
jacob314, line 195: make sure project is still valid after waiting.

src/io/flutter/devtools/DevToolsManager.java
```diff
@@ -187,58 +187,62 @@ private boolean isBazel(Project project) {
...
+        }
+        else {
+          @Nullable final OSProcessHandler handler = getProcessHandlerForPub();
+
+          if (handler != null) {
+            // start the server
+            DevToolsInstance.startServer(handler, instance -> {
+              devToolsInstance = instance;
+
+              devToolsInstance.openPanel(project, uri, contentManager, tabName, pageName);
```
jacob314, line 213: another time gap. Make sure the project is still valid before we open the panel.

src/io/flutter/view/FlutterView.java
```diff
@@ -430,24 +430,29 @@ private void presentDevTools(FlutterApp app, InspectorService inspectorService,
     final DevToolsManager devToolsManager = DevToolsManager.getInstance(app.getProject());
 
     if (devToolsManager.hasInstalledDevTools()) {
-      addBrowserInspectorViewContent(app, inspectorService, toolWindow, isEmbedded);
+      AsyncUtils.invokeLater(() -> addBrowserInspectorViewContent(app, inspectorService, toolWindow, isEmbedded));
```
jacob314, line 433: shouldn't this method instead be asserting that we are on the UI thread? seems on to have it start on some random thread.

src/io/flutter/view/FlutterView.java
```diff
@@ -430,24 +430,29 @@ private void presentDevTools(FlutterApp app, InspectorService inspectorService,
...
-          presentLabel(toolWindow, DEVTOOLS_FAILED_LABEL);
-        }
-      });
+      awaitDevToolsInstall(app, inspectorService, toolWindow, isEmbedded, result);
     }
   }
 
+  protected void awaitDevToolsInstall(FlutterApp app, InspectorService inspectorService, ToolWindow toolWindow, boolean isEmbedded, CompletableFuture<Boolean> installResult) {
+    AsyncUtils.whenCompleteUiThread(installResult, (succeeded, throwable) -> {
+      if (succeeded) {
```
jacob314, line 444: watch out! succeeded is a Boolean so it can be null (when the throwable != null). 
That is why I use the pattern of checking whether the throwable is null first in these callbacks to avoid bugs.
You could also use something like
https://commons.apache.org/proper/commons-lang/javadocs/api-2.4/org/apache/commons/lang/BooleanUtils.html#isTrue(java.lang.Boolean)
or
if (succeeded != null && succeeded)

src/io/flutter/view/FlutterView.java
helin24: Do you mean `addBrowserInspectorViewContent` should assert that we're on the UI thread? I can add that as well, but I think I have to use `invokeLater` here since `presentDevTools` isn't running on the UI thread.

jacob314 (approved):
lgtm

PR #5087, opened 2020-12-02
Remove launch script and devtools script

helin24:
We don't expect the test script to not be available, so this part has been changed to an assert instead of using a backup launch script.

With this change and a previous change to use `runScript` instead of `launchScript` to run apps, the launch script is no longer used.

Similarly, the devtools script is no longer being used, since bazel projects now use a daemon action to launch devtools.


jacob314 (approved):
lgtm

PR #5088, opened 2020-12-03
Update build for M52

stevemessick:
There weren't too many API changes this time.

I added support for unpacking MacOS zip files to get compilation artifacts. This will save money as I need to use a metered data plan to download Android Studio executables. Now I don't have to do it twice.


helin24 (approved):

stevemessick:
The failing tests indicate a class path problem, which I will look into later.

PR #5094, opened 2020-12-04
Stop devtools process on closing

helin24:
This stops the DevTools server when a project is closed (or IntelliJ quits). This change only applies to the external DevTools server, so a separate change is needed for the bazel case.


stevemessick (approved):

devoncarew (approved):
lgtm

Another way to do this would be to have DevToolsManager implement `Disposable`, close any open devtools server instance in the dispose method, and call `Disposer.register(project, devtoolsManager)` somewhere. I don't think this is better than the solution in this PR, just a different way of doing it.

Issue #5063, opened 2020-11-18
stanim template should put the super.initState in the first line


dark-chocolate:
`stanim` creates `initState` like this: 

```dart
@override
void initState() {
  _controller = AnimationController(vsync: this);
  super.initState();
} 
```

But if you read the [docs of initState](https://api.flutter.dev/flutter/widgets/State/initState.html), it clearly mentions that 

>If you override this, make sure your method starts with a call to super.initState().

jacob314:
This should be a trivial fix and will avoid putting users on the wrong path.
Fyii @pq would be nice if we had lints for this and we verified that all our templates were lint free.

Issue #4647, opened 2020-06-23
Build liveTemplates xml from dart code


helin24:
See: https://github.com/flutter/flutter-intellij/pull/4636#discussion_r442335424

PR #5101, opened 2020-12-07
update the live templates

devoncarew:
Some updates to our live templates:

- move the super.initState() call to the first line of the initState() override (fix https://github.com/flutter/flutter-intellij/issues/5063)
- move the system of record for the live templates to separate files in the `resources/liveTemplates` directory; this will make reviewing changes easier. The individual templates are copied into the `resources/liveTemplates/flutter_miscellaneous.xml` file as part of the `./bin/plugin generate` build step (fix https://github.com/flutter/flutter-intellij/issues/4647)


devoncarew:
I see this failing w/:

```
BUILD FAILED in 3m 11s
  
  Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
  Use '--warning-mode all' to show the individual deprecation warnings.
  See https://docs.gradle.org/6.6.1/userguide/command_line_interface.html#sec:command_line_warnings
...
```

We may need to upgrade out gradle scripts? Provision an older version of gradle on the bots?

tool/plugin/lib/plugin.dart
```diff
@@ -589,21 +591,21 @@ https://plugins.jetbrains.com/plugin/uploadPlugin
...
+/// at the end. When the EAP version is released that needs to be updated.
 class GenerateCommand extends ProductCommand {
   final BuildCommandRunner runner;
 
   GenerateCommand(this.runner) : super('generate');
 
-  String get description =>
-      'Generate a valid plugin.xml and .travis.yml for the Flutter plugin.\n'
-      'The plugin.xml.template and product-matrix.json are used as input.';
+  String get description => 'Generate plugin.xml, .travis.yml, and '
```
stevemessick, line 605: Nit: we don't use travis anymore. We generate the github actions file now.

tool/plugin/lib/plugin.dart
```diff
@@ -624,6 +627,48 @@ class GenerateCommand extends ProductCommand {
...
+      final name = p.basenameWithoutExtension(file.path);
+
+      var replaceContents = file.readAsStringSync();
+      replaceContents = replaceContents
+          .replaceAll('\n', '&#10;')
+          .replaceAll('<', '&lt;')
+          .replaceAll('>', '&gt;');
+
+      // look for '<template name="$name" value="..."'
+      final regexp = RegExp('<template name="$name" value="([^"]+)"');
```
stevemessick, line 655: Maybe add a README explaining double quotes are not allowed and new templates have to be manually added to the xml file?

tool/plugin/lib/plugin.dart
```diff
@@ -624,6 +627,48 @@ class GenerateCommand extends ProductCommand {
...
+      final match = regexp.firstMatch(contents);
+      if (match == null) {
+        throw 'No entry found for "$name" live template in ${templateFile.path}';
```
stevemessick, line 658: I don't understand this constraint. It makes it hard to add new live templates. I understand we need to control the size of the pop-up but this means we have to manually add a new one to the xml file.

tool/plugin/lib/plugin.dart
```diff
@@ -624,6 +627,48 @@ class GenerateCommand extends ProductCommand {
...
+      }
+
+      // Replace the existing content in the xml live template file with the
+      // contwnt from the template $name.txt file.
```
stevemessick, line 662: Nit: "content"

stevemessick (approved):
This will make live template management much easier.

My comments are all in the nice-to-have category, not deal-breakers.

stevemessick:
You can ignore the message from Gradle about deprecated features; we always get that. The failure is due to 34 unit tests that fail. It is probably down to two now, since I merged the fix for that problem.

stevemessick:
I restarted Kokoro to see if the tests are fixed.

tool/plugin/lib/plugin.dart
devoncarew: updates

tool/plugin/lib/plugin.dart
devoncarew: Good idea - I'll add one.

tool/plugin/lib/plugin.dart
devoncarew: Yeah, this is only here so that we can provide feedback if an entry isn't found during the copy / in-line stage. We don't have enough info in the txt files to create a full xml entry in the live templates file.

When adding a new template, people will need to create a new entry both in the xml file and create a new txt file. I added that into - how to add a new live template - to the readme.

tool/plugin/lib/plugin.dart
devoncarew: done

PR #5112, opened 2020-12-09
Add devtools service

helin24:
This change is to start running a DevTools server for the duration of a project instead of having it associated with an application. 

Currently we don't use the daemon to serve DevTools for non-bazel projects, but this should be okay once we're using `pub global run devtools` in flutter tools.

### DIFF
--- a/src/io/flutter/run/daemon/DaemonApi.java
+++ b/src/io/flutter/run/daemon/DaemonApi.java
@@ -73,6 +73,10 @@ public class DaemonApi {
     return send("daemon.getSupportedPlatforms", new DaemonGetSupportedPlatforms(projectRoot));
   }
 
+  CompletableFuture<Boolean> daemonShutdown() {
+    return send("daemon.shutdown", new DaemonShutdown());
+  }
+
   CompletableFuture<RestartResult> restartApp(@NotNull String appId, boolean fullRestart, boolean pause, @NotNull String reason) {
     return send("app.restart", new AppRestart(appId, fullRestart, pause, reason));
   }
@@ -401,6 +405,14 @@ public class DaemonApi {
     @Override
     RestartResult parseResult(JsonElement result) {
       return GSON.fromJson(result, RestartResult.class);
+    }
+  }
+
+  private static class DaemonShutdown extends Params<Boolean> {
+    @Nullable
+    @Override
+    Boolean parseResult(@Nullable JsonElement result) {
+      return true;
     }
   }
 
--- a/src/io/flutter/run/daemon/DaemonApi.java
+++ b/src/io/flutter/run/daemon/DaemonApi.java
@@ -74,13 +74,7 @@ public class DaemonApi {
   }
 
   CompletableFuture<Boolean> daemonShutdown() {
-    return send("daemon.shutdown", new Params<Boolean>() {
-      @Nullable
-      @Override
-      Boolean parseResult(@Nullable JsonElement result) {
-        return true;
-      }
-    });
+    return send("daemon.shutdown", new DaemonShutdown());
   }
 
   CompletableFuture<RestartResult> restartApp(@NotNull String appId, boolean fullRestart, boolean pause, @NotNull String reason) {
--- a/src/io/flutter/run/daemon/DaemonApi.java
+++ b/src/io/flutter/run/daemon/DaemonApi.java
@@ -74,7 +74,13 @@ public class DaemonApi {
   }
 
   CompletableFuture<Boolean> daemonShutdown() {
-    return send("daemon.shutdown", new DaemonShutdown());
+    return send("daemon.shutdown", new Params<Boolean>() {
+      @Nullable
+      @Override
+      Boolean parseResult(@Nullable JsonElement result) {
+        return true;
+      }
+    });
   }
 
   CompletableFuture<RestartResult> restartApp(@NotNull String appId, boolean fullRestart, boolean pause, @NotNull String reason) {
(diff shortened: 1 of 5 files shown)


src/io/flutter/run/daemon/DaemonApi.java
```diff
@@ -404,6 +408,14 @@ RestartResult parseResult(JsonElement result) {
     }
   }
 
+  private static class DaemonShutdown extends Params<Boolean> {
```
jacob314, line 411: no need for an inner class. Nust write as a Lambda.

src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -0,0 +1,134 @@
...
+  private DevToolsInstance devToolsInstance;
+
+  @NotNull
+  public static DevToolsService getInstance(@NotNull final Project project) {
+    return ServiceManager.getService(project, DevToolsService.class);
+  }
+
+  private DevToolsService(@NotNull final Project project) {
+    this.project = project;
+    try {
```
jacob314, line 45: have you tested that this code gracefully handles changing your flutter project?

src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -0,0 +1,134 @@
...
+    if (workspace != null) {
+      final String script = workspace.getDaemonScript();
+      if (script != null) {
+        return createCommand(workspace.getRoot().getPath(), script, ImmutableList.of(), androidHome);
+      }
+
+    }
+
+    // Otherwise, use the Flutter SDK.
+    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
```
jacob314, line 93: what happens if the flutter sdk changes once this is already initialized?

src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -0,0 +1,134 @@
...
+        daemonApi.daemonShutdown();
+        if (!process.isProcessTerminated()) {
+          process.destroyProcess();
+        }
+      }
+    });
+  }
+
+  private static GeneralCommandLine chooseCommand(@NotNull final Project project) {
+    final String androidHome = IntelliJAndroidSdk.chooseAndroidHome(project, false);
```
jacob314, line 80: why are we passing android home in. That seems like it has nothing to do with serving devtools. We should either share this code with the regular code to launch the daemon or remove code that is unrelated to the launching a daemon just to serve devtools.

src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -0,0 +1,134 @@
...
+import static com.jetbrains.lang.dart.ide.runner.server.vmService.VmServiceWrapper.LOG;
+
+public class DevToolsService {
+  private class DevToolsServiceListener implements DaemonEvent.Listener {
+  }
+
+  @NotNull private final Project project;
+  private DaemonApi daemonApi;
+  private ProcessHandler process;
+  private DevToolsInstance devToolsInstance;
```
jacob314, line 36: How can a caller of this know whether the devtools instance has started yet or not?
Perhaps vend a CompletableFuture with the devToolsInstance or a boolean indicating the service has started.
Also, how is the service stopped iff a project is closed?

src/io/flutter/run/daemon/DevToolsService.java
helin24: Hmm.. I'm not sure what this would look like, i.e. how does someone change the project? If they open a new project, the project will get its own devtools server.

src/io/flutter/run/daemon/DevToolsService.java
helin24: Ah I didn't understand why this was relevant for DeviceService (where I copied this code from), but I think I can just delete this.

src/io/flutter/run/daemon/DevToolsService.java
helin24: At this point nothing - the server will keep running and presumably there aren't many cases where users need a different version of devtools while working on a project. Are you thinking of some situations where this would be a problem?

src/io/flutter/run/daemon/DaemonApi.java
helin24: I think I'm not doing this correctly - see above (ln 77). A class is required that implements parseResult, is there a way to do that with a lambda?

src/io/flutter/run/daemon/DevToolsService.java
jacob314: if you changed your flutter sdk the version of the devtools server much change so you want to make sure you have the one from the new sdk not the old one.

src/io/flutter/run/daemon/DevToolsService.java
jacob314: if you have two projects open,  having two devtools servers is fine. what I'm worried about is if you open a new project and close your existing project. do we make sure we stop the first devtools server?

helin24: Oh right, I used `ProjectManager.getInstance().addProjectManagerListener` to listen for project closing. In there `daemon.shutdown` is called, which stops the devtools server.

src/io/flutter/run/daemon/DevToolsService.java
helin24: Hmm... I'm not sure how to manage this situation. Do we have examples elsewhere of checking whether the flutter sdk has changed?

I'm also not sure about whether users change their flutter SDK frequently enough to monitor for this? My guess was that SDK would change if users update their flutter version or if they have multiple flutter downloads, but it would also surprise me that the implementation of DevTools would change significantly enough between those versions that a user would need the newest DevTools while in the middle of working on a project. Sounds like there may be some cases I'm not thinking of?

src/io/flutter/run/daemon/DaemonApi.java
jacob314: ah you are right. I missed the extends.

jacob314 (approved):
lgtm

Issue #5034, opened 2020-11-10
out of memory with big parts


benoitskipr:
## Steps to Reproduce

working within big files that have big (or maybe multiple?) `part`  s makes editing laggy until AS runs out of memory, doesn't seem to happen in VS Code.

some_class.dart
```part 'some_class.d.dart';```

changing it to :
```
import 'some_class_part.dart';
export 'some_class_part.dart';
```
instantly makes editing faster, no more lag or out of memory errors

I'm not the only one noticing it : https://stackoverflow.com/questions/64734912/android-studio-running-out-of-memory-while-editing-big-multipart-dart-files

## Version info

[✓] Flutter (Channel unknown, 1.22.2, on Mac OS X 10.15.7 19H15, locale en-BE)
 
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
[✓] Xcode - develop for iOS and macOS (Xcode 12.1)
[✓] Chrome - develop for the web
[!] Android Studio (version 4.1)
    ✗ Flutter plugin not installed; this adds Flutter specific functionality.
    ✗ Dart plugin not installed; this adds Dart specific functionality.
[✓] VS Code (version 1.51.0)
 
[✓] Connected device (2 available)            
    ! Error: Benoit’s iPhone is not connected. Xcode will continue when Benoit’s iPhone is connected. (code -13)

(happens as well on 1.22.2 stable)

alexander-doroshko:
Memory snapshot may help to find memory consumers (Help -> Diagnostics Tools -> Capture Memory Snapshot)

benoitskipr:
@alexander-doroshko I'm missing that last menu item, could it be 'Analyze memory usage' one level up ?

alexander-doroshko:
Maybe Capture Memory Snapshot is only in later IDE versions. 'Analyze memory usage' could help as well (in the menu or right in the Out Of Memory dialog).

benoitskipr:
[memoryusagereport.txt](https://github.com/flutter/flutter-intellij/files/5519366/memoryusagereport.txt)
hope I did it correcly
I'll try to create a repro project
thanks

4uzhoy:
Faced up with that problem today, after updating flutter sdk to **1.22.4**, if downgrade to 
[√] Flutter (Channel stable, **1.20.2,** on Microsoft Windows [Version 10.0.18363.1198], locale ru-RU)
problem disappears, for me this more option than changing 'part' to import-export

[...]

PR #5115, opened 2020-12-10
fix an issue with leaking subscriptions

devoncarew:
- fix an issue with leaking subscriptions
- see https://github.com/flutter/flutter-intellij/issues/5034 for more info

Previously, every call to Action.update() (which happens frequently in IntelliJ), would create and leak a new stream subscription. I don't know that this resolves https://github.com/flutter/flutter-intellij/issues/5034, but it will at least address one issue.


jacob314 (approved):
lgtm

PR #5116, opened 2020-12-11
Open DevTools directly using pub for external projects

helin24:
Eventually we want to use the daemon to start the DevTools server in all cases, but we'll need to start from pub until the daemon method is updated in flutter_tools (and users have an updated version of flutter). It will be easier to always use the DevTools server from this new `DevToolsService` (one server per project) instead of sometimes relying on `DevToolsManager` (one server per app). And having a running server will make it easier to set up a DevTools port while running apps for bazel projects.

This is mostly copied from `DevToolsManager` but simplified since we don't need status messages and everything can run in a background thread.

Next steps will be using this server instead of starting a new one on app startup and when the DevTools action is triggered.


src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -45,6 +53,19 @@ public static DevToolsService getInstance(@NotNull final Project project) {
 
   private DevToolsService(@NotNull final Project project) {
     this.project = project;
+
+    ApplicationManager.getApplication().executeOnPooledThread(() -> {
+      // TODO(helinx): Also use daemon method for later flutter versions where devtools.serve uses pub global run devtools
```
jacob314, line 58: later -> latter
instead of devtools.serve I assume you mean `flutter daemon`?

src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -45,6 +53,19 @@ public static DevToolsService getInstance(@NotNull final Project project) {
 
   private DevToolsService(@NotNull final Project project) {
     this.project = project;
+
```
jacob314, line 56: doesn't this file need an API to query to make sure devtools is done serving? Otherwise we'll launch devtools URLs that fail? a user could run an app very quickly after opening the project and hit a case where devtools isn't fully running yet.

helin24: Yes, I was thinking a method like `getDevToolsUrl` would return a future but wasn't planning to add in this change. Is that what you mean?

src/io/flutter/run/daemon/DevToolsService.java
helin24: I meant later here, i.e. flutter versions after the change in flutter/flutter. I hopefully clarified daemon/devtools.serve in the updated comments

src/io/flutter/run/daemon/DevToolsService.java
jacob314: Yes. That is what I mean.

jacob314 (approved):
lgtm

PR #5121, opened 2020-12-15
Use project DevTools server for embedded browser

helin24:
This changes switches over the embedded browser to using the DevTools server that starts on project open rather than starting an instance of DevTools server associated with an app.

For bazel projects this currently opens a local daemon instance, but this will be changed once the run script is updated.


src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -71,6 +72,17 @@ private DevToolsService(@NotNull final Project project) {
     });
   }
 
+  public DevToolsInstance getDevToolsInstance() {
```
jacob314, line 75: expose the Future directly instead of providing a method that might lock the UI thread.

src/io/flutter/view/FlutterView.java
```diff
@@ -429,35 +438,24 @@ protected void handleJxBrowserInstalled(FlutterApp app, InspectorService inspect
...
+    presentLabel(toolWindow, INSTALLING_DEVTOOLS_LABEL);
+
+    openInspectorWithDevTools(app, inspectorService, toolWindow, isEmbedded);
   }
 
-  protected void awaitDevToolsInstall(FlutterApp app, InspectorService inspectorService, ToolWindow toolWindow, boolean isEmbedded, DevToolsManager devToolsManager) {
+  protected void openInspectorWithDevTools(FlutterApp app, InspectorService inspectorService, ToolWindow toolWindow, boolean isEmbedded) {
     ApplicationManager.getApplication().executeOnPooledThread(() -> {
-      final CompletableFuture<Boolean> result = devToolsManager.installDevTools();
+      final DevToolsInstance instance = DevToolsService.getInstance(myProject).getDevToolsInstance();
```
jacob314, line 449: simpler pattern:
rather than starting a pooled thread just call
`AsyncUtils.whenCompleteUiThread(DevToolsService.getInstance(myProject).getDevToolsInstance(), ...)`
and remove the invokeLater.

src/io/flutter/run/daemon/DevToolsService.java
helin24: I originally wanted this function to manage restarting if the daemon command fails, but that can be done elsewhere too

src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -49,7 +49,7 @@
   @NotNull private final Project project;
   private DaemonApi daemonApi;
   private ProcessHandler process;
-  private CompletableFuture<DevToolsInstance> devToolsInstance = new CompletableFuture<>();
+  public CompletableFuture<DevToolsInstance> devToolsInstance = new CompletableFuture<>();
```
jacob314, line 52: nit: this is java. Keep private and provide a getter.

src/io/flutter/view/FlutterView.java
```diff
@@ -445,16 +445,20 @@ private void presentDevTools(FlutterApp app, InspectorService inspectorService,
   }
 
   protected void openInspectorWithDevTools(FlutterApp app, InspectorService inspectorService, ToolWindow toolWindow, boolean isEmbedded) {
-    ApplicationManager.getApplication().executeOnPooledThread(() -> {
-      final DevToolsInstance instance = DevToolsService.getInstance(myProject).getDevToolsInstance();
+    AsyncUtils.whenCompleteUiThread(DevToolsService.getInstance(myProject).devToolsInstance, (instance, error) -> {
+      // TODO(helinx): Restart DevTools server if there's an error.
+      if (error != null) {
```
jacob314, line 450: make sure the project has not gone away while you were waiting for the devToolsServer to start.

jacob314 (approved):
lgtm

PR #5127, opened 2020-12-16
Revert "Add devtools service (#5112)" and dependencies

helin24:
This reverts commits:
- 492a7196babddef8fd201023023e9131cf129b72
- b5c83b7e402afd9890f99db876b81d2d05dbe6a8
- c932db7872b398a0be4e6e4067e85d06673a0b6d


jacob314:
lgtm

helin24:
The initial problem I was trying to fix here was DevTools opening in Chrome on project open, with flutter/flutter at head and the latest dev release plugin.

I'm actually going to close this revert because this was already fixed by https://github.com/flutter/flutter-intellij/commit/b5c83b7e402afd9890f99db876b81d2d05dbe6a8 - using the old way to start DevTools for non-bazel projects, not using `devtools.serve` through the daemon.

The next changes for non-bazel projects will be
- Changing back to using the daemon and `devtools.serve` (for later flutter versions), but passing along an argument to ensure that the browser isn't opened
- Starting server lazily instead of on project open

PR #5130, opened 2020-12-17
Start DevTools server on first request and handle exceptions

helin24:
We were previously starting a DevTools server on project open; this changes to starting the server when it's first requested (or if a previous attempt to start failed).

This change also adds exception logging and completion for various error conditions while starting the server.

### DIFF
--- a/src/io/flutter/run/daemon/DevToolsService.java
+++ b/src/io/flutter/run/daemon/DevToolsService.java
@@ -39,6 +39,7 @@ import org.jetbrains.annotations.NotNull;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
 
 public class DevToolsService {
   private static final Logger LOG = Logger.getInstance(DevToolsService.class);
@@ -49,7 +50,7 @@ public class DevToolsService {
   @NotNull private final Project project;
   private DaemonApi daemonApi;
   private ProcessHandler process;
-  private CompletableFuture<DevToolsInstance> devToolsInstanceFuture;
+  private AtomicReference<CompletableFuture<DevToolsInstance>> devToolsFutureRef = new AtomicReference<>(null);
 
   @NotNull
   public static DevToolsService getInstance(@NotNull final Project project) {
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -92,23 +100,21 @@ private void setUpWithDaemon() {
...
     catch (ExecutionException e) {
-      e.printStackTrace();
+      logExceptionAndComplete(e);
     }
 
     ProjectManager.getInstance().addProjectManagerListener(project, new ProjectManagerListener() {
       @Override
       public void projectClosing(@NotNull Project project) {
-        devToolsInstance = new CompletableFuture<>();
+        devToolsInstance = null;
```
devoncarew, line 117: Do we need to dispose() the devToolsInstance here?

src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -58,7 +58,19 @@ public static DevToolsService getInstance(@NotNull final Project project) {
 
   private DevToolsService(@NotNull final Project project) {
     this.project = project;
+  }
+
+  public CompletableFuture<DevToolsInstance> getDevToolsInstance() {
+    // Create instance if it doesn't exist yet, or if the previous attempt failed.
+    if (devToolsInstance == null || devToolsInstance.isCompletedExceptionally()) {
+      devToolsInstance = new CompletableFuture<>();
+      startServer();
```
devoncarew, line 67: Consider passing the devToolsInstance as a parameter to startServer() so that we always complete the right future (given that this field is reused over time for new instances).

src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -49,7 +49,7 @@
   @NotNull private final Project project;
   private DaemonApi daemonApi;
   private ProcessHandler process;
-  private CompletableFuture<DevToolsInstance> devToolsInstance = new CompletableFuture<>();
+  private CompletableFuture<DevToolsInstance> devToolsInstance;
```
devoncarew, line 52: Perhaps rename to `devToolsInstanceFuture` for clarity?

devoncarew (approved):
A few comments about potential code style changes.

src/io/flutter/run/daemon/DevToolsService.java
helin24: Is dispose mainly for if an object has dependencies? Since the instance just holds a hostname and port, I think it may not be needed.

src/io/flutter/run/daemon/DevToolsService.java
helin24: I'm not sure if this fully addresses the problem, but I put the future into an atomic reference to ensure that there can only be one at a time (one future, and therefore one process / process handler). It could be problematic if more than one server gets started and the reference to shut it down on close gets lost.

src/io/flutter/run/daemon/DevToolsService.java
devoncarew: Gotcha (yeah, it's just needed for heavier weight things).

PR #5131, opened 2020-12-17
Remove install SDK button from NPW

stevemessick:
Depends on #5070


devoncarew (approved):
lgtm

Some of the technique for in-line progress display here might be useful for the work to show SDK self-init progress.

PR #5138, opened 2020-12-17
Allow devtools to run remotely for bazel (WIP)

helin24:
Instead of lazy loading the DevTools server, I'm instead starting one on project open (for bazel only) so that the port will be available at runtime.

This is suboptimal though (someone could run an app before DevTools starts). Alternatives:
- add a separate script (like `runScript`) to manage any port setup and call this after the server starts (more divergence between bazel and regular flutter tools, too many scripts?)
- wait for DevTools to finish (app running could be slower, would have to move to a pooled thread)


jacob314:
We should avoid diverging the bazel and normal workflows if possible. When they are consistent user feedback from one workflow benefits the other. Lets discuss how to make lazy loading work fine with both workflows.

helin24:
Closing for now, as serving locally is okay unless other requirements come up.

PR #5141, opened 2020-12-21
Use DevTools service for open action

helin24:
This switches to using the project-level DevTools server when opening DevTools from the run console area.

### DIFF
--- a/src/io/flutter/run/OpenDevToolsAction.java
+++ b/src/io/flutter/run/OpenDevToolsAction.java
@@ -67,7 +67,7 @@ public class OpenDevToolsAction extends DumbAwareAction {
         return;
       }
 
-      if (project == null) {
+      if (project.isDisposed()) {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/OpenDevToolsAction.java
```diff
@@ -57,30 +60,23 @@ public void actionPerformed(@NotNull final AnActionEvent event) {
...
         return;
       }
 
-      if (devToolsManager.hasInstalledDevTools()) {
-        devToolsManager.openBrowserAndConnect(myApp, urlString);
-      }
-      else {
-        final CompletableFuture<Boolean> result = devToolsManager.installDevTools();
-        result.thenAccept(o -> devToolsManager.openBrowserAndConnect(myApp, urlString));
+      if (project == null) {
```
jacob314, line 70: you also need to check whether the project has been disposed. I don't think checking whether the project is null does much.

jacob314 (approved):
lgtm

Issue #5139, opened 2020-12-19
UserDataDirectoryAlreadyInUseException


vladostaci:
## What happened
I was just writing my code. Nothing special. It might be related to another issue I'm having which I described here https://github.com/flutter/flutter-intellij/issues/3456#issuecomment-748031624

## Version information

Android Studio `4.1.1` • Flutter plugin `io.flutter 52.1.1` • Dart plugin `201.9245`

Flutter 1.22.5 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 7891006299 (9 days ago) • 2020-12-10 11:54:40 -0800
Engine • revision ae90085a84
Tools • Dart 2.10.4

## Exception

null

```
com.teamdev.jxbrowser.engine.UserDataDirectoryAlreadyInUseException: The user data directory is already in use: C:\Users\user_name\AppData\Roaming\Google\AndroidStudio4.1\plugins\flutter-intellij\jxbrowser\user-data\flutter-project
	at com.teamdev.jxbrowser.engine.internal.EngineImpl.newInstance(EngineImpl.java:209)
	at com.teamdev.jxbrowser.engine.Engine.newInstance(Engine.java:106)
	at io.flutter.jxbrowser.EmbeddedBrowser.<init>(EmbeddedBrowser.java:51)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at com.intellij.serviceContainer.ConstructorInjectionKt.instantiateUsingPicoContainer(constructorInjection.kt:53)
	at com.intellij.serviceContainer.ComponentManagerImpl.instantiateClassWithConstructorInjection(ComponentManagerImpl.kt:660)
	at com.intellij.serviceContainer.ServiceComponentAdapter.createAndInitialize(ServiceComponentAdapter.kt:54)
	at com.intellij.serviceContainer.ServiceComponentAdapter.doCreateInstance(ServiceComponentAdapter.kt:41)
	at com.intellij.serviceContainer.BaseComponentAdapter.getInstanceUncached(BaseComponentAdapter.kt:115)
[...]

PR #5142, opened 2020-12-21
Close embedded browser engine on application close

helin24:
May fix https://github.com/flutter/flutter-intellij/issues/5119, https://github.com/flutter/flutter-intellij/issues/5139

It seems that sometimes the browser fails to start because the directory assigned to JxBrowser is already in use. Closing the engine should stop the use of the directory.


jacob314 (approved):
lgtm

jacob314 (approved):
definitely cleaner.
lgtm

PR #5143, opened 2020-12-21
Use project-level DevTools server for overflow open

helin24:
This is an action available from the Java inspector in the overflow menu.


jacob314 (approved):
lgtm

PR #5144, opened 2020-12-21
Use DevTools service in performance action and delete manager

helin24:
This replaces the open DevTools action in the performance tab and allows us to fully delete DevToolsManager.


src/io/flutter/performance/FlutterPerformanceView.java
```diff
@@ -201,8 +204,21 @@ private void addPerformanceViewContent(FlutterApp app, ToolWindow toolWindow) {
...
     openDevtools.setListener((linkLabel, data) -> {
-      final DevToolsManager devToolsManager = DevToolsManager.getInstance(app.getProject());
-      devToolsManager.openToScreen(app, null);
+      AsyncUtils.whenCompleteUiThread(DevToolsService.getInstance(app.getProject()).getDevToolsInstance(), (instance, ex) -> {
+        if (ex != null) {
+          LOG.error(ex);
+          return;
+        }
+
+        if (app.getProject().isDisposed()) {
```
jacob314, line 213: we should actually move the getProject().isDisposed() checks above logging errors here and elsewhere. If your project has been disposed we don't need to scare the user with an error likely due to the project being closed.

jacob314 (approved):
lgtm

PR #5148, opened 2020-12-22
Send exception if DevTools server fails to start

helin24:
I looked at our exceptions logged to analytics and it seems they probably all pass through `Analytics::sendException`. I wanted to log exceptions from `DevToolsService` since this is replacing a previous system and I'd like to see if there are common error paths.

### DIFF
--- a/src/io/flutter/run/daemon/DevToolsService.java
+++ b/src/io/flutter/run/daemon/DevToolsService.java
@@ -24,7 +24,9 @@ import com.intellij.openapi.project.ProjectManager;
 import com.intellij.openapi.project.ProjectManagerListener;
 import com.intellij.openapi.util.Key;
 import com.intellij.openapi.util.io.FileUtil;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.CharsetToolkit;
+import io.flutter.FlutterInitializer;
 import io.flutter.FlutterUtils;
 import io.flutter.bazel.Workspace;
 import io.flutter.bazel.WorkspaceCache;
@@ -240,6 +242,7 @@ public class DevToolsService {
 
   private void logExceptionAndComplete(Exception exception) {
     LOG.error(exception);
+    FlutterInitializer.getAnalytics().sendException(StringUtil.getThrowableText(exception), false);
     final CompletableFuture<DevToolsInstance> future = devToolsFutureRef.get();
     if (future != null) {
       future.completeExceptionally(exception);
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -240,6 +242,7 @@ private void logExceptionAndComplete(String message) {
 
   private void logExceptionAndComplete(Exception exception) {
     LOG.error(exception);
+    FlutterInitializer.getAnalytics().sendException(StringUtil.getThrowableText(exception), false);
```
jacob314, line 245: We need to first audit these exceptions as part of this cl and change the name so it is clear we are logging the exceptions to analytics. Currently they are including the project name and a file system path both of which should not be included in exceptions that are always logged to analytics.

helin24: I don't think the project name and system path will be logged to analytics - this line makes the exception into text (which only has names of files like `io.flutter.run.daemon.DevToolsService`) and then goes through the `sendPayload` analytics function, which appends info like plugin version. Am I missing analytics getting sent through another path?

helin24: (example of what would go into analytics description field)
```
java.lang.Exception: test exception
	at io.flutter.run.daemon.DevToolsService.logExceptionAndComplete(DevToolsService.java:241)
	at io.flutter.run.daemon.DevToolsService.access$500(DevToolsService.java:46)
	at io.flutter.run.daemon.DevToolsService$2.onTextAvailable(DevToolsService.java:188)
	at jdk.internal.reflect.GeneratedMethodAccessor61.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Metho
```

jacob314: The calls that concern me slightly are 
        logExceptionAndComplete("Unable to find daemon command for project: " + project);
and 
logExceptionAndComplete("Handler was null for command: " + command.toString());
that could expose a file path on their machine.

helin24: Ohhh I see, I'll check through these.

helin24: I just removed the contextual information that could be private. It might be nice to have this information locally logged if users report bugs, but I can always add that back later.

jacob314 (approved):
lgtm

PR #5153, opened 2020-12-28
Track time lapsed to start analysis server and finish indexing

helin24:
We want to find the time lapsed to start the analysis server and finish indexing, and send these values to analytics.

 Example times:
```
Has computed errors: 6890
Indexing has finished: 28634
```

### DIFF
--- a/src/io/flutter/ProjectOpenActivity.java
+++ b/src/io/flutter/ProjectOpenActivity.java
@@ -42,14 +42,13 @@ import org.jetbrains.annotations.NotNull;
 public class ProjectOpenActivity implements StartupActivity, DumbAware {
   public static final ProjectType FLUTTER_PROJECT_TYPE = new ProjectType("io.flutter");
   private static final Logger LOG = Logger.getInstance(ProjectOpenActivity.class);
-  private boolean indexingFinished = false;
 
   public ProjectOpenActivity() {
   }
 
   @Override
   public void runActivity(@NotNull Project project) {
-    project.getService(TimeTracker.class).setProjectOpenTime();
+    TimeTracker.getInstance(project).onProjectOpen();
 
     // TODO(messick): Remove 'FlutterUtils.isAndroidStudio()' after Android Q sources are published.
     if (FlutterUtils.isAndroidStudio() && AndroidUtils.isAndroidProject(project)) {
@@ -78,17 +77,11 @@ public class ProjectOpenActivity implements StartupActivity, DumbAware {
 
     // Report time when indexing finishes.
     DumbService.getInstance(project).runWhenSmart(() -> {
-      if (!indexingFinished) {
-        final Long millisSinceProjectOpen = project.getService(TimeTracker.class).millisSinceProjectOpen();
-        if (millisSinceProjectOpen != null) {
-          FlutterInitializer.getAnalytics().sendEventMetric(
-            "startup",
-            "indexingFinished",
-            millisSinceProjectOpen.intValue()
-          );
-        }
-        indexingFinished = true;
-      }
+      FlutterInitializer.getAnalytics().sendEventMetric(
+        "startup",
+        "indexingFinished",
+        project.getService(TimeTracker.class).millisSinceProjectOpen()
+      );
     });
 
     ApplicationManager.getApplication().executeOnPooledThread(() -> {
--- a/src/io/flutter/ProjectOpenActivity.java
+++ b/src/io/flutter/ProjectOpenActivity.java
@@ -21,6 +21,7 @@ import com.intellij.openapi.project.ProjectTypeService;
 import com.intellij.openapi.startup.StartupActivity;
 import com.intellij.openapi.ui.Messages;
 import icons.FlutterIcons;
+import io.flutter.analytics.TimeTracker;
 import io.flutter.bazel.WorkspaceCache;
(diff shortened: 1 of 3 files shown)


jacob314:
The analysis server doesn't depend on indexing. Indexing is an intellij only concept. If we had the time, we would rewrite the Dart + Flutter plugins for IntelliJ to not depend on IntelliJ indexing at all using the analysis server instead. If the indexing times are much longer than the analysis server time that is an interesting data point. Users running vscode instead would only have to wait for the analysis server time.

src/io/flutter/ProjectOpenActivity.java
```diff
@@ -40,12 +42,15 @@
...
   public static final ProjectType FLUTTER_PROJECT_TYPE = new ProjectType("io.flutter");
   private static final Logger LOG = Logger.getInstance(ProjectOpenActivity.class);
+  private boolean indexingFinished = false;
 
   public ProjectOpenActivity() {
   }
 
   @Override
   public void runActivity(@NotNull Project project) {
+    project.getService(TimeTracker.class).setProjectOpenTime();
```
jacob314, line 52: We usually great a getService method on classes instead that passes in a Project instance. That is a bit cleaner than calling getService on the project itself.

src/io/flutter/ProjectOpenActivity.java
```diff
@@ -71,6 +76,21 @@ public void runActivity(@NotNull Project project) {
     // TODO(messick) Re-enable this after dropping support for 2020.2.
     //excludeAndroidFrameworkDetector(project);
 
+    // Report time when indexing finishes.
+    DumbService.getInstance(project).runWhenSmart(() -> {
+      if (!indexingFinished) {
```
jacob314, line 81: Is this method ever actually called with (indexingFinished == true)?
If you are trying to catch an internal logic error, change this to 
assert(!indexingFinished) to clarify the intent.

src/io/flutter/analytics/TimeTracker.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.intellij.openapi.components.Service;
+import com.intellij.openapi.project.Project;
+
+@Service
+public final class TimeTracker {
+  private final Project project;
+  private Long projectOpenTime;
+
+  public TimeTracker(Project project) {
```
jacob314, line 16: make this like the other services in the project that have a getInstance method that takes a project.

src/io/flutter/analytics/TimeTracker.java
```diff
@@ -0,0 +1,30 @@
...
+    this.project = project;
+  }
+
+  public void setProjectOpenTime() {
```
jacob314, line 20: nit: avoid starting methods that aren't setters with the word set. Maybe call this: onProjectOpen()

src/io/flutter/analytics/TimeTracker.java
```diff
@@ -0,0 +1,30 @@
...
+    this.projectOpenTime = System.currentTimeMillis();
+  }
+
+  public Long millisSinceProjectOpen() {
```
jacob314, line 24: Avoid methods that return boxed numbers. Add an assert that projectOpenTime != null and return a long.
Perhaps return 0 if the project hasn't been opened.

src/io/flutter/dart/FlutterDartAnalysisServer.java
```diff
@@ -77,7 +81,25 @@ public void serverConnected(String s) {
           sendSubscriptions();
         }
       }
+
+      @Override
+      public void computedErrors(String file, List<AnalysisError> errors) {
+        if (!hasComputedErrors) {
+          final Long millisSinceProjectOpen = project.getService(TimeTracker.class).millisSinceProjectOpen();
```
jacob314, line 88: add a separate isProjectOpen instead of using a null Long to indicate unopened.

src/io/flutter/dart/FlutterDartAnalysisServer.java
```diff
@@ -77,7 +81,25 @@ public void serverConnected(String s) {
...
+          if (millisSinceProjectOpen != null) {
+            FlutterInitializer.getAnalytics().sendEventMetric(
+              "startup",
+              "analysisComputedErrors",
+              millisSinceProjectOpen.intValue()
```
jacob314, line 93: if you have to convert the long to an int, return it as an int instead.

src/io/flutter/ProjectOpenActivity.java
helin24: Oh right. I don't need this, was thinking this might run multiple times but that's not the case.

src/io/flutter/dart/FlutterDartAnalysisServer.java
```diff
@@ -84,22 +84,18 @@ public void serverConnected(String s) {
...
-              "startup",
-              "analysisComputedErrors",
-              millisSinceProjectOpen.intValue()
-            );
-          }
+        if (!hasComputedErrors && project.isOpen()) {
+          FlutterInitializer.getAnalytics().sendEventMetric(
+            "startup",
+            "analysisComputedErrors",
+            project.getService(TimeTracker.class).millisSinceProjectOpen()
```
jacob314, line 91: use TimeTracker.getInstance here as well.

jacob314 (approved):
lgtm
with one nit.

PR #5155, opened 2020-12-29
Add field for Dart entrypoint in bazel configuration form

helin24:
This adds the option to use a Dart entrypoint instead of a bazel target. There are cases where the bazel build process will fail with error messages in the console (e.g. when the project has not been set up to indicate what the associated targets are or if an entrypoint is linked to multiple targets). The error messages currently point to relevant documentation but we may be able to make project-specific suggestions with later (internal) changes.

Another next step is to allow for adding Dart entrypoints to saved configuration files that we can use to generate equivalent VSCode configuration files.

Screen Shot 2021-06-09 at 2 47 03 PM

### DIFF
--- a/src/io/flutter/run/bazel/BazelFields.java
+++ b/src/io/flutter/run/bazel/BazelFields.java
@@ -56,7 +56,7 @@ public class BazelFields {
    * The Bazel target or Dart file to invoke.
    */
   @Nullable
-  private final String bazelOrDartTarget;
+  private final String target;
 
   /**
    * Whether or not to run the app with --define flutter_build_mode=release.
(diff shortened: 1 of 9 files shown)


jacob314 (commented):
What is the value of the "use dart file" checkbox? We could tell from a target whether it is a dart file of a build rule so we should just allow both to be entered and validate that the target is one of the allowed types?

helin24:
> What is the value of the "use dart file" checkbox? We could tell from a target whether it is a dart file of a build rule so we should just allow both to be entered and validate that the target is one of the allowed types?

Previously I was thinking we needed this so that the dart field would show a file browser while the original field would not, but just having one field with a file browser works.

src/io/flutter/run/bazel/BazelFields.java
```diff
@@ -139,19 +129,9 @@ public String getAdditionalArgs() {
     return additionalArgs;
   }
 
-
   @Nullable
-  public String getBazelTarget() {
```
jacob314, line 134: nit:  make this just `getTarget()` with a comment that it is either a bazel or dart target.

jacob314 (approved):
lgtm

Issue #5152, opened 2020-12-25
java.lang.UnsatisfiedLinkError attemepting to use Flutter Inspector


rhencke:
## What happened

I opened the Flutter Inspector panel.  UI still is responsive, but panel is stuck showing 'Installing DevTools...'.

## Version information

IntelliJ IDEA `2020.3` • Flutter plugin `io.flutter 52.1.5` • Dart plugin `203.6912`

Flutter 1.25.0-8.1.pre • channel beta • https://github.com/flutter/flutter.git
Framework • revision 8f89f6505b (10 days ago) • 2020-12-15 15:07:52 -0800
Engine • revision 92ae191c17
Tools • Dart 2.12.0 (build 2.12.0-133.2.beta)

## Exception

null

```
java.lang.UnsatisfiedLinkError: /tmp/JxBrowser/7.12/libbrowsercore_toolkit.so: Error relocating /tmp/JxBrowser/7.12/libbrowsercore_toolkit.so: __sbrk: symbol not found
	at java.base/java.lang.ClassLoader$NativeLibrary.load0(Native Method)
	at java.base/java.lang.ClassLoader$NativeLibrary.load(ClassLoader.java:2442)
	at java.base/java.lang.ClassLoader$NativeLibrary.loadLibrary(ClassLoader.java:2498)
	at java.base/java.lang.ClassLoader.loadLibrary0(ClassLoader.java:2694)
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2627)
	at java.base/java.lang.Runtime.load0(Runtime.java:768)
	at java.base/java.lang.System.load(System.java:1837)
	at com.teamdev.jxbrowser.internal.JniLibrary.lambda$new$0(JniLibrary.java:29)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at com.teamdev.jxbrowser.internal.JniLibrary.<init>(JniLibrary.java:28)
	at com.teamdev.jxbrowser.internal.ToolkitLibrary.<init>(ToolkitLibrary.java:38)
	at com.teamdev.jxbrowser.internal.ToolkitLibrary.getInstance(ToolkitLibrary.java:28)
	at com.teamdev.jxbrowser.internal.xz.JniResourceExtractor.extract(JniResourceExtractor.java:50)
	at com.teamdev.jxbrowser.internal.xz.XzExtractor.extract(XzExtractor.java:53)
[...]

PR #5156, opened 2020-12-29
Enable JxBrowser logging with verbose logging option

helin24:
Sometimes verbose logs would be helpful to the JxBrowser team when they're debugging issues (e.g. for https://github.com/flutter/flutter-intellij/issues/5152)

These logs will be available in idea.log files (they will not be reported to our analytics).


jacob314 (approved):
lgtm

PR #5173, opened 2021-01-13
Rename to "Flutter DevTools"

helin24:
This change addresses https://github.com/flutter/devtools/issues/2345

Screen Shot 2021-01-13 at 9 55 41 AM
Screen Shot 2021-01-13 at 10 01 35 AM

### DIFF
--- a/src/io/flutter/run/OpenDevToolsAction.java
+++ b/src/io/flutter/run/OpenDevToolsAction.java
@@ -18,12 +18,12 @@ import io.flutter.devtools.DevToolsUtils;
 import io.flutter.run.daemon.DevToolsService;
 import io.flutter.run.daemon.FlutterApp;
 import io.flutter.utils.AsyncUtils;
-import io.flutter.utils.FlutterModuleUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 public class OpenDevToolsAction extends DumbAwareAction {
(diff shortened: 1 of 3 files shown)


stevemessick:
I think the UI affordance only shows up after the Flutter app is launched. So it won't be "Dart". But the hover text for the Dart icon should have the same name IMO.

You might want to change the description and text in plugin.xml, just for consistency. It is defined twice, and I'm not sure why.

src/io/flutter/run/OpenDevToolsAction.java
```diff
@@ -34,7 +35,11 @@ public OpenDevToolsAction() {
   }
 
   public OpenDevToolsAction(@NotNull final FlutterApp app, @NotNull final Computable<Boolean> isApplicable) {
-    super("Open DevTools", "Open Dart DevTools", FlutterIcons.Dart_16);
+    super(
+      "Open DevTools",
```
kenzieschmoll, line 39: can we make this text the same?
` "Open " + (FlutterModuleUtils.declaresFlutter(app.getProject()) ? "Flutter" : "Dart") + " DevTools",`

jacob314: This is the Flutter plugin so I don't think you can actually run an app unless it declares Flutter.

helin24:
> It is defined twice, and I'm not sure why.

I tried removing it from the `OpenDevToolsAction` but then there is no icon, title, or description. Not sure why..

kenzieschmoll (approved):

PR #5176, opened 2021-01-14
Start DevTools server and pass address to run

helin24:
We want to keep using a DevTools server that persists across app runs (associated with an open project instead), but we also want to pass the address of this server to flutter_tools during `flutter run` so that upcoming features like DevTools links generated in flutter_tools can use the same server.

However, I'm not sure when we should start the project DevTools server:
- If we should start it right before a user wants to run an application, this introduces a delay for the run process starting, and this runs on the UI thread (maybe there's a way to move to other thread?)
- If we start it when a project is opened, this is unnecessary if the user isn't running an app. If the user tries to run an app quickly, this could still result in a delay
- If we don't change the start time, then it will start when the embedded browser tries to display, which would cause the first app run to create an app-affiliated DevTools server (later app runs would use the project-affiliated server)
- (resolved)

We decided to use a modal dialog while waiting for DevTools server to start (shown the first time an app is run):
https://user-images.githubusercontent.com/6379305/105253671-c830cb00-5b34-11eb-8d0b-c43eda84e8bf.mov

(This is the flutter_tools change I'll be gating on: https://github.com/flutter/flutter/pull/73903)
CC @kenzieschmoll

### DIFF
--- a/src/io/flutter/run/SdkFields.java
+++ b/src/io/flutter/run/SdkFields.java
@@ -8,6 +8,7 @@ package io.flutter.run;
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.configurations.RuntimeConfigurationError;
+import com.intellij.openapi.progress.ProgressManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.ArrayUtil;
@@ -27,6 +28,7 @@ import io.flutter.settings.FlutterSettings;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -157,7 +159,19 @@ public class SdkFields {
 
     if (flutterSdk.getVersion().flutterRunSupportsDevToolsUrl()) {
       try {
-        final DevToolsInstance instance = DevToolsService.getInstance(project).getDevToolsInstance().get(30, TimeUnit.SECONDS);
+        final ProgressManager progress = ProgressManager.getInstance();
+
+        final CompletableFuture<DevToolsInstance> devToolsFuture = new CompletableFuture<>();
+        progress.runProcessWithProgressSynchronously(() -> {
+          progress.getProgressIndicator().setIndeterminate(true);
+          try {
+            devToolsFuture.complete(DevToolsService.getInstance(project).getDevToolsInstance().get(30, TimeUnit.SECONDS));
+          }
+          catch (Exception e) {
+            e.printStackTrace();
+          }
+        }, "Starting DevTools", false, project);
+        final DevToolsInstance instance = devToolsFuture.get();
         args = ArrayUtil.append(args, "--devtools-server-address=http://" + instance.host + ":" + instance.port);
       }
       catch (Exception e) {
--- a/src/io/flutter/run/SdkFields.java
+++ b/src/io/flutter/run/SdkFields.java
@@ -19,11 +19,15 @@ import io.flutter.dart.DartPlugin;
 import io.flutter.pub.PubRoot;
 import io.flutter.pub.PubRootCache;
 import io.flutter.run.common.RunMode;
+import io.flutter.run.daemon.DevToolsInstance;
+import io.flutter.run.daemon.DevToolsService;
 import io.flutter.sdk.FlutterCommand;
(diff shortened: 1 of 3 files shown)


stevemessick:
I would start it when the user first runs an app. You can use `ApplicationManager.getApplication().executeOnPooledThread()` to move it off the EDT. Look at how we use `executeOnPooledThread()` for inspiration. You'd probably want to move the remainder of the launch code onto the pooled thread as well.

helin24:
> You'd probably want to move the remainder of the launch code onto the pooled thread as well.

Okay, this is the part I was worried about, wasn't sure if it would be okay to move a bunch of launch code into a thread. But I'll try and see if this is reasonable.

kenzieschmoll:
> I would start it when the user first runs an app. You can use ApplicationManager.getApplication().executeOnPooledThread() to move it off the EDT.

I'd be worried about a race here between DevTools being spun up and the url being passed to flutter tools on run and attach. What if DevTools takes longer to start than the Flutter app? Then the devtools server uri missed the train to be passed into flutter tools on run or attach, and flutter tools may start up a different server because the uri is null when flutter tools looks for it. Seems like we'd have to wait for DevTools to start before we can run the flutter app.

The time to do this should be minimal if we are just pulling from pub, and I would think if the CLI workflow awaits DevTools start, then it would be fine to do this from IDEs as well. See this logic [here](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/devtools_launcher.dart#L71) for some checks I added in devtools_launcher.dart to use an already activated package if available and only attempt pub global activate every 12 hrs.

helin24:
> Seems like we'd have to wait for DevTools to start before we can run the flutter app.

Yes, that's a reasonable concern. I think if I move all/most of the app running code into a non-EDT thread, then it won't feel as clunky to have the app waiting for DevTools to start, particularly if it doesn't take too long. Good to know the CLI is waiting too.

I think I'm actually still starting the project DevTools server from `pub global activate` + `pub global run` directly instead of through a daemon, and I was planning to change to the latter method since the daemon DevTools should be the latest version now too. Sounds like this way may be a little faster since it'll skip activate a lot of the time.

jacob314:
The pub global activate step is probably fine to execute on project load while running the server should be avoided.
We should avoid a race condition by only launching flutter tools once devtools is running. That can be accomplished while keeping the work off the main thread.

helin24:
Can we accept a 1-2 second delay in the start of running an application? (the first time someone runs their app on the project only) From running this a few times on my workbench system, this seems to be the time required to do the `pub global run devtools` part.

If this is an acceptable delay, we could keep it running on the UI thread for now, and I would move `pub global activate` to run once at project open (or maybe every 24 hours if a project is open for a long time, something like that). I'm looking into whether there are alternative APIs we could use to be able to run asynchronously, but if this is possible, it'll require a larger change anyway.

For this change I also need to enhance our versioning checks to distinguish between dev release versions too (e.g. 1.26.0-8.0.pre vs. 1.26.0-11.0.pre)

kenzieschmoll:
IMO that delay is acceptable. It already happens on the CLI, and if it is only paid on the first start, then I don't think the user will really feel the cost.

stevemessick:
You can't have that long a delay on the UI thread. IntelliJ throws an error if something takes too long.

src/io/flutter/run/SdkFields.java
```diff
@@ -150,6 +156,28 @@ public GeneralCommandLine createFlutterSdkRunCommand(
...
+        final ProgressManager progress = ProgressManager.getInstance();
+
+        final CompletableFuture<DevToolsInstance> devToolsFuture = new CompletableFuture<>();
+        progress.runProcessWithProgressSynchronously(() -> {
+          progress.getProgressIndicator().setIndeterminate(true);
+          try {
+            devToolsFuture.complete(DevToolsService.getInstance(project).getDevToolsInstance().get(30, TimeUnit.SECONDS));
+          }
+          catch (Exception e) {
+            e.printStackTrace();
```
stevemessick, line 171: Probably want to log this.

src/io/flutter/run/SdkFields.java
```diff
@@ -150,6 +156,28 @@ public GeneralCommandLine createFlutterSdkRunCommand(
...
+          }
+        }, "Starting DevTools", false, project);
+        final DevToolsInstance instance = devToolsFuture.get();
+        args = ArrayUtil.append(args, "--devtools-server-address=http://" + instance.host + ":" + instance.port);
+      }
+      catch (Exception e) {
+        e.printStackTrace();
```
stevemessick, line 178: Probably want to log this.

stevemessick (approved):
Change those print statements and you're good.

Issue #4588, opened 2020-05-29
issue using --dart-define in run configurations with a value that has a space


samandmoore:
## Steps to Reproduce

* Create a flutter app, the template counter app will do
* Create a run configuration and add the following to the "additional arguments"
```
--dart-define='SOMETHING=cool: thing'
```
* Run the app using the configuration
* Instead of setting the value in the ENV for `SOMETHING='cool: thing'` the run configuration fails with an error like this:
```
Target file "thing'" not found.
```

**Note: this works just fine via the flutter command line**
```
flutter run --dart-define='SOMETHING=cool: thing' -t lib/main.dart
```

And the value is available in the ENV as expected.

If you remove the empty space character between `cool:` and `thing` then it works just fine. Is there some escaping approach for this in Android Studio? Is this an issue with the plugin or Android Studio itself?

## Version info

```
$ flutter doctor -v
[✓] Flutter (Channel unknown, v1.17.0, on Mac OS X 10.15.4 19E287, locale en-US)
    • Flutter version 1.17.0 at /Users/sam/flutter
    • Framework revision d3ed9ec945 (8 weeks ago), 2020-04-06 14:07:34 -0700
    • Engine revision c9506cb8e9
    • Dart version 2.8.0 (build 2.8.0-dev.18.0 eea9717938)


[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
    • Android SDK at /Users/sam/Library/Android/sdk
    • Platform android-29, build-tools 29.0.2
    • Java binary at: /Applications/Android
      Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b4-5784211)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 11.4.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 11.4.1, Build version 11E503a
    • CocoaPods version 1.9.1

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 3.6)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 45.1.1
    • Dart plugin version 192.8052
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b4-5784211)

[✓] VS Code (version 1.44.2)
[...]

PR #5181, opened 2021-01-17
fix args parsing in launch configurations

devoncarew:
- fix args parsing in launch configurations
- fix https://github.com/flutter/flutter-intellij/issues/4588

### DIFF
--- a/src/io/flutter/run/SdkFields.java
+++ b/src/io/flutter/run/SdkFields.java
@@ -11,6 +11,7 @@ import com.intellij.execution.configurations.RuntimeConfigurationError;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.ArrayUtil;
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/SdkFields.java
```diff
@@ -69,6 +70,19 @@ public void setAdditionalArgs(final @Nullable String additionalArgs) {
     this.additionalArgs = additionalArgs;
   }
 
+  public boolean hasAdditionalArgs() {
+    return additionalArgs != null;
+  }
+
+  public String[] getAdditionalArgsParsed() {
+    if (hasAdditionalArgs()) {
+      assert additionalArgs != null;
```
stevemessick, line 79: This assert is redundant but harmless.

stevemessick (approved):

src/io/flutter/run/SdkFields.java
stevemessick: Disregard, I see the getter is @Nullable.

PR #5192, opened 2021-01-20
Add compare function that works for beta versions

helin24:
I think we need to use a beta version to gate the upcoming change for DevTools since the next major release of flutter to include the needed changes won't be out by the time we want the full setup to be working.

I didn't find a built in way to compare these versions, so added this. Let me know if there is an easier way though.

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkVersion.java
+++ b/src/io/flutter/sdk/FlutterSdkVersion.java
@@ -8,10 +8,11 @@ package io.flutter.sdk;
 import com.google.common.annotations.VisibleForTesting;
 import com.intellij.openapi.util.Version;
 import com.intellij.openapi.vfs.VirtualFile;
-import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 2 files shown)


src/io/flutter/sdk/FlutterSdkVersion.java
```diff
@@ -132,6 +149,33 @@ public String fullVersion() {
     return version == null ? "unknown version" : version.toString();
   }
 
+  /**
+   * This considers beta versions when comparing versions. See https://github.com/flutter/flutter/wiki/Flutter-build-release-channels.
+   */
+  public int compareToWithBetaVersion(FlutterSdkVersion otherVersion) {
```
stevemessick, line 155: Rename this "compareTo" and make the class extend Comparable<FlutterSdkVersion>.

It looks like this already will work for the dev channel. If you add a check for the number following "pre" it can work for all channels. My master is currently "1.26.0-2.0.pre.289", for example.

stevemessick (changes requested):
If you don't have time to extend this to check master versions then just change the superclass and method name and add a to-do to about master.

stevemessick (approved):
Thanks!

Issue #5164, opened 2021-01-08
Flutter Inspector font size does not apply correct scale factor on HiDPI displays


raju-bitter:
## Steps to Reproduce

Using two displays with Windows 10, main display is 5k (5120x2880) at 200%, secondary display QHD (2560x1440) at 100%. Installed Android Studio 4.1 with Flutter and Dart plugin. Created new Flutter project, launched app on a 5k display.  Flutter Inspector opened with incorrect font size. The font on the 5k display is double of what it is for the QHD display in on screen size, as seen in the screenshots.
This seems to be a Windows specific problem. Running Android Studio 4.1 on Ubuntu 20.04, the Flutter Inspector font scaling works for both 5k and QHD, even when I drag the Android Studio window from one screen to the other.
When the 5k display is running without scaling at 100%, the font and icon sizes for the Flutter Inspector are rendered correctly.

![Flutter-Inspector-on-5k-Display-with-info](https://user-images.githubusercontent.com/296926/104002901-f9f37a80-51a1-11eb-9313-06a0f54aa42b.PNG)

![Flutter-Inspector-on-QHD-Display-with-info](https://user-images.githubusercontent.com/296926/104002909-feb82e80-51a1-11eb-86db-a6f6b0c707e3.PNG)

## Version info
C:\Users\Raju\dev\flutter\stable\bin\flutter.bat doctor --verbose
[√] Flutter (Channel stable, 1.22.5, on Microsoft Windows [Version 10.0.19041.685], locale en-US)
    • Flutter version 1.22.5 at C:\Users\Raju\dev\flutter\stable
    • Framework revision 7891006299 (4 weeks ago), 2020-12-10 11:54:40 -0800
    • Engine revision ae90085a84
    • Dart version 2.10.4

[√] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at C:\Users\Raju\AppData\Local\Android\sdk
    • Platform android-30, build-tools 30.0.3
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)
    • All Android licenses accepted.

[!] Android Studio (version 4.1.0)
    • Android Studio at C:\Program Files\Android\Android Studio
    X Flutter plugin not installed; this adds Flutter specific functionality.
    X Dart plugin not installed; this adds Dart specific functionality.
[...]

PR #5205, opened 2021-01-22
Update jxbrowser version and set DPI property

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/5164


stevemessick (approved):
Please update flutter-intellij-community.iml before merging.

PR #5207, opened 2021-01-25
Show notification on error with deep link 

helin24:
When an app throws a render overflow error and sends a URL with a deep link to DevTools, we want to parse the widget ID and open in the embedded DevTools window (if the user has enabled this).

Jan-25-2021 15-36-02

I think we were under the impression that navigating to the same URL but with one param changed should not cause a full page refresh, but I haven't found a way to avoid this.

### DIFF
--- a/src/io/flutter/devtools/DevToolsUtils.java
+++ b/src/io/flutter/devtools/DevToolsUtils.java
@@ -54,4 +54,15 @@ public class DevToolsUtils {
     }
     return "http://" + devtoolsHost + ":" + devtoolsPort + "/?" + String.join("&", params);
   }
+
+  public static String findWidgetId(String url) {
+    final String searchFor = "inspectorRef=";
+    final String[] split = url.split("&");
+    for (String part : split) {
+      if (part.startsWith(searchFor)) {
+        return part.substring(searchFor.length());
+      }
+    }
+    return null;
+  }
 }
--- a/src/io/flutter/logging/FlutterConsoleLogManager.java
+++ b/src/io/flutter/logging/FlutterConsoleLogManager.java
@@ -402,7 +402,7 @@ public class FlutterConsoleLogManager {
   private void showDeepLinkNotification(DiagnosticsNode property, String errorSummary) {
     final Notification notification = new Notification(
       FlutterMessages.FLUTTER_NOTIFICATION_GROUP_ID,
-      "Inspect widget",
+      "Inspect error-causing widget",
       errorSummary,
       NotificationType.INFORMATION);
     notification.setIcon(FlutterIcons.Flutter);
--- a/src/io/flutter/logging/FlutterConsoleLogManager.java
+++ b/src/io/flutter/logging/FlutterConsoleLogManager.java
@@ -14,6 +14,11 @@ import com.google.gson.JsonSyntaxException;
 import com.intellij.execution.ui.ConsoleView;
 import com.intellij.execution.ui.ConsoleViewContentType;
 import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationType;
+import com.intellij.notification.Notifications;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.editor.ex.EditorSettingsExternalizable;
 import com.intellij.openapi.editor.impl.softwrap.SoftWrapAppliancePlaces;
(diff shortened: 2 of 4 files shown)


src/io/flutter/devtools/DevToolsUtils.java
```diff
@@ -54,4 +54,15 @@ public static String generateDevToolsUrl(
...
+
+  public static String findWidgetId(String url) {
+    final String searchFor = "inspectorRef=";
+    final String[] split = url.split("&");
+    for (String part : split) {
+      if (part.startsWith(searchFor)) {
+        return part.substring(searchFor.length());
+      }
+    }
+    return null;
```
kenzieschmoll, line 66: instead of doing a string search, should we do URI parsing instead and then grab the query param value at 'inspectorRef'?

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -322,6 +331,11 @@ private void printDiagnosticsNodeProperty(ConsoleView console, String indent, Di
       if (StringUtil.equals("ErrorSpacer", property.getType())) {
         return;
       }
+
+      if (StringUtil.equals("DevToolsDeepLinkProperty", property.getType()) && FlutterSettings.getInstance().isEnableEmbeddedBrowsers()) {
```
kenzieschmoll, line 335: when we see a `DevToolsDeepLinkProperty` does this early return guarantee that we don't show the deep link url in the intellij console?

src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -382,6 +396,23 @@ private void printDiagnosticsNodeProperty(ConsoleView console, String indent, Di
     }
   }
 
+  private void showDeepLinkNotification(DiagnosticsNode property) {
+    final Notification notification = new Notification(
+      FlutterMessages.FLUTTER_NOTIFICATION_GROUP_ID,
+      "Inspect this widget",
+      "Click below to inspect this exception-causing widget in Flutter DevTools",
```
kenzieschmoll, line 403: @csells @InMatrix for input on verbiage

@DanTup sent an example over to me that looked like this:
Screen Shot 2021-01-25 at 2 14 15 PM

I think we should try to use consistent wording between the two IDEs, and I actually like how in this screenshot it just grabs the error message directly from the error, so it's consistent with console output as well.

For the button text, I like "Inspect widget"

src/io/flutter/logging/FlutterConsoleLogManager.java
helin24: Right, the early return skips printing in the console. I could add a short description like "Click on notification to inspect widget" or something like that.

src/io/flutter/logging/FlutterConsoleLogManager.java
helin24: Updated to look like this:
Screen Shot 2021-01-25 at 3 32 42 PM

There's a title as well as the button, let me know if any of you have preferences there.

kenzieschmoll (approved):
lgtm_army

src/io/flutter/devtools/DevToolsUtils.java
helin24: Weirdly I can't seem to get a query string from the URL. e.g. 
```
final URI uri = new URI(urlString);
final String[] paramPairs = uri.getQuery() // this is null
```

maybe because of the `#` after the port? I'm not sure. Certainly it would be nicer to use more standard parsing, if anyone has suggestions on this.

src/io/flutter/logging/FlutterConsoleLogManager.java
kenzieschmoll: maybe change the title text to "Inspect error-causing widget"? 

this wording matches what the error message says. For example,

```
The relevant error-causing widget was:
  Column
  file:///Users/kenzieschmoll/develop/flutter_vignettes/vignettes/spending_tracker/lib/demo.dart:64:26
```

helin24: Sure, like this?

Screen Shot 2021-01-25 at 3 57 02 PM

I'll land with the current lame URL parsing and make updates in a later PR if there's a nicer way.

kenzieschmoll (approved):

src/io/flutter/logging/FlutterConsoleLogManager.java
InMatrix: I'm late to the party. The verbiage looks fine. I like icon in the VS Code version, which signals that this notification is related to an error. For the IntelliJ version, can we go without a title for the message to match the simplicity of the VS Code version? 

A related question: how are we handling multiple exceptions thrown in the same time? Will the user get a bunch of notifications at once? For example, this [Missing Material Widget error](https://github.com/InMatrix/flutter_error_studies/blob/master/lib/errors/missing_material_error.dart) will also generate a RenderFlex Overflow error from the red warning screen on device.

helin24: I'm making another change for the length of time this notification is shown, so I can make the other adjustments there too. 

If there are multiple exceptions, they'll show up stacked like this:
Screen Shot 2021-01-26 at 2 00 22 PM

InMatrix: Will the first error get displayed at the top of the stack or the bottom? The first error is usually relevant to the user's code in my experience.

helin24: Hmm it's the most recent error that gets displayed. When "1 more" is clicked, there's a list shown of all the notifications with their timestamps:
Screen Shot 2021-01-27 at 11 55 39 AM

It may be the IntelliJ default to show the most recent at the top so that new notifications aren't missed. I can check whether this can be adjusted for a particular group of errors.

InMatrix: Thanks for checking. I think we need to try it over a few example errors and determine what makes sense. Flutter framework often throws secondary exceptions down the pipeline caused by the primary error the user can fix in their own code. Maybe there are other filtering strategies we can employ.

Issue #5162, opened 2021-01-05
Flutter Inspector windows do not close after the app being debugged is terminated


jacob314:
## Steps to Reproduce
Run a flutter desktop app such as DevTools (probably any app works)
Stop the app from IntelliJ.
The Flutter Inspector tab remains open. Expected behavior would be for the Flutter Inspector tab to close.
## Version info

[✓] Flutter (Channel unknown, 1.26.0-2.0.pre.157, on macOS 11.0.1 20B29 darwin-x64, locale en-US)
    • Flutter version 1.26.0-2.0.pre.157 at /Users/jacobrichman/git/flutter/1/flutter
    • Framework revision 4b83be6c23 (7 days ago), 2020-12-30 02:09:03 -0500
    • Engine revision 0e5a25d779
    • Dart version 2.12.0 (build 2.12.0-179.0.dev)

[!] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at /Users/jacobrichman/Library/Android/sdk
    • Platform android-30, build-tools 29.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6915495)
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/macos#android-setup for more details.

[!] Xcode - develop for iOS and macOS (Xcode 12.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 12.3, Build version 12C33
    ! CocoaPods 1.9.0 out of date (1.10.0 is recommended).
        CocoaPods is used to retrieve the iOS and macOS platform side's plugin code that responds to
        your plugin usage on the Dart side.
        Without CocoaPods, plugins will not work on iOS or macOS.
        For more info, see https://flutter.dev/platform-plugins
      To upgrade see https://guides.cocoapods.org/using/getting-started.html#installation for
      instructions.

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 4.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
[...]

PR #5218, opened 2021-01-27
Restart DevTools and browser on inspector window reopen

helin24:
There have been cases where DevTools installation gets stuck. This may be fixed with our change to having one DevTools instance per project, but we wanted to offer a way to restart DevTools if users are waiting a long time. 

Jan-27-2021 11-46-00

The logic when a user closes and reopens the window:
- If the DevTools instance has successfully been created, the same instance is returned but the browser is refreshed (this may be useful if users see the frozen wheel or another issue within DevTools)
- If DevTools has completed with a failure, we try to start a new instance
- If DevTools is still in progress, we force a restart by canceling the existing future and trying to start a new instance

Fixes https://github.com/flutter/flutter-intellij/issues/5162

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -5,6 +5,7 @@
  */
 package io.flutter.view;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.intellij.execution.runners.ExecutionUtil;
 import com.intellij.execution.ui.layout.impl.JBRunnerTabs;
 import com.intellij.icons.AllIcons;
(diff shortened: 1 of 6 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -451,32 +455,67 @@ protected void handleJxBrowserInstalled(FlutterApp app, InspectorService inspect
...
+    }
+    this.toolWindowListener.updateOnWindowOpen(() -> {
+      devToolsInstallCount += 1;
+      presentLabel(toolWindow, getInstallingDevtoolsLabel());
+      openInspectorWithDevTools(app, inspectorService, toolWindow, isEmbedded, true);
+    });
+  }
+
+  private String getInstallingDevtoolsLabel() {
+    return "<html><body style=\"text-align: center;\">Installing DevTools... <br><br>If this takes more than 15 seconds, you can restart " +
```
stevemessick, line 479: Move this to FlutterBundle.properties. We only need the English text there so you can leave the HTML here if you want. Line 36 has an example of using a macro.

stevemessick (approved):
Very nice! I'd like to see one small change.

PR #5226, opened 2021-01-29
Modify password store to avoid keyring prompts

helin24:
Since JxBrowser uses Chromium we have some situations where a user opens the embedded browser and is prompted to unlock the keyring in Linux if Chrome hasn't been opened already. We're hoping to avoid this by passing a Chrome option.

### DIFF
--- a/src/io/flutter/jxbrowser/EmbeddedBrowserEngine.java
+++ b/src/io/flutter/jxbrowser/EmbeddedBrowserEngine.java
@@ -13,6 +13,7 @@ import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.util.text.StringUtil;
 import com.teamdev.jxbrowser.engine.Engine;
 import com.teamdev.jxbrowser.engine.EngineOptions;
+import com.teamdev.jxbrowser.engine.PasswordStore;
 import io.flutter.FlutterInitializer;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


src/io/flutter/jxbrowser/EmbeddedBrowserEngine.java
```diff
@@ -36,6 +37,8 @@ public EmbeddedBrowserEngine() {
     final EngineOptions options =
       EngineOptions.newBuilder(SystemInfo.isWindows ? OFF_SCREEN : HARDWARE_ACCELERATED)
         .userDataDir(Paths.get(dataPath))
+        .passwordStore(PasswordStore.BASIC)
+        .addSwitch("--use-mock-keychain")
```
vlad-lubenskyi, line 41: This seems to be a macOS-specific switch and it won't help.

src/io/flutter/jxbrowser/EmbeddedBrowserEngine.java
```diff
@@ -36,6 +37,8 @@ public EmbeddedBrowserEngine() {
     final EngineOptions options =
       EngineOptions.newBuilder(SystemInfo.isWindows ? OFF_SCREEN : HARDWARE_ACCELERATED)
         .userDataDir(Paths.get(dataPath))
+        .passwordStore(PasswordStore.BASIC)
```
vlad-lubenskyi, line 40: When using `BASIC` password store, Chromium stores password without encryption and doesn't require access to the system keyring. I expect this option to help.

helin24:
@vlad-lubenskyi thanks for the additional info!

stevemessick (approved):

jacob314 (commented):
lgtm

PR #5246, opened 2021-02-05
Fix platform selection for stable

stevemessick:


devoncarew (approved):

devoncarew:
@stevemessick - we may consider cherry-picking this fix into the current stable release.

stevemessick:
@devoncarew Already done.

PR #5247, opened 2021-02-05
Fix build for AS canary 5

stevemessick:
The class that implements indexing of Java and Kotlin (for the android module) took a lot of work. You'll notice a lot of methods with a suffix of "1". Those are not needed by 4.1. They are required by 4.2, and in 4.3 they are final.

I was trying to get support for EAP in this PR but I can't compile IntelliJ sources with any recent branch. So that has been delayed.

I think I forgot to update `flutter-intellij.iml`. I'll add that before merging.


stevemessick:
Unit tests are broken until we get support for EAP.

TBR, so the dev channel has canary support next week.

devoncarew (commented):
lgtm

resources/META-INF/plugin.xml
```diff
@@ -21,7 +21,7 @@
 
   <category>Custom Languages</category>
   <version>SNAPSHOT</version>
-  <idea-version since-build="201.7223" until-build="202.8194.7"/>
+  <idea-version since-build="201.7223" until-build="203.*"/>
```
helin24, line 24: Should we be keeping this more specific? I can't remember the details but was .* causing problems for another team?

helin24 (commented):
LGTM

resources/META-INF/plugin.xml
stevemessick: It should be more specific in product-matrix.json but we want the check-in version of plugin.xml to be unbounded for debugging. Good catch, though -- I think I left it unbounded in product-matrix. Will fix later.

PR #1321, opened 2017-09-07
support the --flavor flutter run arg

devoncarew:
- add a text field for the `--flavor` arg to flutter run (fix https://github.com/flutter/flutter-intellij/issues/1289)

@stevemessick 

screen shot 2017-09-07 at 10 59 24 am

PR #5251, opened 2021-02-09
XCode -> Xcode casing

jmagman:
Change from "XCode" to "Xcode" to match the app's correct casing.  See in docs: https://developer.apple.com/xcode/

Screen Shot 2021-02-05 at 3 59 22 PM

Introduced in #1321


stevemessick (approved):
Ignore the failure of unit tests. I just about have a PR to fix that but you don't need to wait -- nothing in this PR affects tests.

PR #5269, opened 2021-02-17
Add analytics for deep link clicked

helin24:


src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -433,6 +433,12 @@ public void actionPerformed(@NotNull AnActionEvent event) {
         if (toolWindow != null && !toolWindow.isVisible()) {
           toolWindow.show();
         }
+
+        FlutterInitializer.getAnalytics().sendEvent(
+          "notification",
+          "deep-link-clicked",
```
kenzieschmoll, line 439: is there anyway to indicate which type of error the notifcation was for? right now all we support are overflow errors but in the future, we'll support more types of errors, and it would be handy to be able to distinguish between them in analytics

helin24: I could pull some information from the error messages from flutter, but I don't think there's a clear indicator of error type in there. The closest thing I see is the error summary from a sibling message, e.g. 'A RenderFlex overflowed by 13 pixels on the right.' For now I could do some string parsing to label these `overflow`, but would it make sense to add a label of this type into the message from flutter tools?

kenzieschmoll: Yeah I think that could make sense once we start supporting other errors. When that time comes, we'll want to make sure that the label from flutter for overflow errors matches this so that analytics are not skewed. This looks good for now though!

kenzieschmoll (approved):

PR #5277, opened 2021-02-20
Permit deep link to open embedded browser

helin24:
Since we now wait for the inspector window to be visible before navigating to DevTools in the embedded browser (instead of navigating on app start), there could be cases where a notification with a deep link is shown before the navigation happens. This change shows the tool window before trying to navigate and waits for the main URL to be set before navigating to the URL with a widget ID.

Feb-19-2021 16-55-37

### DIFF
--- a/src/io/flutter/jxbrowser/EmbeddedBrowser.java
+++ b/src/io/flutter/jxbrowser/EmbeddedBrowser.java
@@ -29,9 +29,11 @@ import com.teamdev.jxbrowser.view.swing.callback.DefaultConfirmCallback;
 import icons.FlutterIcons;
 import io.flutter.FlutterInitializer;
 import io.flutter.settings.FlutterSettings;
+import io.flutter.utils.AsyncUtils;
 import org.jetbrains.annotations.NotNull;
 
-import java.awt.*;
+import java.awt.Dimension;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class EmbeddedBrowser {
@@ -43,7 +45,7 @@ public class EmbeddedBrowser {
   }
 
   private Browser browser;
-  private String mainUrl;
+  private CompletableFuture<String> mainUrlFuture;
 
   private EmbeddedBrowser(Project project) {
     System.setProperty("jxbrowser.force.dpi.awareness", "1.0");
@@ -70,6 +72,8 @@ public class EmbeddedBrowser {
       FlutterInitializer.getAnalytics().sendException(StringUtil.getThrowableText(ex), false);
     }
 
+    resetUrl();
+
     ProjectManager.getInstance().addProjectManagerListener(project, new ProjectManagerListener() {
       @Override
       public void projectClosing(@NotNull Project project) {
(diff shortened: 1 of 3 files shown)


src/io/flutter/jxbrowser/EmbeddedBrowser.java
```diff
@@ -81,6 +85,13 @@ public void projectClosing(@NotNull Project project) {
     });
   }
 
+  /**
+   * This is to clear out a potentially old URL, i.e. a URL from an app that's no longer running.
+   */
+  public void resetUrl() {
+    this.mainUrlFuture = new CompletableFuture<>();
```
jacob314, line 92: Replacing a completable future with another like this has lead to bugs in the past. A reasonable pattern is to only reset the future if it has not yet completed or to complete the existing future with a failure value depending on what you want to do. Otherwise you can end up with strange behavior if someone was waiting on the future you reset.

src/io/flutter/jxbrowser/EmbeddedBrowser.java
```diff
@@ -128,6 +139,12 @@ public void openPanel(ContentManager contentManager, String tabName, String url,
   }
 
   public void updatePanelToWidget(String widgetId) {
-    browser.navigation().loadUrl(mainUrl + "&inspectorRef=" + widgetId);
+    AsyncUtils.whenCompleteUiThread(mainUrlFuture, (url, ex) -> {
+      if (ex != null) {
+        LOG.error(ex);
+        return;
+      }
+      browser.navigation().loadUrl(url + "&inspectorRef=" + widgetId);
```
jacob314, line 147: you probably want to avoid navigating if the url is null which is what you would probably want to have the future return if the  server was never oppened.

src/io/flutter/view/FlutterView.java
```diff
@@ -671,6 +671,7 @@ private void debugActiveHelper(FlutterApp app, @Nullable InspectorService inspec
     toolWindow.setIcon(ExecutionUtil.getLiveIndicator(FlutterIcons.Flutter_13));
 
     if (FlutterSettings.getInstance().isEnableEmbeddedBrowsers()) {
+      EmbeddedBrowser.getInstance(myProject).resetUrl();
```
jacob314, line 674: Comment on why you reset the url here.

jacob314:
This solution gives the best of the both worlds!

src/io/flutter/jxbrowser/EmbeddedBrowser.java
helin24: Okay, we do want to replace the future even if it has completed successfully (since a new URL will have a different vm service URL), but it would make sense to complete with failure if it's not done yet.

src/io/flutter/jxbrowser/EmbeddedBrowser.java
```diff
@@ -144,6 +147,10 @@ public void updatePanelToWidget(String widgetId) {
         LOG.error(ex);
         return;
       }
+      if (url == null) {
```
jacob314, line 150: no need to log with an error if this is something intentional we are causing if the app got shut.

src/io/flutter/jxbrowser/EmbeddedBrowser.java
helin24: completing with null since this is a race condition instead of an error we care about.

PR #5279, opened 2021-02-22
Allow open DevTools from tests

helin24:
I'm not sure if this is actually useful - I think opening DevTools from the console used to be available while running/debugging tests, but at some point we removed it when we were starting DevTools from an app (since tests don't have an associated app and we didn't want to add another way to start DevTools). Since we're serving DevTools separately from apps now, we can add back this option for tests.


stevemessick (approved):

jacob314:
This could be somewhat useful profiling a slow test.

PR #5296, opened 2021-03-03
Get device list after indexing finishes

### LINKED ISSUE #5244: Since version 53.0.5 it won't show the available emulators unless I start them
It was working well in 52.2.5. I could even downgrade the plugin to that version and it worked again. But in 53.0.5 it is not showing the available emulators.

image

I have them correctly created:

image

It only detects them when I start them.

It's not a big issue, but it is there.

## Version info

```
[√] Flutter (Channel beta, 1.25.0-8.3.pre, on Microsoft Windows [Version 10.0.19042.746], locale es-PE)
    • Flutter version 1.25.0-8.3.pre at C:\tools\flutter
    • Framework revision 5d36f2e7f5 (3 weeks ago), 2021-01-14 15:57:49 -0800
    • Engine revision 7a8f8ca02c
    • Dart version 2.12.0 (build 2.12.0-133.7.beta)

[√] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at C:\Users\leozx\AppData\Local\Android\sdk
    • Platform android-30, build-tools 30.0.3
    • Java binary at: C:\Users\leozx\.jdks\openjdk-15.0.1\bin\java
    • Java version OpenJDK Runtime Environment (build 15.0.1+9-18)

      (or visit https://flutter.dev/docs/get-started/install/windows#android-setup for detailed instructions).

[√] IntelliJ IDEA Ultimate Edition (version 2020.3)
    • IntelliJ at C:\Users\leozx\AppData\Local\JetBrains\Toolbox\apps\IDEA-U\ch-0\203.7148.57
    • Flutter plugin version 53.0.5
    • Dart plugin version 203.6912

[!] Connected device
    ! No devices available

! Doctor found issues in 2 categories.
```

### LINKED ISSUE #5248: Device not listed
Hello Everyone,

my device is not list in menu, but when i use the cli I can use it.

Fedora 33 64Bit

Screenshot from 2021-02-06 15-36-29
Screenshot from 2021-02-06 15-35-52

## Version info

[✓] Flutter (Channel beta, 1.25.0-8.3.pre, on Linux, locale en_US.UTF-8)
    • Flutter version 1.25.0-8.3.pre at /home/moritz/dev/flutter
    • Framework revision 5d36f2e7f5 (3 weeks ago), 2021-01-14 15:57:49 -0800
    • Engine revision 7a8f8ca02c
    • Dart version 2.12.0 (build 2.12.0-133.7.beta)

    • Android SDK at /home/moritz/dev/android
    • Platform android-30, build-tools 30.0.3
    • ANDROID_HOME = /home/moritz/dev/android
    • Java binary at: /opt/android-studio/jre/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)

[✓] Chrome - develop for the web
    • Chrome at google-chrome

[✓] Android Studio
    • Android Studio at /opt/android-studio
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)

[✓] IntelliJ IDEA Ultimate Edition (version 2020.3)
    • IntelliJ at /opt/idea-IU-203.6682.168
    • Flutter plugin version 53.0.5
    • Dart plugin version 203.6912

[✓] VS Code (version 1.52.1)
    • VS Code at /usr/share/code
    • Flutter extension can be installed from:
      🔨 https://marketplace.visualstudio.com/items?itemName=Dart-Code.flutter

[✓] Connected device (2 available)
    • SM G770F (mobile) • RF8N21RBH9K • android-arm64  • Android 11 (API 30)
    • Chrome (web)      • chrome      • web-javascript • Google Chrome 87.0.4280.141

• No issues found!

stevemessick:
Wait for indexing to finish before updating the device list. This change is going to affect all daemon state changes. I think it is OK but want someone to double-check that it is not causing downstream problems.

Fixes #5248
Fixes #5244

### DIFF
--- a/src/io/flutter/run/daemon/DeviceService.java
+++ b/src/io/flutter/run/daemon/DeviceService.java
@@ -10,8 +10,10 @@ import com.google.common.collect.ImmutableSet;
 import com.intellij.concurrency.JobScheduler;
 import com.intellij.execution.ExecutionException;
 import com.intellij.openapi.application.ApplicationInfo;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.components.ServiceManager;
 import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.project.DumbService;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.ex.ProjectRootManagerEx;
 import com.intellij.openapi.util.Disposer;
@@ -162,8 +164,11 @@ public class DeviceService {
    * <p>This might mean starting it, stopping it, or restarting it.
    */
   private void refreshDeviceDaemon() {
-    if (project.isDisposed()) return;
-    deviceDaemon.refresh(this::chooseNextDaemon);
+    ApplicationManager.getApplication().executeOnPooledThread(() -> {
+      DumbService.getInstance(project).waitForSmartMode();
+      if (project.isDisposed()) return;
+      deviceDaemon.refresh(this::chooseNextDaemon);
+    });
   }
 
   private void daemonStopped(String details) {


src/io/flutter/run/daemon/DeviceService.java
```diff
@@ -162,8 +164,11 @@ private void fireChangeEvent() {
    * <p>This might mean starting it, stopping it, or restarting it.
    */
   private void refreshDeviceDaemon() {
-    if (project.isDisposed()) return;
-    deviceDaemon.refresh(this::chooseNextDaemon);
+    ApplicationManager.getApplication().executeOnPooledThread(() -> {
+      DumbService.getInstance(project).waitForSmartMode();
+      if (project.isDisposed()) return;
+      deviceDaemon.refresh(this::chooseNextDaemon);
```
jacob314, line 170: Any idea why deviceDaemon.refresh doesn't work robustly in dumbMode?

jacob314:
Another option could be to refresh on load and then again on smart mode. If that doesn't generate UI flicker it could work around the issue while still showing users a valid device list sooner.

jacob314:
lgtm

stevemessick:
> Any idea why deviceDaemon.refresh doesn't work robustly in dumbMode?

No. I was looking at the logs and noticed the daemon was started very early. I know there are limitations to what the plugin can do during indexing and had a hunch this might work.

Perhaps @alexander-doroshko could weigh in? This might not be appropriate, even though it appears to work.

alexander-doroshko:
You can't use index in dumb mode. If you do, you'll see `IndexNotReadyException` thrown. 

I don't know in details what `deviceDaemon.refresh` does. If its implementation doesn't have code like `if (isDumb()) doNothing();` and you don't get `IndexNotReadyException` then the underlying reason of something not working might be different. And the fact that this fix helps might be just a coincidence.

As I understood from #5248, logs prove that the Flutter plugin runs the correct command but receives unexpected result. Unlikely the tool that it runs cares about IDE index. Maybe this fix helps just because it postpones the tool launch? You may experiment with timeout instead of `waitForSmartMode()`. Though I definitely have no idea why timeout or `waitForSmartMode()` may affect the tool behavior.

devoncarew:
I do see this message sequence in the log:

```
starting Flutter device daemon #1: /home/moritz/dev/flutter/bin/flutter daemon 
[<-- [{"event":"daemon.connected","params":{"version":"0.6.0","pid":845769}}]] 
[--> {"method":"device.enable","id":0}] 
[<-- [{"event":"daemon.logMessage","params":{"level":"status","message":"Waiting for another flutter command to release the startup lock..."}}]] 
flutter device daemon #1: Waiting for another flutter command to release the startup lock... 
[<-- [{"event":"daemon.logMessage","params":{"level":"status","message":"Starting device daemon..."}}]] 
flutter device daemon #1: Starting device daemon... 
[<-- [{"id":0}]] 
[<-- [{"event":"device.added","params":{"id":"chrome","name":"Chrome","platform":"web-javascript","emulator":false,"category":"web","platformType":"web","ephemeral":false,"emulatorId":null}}]] 
...
```

As long as we handle the `Waiting for another flutter command to release the startup lock...` reasonable - which I assume we do - that result shouldn't be any issue. The flutter command startup could take a long time, so we'd want to make sure that we don't have any fixed timeout or such in the communication code.

stevemessick:
> Maybe this fix helps just because it postpones the tool launch?

I think that's the reason it helps. On my machine, indexing causes the fan to run at max speed. There's a lot going on. Delaying the daemon until all that has finished seems to help. So, I think this is a good work-around, but for reasons not related directly to the index, but to the work required to produce the index.

Eventually, we should take a closer look at why device-added events are not getting properly recorded.

stevemessick:
We need to do some work in the github workflow. Since the unit tests passed on kokoro I'm satisfied this isn't breaking anything obvious.

PR #5301, opened 2021-03-04
Build EAP and Canary

### LINKED ISSUE #5223: Support 2021.1 EAP
The EAP for `2021.1` is now available; we should update the plugin to support it.

stevemessick:
This includes a significant refactoring that moves Gradle-aware code out of the IntelliJ module and into the Android Studio module. Originally, we hoped to support add-to-app in IntelliJ. The difference in Gradle support between the two products was too great to do that. With the latest EAP, the divergence has grown so large it is impractical to try to compile any common Gradle code.

This doesn't actually compile for EAP. That needs to be a manual process for now. See [this YouTrack issue](https://youtrack.jetbrains.com/issue/IDEA-263477) for discussion.

Fixes #5223


stevemessick:
This is now ready for review.

stevemessick:
TBR to enable dev channel in this weekend's update.

helin24 (commented):
LGTM

PR #5302, opened 2021-03-04
Fix DART_BOT GH action

stevemessick:


tool/github.sh
```diff
@@ -40,8 +40,10 @@ pub get
 if [ "DART_BOT" = "$BOT" ] ; then
 
   # Analyze the Dart code in the repo.
+  chmod 000 flutter-idea # symlinks seem to be causing problems
```
devoncarew, line 43: cc @scheglov (I'm not at all sure that this is related to the recent analysis context changes, but FYI in any case)

devoncarew: @stevemessick - do you have a link to a CI failure (or details about what the problem is)?

stevemessick:
See #5301 

If you run `dart analyze` at the top level it complains about the symlinked files reached through the `flutter-idea/testData` symbolic link. Every PR for the last week or so has been failing this way and we've been ignoring it in favor of yesterday's events.

devoncarew (approved):

PR #5307, opened 2021-03-08
Automate EAP build for dev channel

stevemessick:
It appears that ideaIC-211.6305.21 is actually labeled internally as 211.6222.4. Upgrading the IDE asks to upgrade the Dart plugin. After restarting, there is an error due to the Dart plugin requiring a newer version and the about box still has the 6222.4 version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,6 +8,9 @@ buildscript {
   repositories {
     mavenCentral()
     maven {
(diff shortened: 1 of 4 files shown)


stevemessick:
TBR

helin24 (commented):
LGTM

PR #5325, opened 2021-03-16
Fix the lost SDK on start up problem

### LINKED ISSUE #5313: new projects are not picking up the Flutter SDK to use
New projects are no longer picking up the Flutter SDK to use; we'd previously discovered the correct flutter sdk setting by parsing the .packages file. This code still exists in the plugin, so we'll need to look into why this workflow regressed.

stevemessick:
Fixes #5313 

We used to use `DartPlugin.isDartSdkEnabled(module)` as a quick short-circuit for checking if the Flutter SDK was configured. That isn't reliable (not sure why) anymore. The result is that the first time a project is created after IntelliJ is launched, the project does not have its SDK configured. Projects created after the first one do, however.

This also adds support for the new `.dart_tool/package_config.json` file.

Since the Dart SDK is configured, I tried to find a way to determine its location on disk. You can see in External Libraries that it knows about its content, so the path has to be somewhere. The IntelliJ API does not appear to reveal that information.

@alexander-doroshko If you can take a minute to look at this I'd appreciate any tips on finding the Dart SDK on disk. This approach works, and seems fast enough, but I can't help thinking it is redundant.

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkUtil.java
+++ b/src/io/flutter/sdk/FlutterSdkUtil.java
@@ -19,8 +19,6 @@ import com.intellij.openapi.module.Module;
 import com.intellij.openapi.module.ModuleManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.project.ProjectManager;
-import com.intellij.openapi.roots.ModuleRootManager;
-import com.intellij.openapi.roots.OrderEntry;
 import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.vfs.LocalFileSystem;
@@ -28,7 +26,6 @@ import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.ArrayUtil;
 import com.intellij.util.Url;
 import com.intellij.util.Urls;
-import com.jetbrains.lang.dart.sdk.DartSdk;
 import com.jetbrains.lang.dart.sdk.DartSdkUpdateOption;
 import io.flutter.FlutterBundle;
 import io.flutter.dart.DartPlugin;
@@ -47,9 +44,6 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Set;
-
-import static com.jetbrains.lang.dart.sdk.DartSdkLibUtil.isDartSdkOrderEntry;
-import com.intellij.openapi.roots.*;
 
 public class FlutterSdkUtil {
   /**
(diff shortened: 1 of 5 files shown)


src/io/flutter/utils/FlutterModuleUtils.java
```diff
@@ -361,11 +361,7 @@ public static void setFlutterModuleAndReload(@NotNull Module module, @NotNull Pr
   }
 
   public static void enableDartSDK(@NotNull Module module) {
```
devoncarew, line 363: I just want to confirm that we don't try to set a flutter sdk if an sdk is already configured?

src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -285,23 +328,35 @@ public static String parseFlutterSdkPath(String packagesFileContent) {
...
   }
 
+  private static String extractSdkPathFromUri(String urlString) {
+    if (urlString.startsWith("file:")) {
+      final Url url = Urls.parseEncoded(urlString);
+      if (url == null) {
+        return null;
+      }
+      final String path = url.getPath();
+      // go up three levels for .packages or two for .dart_tool/package_config.json
```
devoncarew, line 349: We could also parameterize this method - the call site knows whether it's 2 deep or 3; either way would work.

devoncarew (approved):

src/io/flutter/sdk/FlutterSdkUtil.java
stevemessick: I agree that a parameter is better; changed.

src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -254,12 +259,50 @@ public static void enableDartSdk(@NotNull final Project project) {
    */
   @Nullable
   public static String guessFlutterSdkFromPackagesFile(@NotNull Module module) {
+    // First, look for .dart_tool/package_config.json
     for (PubRoot pubRoot : PubRoots.forModule(module)) {
-      final VirtualFile packagesFile = pubRoot.getPackagesFile();
+      final VirtualFile packagesFile = pubRoot.getPackageConfigFile();
```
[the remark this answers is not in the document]
stevemessick: If this were Kotlin, then yes. I think we're better off conforming to existing convention, like `getPackagesFile()`.

src/io/flutter/utils/FlutterModuleUtils.java
stevemessick: Good point. I added a test to return immediately if `FlutterSdk.getFlutterSdk(module.getProject())` is not null. Confirmed that new projects do not return.

stevemessick:
Added a patch to modify the JSON parsing for AF.3.1 -- we have 2.7 on the class path and it does not have #parseString.

src/io/flutter/sdk/FlutterSdkUtil.java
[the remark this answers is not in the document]
devoncarew: Sorry - I was referring to the local variable name here - `final VirtualFile packageConfigFile` (not critical for the PR).

alexander-doroshko:
@stevemessick Sorry, I'm offline until 3/29

PR #5349, opened 2021-03-24
Disable bazel hot restart by default

helin24:
We want to default to disabling updating generated files during hot restart and add an option to turn this on for all projects if desired. (A user could enable this feature globally in settings but manually add a flag to disable for a particular project)

This setting is only visible for bazel projects.

### DIFF
--- a/src/io/flutter/run/bazel/BazelFields.java
+++ b/src/io/flutter/run/bazel/BazelFields.java
@@ -285,7 +285,7 @@ public class BazelFields {
     if (!FlutterSettings.getInstance().isEnableBazelHotRestart() && hasDisabledArg) {
       final Notification notification = new Notification(
         FlutterMessages.FLUTTER_NOTIFICATION_GROUP_ID,
-        "Google3 hot restart is disabled by default",
+        "Google3-specific hot restart is disabled by default",
         "You can now remove this flag from your configuration's additional args: " + disableBazelHotRestartParam,
         NotificationType.INFORMATION);
       Notifications.Bus.notify(notification, project);
--- a/src/io/flutter/run/bazel/BazelFields.java
+++ b/src/io/flutter/run/bazel/BazelFields.java
@@ -28,6 +28,7 @@ import io.flutter.run.FlutterDevice;
 import io.flutter.run.common.RunMode;
 import io.flutter.run.daemon.DevToolsInstance;
 import io.flutter.run.daemon.DevToolsService;
+import io.flutter.settings.FlutterSettings;
 import io.flutter.utils.ElementIO;
 import org.jdom.Element;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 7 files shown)


src/io/flutter/run/bazel/BazelFields.java
```diff
@@ -269,6 +274,12 @@ GeneralCommandLine getLaunchCommand(@NotNull Project project,
       commandLine.addParameter(additionalArgsTokenizer.nextToken());
     }
 
+    final String disableBazelHotRestartParam = "--no-enable-google3-hot-reload";
+    if (!FlutterSettings.getInstance().isEnableBazelHotRestart() &&
+        !StringUtil.notNullize(additionalArgs).contains(disableBazelHotRestartParam)) {
```
helin24, line 279: Should I add a message or notification if this flag is in additional args? Could be useful if we want users to stop managing this themselves so we can switch over to default enabled when it's working more consistently in the future.

jacob314: Yes that is a good idea.

src/io/flutter/run/bazel/BazelFields.java
```diff
@@ -269,6 +274,12 @@ GeneralCommandLine getLaunchCommand(@NotNull Project project,
       commandLine.addParameter(additionalArgsTokenizer.nextToken());
     }
 
+    final String disableBazelHotRestartParam = "--no-enable-google3-hot-reload";
```
jacob314, line 277: lets actually swap this around and add a new 
--enable-google3-hot-restart param to the bazel integration.

That way we can change the default in bazel to use the more robust external work flows. Otherwise, only IDE users will benefit from the more robust experience by default.

jacob314 (commented):
Looks good except maybe we should land an alternate parameter for enabling bazel specific reload in g3 so we can swap the g3 default behavior.

src/io/flutter/run/bazel/BazelFields.java
```diff
@@ -274,10 +278,19 @@ GeneralCommandLine getLaunchCommand(@NotNull Project project,
...
     final String disableBazelHotRestartParam = "--no-enable-google3-hot-reload";
-    if (!FlutterSettings.getInstance().isEnableBazelHotRestart() &&
-        !StringUtil.notNullize(additionalArgs).contains(disableBazelHotRestartParam)) {
-      commandLine.addParameter(disableBazelHotRestartParam);
+    final boolean hasEnabledArg = StringUtil.notNullize(additionalArgs).contains(enableBazelHotRestartParam);
+    final boolean hasDisabledArg = StringUtil.notNullize(additionalArgs).contains(disableBazelHotRestartParam);
+    if (!FlutterSettings.getInstance().isEnableBazelHotRestart() && hasDisabledArg) {
+      final Notification notification = new Notification(
+        FlutterMessages.FLUTTER_NOTIFICATION_GROUP_ID,
+        "Google3 hot restart is disabled by default",
```
jacob314, line 288: Nit: this makes it sound like hot restart is disabled rather than just "Google3 specific hot restart"

jacob314 (approved):
lgtm

PR #5369, opened 2021-04-01
Follow up on the isFailure json attribute change, removal of the error attribute to match the Dart IJ Plugin

jwren:
Dart IJ Plugin change: https://github.com/JetBrains/intellij-plugins/commit/9bf7b23b8c8aa8e1ec19ec98c6248d413a84d629


devoncarew (approved):
👍

PR #5426, opened 2021-04-20
Report VM service connection problems

helin24:
We want to see how often different types of internal connection errors are occurring. I could make this bazel-specific but it seems like we might want to know how often these problems are occurring externally also, or at least check on them for one release.


jacob314 (approved):
lgtm

PR #5430, opened 2021-04-22
Add notification for test and run if config warning

helin24:
We want to additionally prompt users with a notification if there is a warning about internal configuration. The message can include a link, so I added simple parsing for a URL to make it clickable.

~~I was originally going to change this for running too but I need some help finding where that text is parsed. If this is a good approach for test and run, then I'll make a separate change for app run/debug.~~

### DIFF
--- a/src/io/flutter/FlutterMessages.java
+++ b/src/io/flutter/FlutterMessages.java
@@ -6,6 +6,7 @@
 package io.flutter;
 
 import com.intellij.notification.Notification;
+import com.intellij.notification.NotificationListener;
 import com.intellij.notification.NotificationType;
 import com.intellij.notification.Notifications;
 import com.intellij.openapi.project.Project;
@@ -17,6 +18,7 @@ import org.jetbrains.annotations.Nullable;
 
 public class FlutterMessages {
   public static final String FLUTTER_NOTIFICATION_GROUP_ID = "Flutter Messages";
+  public static final String BAZEL_CONFIG_WARN_PREFIX = "Configuration warning: ";
 
   private FlutterMessages() {
   }
@@ -37,7 +39,7 @@ public class FlutterMessages {
         FLUTTER_NOTIFICATION_GROUP_ID,
         title,
         message,
-        NotificationType.WARNING), project);
+        NotificationType.WARNING, NotificationListener.URL_OPENING_LISTENER), project);
   }
 
   public static void showInfo(String title, String message, @Nullable Project project) {
--- a/src/io/flutter/FlutterMessages.java
+++ b/src/io/flutter/FlutterMessages.java
@@ -18,6 +18,7 @@ import org.jetbrains.annotations.Nullable;
 
 public class FlutterMessages {
   public static final String FLUTTER_NOTIFICATION_GROUP_ID = "Flutter Messages";
+  // TODO(helinx): We may want to set this string in flutter.json or move this check to the g3 dart plugin.
   public static final String BAZEL_CONFIG_WARN_PREFIX = "Configuration warning: ";
 
   private FlutterMessages() {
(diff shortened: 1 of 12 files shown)


src/io/flutter/utils/UrlUtils.java
```diff
@@ -0,0 +1,22 @@
+package io.flutter.utils;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class UrlUtils {
+    public static String addUrlTags(String input) {
```
jacob314, line 7: nit: add methods should add something to a list. generateHtmlFragmentWithHrefTags.

src/io/flutter/utils/UrlUtils.java
```diff
@@ -0,0 +1,22 @@
...
+        StringBuilder builder = new StringBuilder();
+        for (String token : input.split(" ")) {
+            if (builder.length() > 0) {
+                builder.append(" ");
+            }
+            try {
+                URL url = new URL(token);
```
jacob314, line 14: in general, having code that throws a bunch of exceptions in its normal operation is annoying. Is there a method you can call to check whether a token is a valid URL without trying and catching? If there is not and this is used sparingly, fine to leave as is as IntelliJ already throws tons of exceptions.

testSrc/unit/io/flutter/utils/UrlUtilsTest.java
```diff
@@ -0,0 +1,20 @@
+package io.flutter.utils;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class UrlUtilsTest {
+    @Test
+    public void testAddUrlTags() {
+        assertEquals(
```
jacob314, line 10: nit: add test with multiple links.

src/io/flutter/FlutterMessages.java
```diff
@@ -17,6 +18,7 @@
 
 public class FlutterMessages {
   public static final String FLUTTER_NOTIFICATION_GROUP_ID = "Flutter Messages";
+  public static final String BAZEL_CONFIG_WARN_PREFIX = "Configuration warning: ";
```
jacob314, line 21: Matching on this string could have false positives. Fyi @jwren.
We should consider making the string for bazel configuration warnings more unique and we need to document that this prefix needs to match the bazel tools.
An alternative to this design would be to display all bazel warnings in the g3 dart plugin. That would also allow showing the warning even before you run an app.

helin24: Is there a way to read configuration from flutter.json from within g3, or have flutter.json return a variable from elsewhere in g3? 

Using the g3 dart plugin would be nice too. This maybe wouldn't need to be specific to actions like run/test and would just detect whether the setup is running on mac without remote Dart. 

For now I think I'll just add a todo item. I wanted a message to appear for tools run on the command line as well.

src/io/flutter/utils/UrlUtils.java
helin24: We could use an apache validator instead? It would require importing a new library to our plugin. Otherwise I think this is okay since it's only triggered when the prefix is found.

testSrc/unit/io/flutter/utils/UrlUtilsTest.java
helin24: There's one example of this on line 16; let me know if you're thinking of something else though.

helin24:
Added this notification for run in our process handler. Let me know if there's a better place to put this (if we're already listening to text elsewhere).

src/io/flutter/FlutterMessages.java
jacob314: lets file a bug along with the todo.

jacob314 (approved):
lgtm

PR #5443, opened 2021-04-27
Add option to display all stack traces in debugger log

### LINKED ISSUE #5421: Flutter plugin version 55.15 swallows errors
This is a regression. Please see: https://github.com/flutter/flutter-intellij/issues/4952

You are forcing me to catch my errors and print the stackTrace myself. I appreciate the effort, but I think you are just trying to be "too smart". I want to have the complete stacktrace, always. You can't try to decide for me which part is important, because you can never know that for sure, and even if you are right 99% of the time, the other 1% will create terrible problems and make my team waste countless hours. Stacktrace is free, while my programmers' time is not.

I have broken our app into independent packages/modules, and now you are not giving me the part of the stacktrace that's inside another package (when the error crosses package boundaries). Also, when I use third-party packages I can't know where certain problems happen, so it doesn't help me anymore to fix the problem, or even to allow me to help the package author to solve it. Even when the error is inside of Flutter itself, I do want to see all of the stacktrace.

Flutter's Stacktrace is full of useless info from Gradle and Flutter initialization. The useful part, which are the stacktraces, you are cutting out. Please don't. Please, let me have all of my stacktrace.

stevemessick:
Fixes #5421 by adding an option to display all stack traces in the console log.

### DIFF
--- a/src/io/flutter/settings/FlutterSettings.java
+++ b/src/io/flutter/settings/FlutterSettings.java
@@ -27,6 +27,7 @@ public class FlutterSettings {
   private static final String showOnlyWidgetsKey = "io.flutter.showOnlyWidgets";
   private static final String syncAndroidLibrariesKey = "io.flutter.syncAndroidLibraries";
   private static final String showStructuredErrors = "io.flutter.showStructuredErrors";
+  private static final String includeAllStackTraces = "io.flutter.ncludeAllStackTraces";
   private static final String showBuildMethodGuidesKey = "io.flutter.editor.showBuildMethodGuides";
   private static final String enableHotUiKey = "io.flutter.editor.enableHotUi";
   private static final String enableEmbeddedBrowsersKey = "io.flutter.editor.enableEmbeddedBrowsers";
@@ -108,6 +109,10 @@ public class FlutterSettings {
 
     if (isShowStructuredErrors()) {
       analytics.sendEvent("settings", afterLastPeriod(showStructuredErrors));
+
+      if (isIncludeAllStackTraces()) {
+        analytics.sendEvent("settings", afterLastPeriod(includeAllStackTraces));
+      }
     }
 
     if (showAllRunConfigurationsInContext()) {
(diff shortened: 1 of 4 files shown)


jacob314:
Would it be possible to make the stack traces expandable in the console but collapsed by default? That would be a lot more discoverable than changing a plugin setting. Personally I like the current default but occasionally need to expand a stack trace and can't.

stevemessick:
@jacob314 That sounds like a good topic for a feature request, if you'd like to create one.

jacob314:
Filed an issue.

src/io/flutter/settings/FlutterSettings.java
```diff
@@ -27,6 +27,7 @@
   private static final String showOnlyWidgetsKey = "io.flutter.showOnlyWidgets";
   private static final String syncAndroidLibrariesKey = "io.flutter.syncAndroidLibraries";
   private static final String showStructuredErrors = "io.flutter.showStructuredErrors";
+  private static final String includeAllStackTraces = "io.flutter.ncludeAllStackTraces";
```
jacob314, line 30: nit: can we make this member on the one on line 29 use the word "Key" for consistency?

jacob314 (approved):
lgtm

src/io/flutter/settings/FlutterSettings.java
stevemessick: I should at least change the spelling of ".ncl" to ".incl" but there are others that do not include "Key". I'll just change them all.

marcglasberg:
@stevemessick Hey, I am a bit lost. How do I actually use this?

```
private static final String includeAllStackTraces = "io.flutter.ncludeAllStackTraces";
```

stevemessick:
> How do I actually use this?

There will be a new check box in the Flutter settings, available in M56. That release is scheduled for May 3. I'll include a screen shot in the release notes, which will be published to the flutter-announce mailing list.

PR #5454, opened 2021-04-29
initial null safe conversion of tool/plugin

devoncarew:
- initial null safe conversion of the `tool/plugin` tool


stevemessick (approved):
When I saw 12 files had changed I decided to wait until after M56 was ready before diving in. Turns out, I didn't need to wait, heh.

PR #5456, opened 2021-04-30
Check for new local Dart SDK

helin24:
This fixes the workspace check for bazel projects and is a better solution than https://github.com/flutter/flutter-intellij/pull/5455.


jacob314 (approved):
lgtm

PR #5463, opened 2021-05-05
Run tests with coverage

stevemessick:
This enables the "Run with Coverage" option on the main toolbar, and in the various Run menus. The toolbar button is the right-most one in this screen shot:
Screen Shot 2021-05-07 at 7 27 25 AM

All test modes are supported: all in directory, file, or filtered by name.

This feature re-uses the Java coverage framework. When tests first run you are asked if you want to delete the newly-created coverage data file. Say no! If you delete it you will not be able to see coverage data, even in future runs. (There must be someway to cause it to show, but I have not found it.)
Screen Shot 2021-05-07 at 7 28 07 AM

After running tests, the Cover tool window allows tests to easily be re-run with coverage by using its version of the "Run with Coverage" button.
Screen Shot 2021-05-07 at 7 28 36 AM

There is still work to be done. The Java coverage support includes two things we do not:
- The coverage tool window that displays coverage data in a table
- The ability to generate coverage reports to view in a browser


src/io/flutter/run/coverage/FlutterCoverageAnnotator.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  protected String getFilesCoverageInformationString(@NotNull DirCoverageInfo info) {
+    if (info.totalFilesCount == 0) {
+      return null;
+    }
+    else {
+      return info.coveredFilesCount == 0
+             ? FlutterBundle.message("coverage.string.0.of.1.files.covered", info.coveredFilesCount, info.totalFilesCount)
```
helin24, line 69: I don't understand the intended difference between these two messages. If 0 files are covered, it shows "0 of Y files covered", but if more are covered it shows "X of Y file"? Is it not clear that the 0 case is describing amount of coverage?

src/io/flutter/run/coverage/FlutterCoverageEnabledConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+      throw new RuntimeException(FlutterBundle.message("project.root.not.found"));
+    }
+    myCoverageFilePath = roots.get(0).getRoot().getCanonicalPath() + "/coverage/lcov.info";
+    GuiUtils.invokeLaterIfNeeded(
+      () -> setCurrentCoverageSuite(CoverageDataManager.getInstance(configuration.getProject()).addCoverageSuite(this)),
+      ModalityState.any());
+  }
+
+  public void setCoverageRunner(@Nullable final CoverageRunner coverageRunner) {
+    final String path = myCoverageFilePath;
```
helin24, line 38: Why does `path` get set to `myCoverageFilePath` and then the latter is set to the former? Maybe a comment here would be helpful. Also confusing to see that `myCoverageFilePath` is set again later in the constructor.

helin24 (approved):

src/io/flutter/run/coverage/FlutterCoverageAnnotator.java
stevemessick: TBH I just copied what the Dart plugin did. You're right, it probably needs to be simplified.

src/io/flutter/run/coverage/FlutterCoverageEnabledConfiguration.java
stevemessick: I'll add a comment explaining that the super method clears the value and we want to keep it. It seems like the constructor could just call the super method directly. This code went through quite a few iterations while trying to figure out how to make the Java coverage framework for us. Thanks for catching things that need cleaning up.

PR #5474, opened 2021-05-11
Use --config-only to set up Xcode

### LINKED ISSUE #5411: Open in Xcode no longer needs to do a full build
As mentioned in https://github.com/flutter/flutter-intellij/issues/5408#issuecomment-819118860 there is a new option to prepare a project for Xcode prior to opening it. We need to change the `Open in Xcode` function to use it.

stevemessick:
Fixes #5411.

### DIFF
--- a/src/io/flutter/actions/OpenInXcodeAction.java
+++ b/src/io/flutter/actions/OpenInXcodeAction.java
@@ -22,10 +22,13 @@ import io.flutter.FlutterMessages;
 import io.flutter.FlutterUtils;
 import io.flutter.pub.PubRoot;
 import io.flutter.sdk.FlutterSdk;
+import io.flutter.sdk.FlutterSdkVersion;
 import io.flutter.utils.FlutterModuleUtils;
 import io.flutter.utils.ProgressHelper;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+
+import java.io.IOException;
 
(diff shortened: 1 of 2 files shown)


helin24 (approved):

src/io/flutter/actions/OpenInXcodeAction.java
```diff
@@ -104,9 +111,25 @@ public void processTerminated(@NotNull ProcessEvent event) {
...
+      // Derived from packages/flutter_tools/test/integration.shard/build_ios_config_only_test.dart
+      final VirtualFile ios = pubRoot.getRoot().findChild("ios");
+      if (ios == null || !ios.isDirectory()) return false;
+      final VirtualFile flutter = ios.findChild("Flutter");
+      if (flutter == null || !flutter.isDirectory()) return false;
+      final VirtualFile gen = flutter.findChild("Generated.xcconfig");
+      if (gen == null || gen.isDirectory()) return false;
+      try {
+        final String contents = new String(gen.contentsToByteArray());
+        return contents.contains("DART_OBFUSCATION=true");
```
jmagman, line 125: This is only true in `build_ios_config_only_test.dart` because `--obfuscate` was passed into the integration test.  
https://github.com/flutter/flutter/blob/99472cf072f154a6698c9795918ce480d49ceb57/packages/flutter_tools/test/integration.shard/build_ios_config_only_test.dart#L31
You shouldn't rely on this value being present or set to `true`.

PR #5476, opened 2021-05-11
Fix step-over async call

### LINKED ISSUE #5449: Debugger: Step to the next line in an async function.
Debugging async functions in Dart is super annoying.

I am at line 10. I want to go to line 11:

```
10   var x = await getSomething(); ➜ I am here
11   print(x); ➜ Just want to go here
```

But the `await` means some other code will have the chance to run, and we go to a completely different part of the code.

If I want to go to line 11, I am forced to add a breakpoint there or "force run" to line 11. That's painful. Sometimes I don't even notice there is an await in the middle of the line, and when I try to go the the next line it goes somewhere else and don't even know where I was anymore. I have to restart and pay double attention, which wastes my brain CPU and makes it harder for me to reason about the code.

I was told by Simon Binder that the Dart VM protocol has a "step to the next line in an async function" command (and that is accessible from the observatory), but somehow all IDEs don't use that. 

Please add this to the debugger.

stevemessick:
Fixes #5449


jacob314 (approved):
lgtm

Issue #5400, opened 2021-04-12
Exception: Trying to reset custom component in a presentation


ksilz:
## Steps to Reproduce

Sorry for not being much help here - this just popped up in my IntelliJ event log somewhere along the way.
- I dowgraded Flutter to version 1.22.6 again while in IntelliJ.
- I started the application once more on an Android emulator.
- It may have happened then. 

## Version information

IntelliJ IDEA `2021.1` • Flutter plugin `io.flutter 55.0.5` • Dart plugin `211.6693.108`

Flutter 1.22.6 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 9b2d32b605 (3 months ago) • 2021-01-22 14:36:39 -0800
Engine • revision 2f0af37152
Tools • Dart 2.10.5

## Exception

Trying to reset custom component in a presentation

```
java.lang.Throwable
	at com.intellij.openapi.actionSystem.Presentation.putClientProperty(Presentation.java:403)
	at com.intellij.openapi.actionSystem.Presentation.copyFrom(Presentation.java:376)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.lambda$reflectSubsequentChangesInOriginalPresentation$9(ActionUpdater.java:140)
	at java.desktop/java.beans.PropertyChangeSupport.fire(PropertyChangeSupport.java:341)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:333)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:266)
	at java.desktop/java.beans.PropertyChangeSupport.firePropertyChange(PropertyChangeSupport.java:308)
	at com.intellij.openapi.actionSystem.Presentation.fireBooleanPropertyChange(Presentation.java:335)
	at com.intellij.openapi.actionSystem.Presentation.setVisible(Presentation.java:302)
	at io.flutter.actions.DeviceSelectorAction.updateVisibility(DeviceSelectorAction.java:125)
	at io.flutter.actions.DeviceSelectorAction.lambda$update$2(DeviceSelectorAction.java:119)
	at com.intellij.openapi.application.impl.ApplicationImpl.invokeAndWait(ApplicationImpl.java:415)
	at io.flutter.FlutterUtils.invokeAndWait(FlutterUtils.java:96)
	at io.flutter.actions.DeviceSelectorAction.update(DeviceSelectorAction.java:117)
[...]

PR #5480, opened 2021-05-12
Update device selector on EDT

stevemessick:


helin24 (approved):

stevemessick:
Fixes #5400 and other reports of `Presentation.putClientProperty()` reporting an error.

Issue #5472, opened 2021-05-11
Plugin error at running default new flutter app on M1


alex-pricope:
## What happened

(please describe what you were doing when this exception occurred)

## Version information

IntelliJ IDEA `2021.1.1` • Flutter plugin `io.flutter 56.0.5` • Dart plugin `211.7233`

Flutter 2.2.0-10.2.pre • channel beta • https://github.com/flutter/flutter.git
Framework • revision b5017bf8de (12 days ago) • 2021-04-28 17:09:53 -0700
Engine • revision 91ed51e05c
Tools • Dart 2.13.0 (build 2.13.0-211.13.beta)

## Exception

null

```
java.lang.UnsatisfiedLinkError: /private/var/folders/xx/d3ycs3bd3hl1b0pcw8y7k4200000gn/T/JxBrowser/7.14/libbrowsercore_toolkit.dylib: dlopen(/private/var/folders/xx/d3ycs3bd3hl1b0pcw8y7k4200000gn/T/JxBrowser/7.14/libbrowsercore_toolkit.dylib, 1): no suitable image found.  Did find:
	/private/var/folders/xx/d3ycs3bd3hl1b0pcw8y7k4200000gn/T/JxBrowser/7.14/libbrowsercore_toolkit.dylib: mach-o, but wrong architecture
	/private/var/folders/xx/d3ycs3bd3hl1b0pcw8y7k4200000gn/T/JxBrowser/7.14/libbrowsercore_toolkit.dylib: stat() failed with errno=22
	at java.base/java.lang.ClassLoader$NativeLibrary.load0(Native Method)
	at java.base/java.lang.ClassLoader$NativeLibrary.load(ClassLoader.java:2442)
	at java.base/java.lang.ClassLoader$NativeLibrary.loadLibrary(ClassLoader.java:2498)
	at java.base/java.lang.ClassLoader.loadLibrary0(ClassLoader.java:2694)
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2627)
	at java.base/java.lang.Runtime.load0(Runtime.java:768)
	at java.base/java.lang.System.load(System.java:1837)
	at com.teamdev.jxbrowser.internal.JniLibrary.loadLibrary(JniLibrary.java:53)
	at com.teamdev.jxbrowser.internal.JniLibrary.lambda$new$0(JniLibrary.java:47)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
[...]

PR #5487, opened 2021-05-14
Detect M1 mac and skip JxBrowser

helin24:
If a user is using an M1 mac and the version of IntelliJ native to this chip, then JxBrowser is not currently supported. If we detect this combination then show users the option to open DevTools in the browser instead.

It seems that automatically disabling embedded browser for situations where JxBrowser is not supported results in the setting being somewhat sticky once it is supported, e.g. it seems previously disabling for Big Sur resulted in a much higher proportion of users staying disabled despite no additional errors that would suggest a poorer user experience (maybe this setting sticks around as long as users keep their version of IntelliJ? I'm not sure). So I'm inclined to try prompting users to open in the browser instead. This should be a smaller proportion of users anyway.

Fixes https://github.com/flutter/flutter-intellij/issues/5472

### DIFF
--- a/src/io/flutter/view/FlutterView.java
+++ b/src/io/flutter/view/FlutterView.java
@@ -54,9 +54,7 @@ import io.flutter.inspector.DiagnosticsNode;
 import io.flutter.inspector.InspectorGroupManagerService;
 import io.flutter.inspector.InspectorService;
 import io.flutter.inspector.InspectorSourceLocation;
-import io.flutter.jxbrowser.EmbeddedBrowser;
-import io.flutter.jxbrowser.JxBrowserManager;
-import io.flutter.jxbrowser.JxBrowserStatus;
+import io.flutter.jxbrowser.*;
 import io.flutter.run.FlutterDevice;
 import io.flutter.run.daemon.DevToolsInstance;
 import io.flutter.run.daemon.DevToolsService;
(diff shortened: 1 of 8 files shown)


src/io/flutter/view/FlutterView.java
```diff
@@ -623,12 +627,12 @@ protected void presentClickableLabel(ToolWindow toolWindow, List<LabelInput> lab
...
     for (LabelInput input : labels) {
       if (input.listener == null) {
-        final JLabel descriptionLabel = new JLabel(input.text);
+        final JLabel descriptionLabel = new JLabel("<html>" + input.text + "</html>");
         descriptionLabel.setBorder(JBUI.Borders.empty(5));
         descriptionLabel.setHorizontalAlignment(SwingConstants.CENTER);
         panel.add(descriptionLabel, BorderLayout.NORTH);
       } else {
-        final LinkLabel<String> linkLabel = new LinkLabel<>(input.text, null);
+        final LinkLabel<String> linkLabel = new LinkLabel<>("<html>" + input.text + "</html>", null);
```
jacob314, line 635: adding <html> seems odd. Are they needed? this is still far from valid html with <html> tags added.

tool/plugin/lib/edit.dart
```diff
@@ -136,6 +136,18 @@ List<EditCommand> editCommands = [
     replacement: 'urlClassLoader.addFiles(paths)',
     versions: ['2021.1'],
   ),
+  Subst(
```
jacob314, line 139: @stevemessick is this the best way to do this?
The alternative I can think of is to use reflection to dynamically call the api.

jacob314 (commented):
We should also understand why we accidentally set the sticky opt out instead of making it optional. I expect we changed the value of the setting when we meant to leave it alone but change the value surfaced to users.

tool/plugin/lib/edit.dart
stevemessick: This allows us to avoid reflection.

stevemessick (approved):

src/io/flutter/view/FlutterView.java
helin24: I added this to get text wrapping within the panel. I read that we can also get text wrapping by changing to a text area with some other adjustments, but this seems more straightforward.

helin24:
> We should also understand why we accidentally set the sticky opt out instead of making it optional. I expect we changed the value of the setting when we meant to leave it alone but change the value surfaced to users.

I think this is what happened:
- A few versions ago, Big Sur users had a disabled "embedded browser" field 
- If a user did not adjust any of their other Flutter settings, the embedded browser field does not get a value
- If a user DID adjust any of their other Flutter settings, the embedded browser field was set to false, since we were returning false if Big Sur
- Then we re-enabled the embedded browser field by default, which failed to enable for the second group of users, since they  had a false setting.

This seems to reproduce for me locally. I should have made this setting line conditional on not big sur: https://github.com/flutter/flutter-intellij/pull/5065/files#diff-56079033e7fb39f494b815b798ad070de67300e721b0658082279962ab024389R306

So to fix this, I'll override a false setting for Big Sur users once but allow users to change it back afterwards.

PR #5489, opened 2021-05-17
Xcode config only

stevemessick:
@jmagman This version compares timestamps, as done in flutter_tools. Thanks again!


helin24 (approved):

jmagman (commented):
LGTM!

PR #5490, opened 2021-05-18
Enable embedded browser (once) for Big Sur users

helin24:
When we disabled the embedded browser option for Big Sur users (back when JxBrowser did not support Big Sur yet - see https://github.com/flutter/flutter-intellij/pull/5065), this caused embedded browser to stay disabled if users adjusted unrelated settings (See https://github.com/flutter/flutter-intellij/pull/5487#issuecomment-842517418 for more detail). 

We want to enable embedded browser for Big Sur users by default now, but only once (if a user elects to disable embedded browser in the future, we want to allow this).

### DIFF
--- a/src/io/flutter/settings/FlutterSettings.java
+++ b/src/io/flutter/settings/FlutterSettings.java
@@ -10,6 +10,7 @@ import com.intellij.ide.plugins.IdeaPluginDescriptor;
 import com.intellij.ide.plugins.PluginManagerCore;
 import com.intellij.ide.util.PropertiesComponent;
 import com.intellij.openapi.components.ServiceManager;
+import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.util.registry.Registry;
 import com.intellij.util.EventDispatcher;
 import com.jetbrains.lang.dart.analyzer.DartClosingLabelManager;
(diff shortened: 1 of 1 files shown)


stevemessick (commented):
LGTM for the code.

However, I'm a bit uncertain about the approach. It could offend users who had already disabled the embedded browser. @devoncarew might have some thoughts.

helin24:
> It could offend users who had already disabled the embedded browser.

I'm assuming this is a very small number of users, but perhaps I could run a query to check that this really is the case (I could also check on the number of changes to unrelated settings to see if my hypothesis is correct). Unfortunately there isn't a way to identify that a user has done the disabling. 

I could add a notification that we are doing this switch so that users who have disabled manually are alerted and get instructions how to redo it.

stevemessick:
> I could add a notification that we are doing this switch so that users who have disabled manually are alerted and get instructions how to redo it.

I think that would be adequate. Thanks!

AlexV525:
> I'm assuming this is a very small number of users

Unfortunately I'm the one of these users. And from feedbacks from my community, many developers have disabled it currently. But a notification is good enough for this approach.

helin24:
@AlexV525 that's good to know! Have there been any common reasons for why developers in your community have disabled, particularly for Macs?

AlexV525:
> @AlexV525 that's good to know! Have there been any common reasons for why developers in your community have disabled, particularly for Macs?

AFAIK not only Mac users did this, but other platforms' users did it too. The most common reason for disabling it is that the inspector failed to start, initialize, or inspect widgets in the most of time, and the `chrome_crashpad_helper` consumes a lot memory and failed to be killed. I'm not sure whether these issues have been improved since last time I disable it about nearly half a year ago, IIRC.

AlexV525:
Oh and for Windows users previously they have an issue about the JxBrowser failed to install/start in unknown reasons when using Android Studio.

stevemessick:
@AlexV525 There have been a lot of changes to address the problems you describe. It would help us a lot, if you have time, to turn it on again and see if it works better for you now.

helin24:
@AlexV525 thanks for the feedback! As Steve mentioned, we've made some changes for these problems and we're also actively working on issues reported related to the embedded browser. So if you write up any problems we will look into it.

AlexV525:
I just told developers in our community to enabled the embedding browser and see if I can get any feedbacks.

helin24:
@AlexV525 Great! Thanks for the update.

PR #5494, opened 2021-05-19
Update live templates to assume null-safety

JuYeong0413:
Update IntelliJ live templates to null safety as VS Code snippets are updated in https://github.com/Dart-Code/Dart-Code/commit/2d2f7e222a43392717d6fd064e7c936b517a21b4.

* Add `?` operator to the `Key key` argument
* Update `inh` to match the [docs](https://api.flutter.dev/flutter/widgets/InheritedWidget-class.html)
* Add `late` keyword to the `AnimationController` in `stanim`

Examples:

https://user-images.githubusercontent.com/33684401/118861513-8ca5e200-b917-11eb-9713-f35fc50e0352.mov

https://user-images.githubusercontent.com/33684401/118864419-eeb41680-b91a-11eb-9eb4-7da7aa8bcb71.mov

https://user-images.githubusercontent.com/33684401/118861491-87489780-b917-11eb-95d9-f24e07ef5924.mov

/cc @stevemessick @AlexV525

PR #5496, opened 2021-05-20
Update `stanim` live template

JuYeong0413:
Apply review from https://github.com/flutter/flutter-intellij/pull/5494#discussion_r635689378
I agree with adding `final` for personal preference too, and at the same time I think that being consistent with [the sample code in api docs](https://api.flutter.dev/flutter/animation/AnimationController-class.html) and [VS Code snippet](https://github.com/Dart-Code/Dart-Code/blob/7608e7d217f6b9bd1d4f4370bb941aafe0369115/snippets/flutter.json#L44-L73) is also considerable.

I'd be happy to hear some views about this. /cc @AlexV525 @stevemessick


AlexV525:
I would consult @Hixie with these changes.

resources/liveTemplates/stanim.txt
```diff
@@ -6,12 +6,11 @@ class $NAME$ extends StatefulWidget {
...
 
 class _$NAME$State extends State<$NAME$> with SingleTickerProviderStateMixin {
-  late AnimationController _controller;
+  late final AnimationController _controller = AnimationController(vsync: this);
 
   @override
   void initState() {
     super.initState();
-    _controller = AnimationController(vsync: this);
   }
```
AlexV525, line 14: Once we remove this initializer, the override method will no longer needed.

jonahwilliams:
This will only create the animation controller the first time _controller is accessed. Depending on how the code is updated, the user might not actually access and may be confused when it isn't running. For example, try running the following with the second print commented/uncommented.

```dart
void main() {
  runApp(Foo());
}

class Foo extends StatefulWidget {
  
  State<Foo> createState() => FooState();
}


class FooState extends State<Foo> with SingleTickerProviderStateMixin {
  late final controller = AnimationController(vsync: this, duration: Duration(seconds: 1))
     ..addListener(() {
       print('test');
     })
     ..forward();
  
  void initState() {
    super.initState();
//     print(controller);
  }  
  Widget build(_) => Container();
}

```

AlexV525:
Well that's an amazing case. Ideally I think everyone will bind the controller to an actual widget to use, but from this point of view I think you're right. @JuYeong0413 What do you think?

PR #5514, opened 2021-05-25
Fix edit.dart to clean up its changes

stevemessick:
The `Subst` edit is only for a single change per file. You can use `MultiSubst` for multiple changes. I should have caught this during the code review.


helin24 (approved):
Ah thanks!

PR #5517, opened 2021-05-25
Stop using DevTools URL util method, reorg tests

helin24:
This is to clean up an old util method that we don't really need since adding a `DevToolsUrl` class and move its associated tests.

Also adding a test for finding a widget ID since we didn't have one before.

(does not need to be part of v57)

### DIFF
--- a/src/io/flutter/devtools/DevToolsUrl.java
+++ b/src/io/flutter/devtools/DevToolsUrl.java
@@ -22,7 +22,13 @@ public class DevToolsUrl {
   public String widgetId;
   public Float fontSize;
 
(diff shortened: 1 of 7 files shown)


stevemessick (approved):
DevToolsUrlTest needs to be formatted but otherwise LGTM. If you don't want this in M57 then please wait to merge it.

helin24:
> DevToolsUrlTest needs to be formatted

Do we have a command that does this? Or do you just use Cmd+Alt+L? The latter didn't change anything for me.

stevemessick:
There are some format commands on the `Code` menu. That file has too much indentation.

PR #5520, opened 2021-05-26
update the set of supported IntelliJ versions

devoncarew:
- update the set of supported IntelliJ versions

This drops two older IntelliJ versions (2020.1 and 2020.2).

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [4.1, 4.2, AF.3.1, 2020.4, 2021.1]
+        version: [AF.3.1, 2020.4, 2021.1]
     steps:
       - name: checkout
         uses: actions/checkout@v2
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [AF.3.1, 2020.4, 2021.1]
+        version: [4.2, AF.3.1, 2020.4, 2021.1]
     steps:
       - name: checkout
         uses: actions/checkout@v2
--- a/product-matrix.json
+++ b/product-matrix.json
@@ -2,33 +2,6 @@
   "list": [
     {
       "channel": "stable",
-      "comments": "IntelliJ 2020.1.1, Android Studio 4.1",
-      "name": "Android Studio 4.1",
-      "version": "4.1",
-      "ideaProduct": "android-studio",
-      "ideaVersion": "201.6565218",
-      "dartPluginVersion": "201.7223.99",
-      "sinceBuild": "201.7223",
-      "untilBuild": "201.*",
-      "filesToSkip": [
-        "src/io/flutter/utils/AndroidLocationProvider.java",
-        "flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java"
-      ]
-    },
-    {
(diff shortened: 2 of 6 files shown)


helin24:
Is this for v57? I believe @stevemessick is testing that version now already. 

I'm not sure if we normally change `edit.dart` with these versioning changes, but you may be able to remove old code substitutions as well.

devoncarew:
> Is this for v57? I believe @stevemessick is testing that version now already.

It could be but it's not necessary - we just don't need to publish for these versions anymore.

> I'm not sure if we normally change `edit.dart` with these versioning changes, but you may be able to remove old code substitutions as well.

That's a good idea - I'll take a look at what's no longer used.

devoncarew:
edit.dart is now updated; there were a fair number of changes there so its worth looking over

helin24 (approved):
LGTM, maybe Steve will think of other items. We should wait for him to merge since he may want to keep this out of master until v57 is released.

.github/workflows/presubmit.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [4.1, 4.2, AF.3.1, 2020.4, 2021.1]
+        version: [AF.3.1, 2020.4, 2021.1]
```
stevemessick, line 21: Android Studio 4.2 is the current stable release, so we need to support it.

product-matrix.json
```diff
@@ -1,32 +1,5 @@
...
-      "ideaVersion": "201.6565218",
-      "dartPluginVersion": "201.7223.99",
-      "sinceBuild": "201.7223",
-      "untilBuild": "201.*",
-      "filesToSkip": [
-        "src/io/flutter/utils/AndroidLocationProvider.java",
-        "flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java"
-      ]
-    },
-    {
```
stevemessick, line 18: We can't delete this yet.

stevemessick (changes requested):

product-matrix.json
devoncarew: This should only be used by 2020.1.1 / AS 4.1?

.github/workflows/presubmit.yaml
devoncarew: Ah, this was inadvertent; I'll re-add.

product-matrix.json
devoncarew: Ah, you mean the section below.

tool/plugin/lib/edit.dart
```diff
@@ -175,19 +135,9 @@ class EditAndroidModuleLibraryManager extends EditCommand {
...
-        source = source.replaceAll(
-            "androidProject.init42", "androidProject.initPre41");
-        source = source.replaceAll("PluginManagerCore.getLoadedPlugins(), null",
-            "PluginManagerCore.getLoadedPlugins(), false");
-      }
       if (spec.version.startsWith('4.2')) {
-        source = source.replaceAll(
-            "androidProject.init42", "androidProject.init41");
+        source =
+            source.replaceAll("androidProject.init42", "androidProject.init41");
```
devoncarew, line 140: We might be able to update this as well?

stevemessick: Best to leave that alone for now. I'm hoping to delete it sometime soon.

stevemessick (approved):
I'll publish M57 with support for 4.1. No need to merge this until June.

devoncarew:
Landing this now -

PR #5535, opened 2021-06-01
Clean up unused version for bazel

helin24:
We should remove this from flutter.json since it's no longer being used or updated.


jacob314 (approved):
lgtm

Issue #5557, opened 2021-06-09
With the latest Flutter plugin I don't the DevTools based inspector


jacob314:
<img width="1020" alt="Screen Shot 2021-06-08 at 8 52 10 PM" src="https://user-images.githubusercontent.com/1226812/121290480-88783f00-c89b-11eb-85de-2937a209765c.png">
I was expecting to automatically switch to the DevTools based inspector with the latest version of the Flutter IntelliJ plugin as I am a BigSur user.

Fyi @helin24

helin24:
@jacob314 do you remember if you saw a notification about the embedded browser being turned on?

helin24:
Looking at my change now (https://github.com/flutter/flutter-intellij/pull/5490/files), I'm confused because it looks like maybe it's NOT setting the embedded browser field?? I'll test it out again.

PR #5561, opened 2021-06-09
Change embedded browser setting

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/5557

It looks like the previous change was returning `true` for the embedded browser status once, but not saving it. I'm not sure how I missed this in testing last month; it's possible I built an earlier version with this line while testing the change and then accidentally deleted it before submitting. I verified with this current version that the embedded browser status is set to true.

### DIFF
--- a/src/io/flutter/settings/FlutterSettings.java
+++ b/src/io/flutter/settings/FlutterSettings.java
@@ -38,7 +38,7 @@ public class FlutterSettings {
 
   // TODO(helin24): This is to change the embedded browser setting back to true only once for Big Sur users. If we
   // switch to enabling the embedded browser for everyone, then delete this key.
-  private static final String changeBigSurToTrueKey = "io.flutter.setBigSurToTrueKey";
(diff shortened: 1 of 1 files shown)


src/io/flutter/settings/FlutterSettings.java
```diff
@@ -38,7 +38,7 @@
 
   // TODO(helin24): This is to change the embedded browser setting back to true only once for Big Sur users. If we
   // switch to enabling the embedded browser for everyone, then delete this key.
-  private static final String changeBigSurToTrueKey = "io.flutter.setBigSurToTrueKey";
+  private static final String changeBigSurToTrueKey = "io.flutter.setBigSurToTrueKey2";
```
helin24, line 41: Needed a different key since the earlier one will have been set in v57.

jacob314 (approved):
lgtm

stevemessick (approved):

PR #5570, opened 2021-06-11
update the Inspector to prep for changes to the widget transformer

devoncarew:
- update the Inspector to prep for changes to the widget transformer; remove references to the `parameterLocations` field


jacob314 (approved):
lgtm

PR #5575, opened 2021-06-14
Update build for AS canary

stevemessick:


flutter-studio/src/io/flutter/actions/FlutterNewProjectAction.java_stub
```diff
@@ -85,4 +113,139 @@ public class FlutterNewProjectAction extends AnAction implements DumbAware {
     decorated.setIcon(icon, 1, 0, 0);
     return decorated;
   }
+
+  @SuppressWarnings({"UnstableApiUsage", "unchecked"})
+  private void replaceModuleBuilders() {
+    if (true) return;
```
helin24, line 119: I don't understand the early returns in this file - are these replaced for certain versions?

src/io/flutter/FlutterUtils.java
```diff
@@ -127,7 +127,7 @@ public static boolean isNewAndroidStudioProjectWizard() {
       ApplicationInfo info = ApplicationInfo.getInstance();
       int major = Integer.parseInt(info.getMajorVersion());
       int minor = Integer.parseInt(info.getMinorVersion());
-      return major >= 4 && minor >= 2;
+      return true; //major >= 4 && minor >= 2;
```
helin24, line 130: Is this a real change or for testing?

helin24 (approved):

src/io/flutter/FlutterUtils.java
stevemessick: Both. It's related to your previous question. I need to refactor this since we no longer need to use the old NPW anymore, but it's not high on the list.

PR #5576, opened 2021-06-14
Send caught exceptions as events

helin24:
We want to reserve `Analytics::sendException` for unexpected exceptions rather than for tracking how often different types of caught exceptions occur.

This may be a clumsy usage of the event fields, since I'm combining a label and the exception title into the event action. There is an `eventLabel` field that we don't seem to be using in analytics, but I'm not sure what the abbreviation for this field is - perhaps it would be better to make the event action a description of where the exception comes from (e.g. "embedded-browser-setup") and make the label store the exception title/details.

### DIFF
--- a/src/io/flutter/analytics/Analytics.java
+++ b/src/io/flutter/analytics/Analytics.java
@@ -32,6 +32,7 @@ import java.util.Map;
  */
 public class Analytics {
   public static final String GROUP_DISPLAY_ID = "Flutter Usage Statistics";
+  private static final String EXPECTED_EXCEPTION_CATEGORY = "expected-exception";
 
   private static final String analyticsUrl = "https://www.google-analytics.com/collect";
   private static final String applicationName = "Flutter IntelliJ Plugin";
@@ -112,8 +113,8 @@ public class Analytics {
     sendPayload("timing", args, null);
   }
 
-  public void sendCaughtException(String text) {
-    sendEvent("caught-exception", text);
+  public void sendExpectedException(String location, Throwable throwable) {
+    sendEvent(EXPECTED_EXCEPTION_CATEGORY, location + ":" + throwable.getClass().getName());
   }
 
   public void sendException(String throwableText, boolean isFatal) {
--- a/src/io/flutter/analytics/Analytics.java
+++ b/src/io/flutter/analytics/Analytics.java
@@ -32,7 +32,6 @@ import java.util.Map;
  */
 public class Analytics {
   public static final String GROUP_DISPLAY_ID = "Flutter Usage Statistics";
-  private static final String EXPECTED_EXCEPTION_CATEGORY = "expected-exception";
 
   private static final String analyticsUrl = "https://www.google-analytics.com/collect";
   private static final String applicationName = "Flutter IntelliJ Plugin";
@@ -114,7 +113,7 @@ public class Analytics {
   }
 
   public void sendExpectedException(String location, Throwable throwable) {
-    sendEvent(EXPECTED_EXCEPTION_CATEGORY, location + ":" + throwable.getClass().getName());
+    sendEvent("expected-exception", location + ":" + throwable.getClass().getName());
   }
 
   public void sendException(String throwableText, boolean isFatal) {
--- a/src/io/flutter/analytics/Analytics.java
+++ b/src/io/flutter/analytics/Analytics.java
@@ -112,6 +112,10 @@ public class Analytics {
     sendPayload("timing", args, null);
   }
 
+  public void sendCaughtException(String text) {
+    sendEvent("caught-exception", text);
+  }
+
   public void sendException(String throwableText, boolean isFatal) {
     String description = throwableText;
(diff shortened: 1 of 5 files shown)


src/io/flutter/jxbrowser/EmbeddedBrowser.java
```diff
@@ -71,7 +70,7 @@ private EmbeddedBrowser(Project project) {
       // Skip using a transparent background if an exception is thrown.
     } catch (Exception | Error ex) {
       LOG.info(ex);
-      FlutterInitializer.getAnalytics().sendException(StringUtil.getThrowableText(ex), false);
+      FlutterInitializer.getAnalytics().sendCaughtException("embedded-browser-setup: " + ex);
```
devoncarew, line 73: Here - and for the other jxbrowser related ones - you might instead send an event like:

`sendEvent(JxBrowserManager.ANALYTICS_CATEGORY, "setup:" + ex.getClass().getName());`

src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -263,7 +263,7 @@ private void logExceptionAndComplete(String message) {
 
   private void logExceptionAndComplete(Exception exception) {
     LOG.error(exception);
```
devoncarew, line 265: LOR.error() will also send an enelytics crash event (logging at the error level). I think we can just remove the separate call to `FlutterInitializer.getAnalytics().sendException()`. I'm not sure whether this exception - from `logExceptionAndComplete()` - is user facing or not, or something we should just recover from (like a missing file) or something that indicates that we have a logic error in our code.

stevemessick (approved):
It might be worth checking with an analytics expert about eventLabel. In things like this, the more structure you have, the better. Although, your labels look like they'd be searchable.

src/io/flutter/run/SdkFields.java
```diff
@@ -216,7 +215,7 @@ public GeneralCommandLine createFlutterSdkRunCommand(
       }
       catch (Exception e) {
         LOG.info(e);
-        FlutterInitializer.getAnalytics().sendException(StringUtil.getThrowableText(e), false);
+        FlutterInitializer.getAnalytics().sendCaughtException("devtools: " + e);
```
devoncarew, line 218: I assume we can either switch this to LOG.error() - and not send the exception separately - or keep it at LOG.info() and not send the exception.

It's a fairly big hammer to just catch `Exception`. Are there a few types that a thrown? If so, we should specialize what we do here based on the type of exception.

devoncarew (commented):
Thanks for re-visiting this! Several comments in-line.

src/io/flutter/jxbrowser/EmbeddedBrowser.java
helin24: I wanted to mark all of these as `caught-exception`s, but would that perhaps be better as a label? e.g. eventCategory: `JxBrowserManager.ANALYTICS_CATEGORY`, eventAction: `setup:<Some Exception class>`, eventLabel: `caught-exception` ?

src/io/flutter/run/daemon/DevToolsService.java
helin24: Oops, I fixed most of these `LOG.error` instances in a previous PR but missed this one. These are different errors that can occur during DevTools startup and would result in a more general user-facing message that DevTools failed to start, but I wanted the error details logged to a user's private file in case someone files a bug for any of the particular failure reasons.

src/io/flutter/run/SdkFields.java
helin24: It looks like there are at least two - `ExecutionException` and `TimeoutException` - but I also made this broad since missing an exception type would result in blocking running an app, and we don't need different user experiences based on the exception type.

src/io/flutter/analytics/Analytics.java
```diff
@@ -112,6 +112,10 @@ public void sendTiming(String category, String variable, long timeMillis) {
     sendPayload("timing", args, null);
   }
 
+  public void sendCaughtException(String text) {
```
devoncarew, line 115: `sendCaughtException()` ==> `sendExpectedException`?

We also might want to accept the location and exception as method params so we can combine the uniformly. Something like `(String location, Exception ex)`, and, `sendEvent("expected-exception", location + ":" + ex.getClass().getName());`.

src/io/flutter/jxbrowser/EmbeddedBrowser.java
devoncarew: If we think there will be expected exceptions in more than just the jxbrowser than having an event category for expected exceptions makes sense. The location / message might then be something like `jsbrowser-setup`?

src/io/flutter/run/SdkFields.java
devoncarew: OK, as long we you think we should handle the exception classes uniformly.

devoncarew (approved):
lgtm; adding some ideas for tweaking the method to send events for expected exceptions.

PR #5595, opened 2021-06-25
Dynamic icon preview of non-standard icon packages

### LINKED ISSUE #2334: Show Icons for Custom Icon Packages
Not sure if this is possible, but it would be amazing if custom Icons were displayed in the IDE in the same way the MaterialIcons show up. This means when you're auto-completing, you would see the icon on the right-hand side of the pop-up window. Once it's set, you'd see the icon in the Gutter next to the Breakpoint indicators.

For example, it would be really cool if something like the [`font_awesome_flutter`](https://pub.dartlang.org/packages/font_awesome_flutter) package I've published, or a custom Icon set created by end-users would display a preview of their own icons.

The MaterialIcons preview feature is really handy, so it'd be cool to see this functionality available for all icons!

stevemessick:
Screen Shot 2021-07-12 at 4 08 25 PM

Fixes #2334


mzdm:
As an icon package owner as well, will I need to do some modification in the code and/or letting users know in README how to set this up or will it be automatically?

stevemessick:
@mzdm Good question. I'll have instructions in the release notes next month, with screen shots.
> As an icon package owner as well, will I need to do some modification in the code and/or letting users know in README how to set this up or will it be automatically?

Each project has to configure the icon packs it wants to see previews for. There is a new text field in the preferences/settings page for the Flutter plugin for that. I may be able to eliminate that, and analyze the `pubspec.yaml` but ran out of time to fully investigate that. Adding instructions to your package would help get the word out.

stevemessick:
Here's a screen shot of the new text pane, showing all the icon packages that are known to work. I know of three that do not work, and warn users about them (`flutter_icons`, `flutter_vector_icons`, `material_design_icons_flutter`).

Screen Shot 2021-07-14 at 1 55 29 PM

mzdm:
@stevemessick Great, thanks! I'll be watching the future commits/release notes for instructions. 👍

jwren:
LGTM

stlr00:
Steel doesnt work with my custom icons from FlutterIcon

PR #5619, opened 2021-07-02
Update live template to not use double underscores

### LINKED ISSUE #4577: Don't use extra underscore when creating a State object for a private widget
When using the "stf" template to autogenerate a stateful widget, we typically see something like:

```
class MyWidget extends StatefulWidget {...}

class _MyWidgetState extends State<MyWidget> {...}
```

However, sometimes I want a widget class to be file-private, and in those cases I end up with names like the following:

```
class _MyWidget extends StatefulWidget {...}

class __MyWidgetState extends State<_MyWidget> {...}
```

Notice the double underscore prefixing the `State` subclass. I would still expect only a single underscore, but I end up with two underscores.

In my opinion this is a bug, but if it's not considered a bug then I would still recommend changing this behavior as a feature request.

spkersten:
When using the live template to create a stateful widget whose name starts with an underscore, don't let the associated state class name start with double underscores

Fixes #4577

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,3 +16,4 @@ Letticia Nicoli <letticia.nicoli@gmail.com>
 Alex Li <google@alexv525.com>
 Marcus Tomlinson <themarcustomlinson@gmail.com>
 Juyeong Lee <juyeonglee0413@gmail.com>
+Sander Kersten <spkersten@gmail.com>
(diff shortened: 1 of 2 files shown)


stevemessick (approved):
Thanks!

stevemessick:
If you want credit, please update the AUTHORS file and add it to this PR. I'll leave this open a couple days to give you time.

PR #5622, opened 2021-07-07
Skip Flutter tool windows for non-Flutter projects (WIP)

helin24:
From reading over https://github.com/flutter/flutter-intellij/pull/4281, it looks like these are the goals:
- Do not show any of the tool windows (outline, inspector, performance) for non-Flutter projects (defining this as having a Flutter module or being in a bazel workspace)
- Flutter outline should be available as soon as Flutter project is detected
- Flutter inspector and performance should be available when an app starts running, although they can also be open when app is not running. Once an app has run these tool windows should remain open
- Show windows if add to app module is created - Do we also want to remove tool windows if the only Flutter app module is deleted? (Can be done but maybe not worth the trouble)

### DIFF
--- a/src/io/flutter/FlutterInitializer.java
+++ b/src/io/flutter/FlutterInitializer.java
@@ -118,7 +118,6 @@ public class FlutterInitializer implements StartupActivity {
 
     if (hasFlutterModule || WorkspaceCache.getInstance(project).isBazel()) {
       initializeToolWindows(project);
-      toolWindowsInitialized = true;
     } else {
       project.getMessageBus().connect().subscribe(ProjectTopics.MODULES, new ModuleListener() {
         @Override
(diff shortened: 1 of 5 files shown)


src/io/flutter/FlutterInitializer.java
```diff
@@ -115,6 +116,19 @@ public void runActivity(@NotNull Project project) {
       }
     }
 
+    if (hasFlutterModule || WorkspaceCache.getInstance(project).isBazel()) {
+      initializeToolWindows(project);
+    } else {
+      project.getMessageBus().connect().subscribe(ProjectTopics.MODULES, new ModuleListener() {
+        @Override
+        public void moduleAdded(@NotNull Project project, @NotNull Module module) {
+          if (!toolWindowsInitialized) {
```
stevemessick, line 125: Check if the new module is a Flutter module.

stevemessick (approved):
Looks good after one small change.

Issue #5621, opened 2021-07-05
WriteCommandAction can execute after project disposal


Topbrains:
Thanks for the feedback! If your issue is related to the Flutter framework itself,
please open an issue at
[github.com/flutter/flutter](https://github.com/flutter/flutter/issues/new).

## Steps to Reproduce

_Please tell us what you were doing and what went wrong_

## Version info

Please paste the output of running `flutter doctor -v` here (available from the command
line or from `Tools > Flutter > Flutter Doctor`). It will provide the version of the
Flutter framework as well as of the IntelliJ plugin.

Topbrains:
Please file this bug report at https://github.com/flutter/flutter-intellij/issues/new.

---

## What happened

(please describe what you were doing when this exception occurred)

## Version information

IntelliJ IDEA `2021.1.1` • Flutter plugin `io.flutter 56.0.5` • Dart plugin `211.7233`

Error getting Flutter sdk information.

## Exception

null

```
java.lang.Throwable: Project Project(name=dedupe, containerState=DISPOSE_COMPLETED, componentStore=/Users/rafael/Dart Projects/Command Line/dedupe) (disposed) already disposed
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:161)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:196)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:182)
	at com.intellij.openapi.command.WriteCommandAction.doExecuteCommand(WriteCommandAction.java:297)
	at com.intellij.openapi.command.WriteCommandAction.performWriteCommandAction(WriteCommandAction.java:265)
	at com.intellij.openapi.command.WriteCommandAction.execute(WriteCommandAction.java:246)
	at io.flutter.editor.FlutterSaveActionsManager.lambda$performOrganizeThenFormat$0(FlutterSaveActionsManager.java:153)
	at com.intellij.openapi.application.TransactionGuardImpl.runWithWritingAllowed(TransactionGuardImpl.java:218)
	at com.intellij.openapi.application.TransactionGuardImpl.access$200(TransactionGuardImpl.java:21)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:200)
[...]

PR #5624, opened 2021-07-08
Check for project disposal right before organize executes

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/5621

It looks like since the `WriteCommandAction` is inside `invokeLater`, there's a chance the project could be disposed in between the earlier check and when `execute` is called.

### DIFF
--- a/src/io/flutter/editor/FlutterSaveActionsManager.java
+++ b/src/io/flutter/editor/FlutterSaveActionsManager.java
@@ -171,37 +171,39 @@ public class FlutterSaveActionsManager {
       return;
     }
 
-    if (myProject.isDisposed()) {
-      return;
-    }
-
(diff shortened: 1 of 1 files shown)


src/io/flutter/editor/FlutterSaveActionsManager.java
```diff
@@ -127,30 +127,32 @@ private void performOrganizeThenFormat(@NotNull Document document, @NotNull Virt
...
+
+            // Run this in an invoke later so that we don't exeucte the initial part of performFormat in a write action.
+            //noinspection CodeBlock2Expr
+            ApplicationManager.getApplication().invokeLater(() -> {
+              performFormat(document, file, true);
+            });
+          }
+        }.execute();
+      });
     }
```
stevemessick, line 156: Might as well fix up performFormat(), too.

stevemessick (approved):
Looks good after one more similar change.

PR #5626, opened 2021-07-08
Add analytics for null analyzer during add breakpoint

helin24:
This will log events about analyzer availability while setting breakpoints in bazel projects. This is only one specific case where we suspect there could be breakpoint setting issues, so perhaps there's a more general way to check if a breakpoint has been set successfully?

### DIFF
--- a/src/io/flutter/vmService/VmServiceWrapper.java
+++ b/src/io/flutter/vmService/VmServiceWrapper.java
@@ -367,6 +367,7 @@ public class VmServiceWrapper implements Disposable {
                 final Set<String> mappedCanonicalBreakpoints = new HashSet<>();
                 for (Breakpoint breakpoint : breakpoints) {
                   Object location = breakpoint.getLocation();
+                  // In JIT mode, locations will be unresolved at this time since files aren't compiled until they are used.
                   if (location instanceof  UnresolvedSourceLocation) {
                     final ScriptRef script = ((UnresolvedSourceLocation)location).getScript();
                     if (script != null && libraryUris.contains(script.getUri())) {
--- a/src/io/flutter/vmService/VmServiceWrapper.java
+++ b/src/io/flutter/vmService/VmServiceWrapper.java
@@ -356,8 +356,10 @@ public class VmServiceWrapper implements Disposable {
               @Override
               public void received(Isolate response) {
                 final Set<String> libraryUris = new HashSet<>();
+                final Set<String> fileNames = new HashSet<>();
                 for (LibraryRef library : response.getLibraries()) {
                   libraryUris.add(library.getUri());
+                  fileNames.add(library.getName());
                 }
 
                 final ElementList<Breakpoint> breakpoints = response.getBreakpoints();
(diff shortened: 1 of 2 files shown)


src/io/flutter/vmService/VmServiceWrapper.java
```diff
@@ -343,6 +345,52 @@ else if (errorResponses.size() > 0) {
...
+
+                final Analytics analytics = FlutterInitializer.getAnalytics();
+                final String category = "breakpoint";
+                boolean scriptFound = false;
+                for (Breakpoint breakpoint : breakpoints) {
+                  Object location = breakpoint.getLocation();
+                  if (location instanceof UnresolvedSourceLocation) {
+                    final ScriptRef script = ((UnresolvedSourceLocation)location).getScript();
+                    if (script != null) {
+                      if (libraryUris.contains(script.getUri())) {
```
jacob314, line 370: comment on this logic. Why do we expect some unresolved source locations to match the URIs of scripts we have found?

src/io/flutter/vmService/VmServiceWrapper.java
```diff
@@ -343,6 +345,52 @@ else if (errorResponses.size() > 0) {
...
+                        // Record to analytics that at least one breakpoint has been mapped to a library.
+                        analytics.sendEvent(category, "mapping-found");
```
jacob314, line 372: i'd rather track the count of the # of resolved and unresolved breakpoints than short circuit as soon as we found one successful breakpoint. Naively, something has gone wrong if any of the breakpoints are unresolved and do not match a source locations from the loaded libraries. Perhaps I'm missing something and there is a risk of false positives?

src/io/flutter/vmService/VmServiceWrapper.java
helin24: It seems that locations are only considered "resolved" after a breakpoint has been triggered (although I'm not sure exactly when this happens). At the time that mapping is set up during app startup, all of the breakpoints are unresolved.

src/io/flutter/vmService/VmServiceWrapper.java
helin24: Each set breakpoint generates multiple `Breakpoint` instances that show up here, so I'm not sure how to capture the total number of breakpoints from the user's viewpoint. e.g. for an external project's breakpoint, I see 2 `Breakpoint` instances - one's location has a `script` field with a URI that is associated with a package (this one can be considered "mapped"), the other has a `scriptUri` field instead that appears to be the location of a file on the device.

For a bazel project with a successful breakpoint, there are 3 `Breakpoint` instances - two are similar as the external project, plus one that has a `scriptUri` field that is the non-package location of the file.

So I'm not sure, maybe there's a better way to track a user view of how many breakpoints have been set earlier on in this process?

helin24:
@jacob314 is there a way to identify whether a file is part of a test or not? I just realized there's a problem here when you run an app but also have breakpoints in test files - those will not correspond to a file that's loaded into the library URIs, so we need a way to exclude those from the count of unmapped files.

jacob314:
lgtm

PR #5631, opened 2021-07-13
 Switch EAP to Beta

stevemessick:
I left the edit in place, commented out, since we will need it again when EAP restarts.


stevemessick:
This doesn't work. We still need to build from LATEST-EAP-SNAPSHOT.

PR #5638, opened 2021-07-15
Daemon crash

### LINKED ISSUE #5521: Would like different restart behavior in Flutter Plugin
recently we've noticed several `flutter daemon` crashes that look like repeated failures due to running out of file handles. Because the plugin will restart the daemon in a loop, this has been hammering our crash reporting backend, and likely not providing the user with any actionable feedback or surfacing the crash reports from the tool.

c.f. https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/run/daemon/DeviceDaemon.java#L51

The behavior that I would like is either for a prompt before a restart or at least some variety of backoff. We'd also like to display some notification to the user that the tool has crashed.

If need be the tool could surface the crash report information in a more structured format.

stevemessick:
Fixes #5521

After the first crash (down from 100) an error message is shown in the bottom-right corner of the screen:
Screen Shot 2021-07-15 at 3 26 22 PM

In case there is an intermittent issue, we increase the wait time allowed to start the daemon by 10 seconds.

After five crashes it changes to a modal dialog, displayed in the center of the IDE window. The Flutter plugin isn't usable anyway.
Screen Shot 2021-07-15 at 3 25 45 PM

### DIFF
--- a/src/io/flutter/run/daemon/DeviceDaemon.java
+++ b/src/io/flutter/run/daemon/DeviceDaemon.java
@@ -154,7 +154,7 @@ class DeviceDaemon {
         list = ImmutableList.of("--show-test-device", "daemon");
       }
       else {
-        list = ImmutableList.of("damon");
+        list = ImmutableList.of("daemon");
       }
       return new Command(sdk.getHomePath(), path, list, androidHome);
     }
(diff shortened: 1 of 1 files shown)


src/io/flutter/run/daemon/DeviceDaemon.java
```diff
@@ -411,4 +422,41 @@ public void processTerminated(int exitCode) {
   }
 
   private static final Logger LOG = Logger.getInstance(DeviceDaemon.class);
+
+  // If the daemon cannot be started, display a modal dialog with hopefully helpful
+  // instructions on how to fix the problem. This is q big problem; we really do
```
helin24, line 427: typo: a big problem

helin24 (approved):

PR #5643, opened 2021-07-16
Fix non-static uses of powermock

helin24:


helin24:
Closing in favor of https://github.com/flutter/flutter-intellij/pull/5644

PR #5644, opened 2021-07-16
Remove powermock from FlutterView tests and remove non-stack uses

helin24:


stevemessick (approved):
Wow, that was a lot more work than I expected. I'm gonna RSLGTM this one; the tests run and it looks like more-or-less the same conditions are getting tested.

PR #5656, opened 2021-07-20
Add a field for environment variables to the run config editor

### LINKED ISSUE #5292: Allow passing environment variables to desktop runs
## Request
Allow passing environment variables **to the `flutter run` command itself**.

## Why?
This can be useful for configuring your application using environment variables. While this does not make much sense on mobile platforms, for desktop applications this can be very useful.
The other reason is that currently a GTK incompatibility causes flutter to fail with text input, which however can be solved by using a different GTK version. To achieve this, the `LD_LIBRARY_PATH` could be set, but since the plugin does not allow this, there is no way.

## Version info
```
[✓] Flutter (Channel dev, 1.27.0-8.0.pre, on Linux, locale en_US.UTF-8)
    • Flutter version 1.27.0-8.0.pre at /data/flutter
    • Framework revision b7d4806243 (8 days ago), 2021-02-19 09:22:45 -0800
    • Engine revision 6993cb229b
    • Dart version 2.13.0 (build 2.13.0-30.0.dev)

    • Android SDK at /data/AndroidSDK
    • Platform android-30, build-tools 30.0.3
    • Java binary at: /usr/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_282-b08)

[✓] Chrome - develop for the web
    • CHROME_EXECUTABLE = /usr/bin/chromium

[✓] Linux toolchain - develop for Linux desktop
    • clang version 11.1.0
    • cmake version 3.19.6
    • ninja version 1.10.2
    • pkg-config version 1.7.3

      (or visit https://flutter.dev/docs/get-started/install/linux#android-setup for detailed instructions).

[✓] IntelliJ IDEA Ultimate Edition (version 2020.3)
    • IntelliJ at /home/janrupf/.local/share/JetBrains/Toolbox/apps/IDEA-U/ch-0/203.7148.57
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart

[✓] Connected device (2 available)
    • Linux (desktop) • linux  • linux-x64      • Linux
    • Chrome (web)    • chrome • web-javascript • Chromium 88.0.4324.182 Arch Linux

! Doctor found issues in 1 category.
```

stevemessick:
Fixes #5292

### DIFF
--- a/src/io/flutter/run/FlutterConfigurationEditorForm.java
+++ b/src/io/flutter/run/FlutterConfigurationEditorForm.java
@@ -5,15 +5,20 @@
  */
 package io.flutter.run;
 
+import com.google.common.collect.Lists;
+import com.intellij.execution.configuration.EnvironmentVariablesTextFieldWithBrowseButton;
 import com.intellij.openapi.options.ConfigurationException;
 import com.intellij.openapi.options.SettingsEditor;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.TextFieldWithBrowseButton;
 import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.text.StringUtil;
+import io.flutter.sdk.FlutterSdk;
 import org.jetbrains.annotations.NotNull;
 
 import javax.swing.*;
+import java.util.List;
+import java.util.Set;
 
 import static com.jetbrains.lang.dart.ide.runner.server.ui.DartCommandLineConfigurationEditorForm.initDartFileTextWithBrowse;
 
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/FlutterConfigurationEditorForm.java
```diff
@@ -24,9 +29,25 @@
...
+  private static final List<String> DESKTOP_PLATFORMS =
+    Lists.newArrayList("enable-linux-desktop", "enable-macos-desktop", "enable-windows-desktop", "enable-windows-uwp-desktop");
 
   public FlutterConfigurationEditorForm(final Project project) {
     initDartFileTextWithBrowse(project, myFileField);
+    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(project);
+    if (sdk != null) {
+      final Set<String> platforms = sdk.queryConfiguredPlatforms(true);
+      for (String platform : platforms) {
+        if (DESKTOP_PLATFORMS.contains(platform)) {
```
helin24, line 44: Why are environment variables disabled for desktop platforms?

helin24 (approved):

src/io/flutter/run/FlutterConfigurationEditorForm.java
stevemessick: They are only enabled for desktop platforms. They are disabled (below) if we don't detect at least one desktop.

PR #5660, opened 2021-07-21
Add menu item to Open in AppCode

### LINKED ISSUE #5446: Add "Open in AppCode"
We should add "Open in AppCode" parallel to "Open in Xcode", as suggested in #315. We'd need to be able to detect if AppCode is installed. We'd also need to add the proper project file for AppCode to recognize the project. Also see #4407.

stevemessick:
Add `Open iOS module in AppCode` parallel to the Xcode item.

There are some caveats:
- It requires Spotlight indexing to be enabled.
- The iOS module must have previously been opened and trusted by AppCode

Fixes #5446

### DIFF
--- a/src/io/flutter/actions/OpenInAppCodeAction.java
+++ b/src/io/flutter/actions/OpenInAppCodeAction.java
@@ -13,7 +13,6 @@ import com.intellij.execution.process.ProcessEvent;
 import com.intellij.execution.process.ProcessOutputTypes;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
-import com.intellij.openapi.actionSystem.CommonDataKeys;
 import com.intellij.openapi.actionSystem.Presentation;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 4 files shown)


src/io/flutter/actions/OpenInAppCodeAction.java
```diff
@@ -0,0 +1,117 @@
...
+
+  static {
+    initialize();
+  }
+
+  private static void initialize() {
+    try {
+      // AppCode is installed if this shell command produces any output:
+      // mdfind "kMDItemContentType == 'com.apple.application-bundle'" | grep AppCode.app
+      final GeneralCommandLine cmd = new GeneralCommandLine().withExePath("/bin/bash")
```
helin24, line 42: Does this run even if system is not mac?

helin24 (approved):

src/io/flutter/actions/OpenInAppCodeAction.java
stevemessick: Good catch. There was supposed to be a check for that; I'll add it.

PR #5671, opened 2021-07-25
handle the new event format for widget recount info

devoncarew:
- handle the new event format for widget recount info (`locations`)
-  remove references to `org.jetbrains.yaml` imports - this doesn't compile for me locally

@jacob314

### DIFF
--- a/src/io/flutter/editor/FlutterIconLineMarkerProvider.java
+++ b/src/io/flutter/editor/FlutterIconLineMarkerProvider.java
@@ -33,9 +33,6 @@ import io.flutter.FlutterBundle;
 import io.flutter.utils.IconPreviewGenerator;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.jetbrains.yaml.psi.YAMLCompoundValue;
-import org.jetbrains.yaml.psi.YAMLKeyValue;
-import org.jetbrains.yaml.psi.YamlRecursivePsiElementVisitor;
 
 import javax.swing.*;
 import java.util.*;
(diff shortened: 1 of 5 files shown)


stevemessick:
You can't remove `org.jetbrains.yaml`. See CONTRIBUTING.md, line 34.

devoncarew:
> You can't remove `org.jetbrains.yaml`. See CONTRIBUTING.md, line 34.

Ah, I see that entry. However, it doesn't look like the `YamlAssetMapVisitor` class is used anywhere, and that's the only thing that requires the org.jetbrains.yaml.* imports.

stevemessick:
It isn't used yet. I expect to remove a new text field from preferences by parsing the pubspec.yaml file to find icon package dependencies. I have other ideas for improving pubspec support that would require parsing yaml, too.

devoncarew:
> It isn't used yet. I expect to remove a new text field from preferences by parsing the pubspec.yaml file to find icon package dependencies. I have other ideas for improving pubspec support that would require parsing yaml, too.

devoncarew:
@stevemessick - I reverted the FlutterIconLineMarkerProvider.java change. Assuming this lgtm (after a review), can we get this into the upcoming release? It'll help us make changes to the widget rebuild count event format going forward.

stevemessick:
We're using the snakeyaml library that is part of the platform SDK to do yaml parsing. I thought we were just doing regex matches. Dart has dependencies on both snakeyaml and intellij.yaml.

I'm going to need to clean up the yaml parsing. It should be consistent. I was about to approve your changes when I saw you had reverted them. If you don't want to make another commit, that's fine, but we won't be needing intellij.yaml in the future, I think.

I have not made the release branch yet, so these changes will be in it.

stevemessick (approved):

devoncarew:
> I'm going to need to clean up the yaml parsing. It should be consistent. I was about to approve your changes when I saw you had reverted them. If you don't want to make another commit, that's fine, but we won't be needing intellij.yaml in the future, I think.

OK, sounds good - sounds like we have some room for rationalizing out yaml parsing here

> I have not made the release branch yet, so these changes will be in it.

Great! Will land now.

PR #5676, opened 2021-07-26
Update build for M59

stevemessick:


helin24 (approved):

helin24:
Does plugin.xml also need to be updated?

stevemessick:
> Does plugin.xml also need to be updated?

Yes, the checker bot told me that, too, although not as clearly as you. Done! I'll just let the checks run before committing.

stevemessick:
Unit tests all passed on kokoro but it apparently didn't tell github.

PR #5686, opened 2021-07-30
Allow slow op for 2021.2

stevemessick:
This avoids an error message in 2021.2. The refactoring I'm working on will eliminate the need for this hack.

TBR @helin24


helin24:
LGTM

PR #5704, opened 2021-08-06
Refactor icon preview processing to be incremental

stevemessick:
Prior to this PR, icon preview analysis was performed as a batch operation, which could stall the EDT. This was not a common problem as it ran fast enough when there were only one or two icon packages to analyze. This change converts it to an incremental background process that automatically restarts when a read action gets terminated.

The code analysis hasn't changed, but the order in which operations occur has, in some cases. For example, the field `filesWithNoClasses` is used as the inverse of examining the count of classes in a file. The main difference is the addition of a lot of thread management code.


src/io/flutter/font/FontPreviewProcessor.java
```diff
@@ -77,119 +94,304 @@ public void projectClosed(@NotNull Project project) {
...
+          final WorkItem item = WORK_ITEMS.get(path);
+          if (!processNextItem(project, item)) {
+            if (item.filesWithNoClasses.isEmpty()) {
+              // Finished, clean up and exit.
+              synchronized (this) {
+                if (item == WORK_ITEMS.get(path)) {
+                  WORK_ITEMS.remove(path);
+                }
+                isRunning.set(false);
+                if (System.currentTimeMillis() - startTime > 1000) {
```
helin24, line 128: Is this time arbitrary? Would it be better to always do this?

helin24 (approved):

src/io/flutter/font/FontPreviewProcessor.java
stevemessick: This is a kind of heavyweight action that can cause screen flicker. I don't think we want to do it unless there is a reason. But in my testing, with only a small amount of analysis there is no need to do it. And in my case, there was only one icon that was not displayed, and even that was displayed sometimes. Thanks for the review!

PR #5646, opened 2021-07-16
Switch to JUnit 5

stevemessick:
Fixes #4459

PR #5709, opened 2021-08-12
Update build script for AS canary

stevemessick:
With these changes we can once again build the plugin for Android Studio canary. One big change was switching to a 1.x version of the IntelliJ Gradle plugin. That had a number of breaking changes, described in the migration guide section of the [documentation](https://lp.jetbrains.com/gradle-intellij-plugin/).

I also added another field to the build specs. This allows specifying which version of IntelliJ to use when compiling forms. Hopefully, that will allow us to avoid using the nightly snapshot when the next EAP comes out.

Due to converting some classes to be `final` it was not possible to update the code in `FlutterNewProjectAction` that trimmed the project selection list to only show Flutter. I just deleted all that code.


stevemessick:
One unit test fails. It has nothing to do with this PR, so I want to ignore it. The unit tests need more work, anyway. See #5646.

jwren:
LGTM

PR #5770, opened 2021-09-21
Allow tests in any marked test dir and optionally in sources

### LINKED ISSUE #5542: Android Studio shouldn't show "Build iOS" on Windows
The Build/Flutter menu shows an option to "Build iOS" on windows, but we can only run IOS builds on a Mac.

stevemessick:
Fixes #5542

This checks if a file is in a directory marked as Test Sources Root, even if not named `test`, and adds the gutter icons to run individual tests if it is. It also adds a new setting that will allow files ending with `_test.dart` to be treated as test files if they are in a directory marked as Sources Root, such as `lib`. The new setting is opt-in.

I noticed that many settings were not included in the searchable options, so I added them. I also made some consistency changes, and moved all the embedded strings to the bundle properties file.


stevemessick:
@jwren

jwren (approved):

PR #5777, opened 2021-09-27
Make console stack traces expandable

### LINKED ISSUE #5445: Make stack traces in the console expandable
Currently by default we completely truncate the stack traces for errors after the first error.  Truncating the stack trace for the errors is useful to avoid being overwhelmed by cascading errors but makes it difficult if you need to see a stack trace for an error that wasn't the first.

We work around this by having a setting to display all stack frames but few users will realize the setting exists so if we can solve the problem without requiring users to edit a setting that would be great.
https://github.com/flutter/flutter-intellij/pull/5443/files
If the expandable blocks support in the default IntelliJ console is not sufficient we should probably wait and consider exposing the console from Dart DevTools with JXBrowser to support the functionality.

stevemessick:
Fixes #5445 

If this is first in the list of foldings a stack trace collapses to a single line (the first line in the screen shot). If it is last, another folding may interfere. With it second in the list, a stack trace looks like this:
Screen Shot 2021-09-27 at 2 30 52 PM

That preserves the spirit of the structured stack traces option, I think, but we may need to iterate. If the option to include all stack traces is selected then no folding is done.

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -384,8 +384,9 @@
 
     <applicationService serviceImplementation="io.flutter.jxbrowser.EmbeddedBrowserEngine" overrides="false" />
     <applicationService serviceImplementation="io.flutter.font.FontPreviewProcessor"/>
-    <console.folding implementation="io.flutter.console.FlutterConsoleFolding"/>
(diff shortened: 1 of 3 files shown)


stevemessick:
@jwren

jwren (approved):

PR #5783, opened 2021-09-29
Update jxbrowser to 7.19

stevemessick:
@jwren


stevemessick:
Had to use a bit of cargo-cult magic to push this from my slow network:
> git config ssh.postBuffer 524288000

jwren (approved):

Issue #5796, opened 2021-10-04
Color picker functionality in the IntelliJ gutter is broken by making small code edits


jacob314:
Editing a line after it already has a color values makes the color icon stop being editable for some reason. For example, adding a const where one was missing reproduces this. Wrapping a color value with parens also reproduces the issue.
If you delete the line with the color value and paste it again, then the color picker works correctly so the issue has something to do with updates made when an existing icon already shows in the gutter.

Issue #5780, opened 2021-09-29
Color picker does not work


fctorial:
## Steps to Reproduce

The color picker in the gutter does not work:

https://1drv.ms/u/s!AsFZusAnvKobgSdqPimyQvn4jzbt?e=XZwpKy

stevemessick:
I can't open that link. Please provide details of the problem here. Unless it is due to this issue noted in the release notes:
>Caveat: the color picker appears for all color icons but its selection is ignored if the color is a named Material or Cupertino color.

If that's the problem, you can close this issue because we do not plan to fix it.

fctorial:
> > Caveat: the color picker appears for all color icons but its selection is ignored if the color is a named Material or Cupertino color.

Yes this is it. Apparently this is the source of this statement:

https://groups.google.com/g/flutter-announce/c/V335xbsPWUs/m/14LSp05kAQAJ

Can you describe why this issue is "wont fix"? It is a usability issue and it appears to be within the limits of intellij sdk. (is it?)

jacob314:
I tried out this functionality and it is great when it works but there a few bugs.
# Problems with the current behavior:
1. Editing a line after it already has a color values makes the  color icon stop being editable for some reason. For example, adding a `const` where one was missing reproduces this.
2. Color values like `Colors.red` display a color picker but the value is not saved which is  confusing and makes it seem like our code editing functionality is broken. I think users would prefer to save the color value even if it overrides a nice material color constant with a hex value. We can tackle fixing up color values to reuse existing color constants with a separate lint and quick fix. I would suggest we emit color values as `Color(0xFFAF9299)`. As a bonus we could emit the color as a Colors.foo constant if the exact value matched one of the material colors but that isn't crucial for an MVP. We could also consider adding a lint and quick fix that detects any case where a user is specifying colors as raw values when there are appropriate constants. https://github.com/dart-lang/linter/issues/2983

stevemessick:
When working with named color constants, code completion works great. Making the color picker work equivalently is a lot of work, which was why I said we don't plan to fix it. We'll leave it open, though.

The reason the color picker is displayed for named colors is to provide the preview, and I couldn't find a way to disable the "picker" part of the preview. If IntelliJ adds that, then we can stop displaying the input selector. I think rewriting a named value to a hex value would be a surprising and unwelcome UX, though.

jacob314:
[...]

PR #5816, opened 2021-10-12
Fix issues with color editing.

jacob314:
The color picker now auto-completes to color constants and works
even if the previous version of a color is a color constant.

*List which issues are fixed by this PR. You must list at least one issue.*

https://github.com/flutter/flutter-intellij/issues/5796
https://github.com/flutter/flutter-intellij/issues/5780

https://user-images.githubusercontent.com/1226812/137044657-d3e8091f-122b-4968-91c3-89ad5d2d9984.mp4

https://user-images.githubusercontent.com/1226812/137044679-d8cf19eb-de3e-4a75-bab3-66ce15e5d89b.mp4

### DIFF
--- a/src/io/flutter/dart/DartPsiUtil.java
+++ b/src/io/flutter/dart/DartPsiUtil.java
@@ -6,6 +6,7 @@
 package io.flutter.dart;
 
 import com.intellij.psi.PsiElement;
+import com.intellij.psi.tree.IElementType;
 import com.jetbrains.lang.dart.DartTokenTypes;
 import com.jetbrains.lang.dart.psi.DartArgumentList;
 import com.jetbrains.lang.dart.psi.DartArguments;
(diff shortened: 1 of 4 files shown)


jacob314:
@DanTup you might want to check out these gifs for some inspiration for future support for VSCode.
Some limitations. We currently propose names that might not have the proper const requirements for the call site. E.g. we might propose Colors.red[400] which can't be used in a const call site. Expectation is users would then edit their code to remove const as needed. This might be a bad assumption. We might also insert Colors.red even though the material Colors class is not in scope. The expectation in this case is users can apply the quickfix to add the name into scope. @bwilkerson if we supported this directly in the analysis server, it would be nice to be smarter about these cases applying the appropriate fixes automatically.

DanTup:
@jacob314 thanks! I had a quick look at the APIs in VS Code. We can't really influence the UI much (we just provide data), but the functionality looks reasonable. It shows up in-line rather than in the gutter, but here's how it looks:

Oct-13-2021 14-28-09

There are two parts to the API:

1. Get all the colours and their hex values for the current document (so the inline decorations can be rendered)
2. Given a colors hex value, return an array of the ways in which it can be presented (`Color.red`, `#FF0000`, etc.)

The user can toggle between the representations by clicking on the text. We can also supply "additional edits" with each color that are allowed to modify other parts of the document - that means if the user chose the `Colors.red` version, we could automatically add/update an `import` if it's not already there.

jacob314:
@DanTup that VSCode UI looks fine. We don't have the ability to polish the UI in IntelliJ either so what we are showing is just the IntelliJ default for this behavior as well.

PR #5824, opened 2021-10-15
Remove dependencies on built-in Flutter icons

stevemessick:


stevemessick:
Even though there is one new failing test, I'm going to merge this as-is so it gets into the next dev channel build. I know what's wrong with the test, but don't have time to fix it today. I also know how to repair the other failing tests and plan to fix them next week.

PR #5868, opened 2021-11-23
Build for AS canary

stevemessick:
@jwren 

The `flutter-intellij-community.iml` file is needed to work on the plugin with Android Studio sources.

The studio-specific source trees for `io.flutter.project` and `io.flutter.module` were obsolete. They had to be deleted due to API changes.

The versions all build correctly. The canary version does not work because of a class-not-found error (but it does work in the debugger). The class (`org.jetbrains.idea.maven.wizards.MavenWizardBundle`) is defined in the `flutter-studio` module but it is not found by the Jetbrains-specific NPW code. I'm not sure how to fix that yet; it looks like we'd need to insert something onto the classpath of a module that we don't control. Maybe @alexander-doroshko has some suggestion how to do that? As far as I can tell, the missing class is distributed with IntelliJ but not Android Studio. Normally, Android Studio would not need it. It is used by the IntelliJ NPW framework that the Flutter plugin uses for project creation.


alexander-doroshko:
The behavior of the class loaders has become stricter. Each plugin has its own class loader. Class loaders have strict hierarchy. ClassNotFound may happen even if the class is available in some jar but not available for the class loader in use.

It works differently when running from sources (a single class loader for all the classes), ClassNotFound usually doesn't happen in this case.

To analyze ClassNotFound problem, first of all we need to find the JAR file in the IDE installation in which the corresponding class is located. 

Then we need to see the whole stack trace and analyze in which JAR each class is located. As I understand, classes from Maven plugin will never be available to the Flutter plugin's class loader because Flutter plugin doesn't depend on Maven plugin. 'Legal' use cases should never lead to such situation.

Please email me the stack trace, describe how IDE is started in this experiment and any other info that you think may be useful to analyze the problem. The problem might be not related to class loaders at all.

stevemessick:
Thanks for the suggestion @alexander-doroshko. I found the jar file; it is part of the maven plugin, which is not distributed with Android Studio. As an experiment, I copied the maven plugin from IntelliJ, dropped it into Android Studio, and now the NPW works. So, I'm still not sure how to fix the problem, but I understand it a bit better.

stevemessick:
This is ready for review. The AS canary build is not working yet. I'm consulting with the Android Studio team on potential resolutions for that, so no ETA for canary support yet.

jwren (approved):

PR #5874, opened 2021-12-01
Replace 'flutter pub run' with 'dart devtools'

stevemessick:
@kenzieschmoll I missed the requirement for this to change in the AS plugin, until yesterday.


flutter-idea/src/io/flutter/sdk/FlutterSdkVersion.java
```diff
@@ -59,7 +59,7 @@
   @NotNull
   private static final FlutterSdkVersion MIN_PASS_DEVTOOLS_SDK = new FlutterSdkVersion("1.26.0-11.0.pre");
   @NotNull
-  private static final FlutterSdkVersion MIN_OPTIONAL_PASS_DEVTOOLS_SDK = new FlutterSdkVersion("2.7.0-3.0.pre");
+  private static final FlutterSdkVersion MIN_OPTIONAL_PASS_DEVTOOLS_SDK = new FlutterSdkVersion("2.6.0-12.0.pre.854");
```
kenzieschmoll, line 62: what version of DevTools is included with the flutter version? In DevTools, we will show a warning to users to stop launching from pub if they are on Flutter 2.8.0 or Dart 2.15 or greater. The reasoning behind this is we didn't want to downgrade users by telling them to stop launching from pub. My concern with using Flutter 2.6.0-12.0.pre.854 is that we may downgrade a user's devtools version by doing this. Maybe we should gate instead on flutter 2.8.0, which we know has DevTools 2.8.0 included.

kenzieschmoll (commented):
Generally, I was under the impression the Flutter IntelliJ / AS deferred to the flutter daemon to run DevTools. Is this not the case?

flutter-idea/src/io/flutter/sdk/FlutterSdkVersion.java
stevemessick: `dart devtools` about box shows 2.8 for me. I don't know much about devtools -- all I did was update my master flutter branch to latest and got that.

stevemessick:
> I was under the impression the Flutter IntelliJ / AS deferred to the flutter daemon to run DevTools

It uses `flutter pub run devtools`, which I thought you wanted to have removed. If you don't care about that, I can close this PR w/o merging.

kenzieschmoll:
We do want to get away from pub. I recall @helin24 saying we launched from the flutter daemon. I tagged you in a comment on that doc. 

My comment about which version to gate on still stands, but we will want to land this PR.

kenzieschmoll:
Why was this PR closed?

PR #5907, opened 2022-01-14
[WIP] Remove non-Flutter projects from NPW

stevemessick:
This removes all project types except Flutter from the new project wizard in Android Studio. One of the concerns the AS group has is that the NPW we are using exposes things that don't make sense for Android.

Old
Screen Shot 2022-01-14 at 2 53 40 PM

New
Screen Shot 2022-01-14 at 3 18 42 PM

This uses the `fest` library, which is normally used for functional testing. I'm not sure we want to distribute that as part of the plugin, so I'd like to get feedback on whether this seems like a PR we want to keep.

Until recently, I was trimming the project list when the NPW was constructed. Unfortunately, AS locked down the code I was re-using so I had to remove that.

WIP until the build system is modified to work with `fest`.

### DIFF
--- a/flutter-idea/src/io/flutter/module/FlutterModuleBuilder.java
+++ b/flutter-idea/src/io/flutter/module/FlutterModuleBuilder.java
@@ -5,17 +5,27 @@
  */
 package io.flutter.module;
 
+import static java.util.Arrays.asList;
+import static org.fest.swing.core.matcher.DialogMatcher.withTitle;
+import static org.fest.swing.finder.WindowFinder.findDialog;
+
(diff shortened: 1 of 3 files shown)


stevemessick:
This PR will be deleted. Due to scheduling constraints it may be a while before anyone gets back to this topic. We do not want to distribute a test library in production code.

What we need to do is use this code to find the project list in the debugger. Then trace the widget tree up to the frame and simply write code to traverse from the frame back to the list. We could even go back to the UI we had where all Flutter project types were shown in the project list, rather than being selected from a drop-down list on the second page.

PR #5912, opened 2022-01-24
Add analytics to survey notifications

### LINKED ISSUE #5772: Log user interactions with the survey prompt
Data from Q3 survey suggests that a lot more users responded to the survey from VS Code than IntelliJ, which didn't match the distribution of Flutter's userbase between the two IDEs. We'd like to better understand where the difference came from and fix any bugs that might have prevented users on IntelliJ from seeing the prompt. The preliminary idea is to log events such as showing the prompt, clicking on the survey link, and dismissing the prompt. We should do the same on the VS Code side to enable comparison. 

Cc: @jacob314 @jayoung-lee

stevemessick:
Fixes #5772


pq (approved):
👍

PR #5947, opened 2022-02-08
Use VM for mapping breakpoint files during app run

helin24:
This change updates the VM service files to include the new methods that find resolved URIs and package URIs for files, so that we can use the VM for breakpoint mapping instead of the mapping method in `FlutterPositionMapper` (which does a combination of analyzer and other things). 

Only changes in the first file (`VmServiceWrapper`) are significant; the rest are generated file changes.

~I left some print statements in here for now as I haven't found a way to fix my Java breakpoints.~


flutter-idea/src/io/flutter/vmService/VmServiceWrapper.java
```diff
@@ -433,12 +441,16 @@ public void received(Isolate response) {
                   }
                 }
 
+                System.out.println(String.format("Unmapped count is %s", finalDifference.size()));
```
jacob314, line 444: remove System.out.print here and elsewhere

flutter-idea/src/io/flutter/vmService/VmServiceWrapper.java
```diff
@@ -473,45 +485,105 @@ public void addBreakpoint(@NotNull final String isolateId,
...
-      for (String uri : scriptUris) {
-        myVmService.addBreakpointWithScriptUri(isolateId, uri, line, new AddBreakpointWithScriptUriConsumer() {
-          @Override
-          public void received(Breakpoint response) {
-            breakpointResponses.add(response);
-            breakpointNumbersToCanonicalMap.put(response.getBreakpointNumber(), canonicalBreakpoint);
+      myVmService.lookupPackageUris(isolateId, resolvedUriList, new UriListConsumer() {
+        @Override
+        public void received(UriList response) {
+          final List<String> uris = response.getUris();
```
jacob314, line 501: need to make sure you aren't disposed by the time this callback is triggered.

flutter-idea/src/io/flutter/vmService/VmServiceWrapper.java
```diff
@@ -473,45 +485,105 @@ public void addBreakpoint(@NotNull final String isolateId,
...
+  private String getResolvedUri(XSourcePosition position) {
+    final String url = position.getFile().getUrl();
+
+    if (WorkspaceCache.getInstance(myDebugProcess.getSession().getProject()).isBazel()) {
+      final String root = "google3";
+
+      // Look for a generated file path.
+      final String genFilePattern = root + "/blaze-.*?/(.*)";
+      final Pattern pattern = Pattern.compile(genFilePattern);
+      final Matcher matcher = pattern.matcher(url);
```
jacob314, line 571: I think you can make this more robust by using the bazel package root that should be
someDirectory/google3

that way something wacky like
someDirectory/google3/someDir2/google3
won't false positive.

flutter-idea/src/io/flutter/vmService/VmServiceWrapper.java
```diff
@@ -473,45 +485,105 @@ public void addBreakpoint(@NotNull final String isolateId,
...
+      if (matcher.find()) {
+        final String path = matcher.group(1);
+        return root + ":///" + path;
+      }
+
+      // Look for root.
+      final int rootIdx = url.indexOf(root);
+      if (rootIdx > 0) {
+        return root + ":///" + url.substring(rootIdx + root.length() + 1);
```
jacob314, line 580: nit: just make this "google3:///" as the fact that this google3 uri scheme happens to make the google3 root directory is somewhat arbitrary.

jacob314 (approved):
Fyi @DanTup, it would be nice to add the same bazel project root directory to google3:/// conversion step for VSCode. That will make breakpoints work in google3.

DanTup:
> Fyi @DanTup, it would be nice to add the same bazel project root directory to google3:/// conversion step for VSCode. That will make breakpoints work in google3.

Is this not handled in the VM Service with the new APIs? If not, could you describe exactly what would need doing? I reviewed the `google3` references in this PR, but I'm not sure I entirely understand what's going on. Is it just another special mapping like how we handle `org-dartlang-sdk`?

jacob314:
> > Fyi @DanTup, it would be nice to add the same bazel project root directory to google3:/// conversion step for VSCode. That will make breakpoints work in google3.
> 
> Is this not handled in the VM Service with the new APIs? If not, could you describe exactly what would need doing? I reviewed the `google3` references in this PR, but I'm not sure I entirely understand what's going on. Is it just another special mapping like how we handle `org-dartlang-sdk`?

There is a really unfortunate but hard to avoid edge case where everything in google3 is surfaced in terms of google3:/// instead of true absolute file paths
Basically what has to happen is for all APIs, clients need to replace
/some/path/your/machine/google3/foo/bar/baz
/some/path/your/machine/google3/blaze-*/foo/bar/baz
with
google3:///foo/bar/baz

DanTup:
> Basically what has to happen is for all APIs, clients need to replace
> /some/path/your/machine/google3/foo/bar/baz
> /some/path/your/machine/google3/blaze-*/foo/bar/baz
with
> google3:///foo/bar/baz

How do we know what `/some/path/your/machine/google3` is, and how do we map the other way? (eg. given there are multiple paths/wildcards in your examples above, what does `google3:///` map back to on the way out?)

flutter-idea/src/io/flutter/vmService/VmServiceWrapper.java
helin24: Check that project is not disposed?

jacob314:
> > Basically what has to happen is for all APIs, clients need to replace
> > /some/path/your/machine/google3/foo/bar/baz
> > /some/path/your/machine/google3/blaze-*/foo/bar/baz
> > with
> > google3:///foo/bar/baz
> 
> How do we know what `/some/path/your/machine/google3` is, and how do we map the other way? (eg. given there are multiple paths/wildcards in your examples above, what does `google3:///` map back to on the way out?)
You can determine the /some/path/your/machine/google3 based on the flutter.json file detected for the bazel project.
Good question about the other way out. Now I'm questioning a little bit whether we are handling that case as well as we need to in IntelliJ. I'll open an internal thread and get back to you. I'm not confident we are doing it right today and wonder whether DDS can add an API to help.

PR #5950, opened 2022-02-09
Try SIGINT before SIGKILL for processes

helin24:
We were sending SIGKILL for apps, which is preventing other work from being done that relies on trapping a signal.

CC @chingjun


jacob314 (commented):
Is there a reason not to change the default for all project types? Any time we can stay consistent we reduce the surface area we have to test. If we do change for bazel only, we should note that the change was only for remote workflows not really all bazel workflows.

jacob314 (approved):
lgtm

PR #5972, opened 2022-02-14
Change run config to open one instance of workbench

helin24:
Our run configuration was opening two instances of workbench, and only one of them would connect debug endpoints.


stevemessick (approved):
Thanks!

PR #5974, opened 2022-02-14
Remove old bazel test file mapping code

helin24:
We're now using the VM service instead for file mapping during debugging, so this bazel test code is no longer needed for test breakpoints to work.


jacob314 (approved):
lgtm

PR #5976, opened 2022-02-16
Use time zone in survey calculation

### LINKED ISSUE #5866: Survey announcement is shown when the survey is not supposed to be available
The Q4 survey is supposed to be available from Nov 22, 9:00 AM PST (https://docs.flutter.dev/f/flutter-survey-metadata.json). 

However, the earliest a dev started the Q4 survey from IntelliJ was Nov 21, 7:41 AM PST, based on the Qualtrics record. It's over 24 hours before the planned launch time.

Since the Q4 survey was published on the 19th, the devs should have been able to see the questions and answer them. But this needs to be fixed for the future surveys.

stevemessick:
The javadoc says the original code uses the time zone, but it actually does not. This version does. Everyone in Asia was seeing the survey a day ahead of schedule. There may still be cases where someone testing date-sensitive code will see it before they should, but hopefully this will be a very small number.

Fixes #5866

### DIFF
--- a/flutter-idea/src/io/flutter/survey/FlutterSurvey.java
+++ b/flutter-idea/src/io/flutter/survey/FlutterSurvey.java
@@ -9,12 +9,13 @@ import com.google.gson.JsonObject;
 import org.jetbrains.annotations.NotNull;
 
 import java.time.LocalDate;
+import java.time.ZonedDateTime;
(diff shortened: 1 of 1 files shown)


helin24 (approved):

jayoung-lee:
Thank you @stevemessick !!

PR #5981, opened 2022-02-17
Remove unnecessary(?) arg from gradle properties, update contributing

helin24:
The jvm arg was throwing an error on IntelliJ startup with our plugin code.

Added a couple additional instructions while building the plugin and setting up gpg


stevemessick (approved):
Thanks!

PR #5985, opened 2022-02-18
Add more analytics

stevemessick:
Needs testing, and tests. Some basic work is still not done, too.


flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -0,0 +1,520 @@
...
+  FlutterAnalysisServerListener(@NotNull Project project, DartAnalysisServerService analysisServer) {
+    this.project = project;
+    this.pathToErrors = new HashMap<>();
+    this.pathToErrorTimestamps = new HashMap<>();
+    this.pathToHighlightTimestamps = new HashMap<>();
+    this.pathToOutlineTimestamps = new HashMap<>();
+    this.requestToDetails = new HashMap<>();
+    try {
+      LookupManager.getInstance(project).addPropertyChangeListener(this::onPropertyChange);
+    } catch (NullPointerException e) {
```
jacob314, line 129: what object is null? Prefer to avoid NPE rather than ever catching them.

stevemessick: `LookupManager.getInstance(project)` could be null. I think that only happens during unit tests and it would be better to explicitly check if we're running tests. I'll change it.

stevemessick:
This is ready for review.

Note that due to changes in `Testing` there will be compilation errors when using any IntelliJ older than EAP. The Dart plugin API that is being used is apparently only available in the EAP builds.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -0,0 +1,521 @@
...
+                                 List<CompletionSuggestion> completionSuggestions,
+                                 List<IncludedSuggestionSet> includedSuggestionSets,
+                                 List<String> includedElementKinds,
+                                 List<IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags,
+                                 boolean isLast,
+                                 String libraryFilePathSD) {
+  }
+
+  @Override
+  public void computedErrors(String path, List<AnalysisError> list) {
```
jacob314, line 183: is this getting called for every file in a project?
If it is called for more than each root path, I'm concerned we are logging too much.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -0,0 +1,521 @@
...
+
+  void logE2ECompletionSuccessMS(long e2eCompletionMS) {
+    FlutterInitializer.getAnalytics().sendTiming(E2E_IJ_COMPLETION_TIME, SUCCESS, e2eCompletionMS); // test: logE2ECompletionSuccessMS()
+  }
+
+  void logE2ECompletionErrorMS(long e2eCompletionMS) {
+    FlutterInitializer.getAnalytics().sendTiming(E2E_IJ_COMPLETION_TIME, FAILURE, e2eCompletionMS); // test: logE2ECompletionErrorMS()
+  }
+
+  private void logAnalysisError(@Nullable AnalysisError error) {
```
jacob314, line 383: I'm concerned this is logging too much. If you have 10,000 errors, we are still trying to log 100 events.
What are we trying to achieve by logging the error codes?

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -0,0 +1,521 @@
...
+                                 List<IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags,
+                                 boolean isLast,
+                                 String libraryFilePathSD) {
+  }
+
+  @Override
+  public void computedErrors(String path, List<AnalysisError> list) {
+    assert list != null;
+    list.forEach(this::logAnalysisError);
+    pathToErrors.put(path, list);
```
jacob314, line 186: what happens if you restart the analysis server? Do we get a new event with new errors.
Wish we had a merged Dart and Flutter plugin so we didn't have to duplicate this logic that the analysis errors window already handles.

stevemessick:
@jacob314 My main concern is that we are doing a lot of work that 90% of our users won't see for a year. The new Dart API is (or was -- I have not checked this week) only present in the EAP version of the Dart plugin.

BTW you have to click the Approve button now. LGTM in comments is no longer sufficient to merge. No hurry, I have not yet looked into your questions.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
stevemessick: Not every file, but all Dart files plus some unexpected files. All three `AndroidManifest.xml` files were checked plus `pubspec.yaml`.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -0,0 +1,525 @@
...
+                                 List<IncludedSuggestionRelevanceTag> includedSuggestionRelevanceTags,
+                                 boolean isLast,
+                                 String libraryFilePathSD) {
+  }
+
+  @Override
+  public void computedErrors(String path, List<AnalysisError> list) {
+    assert list != null;
+    List<AnalysisError> existing = pathToErrors.get(path);
+    if (existing != null && existing.equals(list)) {
```
stevemessick, line 186: @jwren Do you think this test makes sense? And will it be expensive?

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
stevemessick: I added that test and asked for an opinion from jwren.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
jacob314: We need to log far fewer events. Logging analytics needs to not be the reason why users have poor performance or complain our tools use too much of their bandwidth. I would suggest changing all these analytics so a single analytic event or a couple of events are logged summarizing the state of all errors reported rather than ever emitting events proportional to the number of errors or files.

stevemessick: I think this will be straight-forward to change. I'll drop the call to `logAnalysisError` for each error and clean up unused elements. We already log the summary info in `serverStatus`.

stevemessick: Done! PTAL

jacob314 (approved):
lgtm

stevemessick:
@jacob314 @jwren After looking at the log file and seeing just how much stuff was getting sent I decided to throttle everything to one transmission per minute. The cumulative error counts are sent when the project is closed, and every two hours, for back-end percentile analysis. Both of those intervals can be adjusted.

.idea/runConfigurations/flutter_idea__runIde_.xml
```diff
@@ -13,7 +13,7 @@
           <option value="runIde" />
         </list>
       </option>
-      <option name="vmOptions" />
```
jacob314, line 16: did you mean to include this change in the CL?

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -122,7 +137,14 @@ public void selectionChanged(@NotNull FileEditorManagerEvent event) {
       public void fileClosed(@NotNull final FileEditorManager source, @NotNull final VirtualFile file) {
       }
     };
-    project.getMessageBus().connect().subscribe(FileEditorManagerListener.FILE_EDITOR_MANAGER, fileEditorManagerListener);
+    messageBusConnection.subscribe(FileEditorManagerListener.FILE_EDITOR_MANAGER, fileEditorManagerListener);
+    messageBusConnection.subscribe(ProjectManager.TOPIC, new ProjectManagerListener() {
+      public void projectClosing(@NotNull Project project) {
+        messageBusConnection.disconnect(); // Do this first tvoid memory leaks when switching pojects.
```
jacob314, line 143: tvoid -> to avoid

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -342,51 +367,85 @@ public void serverStatus(AnalysisStatus analysisStatus, PubStatus pubStatus) {
...
+      if (warningCount > 0) {
+        analytics.sendEventMetric(DAS_STATUS_EVENT_TYPE, WARNINGS, warningCount); // test: serverStatus()
+      }
+      if (hintCount > 0) {
+        analytics.sendEventMetric(DAS_STATUS_EVENT_TYPE, HINTS, hintCount); // test: serverStatus()
+      }
+      if (lintCount > 0) {
+        analytics.sendEventMetric(DAS_STATUS_EVENT_TYPE, LINTS, lintCount); // test: serverStatus()
+      }
+      errorCount = warningCount = hintCount = lintCount = 0;
```
jacob314, line 417: Why do we zero out the counts? This seems wrong. I would expect these #s should match the # of errors reported in the analysis server window. As is, I'm not clear how I would interpret these #s.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -342,51 +367,85 @@ public void serverStatus(AnalysisStatus analysisStatus, PubStatus pubStatus) {
...
+  private void logCompletion(@NotNull String selection, int prefixLength, @NotNull String eventType) {
+    maybeReport(true, (analytics) -> {
+      LOG.debug(eventType + " " + selection + " " + prefixLength);
+      analytics.sendEventMetric(eventType, selection, prefixLength); // test: acceptedCompletion(), lookupCanceled()
+    });
   }
 
   void logE2ECompletionSuccessMS(long e2eCompletionMS) {
-    FlutterInitializer.getAnalytics().sendTiming(E2E_IJ_COMPLETION_TIME, SUCCESS, e2eCompletionMS); // test: logE2ECompletionSuccessMS()
+    maybeReport(true, (analytics) -> {
```
jacob314, line 437: For the completion time, rather than throttling, what you could alternately report a single metric on intellij close that reports the P50, P90, and P95 times for the entire session.
That would make the completion time numbers less noisy than filtering to only report 1 completion per minute.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -456,52 +523,59 @@ private static boolean isDartLookupEvent(@NotNull LookupEvent event) {
...
     @Override
     public void beforeQuickFixInvoked(@NotNull DartQuickFix intention, @NotNull Editor editor, @NotNull PsiFile file) {
-      String path = Objects.requireNonNull(file.getVirtualFile()).getPath();
-      int lineNumber = editor.getCaretModel().getLogicalPosition().line + 1;
-      @SuppressWarnings("ConstantConditions")
-      List<String> errorsOnLine =
-        pathToErrors.containsKey(path) ? pathToErrors.get(path).stream().filter(error -> error.getLocation().getStartLine() == lineNumber)
-          .map(AnalysisError::getCode).collect(Collectors.toList()) : ImmutableList.of();
-      FlutterInitializer.getAnalytics().sendEventMetric(QUICK_FIX, intention.getText(), errorsOnLine.size()); // test: quickFix()
+      maybeReport(true, (analytics) -> {
```
jacob314, line 526: rate limiting quick fixes like this is fine as I doubt users trigger a lot of quick fixes per minute.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -342,51 +367,85 @@ public void serverStatus(AnalysisStatus analysisStatus, PubStatus pubStatus) {
...
-      int hintCount = extractCount(errorCounts, AnalysisErrorType.HINT);
-      FlutterInitializer.getAnalytics().sendEventMetric(DAS_STATUS_EVENT_TYPE, HINTS, hintCount); // test: serverStatus()
-      int lintCount = extractCount(errorCounts, AnalysisErrorType.LINT);
-      FlutterInitializer.getAnalytics().sendEventMetric(DAS_STATUS_EVENT_TYPE, LINTS, lintCount); // test: serverStatus()
+      hintCount = extractCount(errorCounts, AnalysisErrorType.HINT);
+      lintCount = extractCount(errorCounts, AnalysisErrorType.LINT);
+      if (IS_TESTING) {
+        errorCount = warningCount = hintCount = lintCount = 1;
+      }
+      maybeReportErrorCounts();
```
jacob314, line 382: have you manually verified that these #s match what the analyzer window shows?

jacob314:
Rather than rate limiting, I think we should emit summary statistics when the IntelliJ session is closed. The reason is that just rate limiting like this can cause the data to be a bit noisy and skewed. 
For example: right now we'd over sample the very first autocompletion returned right when a user starts typing which might not be representative of autocompletions when users are in the middle of typing. Imagine a user that types for a bit then pauses for a minute to run a build or read code.

Example summary statistics format that I think would work better:
```
{
  event: 'autocompleteE2E',
  P50Time: 73,
  P90Time: 350,
  P95Time: 700 ,
  count: 2000,
}
```
This indicates that the user had 2000 autocomplete events with a P50 time of 73, P90 time of 350, and P95 time of 700.
That way few events are sent to analytics but there is enough data to compute P50, P90, or P95 times across users.
Fyi @jwren who might have some ideas based on how similar problems have been solved in g3.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
```diff
@@ -399,7 +458,10 @@ private void maybeLogInitialAnalysisTime(@NotNull String eventType, @NotNull Str
   }
 
   private void logFileAnalysisTime(@NotNull String kind, String path, long analysisTime) {
-    FlutterInitializer.getAnalytics().sendEvent(kind, DURATION, "", Long.toString(analysisTime)); // test: computedErrors()
+    maybeReport(false, (analytics) -> {
```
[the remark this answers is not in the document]
stevemessick: File analysis time is not throttled. The `false` arg turns it off.

stevemessick:
We can't do all the computation and reporting at exit. IntelliJ enforces a strict, limited time for exit processing (i.e. project close). We can't know what else is going on, so relying on it would potentially limit the data would collect.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
stevemessick: It's been a while, but yes.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
stevemessick: They are accumulated while analysis is active, then sent when analysis is complete. They need to be zeroed so the accumulated values are accurate.

.idea/runConfigurations/flutter_idea__runIde_.xml
stevemessick: No, I missed that. Although I don't think it matters.

stevemessick: Oops, I was wrong. I did add `vmOptions` so removing this empty one is correct.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
stevemessick: See the comment below for why we cannot do it when exiting.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
stevemessick: Done

stevemessick:
> how similar problems have been solved in g3

Percentiles are computed on the server. The same analytic events are used here as are used there.

My statistics is rusty, and my stats book doesn't even mention this, but I'm not sure you'd get the same percentiles if you tried combining percentiles from samples rather than computing the population percentiles on the server.

flutter-idea/src/io/flutter/analytics/FlutterAnalysisServerListener.java
jwren: No right choice.   We can discuss, there are pros and cons for both (as with all things).  If you don't send the information you won't have the signal that users are re-analyzing which is a signal in-itself, but if you do send the information there will be fuller logs with the same information.

jwren (approved):
LGTM

jacob314:
> We can't do all the computation and reporting at exit. IntelliJ enforces a strict, limited time for exit processing (i.e. project close). We can't know what else is going on, so relying on it would potentially limit the data we would collect.

Ok I'm ok with sampling for most things instead. Perhaps I'm being too conservative on how many analytics events we should send.

jacob314:
> 

I agree the median of the median is not the same as the median of all the data but in some ways it is the metric we really want.
For metrics like this what I'm looking for is something I can conceptualize.
For example, the statement I want to make is that for 90% of our users, 90% of their completions are less than 200ms. Framed like that, there is no harm in aggregating locally as it is actually exactly what we want.

jacob314:
lgtm
Lets land this and perhaps iterate on summarizing some of the metrics.

PR #5996, opened 2022-02-25
Use JCEF browser with JxBrowser as backup WIP

helin24:


vlad-lubenskyi:
@helin24 did you reject the idea of using JCEF? Why so?

helin24:
@vlad-lubenskyi we haven't really made a decision yet. We were waiting for all of the platforms we support to have JCEF available including internal ones, which may be close(?), however we're also not sure we want to have to support both of these implementations at once.

vlad-lubenskyi:
Maybe @forantar by any chance knows JetBrains' plans for different platform support?

PR #6005, opened 2022-03-02
Send internal errors to analytics for disconnections

helin24:
This is to help us track internal disconnection errors. Not all errors caught here will be truly the types of errors we'll be looking for, ~but this at least filters out startup messages as an app compiles (the startup messages are intended to go to stderr,~ as stdout is intended to be only what should be parsed by the flutter daemon).

Once we have a sense of what types of post-startup error exist, then we can further refine what messages represent different types of disconnections if needed.

### DIFF
--- a/flutter-idea/src/io/flutter/run/daemon/DaemonConsoleView.java
+++ b/flutter-idea/src/io/flutter/run/daemon/DaemonConsoleView.java
@@ -30,7 +30,6 @@ import org.jetbrains.annotations.NotNull;
  */
 public class DaemonConsoleView extends ConsoleViewImpl {
   private static final Logger LOG = Logger.getInstance(DaemonConsoleView.class);
-  private boolean sendErrorsToAnalytics = false;
   private final Analytics analytics = FlutterInitializer.getAnalytics();
 
   /**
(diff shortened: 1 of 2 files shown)


flutter-idea/src/io/flutter/run/daemon/DaemonConsoleView.java
```diff
@@ -75,6 +83,10 @@ public void print(@NotNull String text, @NotNull ConsoleViewContentType contentT
     }
   }
 
+  public void maybeSendErrorsToAnalytics() {
```
jacob314, line 86: I would remove this method. the name is confusing and the operation should be fast enough to just execute on each print so no need to add some possibly error prone caching.

flutter-idea/src/io/flutter/run/daemon/DaemonConsoleView.java
```diff
@@ -67,6 +72,9 @@ public void print(@NotNull String text, @NotNull ConsoleViewContentType contentT
     if (contentType != ConsoleViewContentType.NORMAL_OUTPUT) {
       writeAvailableLines();
 
+      if (sendErrorsToAnalytics && text.length() > 0) {
+        analytics.sendEvent("potential-disconnect", text);
```
jacob314, line 76: add the 
WorkspaceCache.getInstance(getProject()).isBazel()
check right here and note we are only logging for internal users and this logging needs to not be enabled for all users without being modified to only test for matches for specific known disconnect strings.

flutter-idea/src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -887,6 +887,9 @@ public void onAppDebugPort(@NotNull DaemonEvent.AppDebugPort debugInfo) {
   @Override
   public void onAppStarted(DaemonEvent.AppStarted started) {
     app.changeState(FlutterApp.State.STARTED);
+    if (app.getConsole() instanceof DaemonConsoleView) {
```
jacob314, line 890: remove this check.

jacob314 (approved):
lgtm

PR #6010, opened 2022-03-03
Revert "Use VM for mapping breakpoint files during app run (#5947)"

stevemessick:
Fixes #6002

PR #6044, opened 2022-03-16
Use VM service for mapping breakpoint URIs

helin24:
(Update to https://github.com/flutter/flutter-intellij/pull/5947, https://github.com/flutter/flutter-intellij/pull/6010)

This change updates the VM service files to include the new methods that find resolved URIs and package URIs for files, so that we can use the VM for breakpoint mapping instead of the mapping method in FlutterPositionMapper (which does a combination of analyzer and other things).

The initial change I made did not handle Windows paths and web correctly. The Windows issue has been fixed here, and the web issue has been fixed in flutter (message format for DWDS - https://dart-review.googlesource.com/c/sdk/+/236281, https://github.com/flutter/flutter/commit/1880066dd3a092703e8985e188b8774e40a181b8)

Only changes in the first two files (FlutterSdkVersion, VmServiceWrapper) are significant; the rest are generated file changes.


helin24:
cc @kenzieschmoll

jwren (approved):
LGTM

helin24:
Oops! I still need to update the version check, though 2.13.0 is a reasonable guess. I'll make another update if that's wrong.

PR #6051, opened 2022-03-22
Add jxbrowser log file for easier debugging

helin24:
This seems to produce a different set of logs than when we have logs directed into the normal idea.log file, and these will be easier to send directly to JxBrowser developers who help us with debugging.


vlad-lubenskyi:
I'd suggest setting DEBUG level as default one and make it configurable. The library can produce **a lot** of output at the ALL level

stevemessick (approved):
We should monitor the size of the DEBUG log files. We need to be good citizens, and not use too much unnecessary disk space. But for. now, this is fine.

PR #6073, opened 2022-04-01
Ignore files saved prior to the current event

stevemessick:


helin24:
Is there some context on why this change is being made? When/why are there file changes that don't reach the reload manager for so long?

stevemessick:
If you have a Dart file open when you modify a scratch file then the Dart file triggers a hot reload. If the file was not modified then it is not saved, but still triggers the notification. If it was modified then it got saved during the save event that triggered the notification.

helin24 (approved):

PR #6074, opened 2022-04-01
Update CONTRIBUTING.md

jacob314:
Add missing step to set your JAVA_HOME directory on the command line.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,6 +35,11 @@ file.
   - Determine the directory of your downloaded IntelliJ Community Edition installation 
     (e.g, `IntelliJ IDEA CE.app`, `~/idea-IC-183.4886.37` or 
     `~/Library/Application Support/JetBrains/Toolbox/apps/IDEA-U/ch-0/211.6693.111/IntelliJ IDEA.app`)
+* Set your JAVA_HOME directory.
+    - On Mac, the following works:
+      ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -35,6 +35,11 @@ file.
   - Determine the directory of your downloaded IntelliJ Community Edition installation 
     (e.g, `IntelliJ IDEA CE.app`, `~/idea-IC-183.4886.37` or 
     `~/Library/Application Support/JetBrains/Toolbox/apps/IDEA-U/ch-0/211.6693.111/IntelliJ IDEA.app`)
+* Set your JAVA_HOME directory.
+    - On Mac, the following works:
+      ```
+      export JAVA_HOME=`/usr/libexec/java_home -v 11.0.14.1`
```
helin24, line 41: We should probably make this more general; if someone has a different version of java then this instruction won't work. I normally first run `/usr/libexec/java_home -V` to see what versions of java I have, then put the number after `-v`. (instructions here: https://stackoverflow.com/questions/21964709/how-to-set-or-change-the-default-java-jdk-version-on-macos)

11.0.14.1 works for us now since we're using java version 11 in general (more instructions related below)

jacob314: Updated the instructions to suggest your workflow.

helin24 (approved):

PR #6112, opened 2022-04-15
Add dart devtools for starting server

helin24:
Using `dart devtools` is the faster way to start the server and less complex than the daemon method.

@stevemessick  Can we start deleting the other methods of starting the DevTools server? The oldest method using pub is for versions of Flutter before v1.26, which is from a year ago. 

cc @kenzieschmoll

### DIFF
--- a/flutter-idea/src/io/flutter/run/daemon/DevToolsService.java
+++ b/flutter-idea/src/io/flutter/run/daemon/DevToolsService.java
@@ -146,17 +146,19 @@ public class DevToolsService {
             try {
               final JsonElement element = JsonUtils.parseString(text);
 
-              // params.port
               final JsonObject obj = element.getAsJsonObject();
-              final JsonObject params = obj.getAsJsonObject("params");
-              final String host = JsonUtils.getStringMember(params, "host");
-              final int port = JsonUtils.getIntMember(params, "port");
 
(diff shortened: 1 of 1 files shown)


flutter-idea/src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -115,6 +130,56 @@ private void startServer() {
...
+          final String text = event.getText().trim();
+
+          if (text.startsWith("{") && text.endsWith("}")) {
+            // {"event":"server.started","params":{"host":"127.0.0.1","port":9100}}
+
+            try {
+              final JsonElement element = JsonUtils.parseString(text);
+
+              // params.port
+              final JsonObject obj = element.getAsJsonObject();
```
stevemessick, line 150: In case `dart devtools --machine` changes, you should check for `event:server.started` here.

stevemessick (approved):

stevemessick:
> Can we start deleting the other methods of starting the DevTools server?

Yes. My only concern is for some large project that cannot update to null safety, but they can continue using an old version of the IDE and plugin, in that case.

helin24:
Thanks, I'll try to find info on when it would be reasonable to stop supporting.

PR #6115, opened 2022-04-18
Use devtools script to launch bazel devtools

helin24:
We've been using the daemon to request DevTools server start, but I changed over external DevTools to use `dart devtools` directly instead since that's a little faster and less indirection. This is the equivalent change for bazel workspaces


jwren (commented):
Make sure to do a fair bit of manual testing before landing, but LGTM!

jwren (approved):
LGTM

Issue #6042, opened 2022-03-15
IDE NullPointerException in FlutterWidgetPerf.dispose


filiph:
## What happened

I terminated a Profile run of a macOS Flutter app with the red "Stop" button. IDE works okay, I just got the big red notification and clicked "Report to Google".

## Version information

Android Studio `Bumblebee | 2021.1.1 Patch 2` • Flutter plugin `io.flutter 65.2.2` • Dart plugin `211.7808`

Flutter 2.10.3 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 7e9793dee1 (13 days ago) • 2022-03-02 11:23:12 -0600
Engine • revision bd539267b4
Tools • Dart 2.16.1 • DevTools 2.9.2

## Exception

null

```
java.lang.NullPointerException
	at com.intellij.openapi.util.ObjectTree.runWithTrace(ObjectTree.java:138)
	at com.intellij.openapi.util.ObjectTree.executeAll(ObjectTree.java:168)
	at com.intellij.openapi.util.Disposer.dispose(Disposer.java:153)
	at com.intellij.openapi.util.Disposer.dispose(Disposer.java:141)
	at io.flutter.perf.FlutterWidgetPerf.dispose(FlutterWidgetPerf.java:518)
	at io.flutter.perf.FlutterWidgetPerfManager.updateCurrentAppChanged(FlutterWidgetPerfManager.java:276)
	at io.flutter.utils.StreamSubscription.notify(StreamSubscription.java:43)
	at io.flutter.utils.EventStream.lambda$setValue$0(EventStream.java:78)
	at com.intellij.openapi.application.TransactionGuardImpl.runWithWritingAllowed(TransactionGuardImpl.java:218)
	at com.intellij.openapi.application.TransactionGuardImpl.access$200(TransactionGuardImpl.java:21)
	at com.intellij.openapi.application.TransactionGuardImpl$2.run(TransactionGuardImpl.java:200)
	at com.intellij.openapi.application.impl.ApplicationImpl.runIntendedWriteActionOnCurrentThread(ApplicationImpl.java:787)
	at com.intellij.openapi.application.impl.ApplicationImpl.lambda$invokeLater$4(ApplicationImpl.java:325)
	at com.intellij.openapi.application.impl.FlushQueue.doRun(FlushQueue.java:84)
	at com.intellij.openapi.application.impl.FlushQueue.runNextEvent(FlushQueue.java:133)
[...]

PR #6080, opened 2022-04-04
Fix disposer bug

jacob314:
Take 2. This time it is signed (I hope).

Fixes https://github.com/flutter/flutter-intellij/issues/6042

Filed https://github.com/flutter/flutter-intellij/issues/6077 to track down the broader tech debt issue that we are manually calling dispose() on Disposable objects a fair amount which as a JetBrains antipatern.

Issue #6077, opened 2022-04-02
Track down all manual calls to dispose() in the IntelliJ Plugin


jacob314:
Some classes like `FlutterWidgetPerf`
implement the Disposable interface but have `dispose()`
called manually which is counter to how `dispose()` should be called for a properly functioning IntelliJ plugin.

Mixing manual and automatic calls to dispose leads to confusing bugs like
https://github.com/flutter/flutter-intellij/issues/6042

stevemessick:
`FlutterWidgetPerf.dispose()` is called by `FlutterWidgetPerfManager.dispose()`, which is itself never called due to using the project as a dispose parent: `Disposer.register(project, this);`

We need to fix #6077 before working on this one.

PR #6120, opened 2022-04-19
Fix disposer bug

stevemessick:
This s a (signed) replacement for #6080.

Fixes https://github.com/flutter/flutter-intellij/issues/6042

Filed https://github.com/flutter/flutter-intellij/issues/6077 to track down the broader tech debt issue that we are manually calling dispose() on Disposable objects a fair amount which is a JetBrains anti-pattern.


stevemessick:
Once either this or #6080 lands, we can close every issue labeled "perf-tool".

jacob314:
lgtm

jacob314 (approved):

PR #6129, opened 2022-04-21
Refactor Transport for easier logging while debugging

stevemessick:
The idea here is to change the reference to `HttpTransport` to `NonTransport` while debugging the new analytics code. That way, nothing gets sent to Google Analytics, and we get a good idea of how much data is being sent.

@jacob314 I noticed while doing this that we already have throttling in the analytics. It only sends data once per second.

### DIFF
--- a/flutter-idea/src/io/flutter/analytics/Analytics.java
+++ b/flutter-idea/src/io/flutter/analytics/Analytics.java
@@ -5,6 +5,7 @@
  */
 package io.flutter.analytics;
 
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.project.ProjectManager;
 import com.intellij.openapi.util.SystemInfo;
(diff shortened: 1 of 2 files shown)


flutter-idea/src/io/flutter/analytics/Analytics.java
```diff
@@ -286,5 +283,33 @@ public void send(String url, Map<String, String> values) {
...
+        }
+        postData.append(URLEncoder.encode(param.getKey(), "UTF-8"));
+        postData.append('=');
+        postData.append(URLEncoder.encode(param.getValue(), "UTF-8"));
+      }
+      return postData.toString().getBytes(StandardCharsets.UTF_8);
+    }
+  }
+
+  private static class NonTransport extends HttpTransport {
```
helin24, line 301: Do you mean you will temporarily change code to use this `NonTransport` class when you're debugging analytics code locally? Can you add a comment about that intention here?

helin24 (approved):

flutter-idea/src/io/flutter/analytics/Analytics.java
stevemessick: Done.

PR #6184, opened 2022-05-20
Config for no-response action

stevemessick:
The no-response bot is no longer working. This switches to use the no-response action.


helin24 (approved):

stevemessick:
FlutterWidgetPerfTest.testOverallStatsCalculation_newFormat() may be flakey. It failed the first time on the github run at line 626. @jacob314 I wonder if the delay should just add 1 to the constant? Anyway, second run passed.

I don't know why kokoro failed. It uses the un-tars same file as the github action unit test. The kokoro run claims it cannot find ant, but it was found by the action run. I restarted it and it failed again, so I'm going to ignore it for now.

PR #6208, opened 2022-05-31
Use distributed icons for current SDK

stevemessick:


resources/flutter/README.md
```diff
@@ -1,2 +1,4 @@
 The contents from this directory are copied from
 https://github.com/flutter/tools_metadata/tree/master/resources.
+
+When it is updated, remember to update the version in FlutterSdkVersion.DISTRIBUTED_ICONS.
```
helin24, line 4: I'm a little confused by this instruction - does this mean when the icons are updated in the Flutter SDK, only the latest version of the icons can be used?

helin24 (approved):

resources/flutter/README.md
stevemessick: We only ship one version of the icons. Those icons are generated from the font file obtained from the SDK version in `FlutterSdkVersion.DISTRIBUTED_ICONS`, which will be the latest stable (or possibly beta).

Older SDKs use code that loads individual icons from the font file. That code isn't working too well for Flutter 3.0 and later, so I went back to the old way of showing icon previews.

PR #6224, opened 2022-06-09
[WIP] Update the build for 2022.2 EAP

stevemessick:
Update Gradle and gradlew, the IntelliJ and Kotlin plugins for Gradle, and define a build spec for 2022.2 EAP that uses LATEST_EAP_SNAPSHOT to compile and build the Flutter plugin.

This isn't currently working. Forms do not display, which breaks things like settings, new project creation, and a few other things.


stevemessick:
I asked for help from JetBrains:
https://youtrack.jetbrains.com/issue/WEB-56193/Forms-not-working-in-20222-EAP

stevemessick:
It looks like the plugin has been delayed, so #6225 will remain open for now.

product-matrix.json
```diff
@@ -60,6 +60,18 @@
...
       "untilBuild": "221.*"
+    },
+    {
+      "channel": "dev",
+      "comments": "IntelliJ 2022.2",
+      "name": "2022.2",
+      "version": "2022.2",
+      "ideaProduct": "ideaIC",
+      "ideaVersion": "222.2889.14",
+      "baseVersion": "LATEST-EAP-SNAPSHOT",
```
helin24, line 71: Is this base version going to change over time? Not sure if it matters.

helin24 (approved):

product-matrix.json
stevemessick: It will become something different when the EAP version is released as stable.

PR #6241, opened 2022-06-24
Update .iml for 221

stevemessick:
Trying to build Android Studio from sources with the Flutter plugin.


helin24 (approved):

stevemessick:
I get a Kotlin internal error, reported here:
https://youtrack.jetbrains.com/issue/KT-52960/Kotlin-internal-error-during-IR-lowering

PR #6263, opened 2022-07-14
Log bazel mapping differences from analyzer

helin24:
This should not change behavior, as the old analyzer path mapping is the one that will be used for file mapping. I'm only using the new bazel mapping package to see whether the path generated is the same, logging if not.

### DIFF
--- a/flutter-idea/src/io/flutter/run/FlutterPositionMapper.java
+++ b/flutter-idea/src/io/flutter/run/FlutterPositionMapper.java
@@ -5,6 +5,7 @@
  */
 package io.flutter.run;
 
+import com.android.internal.inputmethod.CancellationGroup;
 import com.google.common.annotations.VisibleForTesting;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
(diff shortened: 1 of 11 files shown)


stevemessick:
It looks like you need to add a dependency to the Gradle script. `com.android.internal.inputmethod` isn't being found.

helin24:
> It looks like you need to add a dependency to the Gradle script. `com.android.internal.inputmethod` isn't being found.

I'm confused about why this is here - it looks like I added it but I'm not sure why it would be used in this file.

flutter-idea/third_party/vmServiceDrivers/org/dartlang/vm/service/VmService.java
```diff
@@ -83,7 +83,7 @@ public class VmService extends VmServiceBase {
   /**
```
stevemessick, line 83: The remaining files were all generated from a script, right? I'm not checking them.

stevemessick (changes requested):
Minor nit in the code, but the Gradle script is broken.

stevemessick (approved):
That's the easy way to resolve compiler errors :)

flutter-idea/third_party/vmServiceDrivers/org/dartlang/vm/service/VmService.java
helin24: correct

flutter-idea/src/io/flutter/run/FlutterPositionMapper.java
```diff
@@ -321,6 +335,20 @@ protected VirtualFile findLocalFile(@NotNull String uri) {
       if (analyzer != null && !isDartPatchUri(remoteUri)) {
         final String path = analyzer.getAbsolutePath(remoteUri);
         if (path != null) {
+          if (fileFuture != null && path.contains("google3")) {
```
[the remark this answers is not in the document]
helin24: True, I'll add a workspace check.

PR #6265, opened 2022-07-18
Check for disposed editor

stevemessick:
Fixes 6264


helin24 (approved):

bernaferrari:
Was this released already? I'm still getting the exception.

stevemessick:
It should be available Sept 1.

PR #6279, opened 2022-08-10
feat: respect embedded emulator settings

### LINKED ISSUE #5917: Native Android vs. Flutter-Android Emulator Launch Inconsistency
There is an inconsistency between a native Android project vs a Flutter-Android project in how the emulator is launched from within Android Studio.

- In a **Native Android** app, clicking the **Run** button first launches the emulator and then starts the app.

- In a **Flutter-Android** app, the user _must_ manually launch the emulator and then start the app (because automatic launching of the emulator fails for some reason.)

## Native Android Project

After clicking on the **Run** button, which is to the right of the device dropdown, Android Studio automatically launches the emulator and then starts the app.  So far so good.

## Flutter-Android Project

Notice how the device dropdown to the left of the  **Run** button is disabled.  Moreover, we have additional dropdown to the left showing `<no devices>`.  **Inconsistency No. 1:** _why have a dropdown that does not work?_

After clicking on the **Run** button, I get:

Not sure why a device was not found as it shows up in the new dropdown, like this:

**Inconsistency No. 2:** _This new dropdown does **not** respect the **Launch in a tool window** option._

So, as a workaround, I must use the **Device Manager** to launch the emulator as follows:

After waiting a few more seconds, the launched emulator finally shows up within the toolbar, replacing the `<no devices>`, like this:

Finally, I can click on the **Run** button and start the Flutter-Android app.

### Expected Behavior

- Clicking on the **Run** button within a **Flutter-Android** app shall first launch the emulator and the start the app, just like it does within a **Native Android** project.

- We should not have the additional dropdown as it confuses a developer coming from the native Android land.

- If we we want to keep this new dropdown, at least, it should respect the **Launch in a tool window** option.

### Version info

```
/home/precor/.local/flutter/bin/flutter doctor --verbose
[✓] Flutter (Channel unknown, 2.8.2-0.0.pre.1, on Ubuntu 20.04.3 LTS 5.13.0-27-generic, locale en_US.UTF-8)
    • Flutter version 2.8.2-0.0.pre.1 at /home/precor/.local/flutter
    • Upstream repository unknown
    • Framework revision 91318852b3 (13 hours ago), 2022-01-27 06:42:52 -0800
    • Engine revision 890a5fca2e
    • Dart version 2.15.1

    • Android SDK at /home/precor/Android/Sdk
    • Platform android-31, build-tools 30.0.3
    • ANDROID_SDK_ROOT = /home/precor/Android/Sdk
    • Java binary at: /opt/android-studio/jre/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.11+0-b60-7590822)

[✓] Android Studio (version 2021.1)
    • Android Studio at /opt/android-studio
    • Flutter plugin version 63.2.2
    • Dart plugin version 211.7798
    • Java version OpenJDK Runtime Environment (build 11.0.11+0-b60-7590822)

[✓] VS Code (version 1.63.2)
    • VS Code at /usr/share/code
    • Flutter extension version 3.32.0

[!] Connected device
    ! No devices available

! Doctor found issues in 1 category.
Process finished with exit code 0
```

**Note:** I have Flutter SDK 2.8.1 installed with minor mods to the `flutter create -t app` template.  Here is my patch: [flutter.zip](https://github.com/flutter/flutter-intellij/files/7955807/flutter.zip)

```
$ flutter emulators
1 available emulator:

Pixel_3a_XL_API_30 • Pixel 3a XL API 30 • Google • android

To run an emulator, run 'flutter emulators --launch <emulator id>'.
To create a new emulator, run 'flutter emulators --create [--name xyz]'.

You can find more information on managing emulators at the links below:
  https://developer.android.com/studio/run/managing-avds
  https://developer.android.com/studio/command-line/avdmanager
```

```
Android Studio Bumblebee | 2021.1.1
Build #AI-211.7628.21.2111.8092744, built on January 18, 2022
Runtime version: 11.0.11+0-b60-7590822 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Linux 5.13.0-27-generic
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 8
Registry: external.system.auto.import.disabled=true
Non-Bundled Plugins: Dart (211.7798), org.jetbrains.kotlin (211-1.6.10-release-923-AS7442.40), io.flutter (63.2.2), org.intellij.plugins.markdown (211.7142.37)
Current Desktop: ubuntu:GNOME
```

pedromassango:
Fix #5917


pedromassango:
@stevemessick the `flutter-idea` project has no acess to `EmulatorSettings` even though I have code completion. Running the plugin works but you try to launch the emulator you will get a crash sayng "EmulatorSettings not found".

I'm struggling to find the correct dependency for that class

stevemessick:
@pedromassango Add this to line 55 of flutter-idea/build.gradle.kts: `"android",`

I'm currently reviewing this PR. After that change, it may be ready for review, but I'm having some trouble launching the IDE.

pedromassango:
Got the same result: class not found

stevemessick:
Yeah, the change I suggested does nothing. The Android plugin is already a dependency. Your code does work when I build a binary plugin and install it. I'm having trouble getting it to work in the debugger, though. I verified that `EmulatorSettings` is in the Android plugin in `lib/android.jar`, and the plugin list include `org.jetbrains.android`, which is its plugin-id in the `plugin.xml` file. I don't understand why the class is not found when running under the debugger.

One thing I noticed is that the tool window starts collapsed. We will need some code to make it visible. See `FlutterPerformanceView.activateToolWindow()` for an example of doing that.

pedromassango:
Could it be that the plugin is not loaded/stripped in Debug mode?

stevemessick:
No, I checked the plugins page. It is listed as "bundled".

pedromassango:
> We will need some code to make it visible. See `FlutterPerformanceView.activateToolWindow()` for an example of doing that.

I was able to open the tool window, however it closes itself right after I open it. Probably some internal code is closing it

stevemessick:
> I was able to open the tool window, however it closes itself right after it opens. Probably some internal code is closing it.

It could be a timing issue. I'd make sure it was opened after everything else had been done, and possibly wrap that code in a call to `Application.invokeLater()`. Sorry but I didn't have a chance to look into the class loading problem today.

pedromassango:
> It could be a timing issue.

Hmm, I'm getting the same behavior after implementing `invokeLater`. I will test again by generating a binary of the plugin

stevemessick:
It looks like your code is missing a few import statements. Does the behavior change after the code is fixed?

pedromassango:
> Does the behavior change after the code is fixed?

I get the same behavior. I was unable to test it as a stand-alone plugin, had some challenges generating the plugin.

stevemessick:
Are you no longer getting this error when running in the debugger?
```java
java.lang.NoClassDefFoundError: com/android/tools/idea/emulator/EmulatorSettings
```
I still do. Very puzzling.

I patched in your changes and built the plugin. I only checked AS 2021.1 but it worked fine. The tool window opened, and after a short delay, the emulator was visible. I'm using macos. What platform are you testing on?

> faced some challenges generating the plugin

Is that related to downloads? Once everything is downloaded it will be re-used, but it uses a lot of bandwidth initially. The GitHub actions build the plugin for each supported platform, so it should "just work".

pedromassango:
> What platform are you testing on?

Mac OS as well. This is crazy 🐒 

>Is that related to downloads? Once everything is downloaded it will be re-used, but it uses a lot of bandwidth initially. The GitHub actions build the plugin for each supported platform, so it should "just work".

Maybe I'm missing something here. How are you generating a new build? I tried to build it mannually following [this](https://plugins.jetbrains.com/docs/intellij/deploying-plugin.html) Jetbrains tutorial.

pedromassango:
> `bin/plugin make`

Thanks. Got the same error after installing the generated plugin from `flutter-intelliJ/releases/release_master/2022.2/`

stevemessick:
I can reproduce that problem. I also tried another IntelliJ version, and got the same result. Now, I think I understand what the problem is. We have a conditional dependency in the Android Studio version on the Android plugin. We don't have that for IntelliJ.

Try adding this to `plugin_template.xml`:
`<depends>org.jetbrains.android</depends>`

Then run `bin/plugin generate` and `bin/plugin make` (or just launch the debugger).

stevemessick:
Adding that dependency works. I can run the IDE in the debugger, launch an emulator, and the emulator tool window appears but is immediately closed. After opening it, the embedded emulator is running. I tried moving the call to `openEmulatorToolWindow()` into the `ProcessAdaptor` but that didn't help. It looks like IntelliJ is closing the window when the emulator tries to use it. You might try adding a delay, but that can be tricky, depending on how fast/slow a user's compute is (and yours is _fast_).

Everything is fine for Android Studio. The tool window stays open and shows the emulator after it is initialized. I don't have time to investigate the difference. If you want to, I suggest looking into the `jetbrains/intellij-community` repo sources. If you don't have time, then I think it would be appropriate to check that the user is not running Android Studio (`FlutterUtils.isAndroidStudio()`) then showing a blue ballon with some text explaining that the tool window has to be re-opened. The tool window will be in the side-bar.

Launching with the debugger can be flakey. I've seen a number of weird issues. Once, it tried to load Android Studio-specific classes. Sometimes it cannot find classes in the Dart plugin. Once the window never opened. And, always the NST library cannot be loaded (ignore that one). If weird things happen, just kill it and launch again.

This is looking good! Thanks for driving it.

pedromassango:
Yes, the class-not-found issue is now fixed, thanks ✅ 

> It looks like IntelliJ is closing the window when the emulator tries to use it.

I was able to make it to work as follow: the emulator starts, the tool window opens and closes itself, after a short period of time (~2sec) the tool windows opens itself again and the emulator is visible and ready. Would that be ok to merge this? not for me (I'm still not happy with that, but open for your PoV).

I will look into it again and see what I can do. otherwise we can go with the ballon approach.

stevemessick:
I think your current behavior is fine since this only happens in IntelliJ and 90%+ of our users use Android Studio. One person suggested adding a one-time balloon pointing out the open-close-open behavior does not happen with Android Studio. I'll leave that up to you. If you'll commit your latest and change it to ready for review I'll merge it. It would be nice to have this in the next release, which is Sept 1. Again, thanks!

pedromassango:
Alright. since I'm still not able to fix the issue, it is fair to merge it like that and prove it on another PR.

resources/META-INF/plugin.xml
```diff
@@ -28,6 +28,7 @@
   <depends>com.intellij.modules.xdebugger</depends>
   <depends>com.intellij.modules.coverage</depends>
   <depends>org.jetbrains.plugins.yaml</depends>
+  <depends>org.jetbrains.android</depends>
```
stevemessick, line 31: I think this change means we will not be able to launch Android Studio in the debugger. But I haven't been able to do that since late last year, due to a Kotlin compiler bug, so that's not a blocker. Anyway, if the bug gets fixed I can add some code to the `plugin` tool to fix up the `plugin.xml` content.

stevemessick (approved):
Thanks! Do you need me to commit this?

pedromassango:
> Do you need me to commit this?

I can merge it. Do I need to what for the `kokoro-mac` task to be completed?

resources/META-INF/plugin.xml
pedromassango: Great. It is just that I wasn’t able to fix the class-not-found with this change

stevemessick:
No need to wait. It will fail.

stevemessick:
@pedromassango Oops! I forgot to ask you to add your "name" to our `AUTHORS` file. Could you create a new PR to do that? It's nice to get credit for one's work, but if you don't want to, it isn't required.

pedromassango:
Oh, nice. [Here](https://github.com/flutter/flutter-intellij/pull/6286) is the PR

stevemessick:
Nope! It is going to fail anyway.

PR #6332, opened 2022-09-29
Revert "Revert to previous version of scorecards."

godofredoc:
Reverts flutter/flutter-intellij#6329

Reverting to use the latest version the problem was actually the workflow using a repo level token rather than the repository one.


godofredoc:
Dependabot created an updated PR

Issue #6351, opened 2022-10-14
Opening New Project from Existing Project Exception


jstorch:
Thanks for the feedback! If your issue is related to the Flutter framework itself,
please open an issue at
[github.com/flutter/flutter](https://github.com/flutter/flutter/issues/new).

## Steps to Reproduce

Opening project from existing open project.  Chose use this window.  Use Alt-T to choose "This"

## What happened

Crash - boom. Got Error Window.

## Version information

Android Studio `Dolphin | 2021.3.1` • Flutter plugin `io.flutter 70.0.3` • Dart plugin `213.7371`

Error getting Flutter sdk information.

## Exception

From the listener io.flutter.jxbrowser.EmbeddedBrowser$1@6f83d90 (class io.flutter.jxbrowser.EmbeddedBrowser$1)

```
java.lang.IllegalStateException: Failed to receive the response.
	at com.teamdev.jxbrowser.internal.rpc.transport.CommonThreadResponseConsumer.await(CommonThreadResponseConsumer.java:55)
	at com.teamdev.jxbrowser.internal.rpc.transport.CommonThreadCallExecutor.execute(CommonThreadCallExecutor.java:29)
	at com.teamdev.jxbrowser.internal.rpc.UniversalServiceConnection.invoke(UniversalServiceConnection.java:183)
	at com.teamdev.jxbrowser.browser.internal.BrowserImpl.tryToClose(BrowserImpl.java:556)
	at com.teamdev.jxbrowser.browser.internal.BrowserImpl.close(BrowserImpl.java:526)
	at com.teamdev.jxbrowser.browser.internal.BrowserImpl.close(BrowserImpl.java:511)
	at io.flutter.jxbrowser.EmbeddedBrowser$1.projectClosing(EmbeddedBrowser.java:88)
	at com.intellij.openapi.project.impl.ProjectManagerImpl$1.projectClosing(ProjectManagerImpl.java:115)
	at com.intellij.util.messages.impl.MessageBusImpl.invokeMethod(MessageBusImpl.java:645)
	at com.intellij.util.messages.impl.MessageBusImpl.invokeListener(MessageBusImpl.java:620)
	at com.intellij.util.messages.impl.MessageBusImpl.access$300(MessageBusImpl.java:33)
	at com.intellij.util.messages.impl.MessageBusImpl$MessagePublisher.executeOrAddToQueue(MessageBusImpl.java:208)
[...]

PR #6353, opened 2022-10-18
Catch browser close exceptions

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/6351

I'm not able to reproduce that issue myself, but it seems appropriate to catch any exceptions from the JxBrowser browser.


stevemessick (approved):
Thanks!

PR #6372, opened 2022-10-25
Disable kokoro until other problems are solved

stevemessick:
The kokoro job fails due to loss of permissions. It isn't doing anything critical, and I don't have time to fix it now, so I'm disabling it.


stevemessick:
Oops, the permission failure happens upstream of these scripts. Deleting...

PR #6389, opened 2022-11-03
Permit attach in bazel context

helin24:
This enables the attach button for bazel projects and calls the correct command for attach. We wanted to add attach to the IDEs since the internal scripts for attach have been updated to make this easier.

CC @chingjun

### DIFF
--- a/flutter-idea/src/io/flutter/actions/AttachDebuggerAction.java
+++ b/flutter-idea/src/io/flutter/actions/AttachDebuggerAction.java
@@ -93,7 +93,6 @@ public class AttachDebuggerAction extends FlutterSdkAction {
   }
 
   public void startCommandInBazelContext(@NotNull Project project, @NotNull Workspace workspace, AnActionEvent event) {
-    // NOTE: When making changes here, consider making similar changes to RunFlutterAction.
     FlutterInitializer.sendAnalyticsAction(this);
 
     RunConfiguration configuration = findRunConfig(project);
--- a/flutter-idea/src/io/flutter/actions/AttachDebuggerAction.java
+++ b/flutter-idea/src/io/flutter/actions/AttachDebuggerAction.java
@@ -105,10 +105,6 @@ public class AttachDebuggerAction extends FlutterSdkAction {
       }
       configuration = settings.getConfiguration();
       if (!(configuration instanceof BazelRunConfig)) {
-        if (project.isDefault() || !FlutterSdkUtil.hasFlutterModules(project)) {
-          return;
-        }
-        showSelectConfigDialog();
         return;
       }
     }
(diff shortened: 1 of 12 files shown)


flutter-idea/src/io/flutter/actions/AttachDebuggerAction.java
```diff
@@ -89,6 +92,40 @@ public void startCommand(@NotNull Project project, @NotNull FlutterSdk sdk, @Nul
     ProgramRunnerUtil.executeConfiguration(env, false, true);
   }
 
+  public void startCommandInBazelContext(@NotNull Project project, @NotNull Workspace workspace, AnActionEvent event) {
```
stevemessick, line 95: Looks like `event` should be `@NotNull`.

flutter-idea/src/io/flutter/run/bazel/BazelFields.java
```diff
@@ -156,6 +156,15 @@ private String getRunScriptFromWorkspace(@NotNull final Project project) {
     return runScript;
   }
 
+  private String getToolsScriptFromWorkspace(@NotNull final Project project) {
+    final Workspace workspace = getWorkspace(project);
+    String toolsScript = workspace == null ? null : workspace.getToolsScript();
+    if (toolsScript != null) {
+      toolsScript = workspace.getRoot().getPath() + "/" + toolsScript;
```
stevemessick, line 163: Does this work on Windows? I think this could happen.

flutter-idea/testSrc/unit/io/flutter/bazel/FakeWorkspaceFactory.java
```diff
@@ -50,6 +51,9 @@ public static Pair.NonNull<MockVirtualFileSystem, Workspace> createWorkspaceAndF
     if (syncScript != null) {
       fs.file("/workspace/" + syncScript, "");
     }
+    if (toolsScript != null) {
+      fs.file("/workspace/" + toolsScript, "");
```
stevemessick, line 55: Another potential Windows issue (not sure about filesystem details, it may be fine).

flutter-idea/testSrc/unit/io/flutter/bazel/FakeWorkspaceFactory.java
```diff
@@ -97,6 +102,7 @@ public static Pair.NonNull<MockVirtualFileSystem, Workspace> createWorkspaceAndF
       "scripts/flutter-test.sh",
       "scripts/flutter-run.sh",
       "scripts/flutter-sync.sh",
+      "scripts/flutter-tools.sh",
```
stevemessick, line 105: ditto

stevemessick (approved):

flutter-idea/src/io/flutter/run/bazel/BazelFields.java
helin24: I can use a separator char instead here

flutter-idea/testSrc/unit/io/flutter/bazel/FakeWorkspaceFactory.java
helin24: This file is only used in tests so I don't think we need to worry about the file path here.

flutter-idea/testSrc/unit/io/flutter/bazel/FakeWorkspaceFactory.java
helin24: Also only for tests

PR #6390, opened 2022-11-03
Change Container to Placeholder in live templates

### LINKED ISSUE #6288: Change snippets to use Placeholder() instead of Container()
See discussion at https://github.com/Dart-Code/Dart-Code/issues/3975#issuecomment-1234507375. Empty `Container()`s may trigger lints, and `Placeholder` seems like an ideal replacement (`const Placeholder()`).

I'm going to change this in VS Code + LSP (https://github.com/Dart-Code/Dart-Code/issues/4136) and it makes sense to keep these the same.

stevemessick:
Fixes #6288

Also updates a couple settings that got changed when I moved to IntelliJ 2022.2 as the dev platform.

@DanTup We can coordinate the actual date that this is released, if you want.

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="11" />
+    <bytecodeTargetLevel target="17" />
   </component>
 </project>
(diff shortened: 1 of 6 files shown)


jwren (approved):

DanTup:
> @DanTup We can coordinate the actual date that this is released, if you want.

For most VS Code users this is coming from the server so will show up based on SDK version and probably not worth trying to sync.

I'll submit the SDK changes shortly, and do the same in the VS Code plugin for users on onlder SDKs that don't yet use server snippets. Thanks!

DanTup:
By the way - the goal of this was to reduce triggering lints, but while testing my change I noticed that without `const` you can also trigger a lint:

Screenshot 2022-11-07 at 13 10 19

So I think we should be using `const Placeholder()`.

Issue #6071, opened 2022-04-01
Use super params in macro-generated code


goderbauer:
The `stless` and friends macros should generate code that uses the new super param feature if it is available. Super parameters are available when the lower SDK bound of the package is at least 2.17.0-0.

The `stless` macro should for example generate the following code:

```dart
class Foo extends StatelessWidget {
  const Foo({super.key});

  @override
  Widget build(BuildContext context) {
    return Container();
  }
}
```

More about super parameters: https://github.com/dart-lang/language/issues/1855
Tracking issue for the flutter framework: https://github.com/flutter/flutter/issues/100575

stevemessick:
We're going to see about moving support for this into the analyzer.

stevemessick:
We're going to see about moving support for this into the analyzer.

PR #6398, opened 2022-11-08
Use super-initializer parameter in live templates

romatallinn:
To avoid having to manually pass each parameter into the super invocation of a constructor, you can use super-initializer parameters to forward parameters to the specified or default superclass constructor. This produces cleaner classes. 

This PR thus updates live templates to use them instead.

[Language Tour | Dart -- Constructors](https://dart.dev/guides/language/language-tour#constructors)
[Language Tour | Dart -- Design](https://dart.dev/guides/language/effective-dart/design#dont-type-annotate-initializing-formals)

Addresses: #6071

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -21,3 +21,6 @@ Sander Kersten <spkersten@gmail.com>
 Pradumna Saraf <pradumnasaraf@gmail.com>
 Pedro Massango <pedromassango.developer@gmail.com>
 Wagner Silvestre <wagner1343@outlook.com>
+
+Dizconto Ltd.
+Roman Sirokov <sirokov@dizconto.com>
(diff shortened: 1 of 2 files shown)


stevemessick:
Thanks for the PR. Unfortunately, this won't work for older versions of the Dart SDK. There are many projects that have not migrated to null-safety yet, so they are still using Flutter 2.12 or so. This change requires Flutter 3.0 or newer. Also, we don't edit the XML file; it is generated (see below).

If you want to make this change, modify `FlutterLiveTemplatesProvider` to check the Flutter SDK version and return a file that works for that version. Unfortunately, getting the SDK version will be difficult, as it depends on the current `Project`. There may be multiple projects open and each could have its own SDK version. I think it can be done using the IntelliJ `WindowManager` API. (You'd need to find the top-most `JFrame` then query it for its project.) You'd also need to do some work in the `tool/plugin` plugin-management tool. It has a `generate` command that generates the live template file from the templates (`*.txt`). It would need to be extended to generate both forms.

Ideally, we'd ask the Dart analysis server to provide the live templates. It knows the SDK version so could provide the proper form. There's been some talk about adding that feature to the LSP API, but the Flutter plugin does not use LSP, so that won't actually help us. Unless we hook up the LSP version, too, which isn't on anyone's radar currently.

stevemessick:
@romatallinn Do you plan to do any more work on this PR? If not, I'll close it.

Issue #6406, opened 2022-11-24
Flutter Inspector - Not Found


politebarista:
Hi! I get the following window (see screenshot) when opening Flutter Inspector. The bug is reproduced only if the "Enable embedding DevTools in the Flutter Inspector tool window" option is selected in the framework settings in Android Studio preferences. If you remove the selection from it, the inspector starts.

<details><summary>Screenshot of Flutter inspector</summary>

![Screenshot 2022-11-24 at 15 07 44](https://user-images.githubusercontent.com/63463572/203727414-095918b6-8c47-4a72-b1fa-3e9e88de8603.png)

</details>

Yes, the flutter version is already outdated, but I don't think it's the case, since I've been using it for some time, and the problem has just appeared.

Using "Invalidate caches and restart" did not help, as did restarting the studio or the computer as a whole.

## Version info

Android Studio version - Android Studio Dolphin | 2021.3.1 Patch 1.

<details><summary>Version of Android Studio plugins</summary>

![Screenshot 2022-11-24 at 15 12 29](https://user-images.githubusercontent.com/63463572/203728355-d99ffc24-1109-421d-945e-83d545d7a2e8.png)

</details>

<details><summary>flutter doctor -v</summary>

[✓] Flutter (Channel unknown, 3.0.3, on macOS 13.0 22A380 darwin-arm, locale en-RU)
    • Flutter version 3.0.3 at /Users/user/flutter
    • Upstream repository unknown
    • Framework revision 676cefaaff (5 months ago), 2022-06-22 11:34:49 -0700
    • Engine revision ffe7b86a1e
    • Dart version 2.17.5
    • DevTools version 2.12.2

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    • Android SDK at /Users/user/Library/Android/sdk
    • Platform android-33, build-tools 33.0.0
    • ANDROID_HOME = /Users/user/Library/Android/sdk
    • ANDROID_SDK_ROOT = /Users/user/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.13+0-b1751.21-8125866)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • CocoaPods version 1.11.3

[✓] Chrome - develop for the web
[...]

PR #6614, opened 2023-02-10
Check version of SDK for forming DevTools URL

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/6406

### DIFF
--- a/flutter-idea/src/io/flutter/run/OpenDevToolsAction.java
+++ b/flutter-idea/src/io/flutter/run/OpenDevToolsAction.java
@@ -17,6 +17,7 @@ import io.flutter.ObservatoryConnector;
 import io.flutter.devtools.DevToolsUrl;
 import io.flutter.run.daemon.DevToolsService;
 import io.flutter.run.daemon.FlutterApp;
+import io.flutter.sdk.FlutterSdk;
 import io.flutter.utils.AsyncUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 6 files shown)


helin24:
CC @DanTup

flutter-idea/src/io/flutter/run/OpenDevToolsAction.java
```diff
@@ -75,8 +76,13 @@ public void actionPerformed(@NotNull final AnActionEvent event) {
 
       final String serviceUrl = myConnector != null && myConnector.getBrowserUrl() != null ? myConnector.getBrowserUrl() : null;
 
+      FlutterSdk flutterSdk = FlutterSdk.getFlutterSdk(project);
+      if (flutterSdk == null) {
+        return;
+      }
```
DanTup, line 82: Do you only support launching DevTools for For Flutter projects? In VS Code we let you use it for Dart-only projects where there might not be a Flutter SDK (so we gate the new URLs on the Dart SDK version rather than Flutter).

flutter-idea/src/io/flutter/sdk/FlutterSdkVersion.java
```diff
@@ -102,6 +102,9 @@ public class FlutterSdkVersion implements Comparable<FlutterSdkVersion> {
   @NotNull
   private static final FlutterSdkVersion MIN_STABLE_MACOS_PLATFORM = new FlutterSdkVersion("3.0.0");
 
+  @NotNull
+  private static final FlutterSdkVersion MIN_MODERN_DEVTOOLS_URL = new FlutterSdkVersion("3.3.0");
```
DanTup, line 106: It may be worth a comment here (or renaming) so it's clear what "modern" means just in case things change again in future and there's a "new modern" format 😄

(in VS Code I named it `supportsDartDevToolsPathUrls` since it's when we use the `PathUrlStrategy`).

DanTup (approved):
The logic and URLs all seem sound to me, though I'm not so familiar with this project so you may still want Steve's review too (it's possible mine won't count towards the check anyway).

stevemessick (approved):
As Danny said, "modern" is relative.

flutter-idea/src/io/flutter/run/OpenDevToolsAction.java
helin24: This is only for Flutter projects. The Dart plugin is managed by JetBrains and doesn't open the same panels

flutter-idea/src/io/flutter/sdk/FlutterSdkVersion.java
helin24: Thanks for the naming suggestion

PR #6638, opened 2023-02-24
View record fields in the debugger

stevemessick:
This also includes a patch to `BoundField` that makes it usable, but I expect that will change with the next update to VM service protocol.

@alexander-doroshko

### DIFF
--- a/flutter-idea/third_party/vmServiceDrivers/org/dartlang/vm/service/element/BoundField.java
+++ b/flutter-idea/third_party/vmServiceDrivers/org/dartlang/vm/service/element/BoundField.java
@@ -42,11 +42,17 @@ public class BoundField extends Element {
    * @return one of <code>String</code> or <code>int</code>
    */
   public Object getName() {
-    final JsonObject elem = (JsonObject)json.get("name");
+    final JsonElement elem = json.get("name");
     if (elem == null) return null;
 
-    if (elem.get("type").getAsString().equals("String")) return elem.getAsString();
-    if (elem.get("type").getAsString().equals("int")) return elem.getAsInt();
+    if (elem.isJsonObject()) return elem.getAsString();
+    if (elem.isJsonPrimitive()) {
+      try {
+        return elem.getAsInt();
+      } catch (NumberFormatException ex) {
+        return elem.getAsString(); // e.g. name is "$1"
+      }
+    }
     return null;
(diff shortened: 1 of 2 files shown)


flutter-idea/third_party/vmServiceDrivers/org/dartlang/vm/service/element/BoundField.java
```diff
@@ -42,11 +42,17 @@ public FieldRef getDecl() {
...
   public Object getName() {
-    final JsonObject elem = (JsonObject)json.get("name");
+    final JsonElement elem = json.get("name");
     if (elem == null) return null;
 
-    if (elem.get("type").getAsString().equals("String")) return elem.getAsString();
-    if (elem.get("type").getAsString().equals("int")) return elem.getAsInt();
+    if (elem.isJsonObject()) return elem.getAsString();
+    if (elem.isJsonPrimitive()) {
+      try {
```
jacob314, line 50: Nit: avoid using try-catch for non-exceptional cases where possible.

jacob314 (approved):
lgtm

stevemessick:
https://github.com/dart-lang/sdk/issues/51514

flutter-idea/third_party/vmServiceDrivers/org/dartlang/vm/service/element/BoundField.java
```diff
@@ -42,11 +42,17 @@ public FieldRef getDecl() {
    * @return one of <code>String</code> or <code>int</code>
    */
   public Object getName() {
-    final JsonObject elem = (JsonObject)json.get("name");
+    final JsonElement elem = json.get("name");
```
devoncarew, line 45: FWIW, this system of record for this file is in https://github.com/dart-lang/sdk/tree/main/pkg/vm_service/java; that lib is generated from the service protocol spec. Any changes here - which aren't upstreamed there - will be written over the next time these files are updated.

stevemessick: I've updated the VM service protocol twice recently, and expect to do so again now that the problems have been addressed.
https://github.com/dart-lang/sdk/issues/51448
https://github.com/dart-lang/sdk/issues/51514

PR #6653, opened 2023-03-02
Update VM service protocol

stevemessick:
This version does not have the bugs in previous updates so no hand editing was required.


devoncarew (approved):
👍

PR #6654, opened 2023-03-02
Delete move.yml

devoncarew:
I _believe_ this config is now defunct.


stevemessick (approved):
It hasn't worked for a long time. However, Hixie did mention that if we wanted to run our own bot they'd not be opposed in principle. It's one of many things I may look into someday. The inability to transfer issues to dart-lang/sdk is a small annoyance.

PR #6681, opened 2023-03-15
Eliminate the dependency on artifacts

stevemessick:
It should be safe to delete the `artifacts` directory after this is merged.

Use IntelliJ 2023.1 for plugin debugging.


tool/plugin/lib/plugin.dart
```diff
@@ -442,13 +442,13 @@ abstract class BuildCommand extends ProductCommand {
           spec.buildForMaster();
         }
 
-        result = await spec.artifacts.provision(
+        /*result = await spec.artifacts.provision(
```
helin24, line 445: Can this be deleted? And same for below

helin24 (approved):

tool/plugin/lib/plugin.dart
stevemessick: Oops, I forgot to clean up the code. Lots to delete.

PR #6699, opened 2023-03-31
Fix BoundField.getName()

stevemessick:
It is used elsewhere in the debugger so I can't just fix it locally.


helin24 (approved):

alexander-doroshko:
Thanks, I've added the same change in `BoundField.getName()` to the Dart plugin.

PR #6757, opened 2023-05-18
Add Short super.key 

DevBlooming:


stevemessick:
@jacob314 @anderdobo We had been holding off making this change (`super.key` constructors) because it is incompatible with older language versions. With Dart 3 out, is it safe to make this change now?

@DevBlooming assuming the answer to the above question is "yes" then please update the AUTHORS file or let me know you don't want to.

stevemessick:
@jacob314 @anderdobo The other way to ask my question is: Do we need to support these code-generating templates for Dart SDKs prior to 2.17.0-0?

stevemessick (approved):

PR #6780, opened 2023-05-31
Revert color preview patch

stevemessick:
This time it is signed. Turns out you cannot create a PR from the GitHub window, since they are not signed.


helin24 (approved):
argh signature!

flutter-idea/src/io/flutter/editor/FlutterColorProvider.java
AlexV525, line 1: And the CHANGELOG entry?

stevemessick: Thanks!

PR #6772, opened 2023-05-30
✨ Support multiple running instance for inspectors

AlexV525:
Resolves #6189.

https://github.com/flutter/flutter-intellij/assets/15884415/b5a3438b-4fbd-4fa8-9027-fd3fe0f11963

## Implementation details

- `Browser` is a singleton in the previous implementation. The PR updates it to be stored as a mapping between the tab name (the unique device name) and the browser instance.
- `Content`s are also cached to ensure they can be added after removing the empty content.

## Additional context

`FlutterPerformanceView` should be able to apply the same update but I didn't investigate it yet.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [ ] I added new tests to check the change I am making, or this PR is [test-exempt].
- [x] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

Issue #6759, opened 2023-05-18
Flutter Inspector: Layout Explorer doesn't fit the screen. Can't see width/height.


marcglasberg:
I would like to obtain the dimensions (width and height) of certain widgets. Previously, this task was easily accomplished using the old Flutter Inspector. However, with the new version, the Layout Explorer has a bug: the whole layout is not entirely visible on the screen, and there is no scrollbar. Even if I resize the screen, the layout just gets bigger and continues out of the screen.

I have attached an image: 

![LayoutInspector](https://github.com/flutter/flutter-intellij/assets/13332110/b96a76d7-be25-47e0-8c17-d7c40671a623)

Secondly, there was a Setting in a previous plugin version that allowed me to switch back to the old Flutter Inspector. I found the older tool better, mainly because it was free from bugs, and was faster to turn on. Could you kindly guide me if there is still an option in the settings to revert to this version?

Lastly, I'd like to propose the addition of a dedicated tab, preferably positioned between the Layout Explorer and the Widget Details Tree, named 'Dimensions.' It would just display the width and height, providing a simpler view for those who may not require the complex layout details. This was present in the old Flutter Inspector.

Thank you.

stevemessick:
@marcglasberg Could you attach your `flutter doctor -v` output? If your Flutter plugin version is 73.1 then you must be using an IDE built on IntelliJ 2023.1. You can use an earlier one, say 2022.3 to get the old inspector, but it is going away with the next general release (June 1). Also, are you using the new or old UI?

@helin24 Is there possibly a scaling issue in that screen shot? Seems odd there are no scroll bars.

@kenzieschmoll Marc has an enhancement request for you.

marcglasberg:
```
[√] Flutter (Channel stable, 3.7.12, on Microsoft Windows [Version 10.0.19044.2965], locale pt-BR)
    • Flutter version 3.7.12 on channel stable at C:\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 4d9e56e694 (4 weeks ago), 2023-04-17 21:47:46 -0400
    • Engine revision 1a65d409c7
    • Dart version 2.19.6
    • DevTools version 2.20.1

[√] Windows Version (Installed version of Windows is version 10 or higher)

[√] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
    • Android SDK at C:\Users\Marcelo\AppData\Local\Android\sdk                 
[...]

PR #6787, opened 2023-06-06
Add JCEF browser

helin24:
We want to add this as an option for users who are experiencing problems with the embedded DevTools window, e.g. https://github.com/flutter/flutter-intellij/issues/6759. I'm not sure if the JCEF browser will be better for cases that are causing problems in the JxBrowser implementation, but I figure we want to start gathering info on how JCEF performs anyway if we want to eventually move away from JxBrowser and its file downloads.

@AlexV525 I've reorganized the management of multiple browsers and content, something you introduced in https://github.com/flutter/flutter-intellij/pull/6772. I pointed out one spot where I'm not sure if my change will work as expected; if you have a chance, please look over/test and let me know.


flutter-idea/src/io/flutter/view/EmbeddedBrowser.java
```diff
@@ -0,0 +1,185 @@
...
+    tab.devToolsUrlFuture.complete(devToolsUrl);
+
+    JComponent component = tab.embeddedTab.getTabComponent(contentManager);
+
+    ApplicationManager.getApplication().invokeLater(() -> {
+      if (contentManager.isDisposed()) {
+        return;
+      }
+
+      contentManager.removeAllContents(false);
```
helin24, line 103: I'm not sure if `setSelectedContent` to `previousContent` from this previous PR needs to be implemented here as well: https://github.com/flutter/flutter-intellij/pull/6772/files#diff-7475425770d21bb97a1d3a0796d3cd02868371cde331681a8812191064a1f6d2R165

What are the implications of not setting selected to previous?

stevemessick:
The Dart plugin version needs to be updated in line 74 of `product-matrix.json` to 232.6734.4.

flutter-idea/src/io/flutter/view/EmbeddedBrowser.java
AlexV525: There is nothing specific here, I wrote this mainly to avoid if `Content`s have different properties between the same tab, so I was thinking it would be safer to remove the previous one then open a new one.

flutter-idea/src/io/flutter/FlutterBundle.properties
```diff
@@ -213,6 +213,8 @@ settings.allow.tests.tooltip=The directory must be marked as a sources root, suc
 settings.font.packages=Font Packages
 settings.show.all.configs=Show all possible run configurations for apps or tests, even if a created configuration already exists
 settings.show.all.configs.tooltip=If there is a defined run configuration to watch a specific test and one to just run it, show both.
+settings.jcef.browser=Use JCEF browser to show embedded DevTools.
```
AlexV525, line 216: I don't see other fields having the ending, so I'm suggesting this.

```suggestion
settings.jcef.browser=Use JCEF browser to show embedded DevTools
```

AlexV525 (approved):
RSLGTM. I've tested the option on IDEA 2023.1 and the functionality seems good.

stevemessick (approved):
LGTM after making the change @AlexV525 suggested (delete the period), and updating the Dart plugin version.

stevemessick (approved):
LGTM after making the change @AlexV525 suggested (delete the period), and updating the Dart plugin version.

stevemessick (approved):
LGTM after making the change @AlexV525 suggested (delete the period), and updating the Dart plugin version.

flutter-idea/src/io/flutter/logging/FlutterConsoleLogManager.java
```diff
@@ -264,7 +266,7 @@ private void processFlutterErrorEvent(@NotNull DiagnosticsNode diagnosticsNode)
           errorSummary = property.getDescription();
         } else if (StringUtil.equals("DevToolsDeepLinkProperty", property.getType()) &&
                 FlutterSettings.getInstance().isEnableEmbeddedBrowsers() &&
```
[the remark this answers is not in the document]
helin24: https://github.com/flutter/flutter-intellij/issues/6789

flutter-idea/src/io/flutter/view/EmbeddedBrowser.java
helin24: Okay, sounds like this should be okay then. Thanks!

PR #6800, opened 2023-06-15
Remove deprecated strong-mode entry in analysis options

parlough:
I saw the old `implicit-casts` option used to be enabled here, so I replaced it with it's more modern equivalent of the `strict-casts` [analysis language mode](https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks).


stevemessick (approved):

stevemessick:
Thanks!

Unfortunately, all commits have to be signed. See this article for more info:
https://gist.github.com/troyfontaine/18c9146295168ee9ca2b30c00bd1b41e

GitHub also has a wiki article on Signing-commits.

parlough:
I forgot about that, thanks for heads up! Should be signed now :)

stevemessick:
Update the AUTHORS file if you want to be credited. Otherwise LGTM.

PR #6845, opened 2023-07-26
Remove isEnableEmbeddedBrowsers setting

### LINKED ISSUE #6789: Remove isEnabledEmbeddedBrowsers setting
It's now set to always be true, so we can delete

stevemessick:
Fixes #6789 

And this clears the M75 milestone :)


helin24 (approved):
Thanks!

PR #6881, opened 2023-08-25
Use ToolEvent for inspect elements

### LINKED ISSUE #6875: Widget inspector doest jump to source code
Hello
After i upgraded to Flutter 3.13, widget inspector stopped jumping to source code,  i tried a new projects and still having the same issue 
___
**DevTools version**: 2.25.0
**Connected Device**:
CPU / OS: arm64 (64 bit) android
Dart Version: 3.1.0
Flutter Version: 3.13.0 / stable
Framework / Engine: efbf63d9c6 / 1ac611c64e
Connected app type: Flutter native (debug build)

helin24:
Fixes #6875

Some things I'm not sure about:
- I'm currently using the custom VM service stream 'ToolEvent' directly. VS Code is using the DAP forwarding of events from this stream; is this a point where it would make sense to use DAP from DDS?
- This skips large portions of previous architecture to listen for changes in selections, e.g. classes like `InspectorObjectGroupManager` - is that okay?

### DIFF
--- a/flutter-idea/src/io/flutter/inspector/InspectorService.java
+++ b/flutter-idea/src/io/flutter/inspector/InspectorService.java
@@ -423,33 +423,35 @@ public class InspectorService implements Disposable {
         break;
       }
       case "ToolEvent": {
-        if ("navigate".equals(event.getExtensionKind())) {
-          JsonObject json = event.getExtensionData().getJson();
+        Optional<Event> eventOrNull = Optional.ofNullable(event);
+        if ("navigate".equals(eventOrNull.map(Event::getExtensionKind).orElse(null))) {
+          JsonObject json = eventOrNull.map(Event::getExtensionData).map(ExtensionData::getJson).orElse(null);
           if (json == null) return;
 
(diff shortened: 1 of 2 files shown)


jacob314 (commented):
This looks correct. All of the complex existing code to deal with object groups, etc can be removed now that you are just directly listening for the tool event. To verify you should test whether navigating to widget creation locations still works correctly after disabling the old code path.

helin24:
> To verify you should test whether navigating to widget creation locations still works correctly after disabling the old code path.
If you mean disabling `notifySelectionChanged` upon `Debug` stream events, then this should be okay, because the old way doesn't work with newer versions of Flutter. I think because selection groups are null.

jwren (approved):

DanTup:
Nice! I think after this ships we could consider dropping the `inspect()` call from inspector, which causes some unexpected evaluations in the VS Code Debug Console (because we can't tell that the `inspect()` wasn't from the users own code).

However, I don't understand why this was necessary to fix the issue. My understanding was that the original change to add ToolEvents was backwards-compatible (because `inspect()` was still also called). Do we know why that stopping working?

PR #6944, opened 2023-10-19
Fix various performance and cleanup issues

jwren:
Java syntax changes include:
- inclusion of @NotNull
- making downcasts explict
- use of amend instead of the string `+` operator


jacob314 (approved):
lgtm

PR #6946, opened 2023-10-19
Use org.hamcrest.MatcherAssert.assertThat instead of the deprecated org.junit.Assert.assertThat calls

jwren:


jacob314 (approved):
lgtm_dog2

PR #6947, opened 2023-10-19
Change ContentFactory.SERVICE.getInstance() to ContentFactory.getInstance()  (#6948)

jwren:
See this for some additional context: https://youtrack.jetbrains.com/issue/IDEA-299177/NoSuchMethodError-when-calling-com.intellij.ui.content.ContentFactory.getInstance


jacob314 (approved):
lgtm
Thanks for taking care of this early.

PR #7059, opened 2023-11-03
Remove JX Browser usages and references

helin24:
This removes all JX Browser code from IJ, so the plugin will no longer download the files and Chromium.

Based on user reports, it seems that JCEF often fails on Linux - I'm not sure if this link will help users: https://intellij-support.jetbrains.com/hc/en-us/articles/360016421559-JCEF-the-missing-libs-problem-on-Linux. Alternatively, using a separate browser is now the recommended fallback.


jwren (approved):
Thank you @helin24!

vlad-lubenskyi:
Thank you guys for trusting JxBrowser! It's been a pleasure :))

helin24:
@vlad-lubenskyi thanks for all of your help!

PR #7065, opened 2023-11-07
Try to use project service for disposal [draft]

helin24:
I'm trying to follow the instructions from https://plugins.jetbrains.com/docs/intellij/disposers.html#choosing-a-disposable-parent for how to properly dispose of objects, because the plugin is leaking memory and I can see in the debugging memory panel that many of our project services (e.g. `FlutterSdkManager`) open a new instance every time a project is opened and doesn't go away when the project is closed. (Is that how project services should work?)

I was hoping that registering dispose actions to a parent disposable that is a project service (`FlutterProjectDisposable` introduced in this change) would make a difference, but I don't see that working.

This other change is more basic because the registered disposer isn't called on app ending: https://github.com/flutter/flutter-intellij/pull/7064. However, in the cases in this PR, I do see that the registered disposers are called on project close; it's just that the project services themselves tend to stick around.

@alexander-doroshko do you have any tips on how to properly manage these classes for disposal?

### DIFF
--- a/flutter-idea/src/io/flutter/FlutterProjectDisposable.java
+++ b/flutter-idea/src/io/flutter/FlutterProjectDisposable.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2023 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter;
+
+import com.intellij.openapi.Disposable;
+import com.intellij.openapi.project.Project;
+import io.flutter.sdk.FlutterSdkManager;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Objects;
+
+public class FlutterProjectDisposable implements Disposable {
+  private final Project project;
+  public static FlutterProjectDisposable getInstance(@NotNull Project project) {
+    return Objects.requireNonNull(project.getService(FlutterProjectDisposable.class));
+  }
+
+  private FlutterProjectDisposable(@NotNull Project project) {
+    this.project = project;
+  }
+
+  @Override
+  public void dispose() {
+    System.out.println("in dispose");
+  }
+}
--- a/flutter-idea/src/io/flutter/dart/FlutterDartAnalysisServer.java
+++ b/flutter-idea/src/io/flutter/dart/FlutterDartAnalysisServer.java
@@ -23,6 +23,7 @@ import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.Consumer;
 import com.jetbrains.lang.dart.analyzer.DartAnalysisServerService;
 import io.flutter.FlutterInitializer;
+import io.flutter.FlutterProjectDisposable;
 import io.flutter.analytics.TimeTracker;
 import io.flutter.utils.JsonUtils;
 import org.dartlang.analysis.server.protocol.*;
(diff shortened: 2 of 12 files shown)


alexander-doroshko:
@helin24 Memory leaks appear if references to unneeded objects are not lost and therefore the JVM garbage collector can't collect unneeded objects.

Disposer/Disposables sometimes help with losing unneeded references but I don't think you need to write a lot of Disposer-related magic in the Flutter plugin.

Let's look at the Disposer concept this way: it builds a tree of Disposable objects and calls child.dispose() method when Disposer.dispose(parent) is called. It means that:
- It doesn't lose references to unneeded objects per se. Effectively, it does nothing useful for you if child.dispose() method is empty
- It is designed to do cleanup inside your child.dispose() _which wouldn't happen otherwise_. You rarely need this in a plugin. Normally, just like in any JVM program, references become unreachable, GC works, that's it
- Rarely do you need to call Disposer.dispose(pluginSpecificObject). You need to understand why you need to call this, otherwise it means you probably don't need to call it
- Never call Disposer.dispose(intelliJPlatformRelatedObject) in a plugin. E.g. never call Disposer.dispose(project). The IntelllJ Platform itself cares about the objects it creates

I'll give you an example when Disposables may be needed in a plugin.

Assume the plugin calls `SomePlatformManager.getInstance(...).addPlatfromRelatedListener(new PluginSpecificListener())`. When should such a listener be removed? If it's never removed, the PluginSpecificListener class instance leaks, and all the references it has in its fields leak. GC can't collect them.

In some cases it's ok, if it's a single app-level listener, which doesn't keep references to anything project-specific. Such listeners may live forever.

If SomePlatformManager is a project-level service (its getInstance() accepts Project) and it doesn't leak, then the GC collects SomePlatformManager with all the registered listeners once the project is closed. Disposables are not needed, all just works.

If SomePlatformManager is an application-level service, it's never collected. But your PluginSpecificListener may be project-specific (has direct or indirect reference to Project in its fields). If PluginSpecificListener is not removed on project close, memory leak appears. 

To solve this problem, many addListener() methods accept parentDisposable. The implementation guarantees that once parentDisposable.dispose() is called (by Platform, not by a plugin!) the listener is removed.

So, addListener(parentDisposable=project, pluginSpecificListener) guarantees that the pluginSpecificListener is deregistered once the project is closed.

Using the Project object as a parent disposable is shown as a warning. This is because of our attempt to make the plugins dynamic. The idea is that the plugin can be installed / uninstalled / updated without closing the project.

The Flutter plugin is not dynamic. It can't be installed / uninstalled / updated without restarting the IDE. So the warnings about using Project as a parent disposable are harmless for you. However, for the sake of clean code, you may use any Flutter-specific project-level service as a parent disposable. Each plugin-specific project-level service is disposed by the Platform both on project close and on plugin unload. You never need to call Disposer.dispose(...) for plugin-specific services.

Memory leaks are hunted using memory snapshots. You find strongly-referenced object, which by design shouldn't be reachable, look for paths from this object to GC roots, and this way find the reason of the leak. That's the universal way. In many cases just accurate code reading may be enough.

The rule of thumb is that for each object you need to clearly understand, whether it is a project-level object or an application-level one. App-level objects shouldn't have references to project-level ones. The opposite is fine. Custom magic with Disposer is rarely (or not at all) needed in the Flutter plugin. Let the standard GC do its job.

helin24:
Closing this; we will be addressing memory leaks through checking references.

PR #7093, opened 2023-11-27
Remove listener for file path on project dispose

helin24:
This class was pointed out as a source of memory leaks, and it seems this is because file paths aren't removed from `listenerCache` on project close. I wanted to check with @alexander-doroshko following the conversation on https://github.com/flutter/flutter-intellij/pull/7065 that this is a reasonable fix.


jwren (approved):

alexander-doroshko:
True, static fields are a constant source of memory leaks.

We usually fight with them by keeping data in a class with the needed lifecycle. For example, in a project-level service.

As a hotfix, the proposed change should work as well.

helin24:
Okay, that makes sense. I'll merge this so that this class will get garbage collection similar to the others, and we can consider moving all of these `project` disposal instances to project level services at a later time.

PR #7095, opened 2023-11-27
Remove 2022.2.x as a supported version for Android Studio

jwren:
https://github.com/flutter/flutter-intellij/issues/7094

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [2022.2, 2022.3, 2023.1, 2023.2]
+        version: [2022.3, 2023.1, 2023.2]
     steps:
       - name: checkout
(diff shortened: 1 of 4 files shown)


helin24 (approved):

AlexV525 (approved):

AlexV525:
BTW is there a policy we can follow to deprecate supports?

jwren:
> BTW is there a policy we can follow to deprecate supports?

Good question.  We should have a few comments in `building.md` to discuss.  Something along the lines of all stable and beta versions of Ultimate, Community and Android Studio, going back to the current stable version ~1 years ago, pending any security risks or other concerns in older versions.  Do you have any additional thoughts here on what we might want to consider?

AlexV525:
> Good question. We should have a few comments in `building.md` to discuss. Something along the lines of all stable and beta versions of Ultimate, Community and Android Studio, going back to the current stable version ~1 years ago, pending any security risks or other concerns in older versions. Do you have any additional thoughts here on what we might want to consider?

Nah, +1 to your statement.

Issue #7100, opened 2023-12-01
Add `ideFeature=` query parameter to DevTools urls


kenzieschmoll:
For DevTools urls that are created and launched / embedded by VS Code, we already pass the parameter `ide=IntelliJ-IDEA` of `ide=Android-Studio`. We'd also like to pass the query parameter `ideFeature=` with the IDE feature that is responsible for opening DevTools.

- `onDebugAutomatic`: when DevTools is opened because the user has the setting enabled to automatically open DevTools when debugging an app. This naming matches what VS Code passes for its equivalent feature.
- `runConsole`: when DevTools is opened in the browser because the Dart icon in run, debug, test consoles was clicked.
- `toolWindow`: when DevTools is opened because the Inspector tool window was manually clicked or opened from the IntelliJ menu.
  - one item I'm not sure of here: can we differentiate when the user manually opened the tool window vs when the tool window was already opened and therefore automatically showed the inspector when a debug session was started?

@helin24

PR #7108, opened 2023-12-04
Ide feature

helin24:
Addresses https://github.com/flutter/flutter-intellij/issues/7100

CC @kenzieschmoll

### DIFF
--- a/flutter-idea/src/io/flutter/view/FlutterView.java
+++ b/flutter-idea/src/io/flutter/view/FlutterView.java
@@ -76,7 +76,7 @@ import java.awt.*;
 import java.util.List;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 @com.intellij.openapi.components.State(
   name = "FlutterView",
@@ -245,7 +245,7 @@ public class FlutterView implements PersistentStateComponent<FlutterViewState>, 
                                               @Nullable InspectorService inspectorService,
                                               ToolWindow toolWindow,
                                               boolean isEmbedded,
-                                              boolean openOnAppLaunch,
+                                              DevToolsIdeFeature ideFeature,
                                               DevToolsInstance devToolsInstance) {
     assert(SwingUtilities.isEventDispatchThread());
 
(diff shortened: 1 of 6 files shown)


flutter-idea/src/io/flutter/view/FlutterView.java
```diff
@@ -275,7 +277,8 @@ private void addBrowserInspectorViewContent(FlutterApp app,
         color,
         UIUtil.getFontSize(UIUtil.FontSize.NORMAL),
         flutterSdkVersion,
-        WorkspaceCache.getInstance(app.getProject())
+        WorkspaceCache.getInstance(app.getProject()),
+        openOnAppLaunch ? DevToolsIdeFeature.ON_DEBUG_AUTOMATIC : DevToolsIdeFeature.TOOL_WINDOW
```
kenzieschmoll, line 281: is it possible that a user can have the openOnAppLaunch setting on, and also manually open the Inspector tool window?

helin24: Hmm yeah a user could open their project, and then before they run their app open the inspector tool window. I can change how we're setting `openOnAppLaunch` to only be true if the tool window was automatically opened.

helin24:
> Open question: can we differentiate when the user manually opened the tool window vs when the tool window was already opened (not automatically opened due to a setting) and therefore automatically showed the inspector when a debug session was started?

The last commit distinguishes these two, but to be clear, let's make sure these examples would be labeled as expected:
1. A user starts with tool window closed but with setting "auto open on app run" true. When they run an app, the tool window opens without their input and DevTools appears --> `ON_DEBUG_AUTOMATIC`
2. The same user stops running their app, but the tool window remains open. They run their app again, and DevTools appears --> `TOOL_WINDOW`
3. The same user stops running their app and closes the tool window. They run their app again, and the tool window opens automatically, DevTools appears --> `ON_DEBUG_AUTOMATIC`

Is that what we're hoping for? Case 2 seems a little odd to me because the tool window doesn't close when an app stops running (should it? probably a bigger change though)

kenzieschmoll:
Case 2 seems fine to me because this scenario could happen for a user regardless of whether they have the on_debug_automatic setting enabled or not.

This makes me think though, should we further distinguish TOOL_WINDOW to be TOOL_WINDOW_RELOAD and TOOL_WINDOW_OPEN ? Which would distinguish when the tool window loads devtools because it was already open vs when the tool window loads devtools because the user clicked on the tool window tab.

helin24:
I think it would be possible to distinguish `TOOL_WINDOW_RELOAD` and `TOOL_WINDOW_OPEN`, but I'm not sure if this is useful for comparing with VS Code - @kenzieschmoll I'll leave it up to you.

helin24:
This will now distinguish the reload and open cases

PR #7214, opened 2024-02-12
Revert JX Browser removal with IDE feature passed along

helin24:
@jwren This now builds and passes tests, though I'll wait to try it with the key before merging.

PR #7215, opened 2024-02-12
Revert JX Browser but ignore IDE feature flag

helin24:
@jwren this is my attempt at doing the same as #7214  but without trying to thread through IDE feature.


helin24:
Want to include feature flags instead

PR #7232, opened 2024-02-20
Make JX Browser the fallback and remove settings option

helin24:
This removes the option to select JCEF or JX Browser in settings for simplicity. We will instead check if JCEF is available first, and if not, download and use JX Browser. Embedded browser failures in IJ (where JCEF is included by default) were very low, so I think it's unlikely users will benefit from having a choice here.

I still need to build this to check whether this is detecting the presence of JCEF accurately.

### DIFF
--- a/flutter-idea/src/io/flutter/FlutterUtils.java
+++ b/flutter-idea/src/io/flutter/FlutterUtils.java
@@ -633,11 +633,10 @@ public class FlutterUtils {
       return null;
     }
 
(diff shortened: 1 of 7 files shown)


helin24:
This works in my testing. However, I also don't recall what JDK with JCEF I was using in the past - currently my version of Android Studio when I select this one:
Screenshot 2024-02-20 at 4 02 45 PM

I get this error: https://youtrack.jetbrains.com/issue/IDEA-317401

And the browser doesn't even appear:
Screenshot 2024-02-20 at 4 03 10 PM

I think this is largely a non-issue since we won't be recommending Android Studio users try to use a JCEF JDK for now. If people hit this issue we can recommend changing back to their default JDK.

jwren (approved):

Issue #7228, opened 2024-02-20
Can't not see the varable's information at breakpoint 


luckysmg:
## Steps to Reproduce

Setup a breakPoint and see the info in var. But everytime will raise error in log console:
```
Cannot invoke "com.google.gson.JsonElement.isJsonPrimitive()" because "elem" is null
```

Here is the img:
<img src="https://github.com/flutter/flutter-intellij/assets/49340347/452d635c-352c-42d2-bc45-231a5bc57e05" width=700 />

## Version info

```
[!] Flutter (Channel stable/v3.3.8.xianyu, 3.3.9-0.0.pre.210, on macOS 14.1.2
    23B92 darwin-x64, locale zh-Hans-CN)
    • Flutter version 3.3.9-0.0.pre.210 on channel stable/v3.3.8.xianyu at
      /Users/wjr/flutter
    ! Upstream repository git@gitlab.alibaba-inc.com:ali-flutter/flutter.git is
      not a standard remote
    • Framework revision a1f0c4291a (6 weeks ago), 2024-01-12 15:32:44 +0800
    • Engine revision 7bc1415569
    • Dart version 2.18.4
    • DevTools version 2.15.0
    • Flutter download mirror http://aliuc-flutter-storage.alibaba.net/uc

[!] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
    • Android SDK at /Users/wjr/Library/Android/sdk
    • Platform android-33, build-tools 33.0.2
    • ANDROID_HOME = /Users/wjr/Library/Android/sdk
    • ANDROID_SDK_ROOT = /Users/wjr/Library/Android/sdk
    • Java binary at: /Library/Internet
      Plug-Ins/JavaAppletPlugin.plugin/Contents/Home/bin/java
    • Java version Java(TM) SE Runtime Environment (build 1.8.0_291-b10)
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/macos#android-setup for
      more details.

[✓] Xcode - develop for iOS and macOS (Xcode 15.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15C500b
    • CocoaPods version 1.14.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[!] Android Studio (version 2023.1)
[...]

PR #7254, opened 2024-03-05
Deprecation of BoundField.getDecl() with NPE check (#7228)

jwren:


alexander-doroshko:
`getDecl()` was used for 'jump to source' feature (from the Variables view in the Debugger tool window), as well as for icon customization. If the VM still provides this info, maybe keep using it?

jwren:
@alexander-doroshko  Yes, this was my first approach, it was pointed out to me that it is deprecated however, so we shouldn't use it going forward.

jacob314 (approved):
lgtm

alexander-doroshko:
> it is deprecated however, so we shouldn't use it going forward

And looks like there's no replacement, is there? It was a nice feature 😢

jacob314:
Fyi @bkonyi for thoughts on whether there is a replacement or should be.

bkonyi:
The `name` and `value` properties in [`BoundField`](https://github.com/dart-lang/sdk/blob/main/runtime/vm/service/service.md#boundfield) should be used instead.

PR #7264, opened 2024-03-11
Use Dart plugin's DevTools instance with DTD

helin24:
DTD and DevTools are now started on project open by the Dart plugin, so it's better for the Flutter plugin to open this existing DevTools service instead of starting its own.

Using the Dart plugin's DevTools applies for any Dart SDK version 2.15+, but the earlier versions of this will not include a link to DTD (only 3.4+)


helin24:
CC @kenzieschmoll @CoderDake

jwren (approved):
LGTM

PR #7287, opened 2024-03-19
Remove some obsolete documentation from docs/building.md

jwren:


helin24:
I have used gsutil also for uploading JX Browser files, and they're at `gs://flutter_infra_release/flutter/intellij/jxbrowser/`. I'm not sure if we should keep the gsutil link for installing it?

jwren:
After chatting with @helin24 , I am closing this issue.  The documentation needs to be updated, not removed.

Issue #6595, opened 2023-01-31
New UI incompatibilities


stevemessick:
The new UI does not show the Project view by default. Our project creation code forces it open. That's probably not great.
Edit: This is fine. New projects are likely going to require interacting with the Project view. Creating a new Java project also opens the Project view.

stevemessick:
The Dart Analysis tool window is completely hidden. You have to use a menu to open it. We should at least make sure its icon shows up in the border.

karvulf:
How did you open it in the menu? @stevemessick

stevemessick:
It's on the `More Tool Windows` (three dots) button.

<img width="342" alt="Screenshot 2023-04-19 at 8 46 35 AM" src="https://user-images.githubusercontent.com/8518285/233130169-f6a62e4b-a8b8-444c-83bd-0cde6d7dedf8.png">

karvulf:
<img width="393" alt="Bildschirm­foto 2023-04-20 um 07 30 12" src="https://user-images.githubusercontent.com/36191072/233267002-f3334448-712b-4512-bcc7-a8ad7bc49096.png">
Interesting, this option is not visible for me. Do you use IntelliJ for Mac or Windows? @stevemessick

stevemessick:
My screen shot was from Android Studio 2022.3 on a Mac. It looks like you are using IntelliJ 2023.1. My `More Tool Windows` does have some Flutter and Dart items in 2023.1, though. Are you sure you have the Dart and Flutter plugins installed correctly? Also check the menu bar: `View > Tool Windows` for Dart and Flutter items.

karvulf:
Ok I have no idea why, but the Analysis Server appeared on the menu button. Finally I can test the new UI, thank you for your help @stevemessick ! :)

c5inco:
Some public bugs filed on Android Studio that relate to the Flutter plugin in the New UI:
1. Buttons like Hot Reload missing from main toolbar by default: https://issuetracker.google.com/294776109
2. Active process indicator on Run button causes layout issues: https://issuetracker.google.com/295119370

For 1) this can be remedied in the plugin by adding the actions to the new toolbar groups in the New UI: Main Toolbar Left/Center/Right. Likely the right considering the previous design.

jwren:
Release 76 is coming out without any attention here, bumping to 77 for the time being.

gnprice:
[...]

PR #7307, opened 2024-04-01
Add panel for DevTools deep links

helin24:
TODOs:
- New icon?
- Move testing over? (testing this logic is pretty annoying)

Screenshot 2024-04-01 at 3 18 19 PM

### DIFF
--- a/flutter-idea/src/io/flutter/deeplinks/DeepLinksViewFactory.java
+++ b/flutter-idea/src/io/flutter/deeplinks/DeepLinksViewFactory.java
@@ -36,14 +36,11 @@ public class DeepLinksViewFactory implements ToolWindowFactory {
             return;
           }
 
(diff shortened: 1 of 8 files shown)


helin24:
@kenzieschmoll let me know if this looks reasonable. Do you have an icon that would be reasonable here? Something that's a style like notifications or device manager would be good.

jwren:
LGTM. Yes, icon for this page as well as others should be updated but we can do this in a future PR (https://github.com/flutter/flutter-intellij/issues/6595)

helin24:
Okay we can land this for now and I'll work on moving over testing and getting rid of similar logic in `FlutterView`.

kenzieschmoll:
Looks good! Agree with Jaime, icon can be addressed in a follow up.

Issue #7323, opened 2024-04-04
Deep links window does not load when there is a running app


kenzieschmoll:
The window opens as expected when there is no app connection.

Repro steps:
1. run a flutter app
2. try to open the deep links tool window
3. Observe "Nothing to show" message
![Screenshot 2024-04-04 at 8 32 51 AM](https://github.com/flutter/flutter-intellij/assets/43759233/07aeaa41-19bc-405e-b91e-7f11d32e6457)

What is odd is that it does load properly as a Tab in the Flutter Inspector view, but I have filed a separate issue about removing the duplicate location for this tool (https://github.com/flutter/flutter-intellij/issues/7320)

![Screenshot 2024-04-04 at 8 31 51 AM](https://github.com/flutter/flutter-intellij/assets/43759233/419ac85f-9d28-4c59-9d33-9761717ea1dc)

@helin24

Issue #7320, opened 2024-04-04
Deep links tab appears in two places


kenzieschmoll:
The deep links page has its own tool window:
![Screenshot 2024-04-04 at 8 21 47 AM](https://github.com/flutter/flutter-intellij/assets/43759233/3b8c5bd5-c50f-49fc-8456-4a449fbe381c)

but it also shows up as a tab in the inspector, which we probably don't want:
![Screenshot 2024-04-04 at 8 18 43 AM](https://github.com/flutter/flutter-intellij/assets/43759233/5fd5a95f-0741-4e07-95d5-30597ffdafb4)

@helin24

helin24:
Fixed by #7325

PR #7325, opened 2024-04-04
Add separate browser windows and save content manager

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/7323

There's some code in embedded browser to manage showing multiple tabs in one window for the situation where a user is running multiple apps and wants to see the inspector for each of those apps, but it was also pushing the deep links tab into the inspector window because I hadn't differentiated windows.

I also added specification of content manager for each tab. I don't know if this is truly necessary given that the content manager is only used when first displaying the content, but this seems more correct.

Uploading deep_links_and_inspector.mp4…


helin24:
Also fixes https://github.com/flutter/flutter-intellij/issues/7320

Issue #6844, opened 2023-07-26
NotificationGroup is scheduled for removal


stevemessick:
Use `<notificationGroup>` extension point to register notification groups and `com.intellij.notification.NotificationGroupManager`.

PR #7337, opened 2024-04-09
Convert hot reload notifications to not use deprecated methods

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/6844


flutter-idea/src/io/flutter/run/FlutterReloadManager.java
```diff
@@ -64,6 +59,7 @@
 import io.flutter.settings.FlutterSettings;
 import io.flutter.utils.FlutterModuleUtils;
 import io.flutter.utils.MostlySilentColoredProcessHandler;
+import org.assertj.core.util.Strings;
```
jwren, line 62: Is this a String utility?  Where is it used below?  Consider org.apache.commons.lang3.StringUtils which we use elsewhere already?

jwren (approved):
Thanks @helin24 !

Issue #7330, opened 2024-04-05
Flutter IntelliJ IDEA 2024.1 -- `ActionUpdateThread.OLD_EDT` is deprecated


chardinson:
## What happened

I don't know

## Version information

IntelliJ IDEA `2024.1` • Flutter plugin `io.flutter 78.5.1` • Dart plugin `241.14768`

Error getting Flutter sdk information.

## Exception

null

```
com.intellij.diagnostic.PluginException: `ActionUpdateThread.OLD_EDT` is deprecated and going to be removed soon. 'io.flutter.actions.ExtractWidgetAction' must override `getActionUpdateThread` and chose EDT or BGT. See ActionUpdateThread javadoc. [Plugin: io.flutter]
	at com.intellij.diagnostic.PluginProblemReporterImpl.createPluginExceptionByClass(PluginProblemReporterImpl.java:23)
	at com.intellij.diagnostic.PluginException.createByClass(PluginException.java:90)
	at com.intellij.diagnostic.PluginException.reportDeprecatedUsage(PluginException.java:125)
	at com.intellij.openapi.actionSystem.ActionUpdateThreadAware.getActionUpdateThread(ActionUpdateThreadAware.java:21)
	at com.intellij.openapi.actionSystem.AnAction.getActionUpdateThread(AnAction.java:199)
```

jacob314:
Use of a deprecated API that needs to be fixed soon so that important ExtractWidgetAction continues to work.

helin24:
We're not using the deprecated code directly in ourplugin. I see this [issue in IDEA](https://youtrack.jetbrains.com/issue/IDEA-344189/com.intellij.diagnostic.PluginException-ActionUpdateThread.OLDEDT-is-deprecated-and-going-to-be-removed-soon-after-restoring) that should be fixed now, but perhaps will remain for a while in 2024.1.

It seems like this is an issue within an IJ class somewhere in the inheritance chain of `DefaultActionGroup`, and should be fixed there rather than us needing to supply a non-deprecated thread type everywhere that we extend the class.

j-fbriere:
Same problem for me.

IntelliJ IDEA `2024.1` • Flutter plugin `io.flutter 78.5.1` • Dart plugin `241.15845`

Flutter 3.19.5 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 300451adae (10 days ago) • 2024-03-27 21:54:07 -0500
Engine • revision e76c956498
[...]

PR #7356, opened 2024-04-17
Declare ActionUpdateThread as EDT for RunFlutterAction

### LINKED ISSUE #7331: Exception in plugin Flutter
Thanks for the feedback! If your issue is related to the Flutter framework itself,
please open an issue at
[github.com/flutter/flutter](https://github.com/flutter/flutter/issues/new).

## Steps to Reproduce

_Please tell us what you were doing and what went wrong_

## Version info

Please paste the output of running `flutter doctor -v` here (available from the command
line or from `Tools > Flutter > Flutter Doctor`). It will provide the version of the
Flutter framework as well as of the IntelliJ plugin.

bc-lee:
From Intellij 2024.1, any actions that doesn't override `getActionUpdateThread` triggers a warning. This is because the default behavior `OLD_EDT` will be removed in the future. To suppress the warning, we need to declare this action as EDT, as it's modifying the presentation in `update()`.

Fixes #7331

## Pre-launch Checklist

- [x] I signed the [CLA].
- [ ] All existing and new tests are passing.

### DIFF
--- a/flutter-idea/src/io/flutter/actions/RunFlutterAction.java
+++ b/flutter-idea/src/io/flutter/actions/RunFlutterAction.java
@@ -13,6 +13,7 @@ import com.intellij.execution.RunnerAndConfigurationSettings;
 import com.intellij.execution.configurations.RunConfiguration;
 import com.intellij.execution.runners.ExecutionEnvironment;
 import com.intellij.execution.runners.ExecutionEnvironmentBuilder;
+import com.intellij.openapi.actionSystem.ActionUpdateThread;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.project.Project;
@@ -124,6 +125,13 @@ public abstract class RunFlutterAction extends AnAction {
     final RunConfiguration config = settings == null ? null : settings.getConfiguration();
     // TODO(pq): add support for Bazel.
     return config instanceof SdkRunConfig && LaunchState.getRunningAppProcess((SdkRunConfig)config) == null;
+  }
+
+  @Override
+  public @NotNull ActionUpdateThread getActionUpdateThread() {
+    // Since we're modifying presentation in update(), we need to run on the EDT, not BGT
+    // See https://plugins.jetbrains.com/docs/intellij/general-threading-rules.html
+    return ActionUpdateThread.EDT;
   }
 
   @Nullable


helin24:
@bc-lee thanks for looking into this. I'm not sure this will fix the full problem, as we have other classes that have shown this error as well. See my comment here: https://github.com/flutter/flutter-intellij/issues/7330#issuecomment-2043523014. 

Potentially we could implement the same thing you're doing here in other places, but I'm also wondering if this problem should be fixed in the IJ platform. Do you have any insight into this?

bc-lee:
The warning generated by Intellij 2024.1 is intentional. JetBrains is pushing all actions to override the `getActionUpdateThread` to ensure that plugin developers specify the thread on which the action should be executed. (See https://youtrack.jetbrains.com/issue/IJPL-143730/Add-assertion-for-OLDEDT-actions and https://github.com/JetBrains/intellij-community/commit/a45d0161609b0ea036252fcc87c10d1bd186f22e for more details) As such, it's necessary for us to address this warning for each of our actions.

This PR is currently in draft because there are several additional actions to fix. I will request a review of this PR once all the warnings have been addressed.

helin24:
@bc-lee Ah thanks so much for clarifying; I didn't realize IJ intended for us to start overriding the function. In that case, yes, supplying overriding threads for all of the affected classes makes sense.

Is this going to be an issue for all classes extending `AnAction`? (20 classes currently)

I'm also uncertain about how to decide EDT vs BGT here. The IJ doc [here](https://plugins.jetbrains.com/docs/intellij/basic-action-system.html#principal-implementation-overrides) suggests that all actions should change `Presentation` on update, but has some other notes on EDT vs. BGT. In our case, since this class uses `Project` during `update`, does that mean this should instead be in a BGT?

bc-lee:
From my search result, we have 85 classes including anonymous classes that extend `AnAction`. See below image for the search result:
Screenshot 2024-04-18 at 08 49 21

Maybe it is not easy to fix them all at once.

helin24:
Yeah, I think it would be reasonable to do one at a time / a few at a time. 

My main outstanding question is whether this change should be EDT or BGT, as I'm not sure that modifying Presentation counts as changing UI "directly".

bc-lee:
It turns out that in most cases, the `getActionUpdateThread` should return `BGT`.

[Official Document](https://plugins.jetbrains.com/docs/intellij/basic-action-system.html#principal-implementation-overrides)
> The preferred method is to run the update on the BGT, which has the advantage of guaranteeing application-wide read access to PSI, the virtual file system (VFS), or project models.

[Jetbrains Platform Slack](https://jetbrains-platform.slack.com/archives/C5U8BM1MK/p1708707369234089?thread_ts=1708694125.773809&cid=C5U8BM1MK)
> Use BGT always and only if you need some data from UI switch to EDT

However, I cannot judge easily if each case is okay to run on the BGT or not. It appears that many of Intellij API's are okay to use both EDT and BGT and if there is a specific requirement, that API has annotation like [@RequiresEdt](https://github.com/JetBrains/intellij-community/blob/b9c3764ff8c5c63edff9184ac76c3d95f47924c6/platform/core-api/src/com/intellij/util/concurrency/annotations/RequiresEdt.java) or `@RequiresBackgroundThread`. Also, the DevKit Plugin can check if the code is running on the correct thread.

I'm closing this PR for now as I cannot give a clear answer to this problem, but if someone can give me some advice on how to handle this, I would appreciate it.

alexander-doroshko:
Hi from JetBrains,

`RunFlutterAction` should be updated in `BGT` because it doesn't manipulate Swing components.

Blind guess is that pretty much all actions in the Flutter plugin should be updated in `BGT`.

helin24:
@alexander-doroshko thanks for the advice!

@bc-lee feel free to reopen this PR. Are you set up to test this?

PR #7362, opened 2024-04-17
Add the new verify checks to the CHECK_BOT

jwren:


jwren:
Closing PR, the verify Gradle tasks currently have mostly redundant processes as compared to the `YYYY.{1,2,3}` builds, especially for the additional amount of time required.

PR #7363, opened 2024-04-17
Fix the Kokoro bot by having it build instead of test (since we don't have an IC in our product matrix.)

jwren:


helin24:
Sorry what is an IC target?

jwren:
> Sorry what is an IC target?

IntelliJ type, see https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#intellij-extension-type.  Kokoro was/is broken due to the fact that the AS version lookup is a miss for the test execution. Looking at the history of this repo it looks like the test does work on the Kokoro bot when we are building a version for Community (`IC`).

helin24 (approved):

PR #7417, opened 2024-05-09
Use embed=one for DevTools URL

### LINKED ISSUE #7403: Migrate existing usages of `embed=true` query parameter for DevTools screens to `embedMode=one`
As of https://github.com/flutter/devtools/pull/7710, which was rolled into the Dart SDK in https://dart-review.googlesource.com/c/sdk/+/365620, DevTools expects a new format for setting embedded mode. We will still need to use the legacy query parameter for Dart SDK versions before https://dart-review.googlesource.com/c/sdk/+/365620.

There are now two embedded modes that can be set through a query parameter: `embedMode=one` (for embedding a single screen, which is the current behavior) or `embedMode=many` (for embedding many screens, which is newly supported behavior added for https://github.com/flutter/flutter-intellij/issues/7195).

The change was made to be backwards compatible with `embed=true`, but we should migrate to the new format so that we can remove the legacy handlers.

helin24:
Fixes #7403

### DIFF
--- a/flutter-idea/src/io/flutter/devtools/DevToolsUrl.java
+++ b/flutter-idea/src/io/flutter/devtools/DevToolsUrl.java
@@ -29,6 +29,7 @@ public class DevToolsUrl {
   private final FlutterSdkUtil sdkUtil;
 
   private final boolean canUseDevToolsPathUrl;
+  private final boolean canUseEmbedOne;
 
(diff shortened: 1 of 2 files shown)


jwren (approved):

helin24:
CC @kenzieschmoll I'm gating on `3.23.0-0.1.pre` which has Dart 3.5.0.

Issue #7398, opened 2024-05-02
Remove the Flutter outline 


kenzieschmoll:
The IntelliJ Structure view provides this functionality so we do not need to maintain our own.

helin24:
The main difference I see after looking at a few files is having some "wrap with..." buttons here:
![Image](https://github.com/flutter/flutter-intellij/assets/6379305/ab99f53b-8c04-42ab-a220-cf4db2f47d86)

I don't see this functionality available elsewhere, but are we okay with not having this any more?

CC @jwren

jwren:
Yes, this issue was opened when we thought there was no difference in functionality between the Structure view and the Flutter Outline.  I suspect that with time functionality like this might come from DevTools and DTD, but for the time being, we should not remove it.

helin24:
Okay, thanks for clarifying. For future reference, another thing that I found that is part of the flutter outline is this inline editing panel: https://github.com/flutter/flutter-intellij/pull/4160

kenzieschmoll:
> I don't see this functionality available elsewhere, but are we okay with not having this any more?

Aren't all these functions available as quick fixes?

kenzieschmoll:
Also, the outline view doesn't seem to load for me. We should add analytics to this view to see if many users are using it:
![Screenshot 2024-05-09 at 3 31 11 PM](https://github.com/flutter/flutter-intellij/assets/43759233/061f2f9b-b4a4-4a0c-8589-bafc3dc30ffc)

kenzieschmoll:
> Also, the outline view doesn't seem to load for me

Ah, looks like this view is configured to only be enabled when there is a running app, which is a bug. The outline view + Hot UI for that matter should work with now connected app.

helin24:
We do have data on the usage of the outline, last week numbers below:
- 1,000 unique users
- 8,200 usages of jump to source from the preview
- 2,000 usages of any of the shortcut actions
- 3,500 users who have "enabled hot UI" in settings, which allows them to access the inline editing panel. I don't know how many are actually using the panel.

I didn't realize we could open the quick fix menu from the editor - knowing that makes me less reluctant to remove the outline view.

I'm able to see the outline view for some classes in DevTools, e.g. `lib/src/screens/inspector/inspector_tree_controller.dart` without having an app running, so I'm not sure what's going wrong in your situation. I think at first when I opened a hello flutter app, it wasn't working though, so I believe it's not consistent.

[...]

PR #7418, opened 2024-05-09
Remove outline view (draft)

helin24:
Addresses https://github.com/flutter/flutter-intellij/issues/7398, but we may skip doing for now.


helin24:
Closing for now. May consider later on but not at the moment.

PR #7426, opened 2024-05-14
Fix background color

### LINKED ISSUE #7319: DevTools pages do not load in the correct theme
Screenshot 2024-04-04 at 8 06 55 AM

Screenshot 2024-04-04 at 8 11 55 AM

@helin24 are we properly passing the ide theme query parameters? 
CC @DanTup I thought you fixed a similar issue in the VS Code extension but I couldn't find a link. Can you share what the URL looks like on the VS Code side?

helin24:
Fixes #7319 

This became a larger change than I originally intended, but I figured refactoring the `DevToolsUrl` class would be helpful for upcoming work on integrating more DevTools panels.

Summary of changes:
- Add param `&theme=[light/dark]` to fix the original issue
- Use builder for making DevToolsUrl class
- Use utility class for finding the IDE's background color, light/dark mode, and font size
- Only send theming information when browser is embedded


jwren (approved):

jwren:
Looks great!

FYI -- @kenzieschmoll

Issue #2932, opened 2018-12-08
consider checking for flutter framework updates


devoncarew:
We should consider checking for flutter framework updates as part of the IntelliJ 'check for updates' action.

pq:
FWIW I did some poking around (and confirmed w/ Alex) and there doesn't seem to be a way to easily hook into the update check.  The best angle is probably to do our own update check the way we do in the Dart Plugin.

Popping up, I do think this would be very handy since so many of our exchanges with folks begin with "could you update you SDK and try again?"  I'll look a little deeper.

@devoncarew feel free to re-assign if you're already tackling. 👍

devoncarew:
Cool, thanks for looking into it. I think android studio has custom behavior here, but they may replace the action completely (which isn't something we're likely to want to do).

jwren:
With recent changes and going forward, the Flutter IJ plugin will be required to work with recent SDK versions, see recent changelog for details.

PR #7463, opened 2024-05-28
Remove the SDK minimum validation notification in the Flutter Plugin

jwren:
Race conditions where the sdk hasn't been cached resulting in the notification provides more of a distraction than value. (The Flutter SDK being tested here is pre-`0.0.12`.)

This work should be replaced with something such as https://github.com/flutter/flutter-intellij/issues/2932


helin24 (approved):

helin24:
@itsjustkevin this is another instance where it would be good to start explicitly deprecating support for old versions of Flutter

PR #7472, opened 2024-05-29
Add default unknown value if getting IDE value fails

helin24:


jwren (approved):

flutter-idea/src/io/flutter/devtools/DevToolsUrl.java
```diff
@@ -160,7 +160,8 @@ else if (flutterSdkVersion != null) {
   public String getUrlString() {
     final List<String> params = new ArrayList<>();
 
-    params.add("ide=" + sdkUtil.getFlutterHostEnvValue());
+    String ideValue = sdkUtil.getFlutterHostEnvValue();
+    params.add("ide=" + (ideValue == null ? "IntelliJPluginUnknown" : sdkUtil.getFlutterHostEnvValue()));
```
kenzieschmoll, line 164: can we use `ideValue` instead of `sdkUtil.getFlutterHostEnvValue()` on this line?

PR #7495, opened 2024-06-07
Add code for getting consent for unified analytics

helin24:
I'm leaving this turned off for now as it will show users a notification asking if they consent. Turn on after PR for trying to send existing analytics through DTD with unified analytics.

### DIFF
--- a/flutter-idea/src/io/flutter/analytics/UnifiedAnalytics.java
+++ b/flutter-idea/src/io/flutter/analytics/UnifiedAnalytics.java
@@ -7,32 +7,34 @@ package io.flutter.analytics;
 
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
-import com.intellij.ide.BrowserUtil;
 import com.intellij.notification.Notification;
 import com.intellij.notification.NotificationType;
 import com.intellij.notification.Notifications;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
 import com.jetbrains.lang.dart.ide.toolingDaemon.DartToolingDaemonConsumer;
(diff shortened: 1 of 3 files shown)


helin24:
CC @kenzieschmoll @bwilkerson

jwren (approved):

flutter-idea/src/io/flutter/analytics/UnifiedAnalytics.java
```diff
@@ -0,0 +1,232 @@
...
+          finalResult.complete(false);
+        }
+      });
+      Notifications.Bus.notify(notification, project);
+      System.out.println("Ran notification");
+    });
+    return finalResult;
+  }
+
+  private String getToolName() throws Exception {
```
kenzieschmoll, line 169: could be beneficial to add a link to the unified_analytics package library where these values are defined so that it is clear where the strings came from.

flutter-idea/src/io/flutter/analytics/UnifiedAnalytics.java
```diff
@@ -0,0 +1,232 @@
...
+  @NotNull final FlutterSdkUtil flutterSdkUtil;
+
+
+  public UnifiedAnalytics(@NotNull Project project) {
+    this.project = project;
+    this.dtdService = DartToolingDaemonService.getInstance(project);
+    this.flutterSdkUtil = new FlutterSdkUtil();
+  }
+
+  public void manageConsent() {
```
kenzieschmoll, line 45: Does this code play well with the existing analysis server integration with unified_analytics?

helin24: It sounds like once IntelliJ is marked as consented in the analytics file, then analysis server will be able to use this? I'll have to check with @bwilkerson whether this works once we turn this on, but that will have to wait until after I return.

flutter-idea/src/io/flutter/analytics/UnifiedAnalytics.java
helin24: Definitely, I'll add this in a next PR

PR #7569, opened 2024-07-18
Update icons

helin24:
New UI dark mode:
Screenshot 2024-07-18 at 10 03 51 AM

New UI light mode:
Screenshot 2024-07-18 at 10 03 38 AM

Old UI: 
Screenshot 2024-07-18 at 10 04 35 AM

Fixes https://github.com/flutter/flutter-intellij/issues/6595


jwren (approved):

c5inco:
Hello! I'm the UX Design lead on Android Studio. Just saw this icons change - nice work! I would though advise making some tweaks to the icons to better match the New UI icon style for the tool windows, as the line stroke is different between 20x20 versions and 16x16 versions. You can find more guidance on this in the Figma Community file: https://www.figma.com/community/file/1227729570033544559

deakjahn:
@helin24 If we might ask, is there a specific reason why all tool windows use the same generic Flutter logo? Some overlay giving a hint of what each does would certainly go a long way if anybody happens to have more than one present in a project. Yes, I'm aware of the beta labels but beta or not, they don't look very useful, given the space constraints.

I would even say that the opposite would be better, having some usual, immediately recognizable icon for Perfomance or Outline or whatever, then adding a smaller Flutter overlay to it rather than the opposite. We all happen to know that we're working on a Flutter project, so there's no real need to convey that info so prominently in the icons. What we want to know is which one to click on when we want a specific functionality. Just for an example, reusing the AS Profile icon for Flutter Performance would be just perfect in my book, with a smaller Flutter distinction added.

helin24:
> @helin24 If we might ask, is there a specific reason why all tool windows use the same generic Flutter logo? Some overlay giving a hint of what each does would certainly go a long way if anybody happens to have more than one present in a project. Yes, I'm aware of the beta labels but beta or not, they don't look very useful, given the space constraints.

This PR is part of work underway to replace the generic Flutter logo - for v81, the icons will probably look similar to the screenshot in this PR description. The remaining Flutter logos for performance and outline is because we are planning to remove or redo those two tool windows in later releases.

helin24:
@c5inco these are updated following your guidance for dark/light, normal/compact mode: https://github.com/flutter/flutter-intellij/pull/7574

deakjahn:
OK, thanks, I'm looking forward to it.

PR #7581, opened 2024-07-29
Make performance tab warning text red

helin24:
Follow up to https://github.com/flutter/flutter-intellij/pull/7577

Screenshot 2024-07-29 at 10 45 47 AM


kenzieschmoll:
@jwren may know if there is an accessible error / warning color that is used elsewhere in the IntelliJ plugin that we could re-use here (the current red may be a bit jarring), but generally LGTM.

kenzieschmoll (approved):

helin24:
I updated to be a less obnoxious shade of red, which we use in light mode for the bars in the performance page: 
image

Screenshot 2024-07-29 at 11 41 36 AM

This one seems reasonable for both light and dark mode so I'll go with this color.

jwren:
Change LGTM, this is temporary code.  The difference in the color red is fine with me as the point is to draw attention to it.

jwren (approved):

Issue #7626, opened 2024-08-22
Do not reload Tool Windows when theme is changed


kenzieschmoll:
Performing a full reload of the DevTools tool window when the IDE theme is changed is disruptive to the user's workflow and may cause them to lose any debugging data they were looking at before the theme change.

Now that we have DTD to communicate over, we should send an event when the IDE theme changes that DevTools can listen for. DevTools can then update itself for the new IDE theme values.

helin24:
There may be some situations where DTD isn't available, e.g. internal usages. Do you have an opinion on whether we should continue to update theme with a refresh or ignore theme changes in those cases? (or something else)

kenzieschmoll:
If DTD is not available, I recommend we do not update the theme. Having a mismatched theme is a better result than the user potentially losing state in their tool if we force refresh IMO.

helin24:
With the latest changes to DevTools this issue is now resolved

PR #7638, opened 2024-09-05
Remove automatic refresh of inspector window on theme change and check during manual refresh

helin24:
This change:
- Removes an automatic refresh of the inspector tool window when theming changes
- Adds check for theming during manual refresh of all of the DevTools windows


jwren (approved):

helin24:
Partially addresses https://github.com/flutter/flutter-intellij/issues/7626 CC @kenzieschmoll

Issue #7621, opened 2024-08-21
Tool window icons should be monochrome when selected


kenzieschmoll:
![Screenshot 2024-08-21 at 4 46 41 PM](https://github.com/user-attachments/assets/82cde343-12cf-4ee8-9a5d-3ab29c9ac48e)

The Flutter icon is not visible when the tool window is selected because it blends into the blue background. When selected we should use a version of the icon that is monochrome.

c5inco:
See this comment: https://github.com/flutter/flutter-intellij/pull/7574#issuecomment-2311515115

PR #7640, opened 2024-09-06
Use standard icon colors so that selected icons are white

helin24:
I changed the colors of the Flutter insets within the SVG files. The light icons did not change much, because the blue used before was pretty close to the standard blue.

The dark versions' Flutter insets were previously an aqua shade, and I'm not sure if it would be better to change to teal or blue. For now I've chosen teal because it seemed a little brighter, but neither seems ideal. For reference, the colors are coming from https://www.figma.com/design/jYJmUxEUdob9FxsCxfRTDb/Int-UI-Kit-(Community) and we can use the colors that are identified as being for icons.

| Original | Blue | Teal |
|---------|------|----|
|Screenshot 2024-09-06 at 12 59 41 PM|Screenshot 2024-09-06 at 1 10 43 PM|Screenshot 2024-09-06 at 1 00 52 PM|
|Screenshot 2024-09-06 at 1 05 11 PM|Screenshot 2024-09-06 at 1 09 08 PM|Screenshot 2024-09-06 at 1 05 40 PM

Fixes https://github.com/flutter/flutter-intellij/issues/7621


kenzieschmoll:
I think the Blue looks more like Flutter colors than the Teal.

helin24:
Okay, I'll switch to blue then. Plus it's a little weird to be using blue for light mode and then teal for dark mode.

jwren (approved):

PR #7647, opened 2024-09-11
Send theme changed event over DTD

helin24:
Addresses https://github.com/flutter/flutter-intellij/issues/7626

### DIFF
--- a/flutter-idea/src/io/flutter/FlutterInitializer.java
+++ b/flutter-idea/src/io/flutter/FlutterInitializer.java
@@ -7,6 +7,7 @@ package io.flutter;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
 import com.intellij.ProjectTopics;
 import com.intellij.ide.BrowserUtil;
 import com.intellij.ide.plugins.IdeaPluginDescriptor;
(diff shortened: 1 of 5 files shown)


flutter-idea/src/io/flutter/FlutterInitializer.java
```diff
@@ -255,6 +276,83 @@ private void setUpDtdAnalytics(Project project) {
...
+      if (lastScheduledThemeChangeTime.get() != requestTime) {
+        System.out.println("this request is being dropped:  " + requestTime);
+        System.out.println("last scheduled time is: " + lastScheduledThemeChangeTime.get());
+        // A more recent request has been set, so drop this request.
+        return;
+      }
+
+      System.out.println("this request is being sent:  " + requestTime);
+      final JsonObject params = new JsonObject();
+      params.addProperty("eventKind", "themeChanged");
```
helin24, line 313: @DanTup do these fields follow the same conventions as editor events that you're sending from VS Code?

DanTup: At a glance, this looks the same to me :)

My code is here:

https://github.com/Dart-Code/Dart-Code/blob/1dfe247fb5a4d0fbed25283d80067c5ce77ee286/src/shared/services/tooling_daemon.ts#L190-L198

helin24: Okay great - mainly I was concerned because I couldn't tell from your code whether "themeChanged" should be uppercase etc.

DanTup: Ah, that comes from the enum (and your casing is correct):

https://github.com/Dart-Code/Dart-Code/blob/1dfe247fb5a4d0fbed25283d80067c5ce77ee286/src/shared/services/tooling_daemon_services.ts#L150

helin24: thanks!

kenzieschmoll (approved):

PR #7650, opened 2024-09-13
Add notification when SDK is unsupported

helin24:
Analogous to https://github.com/Dart-Code/Dart-Code/pull/5257


jwren:
@helin24 Thanks for this, land it as-is and I'll make sure that the correct version is set before it rolls out to users.

jwren (approved):

PR #7676, opened 2024-10-02
Add a way for developers of DevTools to use Flutter plugin with locally served DevTools

helin24:
This offers a way to use a development version of DevTools (whatever `devtools_tool serve` runs) while using a standard Flutter plugin. This should be helpful when testing DevTools-IJ interactions.

Typically, the Dart plugin starts DTD and then also starts DevTools, passing in the DTD URL. The Flutter plugin will then use this instance of DevTools to access embedded tool windows. When using this locally served DevTools, the Dart plugin behaves the same way, but the Flutter plugin starts its own instance of local DevTools using the same DTD URL. (Note that there could be some side effects from having two DevTools instances started, connected to the same DTD instance.)

Instructions for using this:
1. Set the registry key to your DevTools directory: Go to Help > Find action > Registry > Find "flutter.local.devtools.dir" and set to your DevTools directory. (e.g. `Users/helinx/Documents/devtools`). If you want to add additional args, put these in the option "flutter.local.devtools.args".
2. Restart IntelliJ
3. You should be able to open Help > Show log in finder > Open idea.log; in here with the registry key set as above, then you should see output like "DevTools startup: <various messages that come from running `devtools_tool serve`>"
4. To turn off using local DevTools, go back to the registry key and clear out the setting.

CC @kenzieschmoll @elliette

### DIFF
--- a/flutter-idea/src/io/flutter/run/daemon/DevToolsService.java
+++ b/flutter-idea/src/io/flutter/run/daemon/DevToolsService.java
@@ -5,7 +5,7 @@
  */
 package io.flutter.run.daemon;
 
-import  com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableList;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonSyntaxException;
--- a/flutter-idea/src/io/flutter/run/daemon/DevToolsService.java
+++ b/flutter-idea/src/io/flutter/run/daemon/DevToolsService.java
@@ -39,6 +39,9 @@ import org.jetbrains.annotations.NotNull;
 
 import java.io.File;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


flutter-idea/src/io/flutter/run/daemon/DevToolsService.java
```diff
@@ -119,12 +123,38 @@ private void startServer() {
...
       if (dartDevToolsSupported) {
+        // This condition means we can use `dart devtools` to start.
         final WorkspaceCache workspaceCache = WorkspaceCache.getInstance(project);
         if (workspaceCache.isBazel()) {
+          // This is only for internal usages.
           setUpWithDart(createCommand(workspaceCache.get().getRoot().getPath(), workspaceCache.get().getDevToolsScript(),
                                       ImmutableList.of("--machine")));
         }
         else {
+          final String localDevToolsDir = Registry.stringValue(LOCAL_DEVTOOLS);
```
kenzieschmoll, line 134: would it also be possible to specify arguments that should be passed to the `devtools_tool serve` command? For example, VS Code allows a list of "args" to be specified where we can add any additional flags that need to be passed (like `--wasm`) for example.

helin24: Sure, I think I could add a separate registry field for those

kenzieschmoll (commented):
This is fantastic, Helin! Thanks for adding.

kenzieschmoll (approved):

PR #7710, opened 2024-10-15
Continued removal of deprecated or dead code

jwren:
Continuation of https://github.com/flutter/flutter-intellij/pull/7706

PR #7711, opened 2024-10-15
Continued removal of deprecated or dead code

jwren:
This PR removes the EditorMouseEventService.java and EditorPositionService.java project services

Continuation of https://github.com/flutter/flutter-intellij/pull/7710 and https://github.com/flutter/flutter-intellij/pull/7706


kenzieschmoll (commented):
Will these services be necessary for the property editor? We can always add them back if so, but we'll need functionality to get the widget at the current cursor position so that DevTools can ask the analysis server for the editable properties of this widget.

jwren:
Services yes, but not these.  These mouse event listeners/handlers are compute the widget in context using the Flutter Outline computed information which is not as robust as the upcoming DTD & DevTools work.

kenzieschmoll (approved):

PR #7713, opened 2024-10-15
Gracefully handle perf tool window not being created

parlough:
A workaround for #7691, not accessing the tool window if it wasn't created due to `isApplicableAsync` being false for the SDK version. Does it in this listener so if the SDK version changes, the event still can be handled properly.


jwren:
LGTM, thank you @parlough

jwren (approved):

Issue #7670, opened 2024-09-28
IntelliJ Platform Gradle Plugin (2.x) migration / Flutter plugin IDE (IntelliJ IDEA 2024.3 EAP)


yasergh:
Thanks for the feedback! If your issue is related to the Flutter framework itself,
please open an issue at
[github.com/flutter/flutter](https://github.com/flutter/flutter/issues/new).

## Steps to Reproduce

_Please tell us what you were doing and what went wrong_

## Version info

Please paste the output of running `flutter doctor -v` here (available from the command
line or from `Tools > Flutter > Flutter Doctor`). It will provide the version of the
Flutter framework as well as of the IntelliJ plugin.

jwren:
I started to look at this, but versions of the currently required Kotlin plugin aren't being pulled down.

```
{
      "channel": "stable",
      "comments": "IntelliJ 2024.3",
      "name": "2024.3",
      "version": "2024.3",
      "isUnitTestTarget": "false",
      "ideaProduct": "idea",
      "ideaVersion": "243-EAP-SNAPSHOT",
      "baseVersion": "243.16718.32",
      "androidPluginVersion": "243.16718.32",
      "dartPluginVersion": "243.16718.30",
      "sinceBuild": "243.16718.30",
      "untilBuild": "243.*"
    }
```

```Cannot find builtin plugin 'Kotlin' for IDE: /Users/jwren/.gradle/caches/modules-2/files-2.1/...```

jwren:
In order to continue to roll to new platform versions, starting with `2024.3`, a migration to IntelliJ Platform Gradle Plugin (2.x) migration is required:
- https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-migration.html
- https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-faq.html
- https://github.com/JetBrains/intellij-platform-gradle-plugin/releases
- https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin.html

jwren:
Currently blocked by not being able to add dependencies into the Flutter IJ plugin, see issue opened here on the Gradle IJ project team: https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/1810.

jwren:
Closing out as finished, however some cleanup activities will continue in this space.

ThomasDevApps:
[...]

PR #7715, opened 2024-10-16
Remove a modification for the idea 2023.2 version of the platform

jwren:


flutter-studio/build.gradle.kts
```diff
@@ -52,15 +52,12 @@ intellij {
...
   // see https://github.com/flutter/flutter-intellij/issues/7145
-  if(ide == "android-studio") {
+  if (ide == "android-studio") {
     pluginList.add("org.jetbrains.android")
   } else if (ide == "ideaIC") {
     pluginList.add("org.jetbrains.android:$androidVersion")
   }
 
-  if (ideVersion != "2023.2") {
-    pluginList.add(smaliPlugin)
```
parlough, line 62: So is this plugin not needed anymore when running in Android Studio?

jwren: This logic was used when building the Flutter plugin for the `2023.2` IDEA version of the product, at this point however we won't be pushing out a version of the Flutter plugin for `2023.2`: https://github.com/flutter/flutter-intellij/blob/07610c1301ce8792975d10274266abc97e4b2699/product-matrix.json

kenzieschmoll: This logic looks like we wanted the `smaliPlugin` for all versions that were **not** 2023.2, including newer versions. So does that mean we should be changing this from
```
if (ideVersion != "2023.2") {
  pluginList.add(smaliPlugin);
}
```
to
```
pluginList.add(smaliPlugin);
```

jwren:
Closing as all the `build.gradle.kts` files are being revisited due to https://github.com/flutter/flutter-intellij/issues/7670

PR #7760, opened 2024-11-04
Remove GA Analytics

jwren:


kenzieschmoll (commented):
Do we really want to remove all this? I thought the plan was to replace legacy analytics with unified analytics events. Removing all the analytics event send sites would require that we add them all back with unified analytics instead of just swapping the logic to send through unified analytics package instead of through the legacy GA.

jwren:
Acknowledged and we can discuss more if you want.  The counter argument though is that the Analytics API was designed with a GA-first mind set, and especially as time went on, the APIs were not called consistently, nor everywhere where we want them called from.  I imagine that this PR will be used when the better analytics goes in, but now we'll have a clean history to improve code health and ultimately a better analytics story for the plugin.

kenzieschmoll:
Ack. Do we still show the unified analytics prompt coming from the analysis server after this change? CC @bwilkerson As long as we are not removing that logic, then it's probably fine to remove all this.

jwren:
The logic to notify the IDEs in a generic way for analytics goes through the DAS (non-flutter) protocol via the Dart plugin for IntelliJ.  Relevant links:

- https://github.com/JetBrains/intellij-plugins/commit/7295e00684e2b12bc6ed8f43f9977f87d03cdf9e#diff-b2fa6867f6081548120831a6768ae898e285bd94ed032f3fc9a5713ab524251fR2176 & https://github.com/JetBrains/intellij-plugins/pull/888
- https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#request_server.setClientCapabilities

kenzieschmoll (approved):

jwren:
Note: don't commit yet, the Flutter preference page has an issue with this PR, it needs to be fixed before this can land

Issue #7778, opened 2024-11-11
Remove the Flutter Outline


jwren:
The Flutter Outline view provides two pieces of now redundant information and real estate in the Flutter IJ plugin:

1. Outline of the Flutter code -- this can be found in the Structure view
2. Buttons to quickly refactor widgets -- these can all be found via the `Alt+Enter` keyboard shortcut in the code editor

PR #7779, opened 2024-11-11
Comment out the usage of the Flutter Outline view

jwren:
This resolves https://github.com/flutter/flutter-intellij/issues/7778

### DIFF
--- a/resources/META-INF/plugin_template.xml
+++ b/resources/META-INF/plugin_template.xml
@@ -327,8 +327,8 @@
     <codeInsight.lineMarkerProvider language="Dart" implementationClass="io.flutter.editor.FlutterIconLineMarkerProvider"/>
     <errorHandler implementation="io.flutter.FlutterErrorReportSubmitter"/>
 
-    <toolWindow id="Flutter Outline" anchor="right" icon="FlutterIcons.FlutterDefault"
-                factoryClass="io.flutter.preview.PreviewViewFactory"/>
+<!--    <toolWindow id="Flutter Outline" anchor="right" icon="FlutterIcons.FlutterDefault"-->
+<!--                factoryClass="io.flutter.preview.PreviewViewFactory"/>-->
     <projectService serviceImplementation="io.flutter.preview.PreviewView" overrides="false"/>
 
     <toolWindow id="Flutter Inspector" anchor="right" icon="FlutterIcons.DevToolsInspector"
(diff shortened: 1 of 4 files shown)


bwilkerson (approved):

jwren:
Note: Don't commit this until all `82.x` point releases are done

resources/META-INF/plugin_template.xml
```diff
@@ -327,8 +327,8 @@
     <codeInsight.lineMarkerProvider language="Dart" implementationClass="io.flutter.editor.FlutterIconLineMarkerProvider"/>
     <errorHandler implementation="io.flutter.FlutterErrorReportSubmitter"/>
 
-    <toolWindow id="Flutter Outline" anchor="right" icon="FlutterIcons.FlutterDefault"
-                factoryClass="io.flutter.preview.PreviewViewFactory"/>
+<!--    <toolWindow id="Flutter Outline" anchor="right" icon="FlutterIcons.FlutterDefault"-->
+<!--                factoryClass="io.flutter.preview.PreviewViewFactory"/>-->
```
kenzieschmoll, line 331: how much work is it to commend out the tool window content instead so that a message can be displayed instructing users to open the Structure view?

jwren: It would be removing the current JComponent which holds the outline and buttons with a message and link to go use the provided Structure view.

We decided to remove the outline, this PR, but I didn't know if we decided on having a few simply point devs to a different window was a net-positive.

kenzieschmoll: IMO it would be less jarring for users to have an explanation for why this tool window disappeared than it would be for the tool window to be completely gone. At least for one release cycle would be my preference. @anderdobo what do you think?

Options:
1) completely remove the flutter outline view tool window in IntelliJ / Android Studio
2) remove the outline view content, but leave the tool window. Replace the existing outline content with a message that instructs users to open the IntelliJ structure view to see the tree structure of their code, and to use alt + enter to view a list of available refactorings for a widget.

anderdobo: I agree with #2. We should remove the outline view but have messaging in the tool window that guides developers.

jwren:
@kenzieschmoll @anderdobo 

Updated and included information from https://www.jetbrains.com/guide/java/tips/context-actions/, see image:

Screenshot 2024-11-18 at 2 54 52 PM

PR #7789, opened 2024-11-15
Removal of all gnu.trove.* deps from the Flutter plugin

jwren:
These are deprecated references caught by the `./bin/plugin verify` tool

### DIFF
--- a/flutter-idea/src/io/flutter/perf/FlutterWidgetPerf.java
+++ b/flutter-idea/src/io/flutter/perf/FlutterWidgetPerf.java
@@ -5,6 +5,7 @@
  */
 package io.flutter.perf;
 
+import com.android.tools.idea.io.netty.util.collection.IntObjectHashMap;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.HashMultimap;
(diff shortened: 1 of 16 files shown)


flutter-idea/src/io/flutter/perf/FlutterWidgetPerf.java
```diff
@@ -5,6 +5,7 @@
  */
 package io.flutter.perf;
 
+import com.android.tools.idea.io.netty.util.collection.IntObjectHashMap;
```
parlough, line 8: I'm guessing this import and class (here and elsewhere) might not be available in IntelliJ if the android plugin isn't available. Consider using the fastutil `Int2ObjectMap`. I believe that should be available in the base platform.

jwren: `Int2ObjectMap` is abstract. Follow-on work could resolve this and not depend on com.android.tools.idea.io.netty.util.collection`.

parlough (approved):

bwilkerson (approved):

PR #7807, opened 2024-11-20
Avoid internal AS collections

parlough:
The `Int2ObjectOpenHashMap` implementation is the most common implementation I found across the IntelliJ platform, so went with that for its general usage characteristics.


jwren (approved):
Thanks @parlough

Issue #7830, opened 2024-11-27
Migrate instances of EditorNotifications.Provider to the new API (-> interface of EditorNotificationProvider)


jwren:
From the verify cli command:

```
      #Deprecated class com.intellij.ui.EditorNotifications.Provider reference
          Deprecated class com.intellij.ui.EditorNotifications.Provider is referenced in io.flutter.editor.NativeEditorNotificationProvider.<init>(Project)
          Deprecated class com.intellij.ui.EditorNotifications.Provider is referenced in io.flutter.samples.FlutterSampleNotificationProvider
          Deprecated class com.intellij.ui.EditorNotifications.Provider is referenced in io.flutter.inspections.SdkConfigurationNotificationProvider
          Deprecated class com.intellij.ui.EditorNotifications.Provider is referenced in io.flutter.editor.NativeEditorNotificationProvider
          Deprecated class com.intellij.ui.EditorNotifications.Provider is referenced in io.flutter.editor.FlutterPubspecNotificationProvider.<init>(Project)
          Deprecated class com.intellij.ui.EditorNotifications.Provider is referenced in io.flutter.inspections.SdkConfigurationNotificationProvider.<init>(Project)
          Deprecated class com.intellij.ui.EditorNotifications.Provider is referenced in io.flutter.editor.FlutterPubspecNotificationProvider
          Deprecated class com.intellij.ui.EditorNotifications.Provider is referenced in io.flutter.samples.FlutterSampleNotificationProvider.<init>(Project)
```

PR #7842, opened 2024-12-03
Migrate the FlutterSampleNotificationProvider to the new EditorNotificationProvider API (#7840)

jwren:
This is progress on https://github.com/flutter/flutter-intellij/issues/7830


flutter-idea/src/io/flutter/samples/FlutterSampleNotificationProvider.java
```diff
@@ -6,87 +6,98 @@
...
 
     // Run the code to query the document in a read action.
     final List<FlutterSample> samples = ApplicationManager.getApplication().
       runReadAction((Computable<List<FlutterSample>>)() -> {
-        //noinspection CodeBlock2Expr
         return getSamplesFromDoc(flutterPackagePath, document, filePath);
       });
 
-    return samples.isEmpty() ? null : new FlutterSampleActionsPanel(samples);
+    if (samples != null && !samples.isEmpty()) {
```
pq, line 94: Can `samples` be null here or can you skip this check? (It looks like `getSamplesFromDoc` at worst returns an empty list?)

flutter-idea/src/io/flutter/samples/FlutterSampleNotificationProvider.java
```diff
@@ -6,87 +6,98 @@
...
+      return new FlutterSampleActionsPanel(samples);
+    }
+    return null;
   }
 
-  private List<FlutterSample> getSamplesFromDoc(String flutterPackagePath, Document document, String filePath) {
+  private List<FlutterSample> getSamplesFromDoc(@NotNull String flutterPackagePath, @NotNull Document document, @NotNull String filePath) {
```
pq, line 100: Can you mark this `@NotNull`?

pq (approved):

flutter-idea/src/io/flutter/samples/FlutterSampleNotificationProvider.java
jwren: Done.

flutter-idea/src/io/flutter/samples/FlutterSampleNotificationProvider.java
pq: Ah! So maybe this is new Java. I would have expected `ApplicationManager.getApplication(). runReadAction` to throw an NPE if `getApplication` returns null, but cool!

Issue #7843, opened 2024-12-03
New simple IJ inspector for the `analysis_options.yaml` file


jwren:
New inspector idea, on `analysis_options.yaml` files, have a banner (that is easy to dismiss), to point users to documentation at https://dart.dev/tools/analysis

jwren:
PR closed, putting a pin in this until we make a decision on how to proceed.

PR #7858, opened 2024-12-10
New simple IJ inspector for the analysis_options.yaml file

jwren:
Resolves this issue, https://github.com/flutter/flutter-intellij/issues/7843 (if we want to include it).


jwren:
Closing, putting a pin in this until we make a decision on how to proceed.

Issue #7810, opened 2024-11-22
Test with coverage disabled


Bogala:
Hello,

I can't launch Tests with Coverage since my last plugin update : 
- the option is not available
- when I add manually shortcut tu Run "Tests in test" with coverage, the button is disabled
![image](https://github.com/user-attachments/assets/732c3efb-0922-4662-89d4-35b2dfb35b26)

## Steps to Reproduce

I have tried :
- File | Invalidate Caches | Invalidate and Restart
- steps suggested by File | Repair IDE
- Manually re-generate project configuration settings
- Reinstall plugins
- Reinstall IDE (IntelliJ IDEA Ultimate
- With other IDE (Android Studio)

### Flutter Doctor

Flutter (Channel stable, 3.24.5, on macOS 15.1.1 24B91 darwin-arm64, locale en-FR)
    • Flutter version 3.24.5 on channel stable
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision dec2ee5c1f (), 2024-11-13 11:13:06 -0800
    • Engine revision a18df97ca5
    • Dart version 3.5.4
    • DevTools version 2.37.3

Xcode - develop for iOS and macOS (Xcode 16.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 16B40
    • CocoaPods version 1.15.2

IntelliJ IDEA Ultimate Edition (version 2024.3)
    • IntelliJ at /Users/bfontaine/Applications/IntelliJ IDEA Ultimate.app
    • Flutter plugin version 82.2.4
    • Dart plugin version 243.22562.3

### IDE Configuration

IntelliJ IDEA 2024.3 (Ultimate Edition)
Build #IU-243.21565.193, built on November 13, 2024
Licensed to Benoit FONTAINE
Subscription is active until December 6, 2024.
Runtime version: 21.0.5+8-b631.16 aarch64 (JCEF 122.1.9)
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Toolkit: sun.lwawt.macosx.LWCToolkit
macOS 15.1.1
GC: G1 Young Generation, G1 Concurrent GC, G1 Old Generation
Memory: 4096M
Cores: 10
Metal Rendering is ON
Registry:
  suggest.all.run.configurations.from.context=true
  ide.experimental.ui=true
  i18n.locale=
  terminal.new.ui=true
Non-Bundled Plugins:
  com.intellij.apacheConfig (243.21565.208)
[...]

PR #7904, opened 2025-01-21
7810 test coverage regression fix

jwren:
Fix the test coverage regression, this resolves https://github.com/flutter/flutter-intellij/issues/7810.


bwilkerson (approved):
There are several file here that appear to be new, but they have a copyright of 2021. That looks wrong to me, but if it conforms to the norm for this repo then no problem. Just mentioning in case it was a copy/paste error.

PR #7924, opened 2025-02-04
Add a flutter pub add action

jakemac53:
Opening in case we are interesting in actually upstreaming this - I mostly just did it to get familiar with how actions are set up and plumbed through.

Using the action opens a dialog like this:

Screenshot 2025-02-04 at 1 16 26 PM

It does do some validation that a non-empty package name was input, but that is it today:

Screenshot 2025-02-04 at 1 16 59 PM

And it does ensure you actually clicked "OK" before actually running any commands.


jwren (approved):
LGTM

Revert out the `gradle.properties` changes.  Also, you'll have to run `./bin/plugin generate` to get the github presubmit action to pass.

flutter-idea/src/io/flutter/sdk/FlutterCommand.java
```diff
@@ -272,6 +272,7 @@ enum Type {
     CONFIG("Flutter config", "config"),
     CREATE("Flutter create", "create"),
     DOCTOR("Flutter doctor", "doctor", "--verbose"),
+    PUB_Add("Flutter pub add", "pub", "add"),
```
jakemac53, line 275: I was a bit confused why this wasn't all caps so idk if that matters or if you know where its coming from

jakemac53:
> Revert out the `gradle.properties` changes. Also, you'll have to run `./bin/plugin generate` to get the github presubmit action to pass.

Done 👍

flutter-idea/src/io/flutter/sdk/FlutterCommand.java
jakemac53: Oh I just realized I typed this manually LOL, updating to match the others.

PR #7928, opened 2025-02-04
ignore gradle.properties and .idea/misc.xml to reduce friction creating prs

jakemac53:
These hold configuration based on the current SDKs you are building with, and **probably** don't make sense to commit :).


jwren (approved):

jakemac53:
It looks like some of the presubmits need these files to exist, do you know if there is a command or something that can be used to generate them?

jwren (approved):

PR #7931, opened 2025-02-05
Replace deprecated ComboBoxWithBrowserButton

keertip:
As per https://plugins.jetbrains.com/docs/intellij/built-in-button.html#browse


jwren (approved):
Thanks @keertip This PR LGTM, we need to test that it runs through the verifier successfully and doesn't have any regressions in the built runtime of the product before landing.

PR #7933, opened 2025-02-05
Match IntelliJ's zoom setting in the embedded DevTools panels

elliette:
Work towards https://github.com/flutter/devtools/issues/8243

Screenshot 2025-02-05 at 3 13 22 PM

After a user updates the zoom in their settings, the next time the DevTools panels are loaded, the new zoom setting will be applied. They can also trigger an immediate refresh of the embedded panel to get the new zoom setting.


flutter-idea/src/io/flutter/view/EmbeddedJcefBrowser.java
```diff
@@ -33,6 +33,11 @@ public void close() {
 
   }
 
+  @Override
+  public void matchIdeZoom() {
+
```
jwren, line 38: Is this intentionally empty?

jwren (approved):
LGTM! Thank you @elliette!

flutter-idea/src/io/flutter/view/EmbeddedJcefBrowser.java
elliette: Chatted offline, this is an experimental browser, does not need implementation here.

Issue #7947, opened 2025-02-20
Making Android dependencies optional


itviewer:
eg, if I only want to develop desktop applications or iOS applications

Issue #7205, opened 2024-02-06
Cannot install Flutter plugin for Rider


MeltdownInteractive:
## Steps to Reproduce

Install the Flutter plugin in Rider 2023.3.3 from 
https://plugins.jetbrains.com/plugin/9212-flutter/versions/stable

When installing the plugin, an error message is shown in Rider, and the Flutter plugin is disabled.
_Requires plugin 'org.jetbrains.android' to be installed_

There is no further information about which plugin this refers to, and when asked to install the dependencies, nothing happens. 
There is no option to enable the selected plugin. 

The only 'Android' plugin I can find for Jetbrains is this one, yet it is not compatible with Rider.
https://plugins.jetbrains.com/plugin/22989-android

## Version info
JetBrains Rider 2023.3.3
Build #RD-233.14015.60, built on January 19, 2024
Licensed to Greg Quinn
Subscription is active until February 12, 2024.
Runtime version: 17.0.9+7-b1087.11 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Windows 11.0
.NET Core v7.0.10 x64 (Server GC)
GC: G1 Young Generation, G1 Old Generation
Memory: 4084M
Cores: 32
Registry:
  editor.config.csharp.support=true
  ide.experimental.ui=true
Non-Bundled Plugins:
  com.intellij.resharper.StructuredLogging (2023.3.2.321)
  Dart (233.13763.5)
  com.intellij.resharper.azure (3.50.0.1755-2023.3)

rickyoleary:
This works, [as per this suggestion](https://github.com/flutter/flutter-intellij/issues/6809). I just tried it, strange it's not supported by default.

Alex-Grasley:
The suggestion in @rickyoleary's comment worked for me too. This would be great is we can get official support for this as using a single IDE for front end and back end development would be amazing.

Issue #6809, opened 2023-06-29
Allowing the plugin on other IDEs such as JetBrains Rider


waseemw:
I've been doing this for months now, all I have to do is:
1. Download the flutter plugin from https://plugins.jetbrains.com/plugin/9212-flutter
2. Use 7zip to open it and modify the *-idea-*.jar file
3. Open the META-INF/plugin.xml inside
4. Add `optional="true"` to `<depends>org.jetbrains.android</depends>`
5. End result would be `<depends optional="true">org.jetbrains.android</depends>`
6. Close and click 'yes' on the 7zip prompt to modify the original zip file.
7. In JetBrains Rider, install the plugin from disk and chose the zip file

I'm wondering why is it being limited to Android Studio and Intellij IDEA if it works just fine (including debugging) on JetBrains Rider for example?

stevemessick:
We have always supported IntelliJ and Android Studio. If the community would like to add support for other platforms, we'd be happy to review a PR, but it would need to be supported by the original authors.

MeltdownInteractive:
> We have always supported IntelliJ and Android Studio. If the community would like to add support for other platforms, we'd be happy to review a PR, but it would need to be supported by the original authors.

@stevemessick 
Can you just not change the builds so the `optional ="true"` fix mentioned above is in the plugin.xml, so we don't need to go in and edit the XML each time we update the plugin?

stevemessick:
@MeltdownInteractive I'm no longer involved with this project.

PR #7949, opened 2025-02-21
⚙️ Flip `org.jetbrains.android` to optional

AlexV525:
> [!TIP]
> To try out the plugin on other IDEs, head over to https://github.com/AlexV525/flutter-intellij/releases and download the latest release!

Resolves https://github.com/flutter/flutter-intellij/issues/6809
Resolves https://github.com/flutter/flutter-intellij/issues/7205
Resolves https://github.com/flutter/flutter-intellij/issues/7947

Tested with PyCharm 2024.3.2

9c500df58e0a9cb95686c54b8d545cbf

## Pre-launch Checklist

- [x] I signed the [CLA].

[Contributor Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md
[breaking change policy]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md

### DIFF
--- a/resources/META-INF/android-contribs.xml
+++ b/resources/META-INF/android-contribs.xml
@@ -1,0 +1,3 @@
+<!-- Defines Android-supported IDE-specific contributions and implementations. -->
+<idea-plugin>
+</idea-plugin>
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,7 +13,7 @@ ChangJoo Park <pcjpcj2@gmail.com>
 Amol Grover <frextrite@gmail.com>
 Jeremy Judeaux <jeremy.judeaux@volune.net>
 Letticia Nicoli <letticia.nicoli@gmail.com>
-Alex Li <google@alexv525.com>
+Alex Li <alexv.525.li@gmail.com>
 Marcus Tomlinson <themarcustomlinson@gmail.com>
 Juyeong Lee <juyeonglee0413@gmail.com>
 Hrishikesh Kadam <hrkadam.92@gmail.com>
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -29,7 +29,6 @@
   <depends>com.intellij.modules.lang</depends>
   <depends>com.intellij.modules.xdebugger</depends>
   <depends>org.jetbrains.plugins.yaml</depends>
-  <depends>org.jetbrains.android</depends>
   <depends>Dart</depends>
   <depends>Git4Idea</depends>
 
@@ -42,6 +41,9 @@
 
   <!-- Contributes Gemini Code Assist extensions. -->
   <depends optional="true" config-file="gemini-contribs.xml">com.google.tools.ij.aiplugin</depends>
+
+  <!-- Contributes Android-supported features and implementations. -->
+  <depends optional="true" config-file="android-contribs.xml">org.jetbrains.android</depends>
 
   <!-- Contributes Android Studio-specific features and implementations. -->
   <!--suppress PluginXmlValidity -->
--- a/resources/META-INF/plugin_template.xml
+++ b/resources/META-INF/plugin_template.xml
@@ -27,7 +27,6 @@
   <depends>com.intellij.modules.lang</depends>
   <depends>com.intellij.modules.xdebugger</depends>
   <depends>org.jetbrains.plugins.yaml</depends>
-  <depends>org.jetbrains.android</depends>
   <depends>Dart</depends>
   <depends>Git4Idea</depends>
 
@@ -40,6 +39,9 @@
 
   <!-- Contributes Gemini Code Assist extensions. -->
   <depends optional="true" config-file="gemini-contribs.xml">com.google.tools.ij.aiplugin</depends>
+
+  <!-- Contributes Android-supported features and implementations. -->
+  <depends optional="true" config-file="android-contribs.xml">org.jetbrains.android</depends>
 
   <!-- Contributes Android Studio-specific features and implementations. -->
   <!--suppress PluginXmlValidity -->


CaiJingLong:
image

WebStorm ready.

xkeyC:
👍 That's a great update, which allows us to use JetBrains Community Edition IDEs for coding, as not everyone needs Android support (they can try Android Studio for that).

AlexV525:
Turns out the built artifacts on Windows will fail with specific projects but not related.

README.md
```diff
@@ -3,8 +3,9 @@
 [![Latest plugin version](https://img.shields.io/jetbrains/plugin/v/9212)](https://plugins.jetbrains.com/plugin/9212-flutter)
 [![Build Status](https://github.com/flutter/flutter-intellij/workflows/presubmit/badge.svg)](https://github.com/flutter/flutter-intellij/actions?query=branch%3Amaster+workflow%3Apresubmit)
 
-An IntelliJ plugin for [Flutter](https://flutter.dev/) development. Flutter is a multi-platform
-app SDK to help developers and designers build modern apps for iOS, Android and the web.
+An IntelliJ plugin for [Flutter](https://flutter.dev/) development.
+Flutter is a multi-platformFlutter is Google's SDK for crafting beautiful,
```
mossmana, line 7: Nit: Did you mean to remove "Flutter is a multi-platform"? Right now there's no space between platform and Flutter.

The updates to the README look good other than the above nit. However, I might split out the changes to the README into a separate PR as they seem unrelated to the the changes to make org.jetbrains.android optional and could be merged separately.

AlexV525: Thanks for the catch. I was trying to copy the latest description from the Flutter repo but seems not make it clean. I'll split changes to another PR later.

AlexV525: Done revert

AlexV525:
@jwren

pq:
This is cool. Since it raises a kind of product question I'll have to defer to folks with more context.

Specifically: are we signing on to support these additional IDEs?

/fyi @anderdobo

AlexV525:
For everyone who wants to try the plugin on other IDEs, I've setup the release workflow with my fork, and you can download the plugin through GitHub Release: https://github.com/AlexV525/flutter-intellij/releases

helin24:
I've been looking over IJ documentation to make sure I understand how extensions work and the consequences of having this dependency.

One thing that I want to clarify - it seems our current setup with only `<depends>org.jetbrains.android</depends>` is wrong if we aren't also including `config-file="idea-contribs.xml"`, right? If a user does have `org.jetbrains.android` they will still be missing the things in the config file. Unless the contents of the config file are only needed for Android Studio and not for situations where a user is using IntelliJ with the android plugin?

helin24:
@AlexV525 I talked to Jaime today about this change. Does `org.jetbrains.android` and `com.intellij.modules.androidstudio` contain the same classes? In which case could we include `studio-contribs.xml` when either are available?

AlexV525:
@helin24 Thanks for digging in!

> Does `org.jetbrains.android` and `com.intellij.modules.androidstudio` contain the same classes?

They seem similar, but `com.intellij.modules.androidstudio` has more content than `org.jetbrains.android`.
3512d643fcfeac96730082eebf4096c5

> In which case could we include `studio-contribs.xml` when either are available?

If I understand you correctly, we might also put the following declaration:
```xml
<depends optional="true" config-file="studio-contribs.xml">com.intellij.modules.androidstudio</depends>
```
Correct?

helin24:
We already have:

```
<depends optional="true" config-file="studio-contribs.xml">com.intellij.modules.androidstudio</depends>
```

But I was wondering if we could add:

```
<depends optional="true" config-file="studio-contribs.xml">org.jetbrains.android</depends>
```

instead of using a new config file `android-contribs.xml`.

However, if the contents of the android studio module and the android plugin are different, then we're risking some files missing or versioning issues. Unless the functionality we're using from the android studio module is so basic that it's likely to be available in both? Having not looked into the details, I have no idea.

It seems odd to me that we would check for `org.jetbrains.android` as a dependency but then add no additional functionality as a result though.

@AlexV525

AlexV525:
@helin24 

> But I was wondering if we could add:
> 
> ```
> <depends optional="true" config-file="studio-contribs.xml">org.jetbrains.android</depends>
> ```
> 
> instead of using a new config file `android-contribs.xml`.

> It seems odd to me that we would check for `org.jetbrains.android` as a dependency but then add no additional functionality as a result though.

I guess `~.android` stands its place because platform plugins are no longer bundled with editors (there is a standalone *Android* plugin in the Android Studio).

I haven't take a look about which dep we use for Android integrations. It seems reasonable that we rely on `~.androidstudio` only since `~.android` looks like a subset of `~.androidstudio`.  https://github.com/JetBrains/android
(Look like theoratically they are the same).

So we may be able remove the `~.android` without breaking anything.

helin24:
@AlexV525 

Hmm.. are you saying now that perhaps we don't need this line entirely? (i.e. are you referring to `org.jetbrains.android` with `~.android`?)

```
<depends optional="true" config-file="studio-contribs.xml">org.jetbrains.android</depends>
```

My understanding is this:
- If the flutter plugin is installed into Android Studio, then the dependency `com.intellij.modules.androidstudio` is present, because this is bundled into Android Studio. And so therefore we want to include `studio-contribs.xml` to have the functionality that depends on Android Studio-related classes.
- If the flutter plugin is installed into IntelliJ IDEA, then the dependency `com.intellij.modules.androidstudio` is not present. However, we want to check for the dependency `org.jetbrains.android`, which is the Android plugin. If `org.jetbrains.android` is present, then (maybe) we should also include `studio-contribs.xml`.
- Perhaps there are cases where someone would be in Android Studio and also have the Android plugin? In which case it sounds like based on your investigation, this is redundant for them (unrelated to the flutter plugin).

Maybe the outstanding question now is what functionality the line in question:

```
<depends optional="true" config-file="studio-contribs.xml">org.jetbrains.android</depends>
```

contributes (or could contribute, with `studio-contribs.xml` included) to IntelliJ IDEA users.

Let me know if my understanding is in line with your thinking.

AlexV525:
> Maybe the outstanding question now is what functionality the line in question:
> 
> ```
> <depends optional="true" config-file="studio-contribs.xml">org.jetbrains.android</depends>
> ```
> 
> contributes (or could contribute, with `studio-contribs.xml` included) to IntelliJ IDEA users.
> 
> Let me know if my understanding is in line with your thinking.

@helin24 : I think you are correct. But how would I test this using the current configuration? I've tried many times to `runIde` with IC but always starts an Android Studio instance.

helin24:
@AlexV525 Are you able to build the plugin with `bin/plugin make`? Then you would have a .zip file that you could install to an instance of IDEA.

helin24:
@AlexV525 Are you able to build the plugin with `bin/plugin make`? Then you would have a .zip file that you could install to an instance of IDEA. Let me know if you need some additional info from me on how to make that work

AlexV525:
@helin24 Sorry for the delay. I've tested with `studio-contribs` on IDEA. All actions are working properly like in Android Studio, but none of the extensions are working as far as I can tell. So I've created a new contribs specifically for the base Android plugin that only supports existing `<actions>` config.

helin24:
~~Thanks for testing it! That's fine to have a separate file; looking at it again, I think the extensions should only be for Android Studio because it includes things like the new project wizard which we only want to override in Android Studio, not in IntelliJ.~~

~~In that case though, is the actions code exactly the same? I'm hoping we could have `studio-contribs.xml` depend on `android-contribs.xml` then.~~

This seems wrong. Let me write a new comment.

helin24:
Okay, so I see that the actions are for new project wizard, which seem like they shouldn't be needed for IntelliJ. When the Android plugin is installed in IntelliJ, what does new project wizard look like? Does going to File > New project pull up something like this:
Screenshot 2025-07-23 at 9 51 03 AM

Or something like this:
Screenshot 2025-07-23 at 9 51 23 AM

If it's the first scenario, then I think we actually don't need any of the actions that are currently in `android-contribs.xml`, in which case we could just make it optional without adding a new config file. 

Unless we can use the extensions listed in `studio-contribs.xml` when the android plugin is installed (and want to), why are we even including `org.jetbrains.android`?

Sorry for this extended discussion; as you can tell I have very little familiarity with the Android parts of this plugin and I want to make sure we're doing the right thing.

helin24:
I'm also curious, what happens here when the `org.jetbrains.android` dependency is optional: https://github.com/flutter/flutter-intellij/blob/4c7dcef8c9b3ac15bee177709a5cb0e65fb9350a/src/io/flutter/ProjectOpenActivity.java#L103

This file imports `org.jetbrains.android.facet.AndroidFrameworkDetector` and the linked line uses that class. So if the `org.jetbrains.android` plugin is optional, why does this code compile anyway? I guess at runtime the code that uses `AndroidFrameworkDetector` never runs since we check for the presence of the plugin.

AlexV525:
> Sorry for this extended discussion; as you can tell I have very little familiarity with the Android parts of this plugin and I want to make sure we're doing the right thing.

No problem at all! There are always some small changes that expose some unthoughtful history. :)

> Okay, so I see that the actions are for new project wizard, which seem like they shouldn't be needed for IntelliJ. When the Android plugin is installed in IntelliJ, what does new project wizard look like? Does going to File > New project pull up something like this:

> If it's the first scenario, then I think we actually don't need any of the actions that are currently in `android-contribs.xml`, in which case we could just make it optional without adding a new config file.

It's the first scenario. Only a new entry will display at this place. Clicking it will navigate to the normal project creation panel.
image

The problem is, if we don't specify a config file:
image
That's why I started with an empty contribs.

(P.S. I also found a bug that when the Flutter plugin is installed without the Android plugin installed, clicking "File > New > Project" will result in non-responsive.)

> Unless we can use the extensions listed in `studio-contribs.xml` when the android plugin is installed (and want to), why are we even including `org.jetbrains.android`?

There are a couple of businesses that still rely on the plugin:
https://github.com/flutter/flutter-intellij/blob/dec96dc5e19ea756e12a098da2c6474eda3d6320/src/io/flutter/FlutterInitializer.java#L163-L165

> I'm also curious, what happens here when the `org.jetbrains.android` dependency is optional:
> 
> https://github.com/flutter/flutter-intellij/blob/4c7dcef8c9b3ac15bee177709a5cb0e65fb9350a/src/io/flutter/ProjectOpenActivity.java#L103
> 
> This file imports `org.jetbrains.android.facet.AndroidFrameworkDetector` and the linked line uses that class. So if the `org.jetbrains.android` plugin is optional, why does this code compile anyway? I guess at runtime the code that uses `AndroidFrameworkDetector` never runs since we check for the presence of the plugin.

Because it requires us (the author of flutter-intellij) to install the plugin at our compile time, rather than end users.

AlexV525:
> This file imports `org.jetbrains.android.facet.AndroidFrameworkDetector` and the linked line uses that class. So if the `org.jetbrains.android` plugin is optional, why does this code compile anyway? I guess at runtime the code that uses `AndroidFrameworkDetector` never runs since we check for the presence of the plugin.

I suspect this is the reason I ran into non-responsive interface.

resources/META-INF/android-contribs.xml
```diff
@@ -0,0 +1,5 @@
+<!-- Do not edit; instead, modify studio-contribs_template.xml, and run './bin/plugin generate'. -->
```
helin24: Delete the first line here since we don't need a template to generate this file

AlexV525: Done editing

helin24:
@AlexV525 can you investigate changing this method to not use the Android dependency? https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/FlutterUtils.java#L92

I think it will be something like checking `ApplicationInfo.getInstance().getFullApplicationName()`.

I just want to get rid of the error that occurs during File > New Project when a user is on IntelliJ IDEA and has the Flutter plugin installed but not the Android plugin.

AlexV525:
> @AlexV525 can you investigate changing this method to not use the Android dependency? https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/FlutterUtils.java#L92
> 
> I think it will be something like checking `ApplicationInfo.getInstance().getFullApplicationName()`.
> 
> I just want to get rid of the error that occurs during File > New Project when a user is on IntelliJ IDEA and has the Flutter plugin installed but not the Android plugin.

Sure I could try but no time guarantee for now :)

helin24:
@AlexV525 no problem - if you can't get to it next week either I'll work on it or we may revert this change for this release and try again for next release.

Issue #7911, opened 2025-01-31
CONTRIBUTING.md guide feedback


kenzieschmoll:
Filing this issue as a friction log that I encountered while trying to set up my development environment:

> 1. Set your JAVA_HOME directory in your environment.

It is unclear how to do this. Should go in a `.bash_profile`, `.zshrc`, etc. file? Should this be done from the IDE? More clear instructions needed.

> 2. Set your FLUTTER_SDK directory to point to /path/to/flutter.
> 3. Also set your DART_SDK directory to /path/to/flutter/bin/cache/dart-sdk.
> 4. Ensure both DART_SDK, FLUTTER_SDK and JAVA_HOME are added to the PATH in the shell initialization script that runs at login. (not just for the one used for every interactive shell).

Where do you set these directories? Are these supposed to be environment variables in your `.bash_profile`, `.zshrc`, etc.? Are these supposed to be set in the IDE? How do you know where to add this to satisfy the requirement of "...shell initialization script that runs at login. (not just for the one used for every interactive shell)"? More explicit instructions needed.

> Determine the directory of your downloaded IntelliJ IDEA installation. e.g.

How do you do this? Is there an easy way to find out where your IDEA installation directory is?

> If you didn't see the popup at the first start, delete & re-clone the repo and try again.

Is there a better way to re-run this than deleting the repository and starting over? If this doesn't succeed on the first try, this is painful to have to keep deleting the repo to reproduce this step.

> 8. In the "Project Structure" dialog (File | Project Structure):
> - Select "Platform Settings > SDKs", click the "+" sign at the top "Add New SDK (Alt+Insert)", then select "Add JDK...".
>   - Point it to the directory of the jbr which is under the IDEA's content (e.g. IntelliJ IDEA CE.app/Contents/jbr).

Should this be "Add JDK from disk"? Also, it is unclear where this is "Point it to the directory of the jbr which is under the IDEA's content (e.g. IntelliJ IDEA CE.app/Contents/jbr).".  I do not know where to find this path on my machine.

> Select "Platform Settings > SDKs", click the "+" sign at the top "Add New SDK (Alt+Insert)", then select "Add IntelliJ Platform Plugin SDK...".

There is no option for "Add IntelliJ Platform Plugin SDK"
![Image](https://github.com/user-attachments/assets/ea97c146-15c0-4158-bf6f-349e03dc74fd)

[...]

PR #7960, opened 2025-03-05
Update CONTRIBUTING steps

elliette:
Fixes https://github.com/flutter/flutter-intellij/issues/7911

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -212,14 +212,20 @@ name and contact info to the [AUTHORS](AUTHORS) file.
 1. One-time Dart plugin install - first-time a new IDE is installed and run you will need to install the Dart plugin.
     - Find `Plugins` (in "File | Settings | Plugins") and install the Dart plugin, then restart the IDE if needed.
 
-2. Build the project using `Build` | `Build Project`.
+2. Add a license key for JX Browser
+    - **[Note]** This is only required if you are making changes to the embedded views.
+    - Make a copy of `jxbrowser.properties.template` and rename it as `jxbrowser.properties`
+    - Add a [license key](https://teamdev.com/jxbrowser/docs/guides/introduction/licensing/) for JX Browser
+    - ** [Googlers only]** License key can be found in Valentine.
 
-3. Try running the plugin; select the `flutter-intellij [runIde]` run config then click the Debug icon.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -52,79 +52,174 @@ name and contact info to the [AUTHORS](AUTHORS) file.
...
     - [IntelliJ Downloads](https://www.jetbrains.com/idea/download/)
     - Either the community edition (free) or Ultimate will work.
-    - Determine the directory of your downloaded IntelliJ IDEA installation. e.g.
-        * `IntelliJ IDEA CE.app` (macOS)
-        * `~/Library/Application Support/JetBrains/Toolbox/apps/IDEA-U/ch-0/231.8109.175/IntelliJ IDEA.app` (macOS)
-        * `~/idea-IC-231.8109.175` (Linux)
-        * `X:\path\to\your\IDEA-U\ch-0\231.8109.175` (Windows after installed)
-6. Start the IntelliJ IDEA with the `flutter-intellij` project.
+
+2. [PLACHOLDER] Run [command?] to generate the `gradle.properties` file.
```
elliette, line 115: @jwren What is the command we need to run now that `gradle.properties` is gitignored?

jwren: Any of the ./bin/plugin commands work, many of them have a flag to specify which version.  As we discussed yesterday, ideally, we would have a new command where all it does is update and create this file

jwren: Any of the ./bin/plugin commands work, many of them have a flag to specify which version.  As we discussed yesterday, ideally, we would have a new command where all it does is update and create this file

jwren (approved):

CONTRIBUTING.md
elliette: Hmm I just tried testing this by deleting `gradle.properties` and running `./bin/plugin generate` but it didn't create it. I then first created `gradle.properties` and ran `./bin/plugin generate` and it didn't populate it with anything. 

Should I add instructions for manually creating a `gradle.properties` file instead? At last until we have a command to create this file

jwren: For the time being, yes, include the gradle.properties that is currently working for you.

PR #7964, opened 2025-03-10
Add an optional dependency on the Gemini Code Assist IJ plugin

jwren:


build.gradle.kts
```diff
@@ -90,6 +90,7 @@ dependencies {
       "org.jetbrains.kotlin",
       "org.jetbrains.plugins.gradle",
       "org.intellij.intelliLang",
+      "com.google.tools.ij.aiplugin",
```
kenzieschmoll, line 93: should this be in the next if statement, only if the product is android studio?

flutter-idea/build.gradle.kts
```diff
@@ -70,6 +70,7 @@ dependencies {
       "org.jetbrains.kotlin",
       "org.jetbrains.plugins.gradle",
       "org.intellij.intelliLang",
+      "com.google.tools.ij.aiplugin",
```
kenzieschmoll, line 73: same here and below

build.gradle.kts
jwren: If we can provide the support to all users without additional work we should.  It may be the case that some support is added through here, and other support will be AS specific.

build.gradle.kts
```diff
@@ -91,6 +91,10 @@ dependencies {
       "org.jetbrains.plugins.gradle",
       "org.intellij.intelliLang",
     )
+    // The "com.google.tools.ij.aiplugin" is not avaliable for 2023.3:
+    if (!ideaVersion.startsWith("2024.3")) {
```
kenzieschmoll, line 95: this seems fragile. What if the user is on 2024.4 or if we want to support this for versions greater than 2024.3 in the future? Can we add some logic to parse the version so that we can do a >= check instead?

kenzieschmoll: Idk if this version is following semver strictly, but in case it is useful, here is the dart code we use to parse semantic versions and compare them in DevTools: https://github.com/flutter/devtools/blob/master/packages/devtools_shared/lib/src/utils/semantic_version.dart/#L10

jwren: This was supposed to be `2023.3`.... We have logic in our Java code, this is in the Kotlin code, which is not a full-fledged setup for Kotlin either.  I was just going back and forth with @mossmana, given the age of `2023.3, we are removing it so that when the next version of the Flutter IJ plugin has any features added in this space

PR #7971, opened 2025-03-14
Auto select first device when there is no selection yet

### LINKED ISSUE #7970: "Run" widget is unable to pick suitable device (like `flutter run` does)
I guess for projects where starting an emulator is part of the workflow anyway, this doesn't matter much since the emulator startup does make the widget pick up on its availability.

However for desktop projects, the target platform is available without additional steps and selecting one can be automated.

Lootwig:
This change automates the requirement to manually select a device, currently a mandatory manual step when not running from the command line (`flutter run` is capable of autonomously selecting a matching device)

Fixes #7970.

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] All existing and new tests are passing.


AlexV525:
*(Personal suggestion)* Since there is no test in the PR, could you attach a screen recording of how it runs?

jwren:
@Lootwig --- thanks for the PR and sorry for the delay.  Choosing the first option in the list is a good idea, and probably better than doing nothing, which is what we currently do, however what is the ordering of this list?  Shouldn't we be opinionated about which type of run configuration, i.e. when getting started I always choose "Web" as I always have chrome open and on my machine, but perhaps this is wrong for the average Flutter developer.

Lootwig:
> @Lootwig --- thanks for the PR and sorry for the delay. Choosing the first option in the list is a good idea, and probably better than doing nothing, which is what we currently do, however what is the ordering of this list? Shouldn't we be opinionated about which type of run configuration, i.e. when getting started I always choose "Web" as I always have chrome open and on my machine, but perhaps this is wrong for the average Flutter developer.

while the ordering *might* even follow some semi-predictable pattern depending on how the different platforms are being detected, I believe the need for an (immediate) customization is overshooting the target. The majority of use cases (?) involves the dev _launching_ their target platform as part of their workflow, either manually or via startup task in the IDE - both of which are enough to trigger the plugin's detection of a "newly arrived" platform resulting in the convenient auto-selection.

Yes, this implementation may introduce a _new_ group of users, deploying between 7-58 build targets at the same time to the concept of "people usually don't have to click this button every time, you know".. doesn't sound like a bad problem to have, though ^^

jwren (approved):

Issue #8037, opened 2025-04-09
org.jetbrains.intellij.platform version update to `2.5.0`


jwren:
https://plugins.gradle.org/plugin/org.jetbrains.intellij.platform -- the version can be bumped in the Flutter IJ plugin to `2.5.0`

PR #8038, opened 2025-04-09
Update the `org.jetbrains.intellij.platform` version to `2.5.0`

jwren:
See https://plugins.gradle.org/plugin/org.jetbrains.intellij.platform

This resolves https://github.com/flutter/flutter-intellij/issues/8037

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -32,8 +32,7 @@ plugins {
   // https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin.html
   // https://github.com/JetBrains/intellij-platform-gradle-plugin/releases
   // https://plugins.gradle.org/plugin/org.jetbrains.kotlin.jvm
-  // TODO(jwren) "2.2.0" can't be used to build 2023.3, as soon as we don't support this version, update the version here and elsewhere:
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -32,8 +32,7 @@ plugins {
   // https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin.html
   // https://github.com/JetBrains/intellij-platform-gradle-plugin/releases
   // https://plugins.gradle.org/plugin/org.jetbrains.kotlin.jvm
-  // TODO(jwren) "2.2.0" can't be used to build 2023.3, as soon as we don't support this version, update the version here and elsewhere:
-  id("org.jetbrains.intellij.platform") version "2.4.0"
+  id("org.jetbrains.intellij.platform") version "2.5.0"
```
pq, line 35: Aside: would it make sense to store this in the product matrix file to keep related configuration in one place?

pq (approved):

build.gradle.kts
jwren: If we have too I suppose, I sure hope that we don't have to have different settings here for each version shipped.  As you see in the comment, I have been predicating our support of older versions on us not fragmenting the number of inputs even more.

PR #8046, opened 2025-04-11
Fix per-project triggering of "runIde" task

### LINKED ISSUE #8045: `runIde` task is launched twice when using predefined run configuration
Both projects' `runIde` tasks are triggered (and thus opening two IDEs) when using the current pre-defined run configuration.

Lootwig:
Fixes #8045


jwren (approved):

pq (commented):
Amazing. Thank you!

Issue #7980, opened 2025-03-24
"flutter.setSubscriptions" call never forgets a file (aka will ask about an ever growing number of files)


jensjoha:
In IntelliJ, opening a project with `flutter` in the `environment` in `pubspec.yaml`, e.g. something like

```
name: issue_whatnot
publish_to: "none"

environment:
  flutter: ">=3.29.0 <4.0.0"
  sdk: ">=3.7.0 <4.0.0"
```

opening a file sends a `flutter.setSubscriptions` request looking something like this:

```
{
  "id": "42",
  "method": "flutter.setSubscriptions",
  "params": {
    "subscriptions": {
      "OUTLINE": [
        "file:///path/to/file_1.dart"
      ]
    }
  }
}
```

Say we close the file and open another, it now sends something like this:

```
{
  "id": "42",
  "method": "flutter.setSubscriptions",
  "params": {
    "subscriptions": {
      "OUTLINE": [
        "file:///path/to/file_1.dart",
        "file:///path/to/file_2.dart"
      ]
    }
  }
}
```

closing that and opening a third, closing that and opening a forth etc a bunch of times and you'll end up with something like

```
{
  "id": "42",
  "method": "flutter.setSubscriptions",
  "params": {
    "subscriptions": {
      "OUTLINE": [
        "file:///path/to/file_1.dart",
        "file:///path/to/file_2.dart",
        ...
        "file:///path/to/file_n.dart",
      ]
    }
  }
}
```

at which point the analyzer will start to become slow:

```
5 ms (+ 22045 ms)	completion.getSuggestions2
23121 ms (+ 100 ms)	edit.getAssists
```

Notice that if I open a file I've already opened once it doesn't send the request, I believe because of this:

https://github.com/flutter/flutter-intellij/blob/38e78f0d9fac4ed3d35b59e8445c8f415e4f09c7/flutter-idea/src/io/flutter/dart/FlutterDartAnalysisServer.java#L140-L150.

(i.e. `sendSubscriptions` is only called i it actually adds the path)

I'm guessing something goes wrong in these lines:

https://github.com/flutter/flutter-intellij/blob/38e78f0d9fac4ed3d35b59e8445c8f415e4f09c7/flutter-idea/src/io/flutter/editor/ActiveEditorsOutlineService.java#L125-L152

[...]

PR #8057, opened 2025-04-14
Fix `FlutterOutline` service infinite subscriptions

pq:
Fixes an issue where we never unsubscribe outline listeners.

The rub is that `getLocalFileUri` is being called on a path that is already a URI and is producing a bogus path which has no associated listener registered (because it’s bogus).

In general the issue here is a symptom of URIs being encoded in strings that are never validated.

The quick fix is to not “uri-ify” an already valid UR string. Longer term we should consider a deeper audit of places we’re passing URIs around. It’s possible there are other related issues. If/when we do this, Alex shared some [interesting thoughts from his work on the Dart plugin](https://github.com/flutter/flutter-intellij/issues/7980#issuecomment-2785510617) that’d be worth synthesizing.

Fixes: https://github.com/flutter/flutter-intellij/issues/7980

## Pre-launch Checklist

- [x] I signed the [CLA].

[Contributor Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md
[breaking change policy]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md

### DIFF
--- a/flutter-idea/src/io/flutter/editor/ActiveEditorsOutlineService.java
+++ b/flutter-idea/src/io/flutter/editor/ActiveEditorsOutlineService.java
@@ -132,8 +132,7 @@ public class ActiveEditorsOutlineService implements Disposable {
       }
 
       for (final String path : obsoletePaths) {
-        final String filePathOrUri = getAnalysisServer().getAnalysisService().getLocalFileUri(path);
-        final FlutterOutlineListener listener = outlineListeners.remove(filePathOrUri);
+        final FlutterOutlineListener listener = outlineListeners.remove(path);
         if (listener != null) {
           getAnalysisServer().removeOutlineListener(FileUtil.toSystemDependentName(path), listener);
         }


pq:
@jwren: holding off on changelog changes until https://github.com/flutter/flutter-intellij/pull/8056 lands, or you can edit that directly?

I'd add a line like:

    - Fix Flutter Outline View event over-subscriptions (#7980)

(Or similar)

jwren:
Thanks @pq -- I'll update the changelog PR.  This change was successful in the presubmits, landing it now.

pq:
(Updated branch.)

flutter-idea/src/io/flutter/editor/ActiveEditorsOutlineService.java
```diff
@@ -132,8 +132,7 @@ private void updateActiveEditors() {
       }
 
       for (final String path : obsoletePaths) {
-        final String filePathOrUri = getAnalysisServer().getAnalysisService().getLocalFileUri(path);
-        final FlutterOutlineListener listener = outlineListeners.remove(filePathOrUri);
+        final FlutterOutlineListener listener = outlineListeners.remove(path);
```
jwren, line 135: As a follow up, reference the issue that is "fixed" with the line of code written like this, to allow more transparency to what is going on and the sensitive nature of this particular code.

Issue #8034, opened 2025-04-09
Fix the verifier setup in gradle, and update to use the latest verifier jar version


jwren:
The version needs to be updated for sure.  https://github.com/JetBrains/intellij-plugin-verifier/releases/tag/1.384

I think that the issue that broke it is two fold: some of the refactoring for 2025.1 broke the ability for it to run for the Flutter IJ plugin.

Also, when I updated the version the last time around, the jar wasn't in the PR, glancing at the `.gitignore` file, I think this was the issue.  https://github.com/flutter/flutter-intellij/pull/7871

cc @pq

jwren:
Related: https://youtrack.jetbrains.com/issue/MP-7574/Verifier-home-runner-.pluginVerifier-extracted-plugins-existence-error

This PR landed a few weeks after this issue was resolved, see history here: https://github.com/flutter/flutter-intellij/pull/8078.

I think that updating to a recent version, `1.388`, of the IJ Verifier resolved this issue.

PR #8077, opened 2025-04-17
Update the verifier to 1.384 and make changes in the build.gradle.kts to fix the IJ Verifier for the Flutter plugin

jwren:
This resolves https://github.com/flutter/flutter-intellij/issues/8034


.gitignore
```diff
@@ -13,7 +13,6 @@ flutter-gui-tests/.gradle/
 flutter-gui-tests/guitest.log
```
pq, line 13: Aside: we can remove this now too.

pq (approved):

PR #8108, opened 2025-04-23
[Property Editor] Enable Property Editor panel

elliette:
property_editor_intellij

- If the Flutter version does not support the Property Editor, no panel is shown. 
- If the Dart plugin version does not support the Property Editor, the following message is shown:

Screenshot 2025-04-23 at 2 27 59 PM


flutter-idea/src/io/flutter/sdk/FlutterSdkVersion.java
```diff
@@ -47,7 +47,7 @@ public final class FlutterSdkVersion implements Comparable<FlutterSdkVersion> {
   private static final FlutterSdkVersion MIN_SUPPORTS_DTD = new FlutterSdkVersion("3.22.0");
 
   @NotNull
-  private static final FlutterSdkVersion MIN_SUPPORTS_PROPERTY_EDITOR = new FlutterSdkVersion("3.29.0");
+  public static final FlutterSdkVersion MIN_SUPPORTS_PROPERTY_EDITOR = new FlutterSdkVersion("3.32.0-0.1.pre");
```
jwren, line 50: LGTM.  I will leave it to you if this changes at all with cherry picks.

jwren (approved):

helin24:
Lgtm!

I thought that the flutter plugin version requires a certain dart plugin version, and so it would be possible to control what dart plugin we have without an extra check (which gets confusing since we also have dart sdk checks elsewhere). I see in testing you have this scenario though - is this also going to occur for normal users?

elliette:
> I thought that the flutter plugin version requires a certain dart plugin version, and so it would be possible to control what dart plugin we have without an extra check (which gets confusing since we also have dart sdk checks elsewhere). I see in testing you have this scenario though - is this also going to occur for normal users?

Oh that would be great if that is the case! Do you know where we tie the flutter plugin version to the dart plugin version?

helin24:
The files involved are `gradle.properties` and `product-matrix.json`, for example changed in https://github.com/flutter/flutter-intellij/pull/7462/files

But I know this can be tricky to update and make sure all the versions are correct. @jwren has more experience testing different build combinations.

elliette:
> But I know this can be tricky to update and make sure all the versions are correct. @jwren has more experience testing different build combinations.

Chatted offline with @jwren - we are hoping to have this version of flutter-plugin support dart plugin versions that both include and don't include the required property editor changes, therefore will leave as is with the warning message.

elliette:
> But I know this can be tricky to update and make sure all the versions are correct. @jwren has more experience testing different build combinations.

Chatted offline with @jwren - we are hoping to have this version of flutter-plugin support dart plugin versions that both include and don't include the required property editor changes, therefore will leave as is with the warning message.

Issue #8098, opened 2025-04-23
[process] Establish Java/Kotlin coding standards


pq:
For Dart contributions, we can point folks to the [Dart style guide](https://dart.dev/guides/language/effective-dart) as a North star. We should establish what we want to do for Java/Kotlin and communicate that in the PR template and Contributing guides as well. (See: #8062).

helin24:
Following up on https://github.com/flutter/flutter-intellij/pull/8365#discussion_r2226069098, I'm thinking it would make sense to avoid * (full package) imports and instead commit to importing single classes. The main benefits I see are that single class imports reduce opportunity for naming conflicts (e.g. if two packages end up having classes that are named the same) and they could make compile a little faster (since compiler does not need to search packages for the right classes to import). The package import can be easier if we are using many classes in a package, but since our IDE tends to manage imports pretty easily I don't think this is a problem.

One reference: https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad (the AI summary from searching "single class vs package import" also seems informative)

Issue #7718, opened 2024-10-16
Migrate or remove all API usages that will be removed with time


jwren:
Many of the errors reported in the JetBrains marketplace, https://plugins.jetbrains.com/plugin/9212-flutter/edit/versions/stable/615584, are for deprecated issues which **will be removed**, these usages in the Flutter plugin need to be migrated or removed.

jwren:
The pre-requisite for this was to migrate to Gradle IntelliJ platform `2.1.x` to have the updated verifier which is done, this work will not be finished for this upcoming Milestone, bumping to the next one.

jwren:
Some additional work in https://github.com/flutter/flutter-intellij/pull/8018

jwren:
Latest report from the marketplace continues to show progress and opportunity: https://plugins.jetbrains.com/plugin/9212-flutter/edit/versions/stable/712464.  Also related, the verifier is currently not working see https://github.com/flutter/flutter-intellij/issues/8034 for status.

pq:
@jwren, @helin24: working through these, there are some internal bits for which there are not (and I expect may never be) migration paths (for example `PlatformUtils.getPlatformPrefix()` and `module.setModuleType(..)`). For these cases I'm inclined to ignore the inspection and call out to an issue tracking "necessary internal dependencies". How does that sound to you all?

jwren:
@pq -- If they are marked as APIs that will **removed** from the platform, then no we can't just ignore them unfortunately.  In the case here of `PlatformUtils.getPlatformPrefix()`, this API is marked as **internal** (not for removal, see https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/util/PlatformUtils.java#L69).  We should create a new issue and link it here, as well other locations where internal APIs are being used, and then also in the verifier logic where we have `VerifyPluginTask.FailureLevel.INTERNAL_API_USAGES` commented out.

jwren:
Closing this issue since https://github.com/flutter/flutter-intellij/issues/8361 tracks the rest of the work for `SCHEDULED_FOR_REMOVAL_API_USAGES `, thanks @pq.

PR #8121, opened 2025-04-28
[CQ] API: migrate off unstable `getInstanceOrNull`

pq:
This method is marked as unstable and slated for possible removal.

image

That said, this replacement doesn't totally preserve the current semantics.

@helin24: is there any reason we don't want to prompt a call to the application manager here?

Specifically, we'll now cause this to happen (if `ourInstance` is `null`):

```java
(ProgressManager)ApplicationManager.getApplication().getService(ProgressManager.class);
```

See: #https://github.com/flutter/flutter-intellij/issues/7718

---

<details>

</details>

### DIFF
--- a/flutter-idea/src/io/flutter/view/FlutterView.java
+++ b/flutter-idea/src/io/flutter/view/FlutterView.java
@@ -198,10 +198,8 @@ public class FlutterView implements PersistentStateComponent<FlutterViewState>, 
             });
           }));
       };
-      final ProgressManager progressManager = ProgressManager.getInstanceOrNull();
-      if (progressManager != null) {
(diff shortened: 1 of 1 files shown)


helin24:
I've never thought about this very hard, so I'm not sure. If there isn't an instance of `ProgressManager` and we call `getInstance`, will that throw an error we need to catch? Requesting the service if it doesn't exist seems fine to me, though I suppose I intended the progress indicator to be optional.

helin24 (approved):

pq:
Thanks! I *think* this is a safe call. I just wanted to make sure you hadn't picked the `OrNull` variation intentionally.

I'll do some more poking around to see if we need to do any guarding my instinct is not...

Issue #8100, opened 2025-04-23
[cq] Migrate startup initializers to use the `ProjectActivity` API


pq:
The current approach is deprecated, unstable and subject to removal.

Specifically, we need to update:

- [x] `FlutterInitializer` (https://github.com/flutter/flutter-intellij/pull/8200)
- [x] ~`FlutterSdkManager`~ and
- [ ] `FlutterStudioStartupActivity`

(See stacktrace below.)

```
2025-04-23 08:23:25,396 [  81501]   WARN - #c.i.i.s.i.StartupManagerImpl - Migrate io.flutter.FlutterInitializer to ProjectActivity [Plugin: io.flutter]
com.intellij.diagnostic.PluginException: Migrate io.flutter.FlutterInitializer to ProjectActivity [Plugin: io.flutter]
	at com.intellij.ide.startup.impl.StartupManagerImpl.doRunPostStartupActivities(StartupManagerImpl.kt:293)
	at com.intellij.ide.startup.impl.StartupManagerImpl.access$doRunPostStartupActivities(StartupManagerImpl.kt:70)
	at com.intellij.ide.startup.impl.StartupManagerImpl$runPostStartupActivities$3$2.invokeSuspend(StartupManagerImpl.kt:196)
	at com.intellij.ide.startup.impl.StartupManagerImpl$runPostStartupActivities$3$2.invoke(StartupManagerImpl.kt)
	at com.intellij.ide.startup.impl.StartupManagerImpl$runPostStartupActivities$3$2.invoke(StartupManagerImpl.kt)
	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:62)
	at kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(Builders.common.kt:163)
	at kotlinx.coroutines.BuildersKt.withContext(Unknown Source)
	at com.intellij.ide.startup.impl.StartupManagerImpl$runPostStartupActivities$3.invokeSuspend(StartupManagerImpl.kt:195)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:608)
[...]

Issue #6953, opened 2023-10-20
Components API deprecation - migrate all ProjectManagerListener.projectOpened() to backgroundPostStartupActivity extensions


jwren:
See https://plugins.jetbrains.com/docs/intellij/plugin-components.html#comintellijpoststartupactivity for notice and documentation on the deprecation intentions of Components from JetBrains.

Migration forward has different directions before and after `2023.1`, if we can, it would be prudent to wait until we are only supporting this major platform as a minimum version.

---

- [ ] `FlutterProjectStructureDetector`
- [x] https://github.com/flutter/flutter-intellij/pull/8123

---

See also: https://github.com/flutter/flutter-intellij/issues/7718

jwren:
The pre-requisite for this was to migrate to Gradle IntelliJ platform `2.1.x` to have the update verifier which is done, this work will not be finished for this upcoming Milestone, bumping to the next one.

pq:
(Updated description w/ API calls to migrate.)

jwren:
Closing this issue as https://github.com/flutter/flutter-intellij/issues/8361 is now tracking the verifier enums and status.

PR #8123, opened 2025-04-28
[CQ] API: migrate `FlutterSdkManager` off `projectOpened`

pq:
Move `FlutterSdkManager` (unstable) post-project open initialization to a (supported) `ProjectActivity`.

See: https://github.com/flutter/flutter-intellij/issues/8100, https://github.com/flutter/flutter-intellij/issues/6953

---

<details>

</details>


jwren:
This resolves the single call to `projectOpened`, but now we have the functionality split between two files, `FlutterSdkManager` and `FlutterSdkManagerStartupActivity.kt`, wouldn't it be more prudent to migrate the entire file over so that 1) we have one file where this logic is all shared, and 2) we don't increase the number of entries in the `plugin.xml` files?

With the bootstrappers, the loading in of the extensions isn't always consistent.  This will potentially change the order of operations for all users or some users.

pq:
Good questions!

I don't think we want (all of) the `FlutterSdkManager` registered as a startup activity and am inclined to think that `checkForFlutterSdkChange` should never have been in it the first place but I'm happy to discuss. (I'm not even sure it can be since it's primarily a project **_service_**.)

As for duplicate entries, even if the class were shared, it would need two entries in the xml for `projectService` and `backgroundPostStartupActivity` registrations so I don't see a way to reduce that. Or maybe you meant something else?

One thought is that we could make this a more general `FlutterSdkStartupActivity` so that other post startup activity logic could be hosted here too. (Another reason not to make `FlutterSdkManager`  a `ProjectActivity`.)

Happy to do that when I tackle the next `projectOpened` migration if it sounds good to you.

jwren (approved):

PR #8150, opened 2025-05-02
Some nullability and documentation cleanup in package io.flutter.dart

jwren:
---


pq:
Cool! This may conflict w/ https://github.com/flutter/flutter-intellij/pull/8139 which I was holding but maybe I can land it so we can move ahead with more cleanup?

pq (approved):

PR #7929, opened 2025-02-04
[CQ] remove (stale) deploy `plugin` command

pq:
## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [x] I signed the [CLA].

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md

PR #8155, opened 2025-05-06
Revert deletion of deploy command to fix dev build

helin24:
I think this should fix the dev build, which stopped uploading back in February. It uses the command `./bin/plugin deploy --channel=dev` (see `tool/kokoro/deploy.sh`). The deletion happened in https://github.com/flutter/flutter-intellij/pull/7929.

I can't test it from an internal command until these changes are in the main branch though.

It could be we no longer need the dev build, but this seems like an easy enough change.


jwren:
LGTM, CC @pq who removed this after a conversation we had. I did not know at the time that Kokoro uses the `deploy` command.

jwren (approved):

pq:
This is great to know. Thanks for the fix and for looping me in!

PR #8160, opened 2025-05-07
[CQ] Cleanup in the singleton utils package

jwren:


helin24:
I'm surprised this doesn't cause failing tests - is it needed for checking that the right files are "downloaded" etc in the JX browser tests?

Piinks:
Hey @jwren! Greetings from stale PR triage. 👋  Is this still on your radar?

Issue #8144, opened 2025-04-30
Infinite spinner while trying to load property editor


kenzieschmoll:
Successful load (remove .txt to load in Chrome DevTools):

[successful_load.har.txt](https://github.com/user-attachments/files/19983979/successful_load.har.txt)

@jwren @elliette

jwren:
Attaching a spinning ball case:

kenzieschmoll:
> Attaching a spinning ball case:

I think the attachment is missing. I had to add .txt to the file extension to get github to accept it.

DanTup:
I can reproduce this in Android Studio. I saw it consistently yesterday, however today I was able to load the property editor once, however it usually fails. When it fails, the property editor connects to DTD and connects to the Service stream, but it is not told about any services (like the Editor services), and therefore doesn't initialize.

I enabled the analysis server instrumentation log and confirmed that Android Studio was calling `connectToDtd`. However, it appears that it's connecting to a different version to the one sent to the analysis server:

![Image](https://github.com/user-attachments/assets/73eec0ac-bf91-412a-bef8-40879daf0cf3)

When I review my running processes, I can see a single DTD instance, and I can see that DevTools was started with a DTD URI that matches the one the analysis server connected to.. However I also see another DevTools running, that does not have a `--dtd-uri`:

![Image](https://github.com/user-attachments/assets/1c596482-10c9-4c89-ad78-f045cca66238)

Based on this, my current theory (with zero knowledge of how the plugins work) is that Android Studio is somehow spawning _two_ DevTools servers. One is correctly being provided a DTD URI (for the DTD that Android Studio started), but the second is not, and is therefore spawning its own DTD, and that's the server being used.

Since this is intermittent, I wonder if there is a race here, and when it works, only one DevTools server was spawned (or, both were spawned after DTD was started and therefore were provided the same/correct DTD URI).

DanTup:
I missed it in the list before, but I do also see the second instance of DTD in the process list, but it is spawned as `dartaotruntime.exe` instead of `dart.exe` (my assumption is that this is the one spawned by DevTools).

![Image](https://github.com/user-attachments/assets/e951bcbb-e67a-4ac7-8ea6-2ebb6f03b5fe)

I don't think I know enough about the plugins to debug this any further, but perhaps it will be useful to someone more familiar with them.

DanTup:
[...]

PR #8177, opened 2025-05-09
Wait for the Dart plugin to start the DevTools server with retries 

elliette:
Fixes https://github.com/flutter/flutter-intellij/issues/8144

This PR:
* Makes sure we only ever have one connected DevTools server
* Tracks the start-up of the DevTools server using the [Progress API](https://plugins.jetbrains.com/docs/intellij/background-processes.html) instead of `executeOnPooledThread` (logic pulled into `DevToolsServerTask.java`). This allows us to display the progress and cancel the server startup on `forceRestart`.
* Retries waiting for the Dart plugin initiated DevTools server until a timeout is reached
* If the Dart plugin DevTools is not ready by the timeout, switching to a new Flutter panel will re-trigger starting the server

### Server start-up status:
Screenshot 2025-05-09 at 12 40 31 PM

### Message if waiting for the server timed out:
Screenshot 2025-05-09 at 1 26 55 PM

### Loaded Property Editor:
Screenshot 2025-05-09 at 12 39 37 PM


flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
```diff
@@ -0,0 +1,399 @@
...
+      // https://github.com/flutter/flutter-intellij/blob/master/CONTRIBUTING.md#developing-with-local-devtools
+      final String localDevToolsDir = Registry.stringValue(LOCAL_DEVTOOLS_DIR);
+      if (!localDevToolsDir.isEmpty()) {
+        progressIndicator.setFraction(60);
+        progressIndicator.setText2("Starting local server");
+        setUpLocalServer(localDevToolsDir);
+      }
+
+      // If the Dart plugin does not start DevTools, then call `dart devtools` to start the server.
+      final Boolean dartPluginStartsDevTools = true;
```
elliette, line 115: Is there a Dart plugin version we want to gate this on? Or was this from long ago enough that it is safe to remove? 

Here's when the Dart Plugin began setting the host and port: https://github.com/JetBrains/intellij-plugins/commit/91822865c5a97915d68f0345486b53fe8cf4d17c

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
```diff
@@ -0,0 +1,399 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
+
+class DevToolsServerTask extends Task.Backgroundable {
+  private static final Logger LOG = Logger.getInstance(DevToolsServerTask.class);
```
pq, line 61: Consider marking this and the following initialized fields `@NotNull`? (That will help with null-check inspections down-stream.)

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
```diff
@@ -0,0 +1,399 @@
...
+  public static final String LOCAL_DEVTOOLS_ARGS = "flutter.local.devtools.args";
+  private @NotNull Project project;
+  private final AtomicReference<CompletableFuture<DevToolsInstance>> devToolsFutureRef;
+
+  private DaemonApi daemonApi;
+  private ProcessHandler process;
+
+  public DevToolsServerTask(@NotNull Project project,
+                            @NotNull String title,
+                            AtomicReference<CompletableFuture<DevToolsInstance>> devToolsFutureRef) {
```
pq, line 72: Is this certain to be non-null? If so, could you annotate it `@NotNull` and consider the same for `devToolsFutureRef`?

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
```diff
@@ -0,0 +1,399 @@
...
+
+  @Override
+  public void run(@NotNull ProgressIndicator progressIndicator) {
+    try {
+      progressIndicator.setFraction(30);
+      progressIndicator.setText2("Init");
+
+      // If DevTools is not supported, start the daemon instead.
+      final boolean dartDevToolsSupported = dartSdkSupportsDartDevTools();
+      if (!dartDevToolsSupported) {
```
pq, line 86: Consider logging this?

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
```diff
@@ -0,0 +1,399 @@
...
+        setUpWithDart(createCommand(workspaceCache.get().getRoot().getPath(), workspaceCache.get().getDevToolsScript(),
+                                    ImmutableList.of("--machine")));
+        return;
+      }
+
+      // This is only for development to check integration with a locally run DevTools server.
+      // To enable, follow the instructions in:
+      // https://github.com/flutter/flutter-intellij/blob/master/CONTRIBUTING.md#developing-with-local-devtools
+      final String localDevToolsDir = Registry.stringValue(LOCAL_DEVTOOLS_DIR);
+      if (!localDevToolsDir.isEmpty()) {
```
pq, line 108: Maybe we can add this to the log too? (May help with debugging down the road.) I'm thinking `LOG.info(...)`.

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
```diff
@@ -0,0 +1,399 @@
...
+      final Version version = Version.parseVersion(dartSdk.getVersion());
+      assert version != null;
+      return version.compareTo(2, 15, 0) >= 0;
+    }
+    return false;
+  }
+
+  @Override
+  public void onThrowable(@NotNull Throwable error) {
+    super.onThrowable(error);
```
pq, line 152: (Nit.) It looks like this call just logs the error to the `Task.class` logger. Since we're doing our own logging in `cancelWithError`, maybe not necessary to call `super.onThrowable(..)`?

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
```diff
@@ -0,0 +1,399 @@
...
+    args.add("--dtd-uri=" + dtdUri);
+    final String localDevToolsArgs = Registry.stringValue(LOCAL_DEVTOOLS_ARGS);
+    if (!localDevToolsArgs.isEmpty()) {
+      args.addAll(Arrays.stream(localDevToolsArgs.split(" ")).toList());
+    }
+
+    setUpInDevMode(createCommand(localDevToolsDir, "dt", args));
+  }
+
+  private void setUpInDevMode(GeneralCommandLine command) {
```
pq, line 173: Can we annotate this arg `@NotNull` and mark the return of `createCommand` as `@NotNull`?

pq (approved):
Just some nits really. This looks great. Looking forward to testing it out! 🎉

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
pq: I'll defer to @jwren but I don't think we're compatible with any plugins that go that far back so I think not?

flutter-idea/src/io/flutter/devtools/AbstractDevToolsViewFactory.java
```diff
@@ -98,19 +103,36 @@ public void createToolWindowContent(@NotNull Project project, @NotNull ToolWindo
...
+
+  private void loadDevToolsInEmbeddedBrowser(@NotNull Project project,
+                                             @NotNull ToolWindow toolWindow,
+                                             @NotNull FlutterSdkVersion flutterSdkVersion) {
     AsyncUtils.whenCompleteUiThread(
       DevToolsService.getInstance(project).getDevToolsInstance(),
       (instance, error) -> {
+        viewUtils.presentLabel(toolWindow, "Loading " + getToolWindowTitle() + "...");
+
         // Skip displaying if the project has been closed.
```
helin24, line 121: rewrite this comment

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
jwren: No, don't gate it on the Dart Plugin version.

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
```diff
@@ -0,0 +1,399 @@
...
+            process.destroyProcess();
+          }
+        }
+      }
+    });
+  }
+
+  private CompletableFuture<DevToolsInstance> checkForDartPluginInitiatedDevToolsWithRetries(ProgressIndicator progressIndicator)
+    throws InterruptedException {
+    progressIndicator.setText2("Waiting for Dart Plugin");
```
jwren: Nit -- perhaps "Waiting for DTD Server to load"

jwren (approved):

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
elliette: Good point, done! (And other @NotNull instances)

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
elliette: Yes, done!

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
elliette: Done

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
elliette: Ahh good point, removed

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
elliette: Yes, done

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
elliette: Done, switched to "Waiting for server with DTD"

helin24 (approved):
thanks! looks great!

flutter-idea/src/io/flutter/run/daemon/DevToolsServerTask.java
pq: Nice.

flutter-idea/src/io/flutter/devtools/AbstractDevToolsViewFactory.java
elliette: Oops autosubmit beat me to it! I will open a follow up PR monday

PR #8253, opened 2025-06-06
[draft] Add a separate file for plugin logs

helin24:
WIP example for logging separately from idea.log. Doc upcoming

### DIFF
--- a/.idea/runConfigurations/flutter_intellij__runIde_.xml
+++ b/.idea/runConfigurations/flutter_intellij__runIde_.xml
@@ -1,5 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="flutter-intellij [runIde]" type="GradleRunConfiguration" factoryName="Gradle">
+    <log_file alias="flutter" path="$PROJECT_DIR$/build/idea-sandbox/AI-AI-251.23774.16.2511.13338727/log/flutter.log" />
     <ExternalSystemSettings>
       <option name="executionName" />
       <option name="externalProjectPath" value="$PROJECT_DIR$" />
--- a/flutter-idea/src/io/flutter/logging/PluginLogHandler.java
+++ b/flutter-idea/src/io/flutter/logging/PluginLogHandler.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2025 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.logging;
+
+import com.intellij.openapi.application.PathManager;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+
+public class PluginLogHandler extends Handler {
+  private static final String LOG_FILE_NAME = "flutter.log";
+  public PluginLogHandler() {
+  }
+
+  @Override
+  public void publish(LogRecord record) {
+    // Write to a file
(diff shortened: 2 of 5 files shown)


flutter-idea/src/io/flutter/logging/PluginLogHandler.java
```diff
@@ -0,0 +1,52 @@
...
+  @Override
+  public void publish(LogRecord record) {
+    // Write to a file
+    final String logPath = PathManager.getLogPath();
+    System.out.println("Log path: " + logPath);
+    File logFile = new File(logPath + File.separatorChar + LOG_FILE_NAME);
+
+    try {
+      if (!logFile.exists()) {
+        logFile.createNewFile(); // Create the file if it doesn't exist
```
pq: Does this need to be wrapped in a write action or anything?

helin24:
At the moment, this is the output in `flutter.log` (any comments on log format?): 

```
2025-06-27 09:34:38 io.flutter.run.daemon.DevToolsServerTask [INFO   ] Finding or starting DevTools  
2025-06-27 09:34:39 io.flutter.FlutterInitializer [INFO   ] Executing Flutter plugin startup for project: hello_external_2  
2025-06-27 09:34:39 io.flutter.FlutterInitializer [INFO   ] Flutter module has been found for project: hello_external_2  
```

flutter-idea/src/io/flutter/logging/PluginLogHandler.java
helin24: resolving since code has changed and I'm no longer using a custom handler

jwren (approved):

.idea/runConfigurations/flutter_intellij__runIde_.xml
```diff
@@ -1,5 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="flutter-intellij [runIde]" type="GradleRunConfiguration" factoryName="Gradle">
+    <log_file alias="flutter" path="$PROJECT_DIR$/build/idea-sandbox/AI-AI-251.23774.16.2511.13338727/log/flutter.log" />
```
pq, line 3: Should this be checked in? Will it be the same for everyone? Or is this specific to the build you have on your machine?

flutter-idea/src/io/flutter/logging/PluginLogger.java
```diff
@@ -0,0 +1,37 @@
...
+  private static final FileHandler fileHandler;
+  static {
+    final String logPath = PathManager.getLogPath();
+    try {
+      fileHandler = new FileHandler(logPath + File.separatorChar + LOG_FILE_NAME, 1024 * 1024, 1);
+    }
+    catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+    System.setProperty("java.util.logging.SimpleFormatter.format",
```
pq, line 28: Odd that you can't set format properties programmatically. (I see you can use a logging properties file but that's no better.)

Anyway: odd!

.idea/runConfigurations/flutter_intellij__runIde_.xml
helin24: Oops I missed this comment. I don't know what to do about this - this may be the same on all of our machines, but it looks like a very specific version of Android Studio. I guess this would update depending on what version we're building based on the product matrix?

flutter-idea/src/io/flutter/logging/PluginLogger.java
helin24: I agree, super weird. It could be we'll find a better way to do this but I did not when I was working on it.

PR #8268, opened 2025-06-13
Add a forceLoad option to property editor

helin24:
The property editor was sometimes failing to load. I think this was the repro:
1. Open a project (property editor will load)
2. Close the project
3. Open the project again (property editor fails to load)

It seems the static var is set to null only on IJ first startup, not on project open. There may be a better way to fix this but I'm not sure what it is yet.


jwren (approved):

pq (commented):
Nice catch!

PR #8271, opened 2025-06-18
Change the verifier to use ides { recommended() }, and disable our check on experimental usages.

jwren:
The experimental usage is due to a call to `com.intellij.ui.BadgeIcon`


jwren:
Ignore the failing presubmit, this change did not cause it.

PR #8278, opened 2025-06-23
Consolidate gradle build files

helin24:
This removes the gradle build files from the `flutter-idea` and `flutter-studio` directories and adds all sources to the top level directory.

I tested that:
- `bin/plugin make` runs and the resulting plugins can be installed and work
- `bin/plugin verify` runs successfully
- `bin/plugin test` runs successfully


jwren (approved):
LGTM!  Thanks @helin24!

PR #8288, opened 2025-06-26
Add a table about SDK support in README

helin24:
This links to the flutter.dev site and provides a table similar to what Danny added: https://dartcode.org/sdk-version-compatibility/

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,17 @@ A brief summary of the [getting started guide](https://docs.flutter.dev/tools/an
 - install the [Flutter SDK](https://docs.flutter.dev/get-started/install)
 - run `flutter doctor` from the command line to verify your installation
 - ensure you have a supported IntelliJ development environment; either:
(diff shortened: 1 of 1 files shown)


pq (approved):
I LOVE this!

Thinking ahead, I wonder if we couldn't codegen the table? Is there somewhere else the compatibility info is captured that we can pull from?

(My concern is the table getting stale or out of sync.)

helin24:
Yes, some sort of codegen would be nice, ideally run each time we release. I think we would want to move the version we want to warn users on and the oldest supported version to a configuration file

pq:
> Yes, some sort of codegen would be nice, ideally run each time we release. I think we would want to move the version we want to warn users on and the oldest supported version to a configuration file

Good idea! Maybe open an issue to keep track of the idea?

helin24:
Already done :) #8289

PR #8290, opened 2025-06-26
Remove code for managing daemon startup of DevTools

helin24:
This was for versions of Dart SDK < v2.15 which is very outdated


pq (approved):
Sweet!

Issue #8292, opened 2025-06-27
[CQ] consider null-safety in the plugin verifier


pq:
Follow-up from https://github.com/flutter/flutter-intellij/issues/8291.

As we clean up the plugin, we could consider enabling the inspection in the verifier and adding filtering to only apply to identified "clean" directories. That would prevent any regressions from getting introduced accidentally.

Issue #8291, opened 2025-06-27
[CQ] ☂  fix nullability problems


pq:
A tracking issue for fixing nullability problems.

![Image](https://github.com/user-attachments/assets/5dfe20ed-185c-4bee-b389-a897de4ce40e)

PR #8293, opened 2025-06-27
[CQ] fix nullability problems for `flutter/actions`

pq:
Fix nullability problems for all the actions in `src/io/flutter/actions/`.

See https://github.com/flutter/flutter-intellij/issues/8291.

If we pursue https://github.com/flutter/flutter-intellij/issues/8292, we could mark `src/io/flutter/actions/` as null-"clean".

---

<details>

</details>

### DIFF
--- a/flutter-idea/src/io/flutter/actions/AttachDebuggerAction.java
+++ b/flutter-idea/src/io/flutter/actions/AttachDebuggerAction.java
@@ -125,7 +125,7 @@ public class AttachDebuggerAction extends FlutterSdkAction {
   }
 
   @Override
-  public void update(AnActionEvent e) {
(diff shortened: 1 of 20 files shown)


pq:
Exactly. It's really unfortunate that the label is so general. It'd be way more useful IMO if it said "nullability issue" or similar.

As a rule, I try and avoid using it but in some cases (when for example, you're interacting w/ unannotated libraries) there's not a great alternative. (I'm on the fence about whether `assert`s are any better or just add more noise.)

FWIW: if we ever do get to a null-safe world, I'd love to do another round of cleanup that evaluates all the suppressions and cleans them up.

That may be majorly wishful thinking though!

PR #8316, opened 2025-07-08
Remove the use of the now deprecated `WriteCommandAction.Simple` usages

jwren:


helin24 (approved):

jwren:
Verified with the latest that this change does not break formatting on save, landing now.

PR #8317, opened 2025-07-08
Remove usage of deprecated Notification.setListener(..)

jwren:
Also: cleanup of nullability parameters in `FlutterMessages.java`

### DIFF
--- a/flutter-idea/src/io/flutter/FlutterMessages.java
+++ b/flutter-idea/src/io/flutter/FlutterMessages.java
@@ -6,7 +6,6 @@
 package io.flutter;
 
 import com.intellij.notification.Notification;
-import com.intellij.notification.NotificationListener;
 import com.intellij.notification.NotificationType;
 import com.intellij.notification.Notifications;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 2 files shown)


flutter-idea/src/io/flutter/FlutterMessages.java
```diff
@@ -23,24 +22,24 @@ public class FlutterMessages {
...
   }
 
-  public static void showWarning(String title, String message, @Nullable Project project) {
+  public static void showWarning(@NotNull String title, @NotNull String message, @Nullable Project project) {
     Notifications.Bus.notify(
       new Notification(
         FLUTTER_NOTIFICATION_GROUP_ID,
         title,
         message,
-        NotificationType.WARNING).setListener(NotificationListener.URL_OPENING_LISTENER), project);
```
helin24, line 40: Is a listener already set for this without using the deprecated fn? Or are we deciding to not listen for URL opens?

helin24 (approved):

Darrylwin:
good job bro. But U're sure this simple delete have not bad effect on the code by the way ?

flutter-idea/src/io/flutter/FlutterMessages.java
jwren: Ack, I didn't see any negative effects on the core UX here of a user being exposed to the notification.  If there is some piece I am missing we can prioritize and implement with non-deprecated APIs.

PR #8319, opened 2025-07-08
Remove deprecated constructor java.net.URL.<init>(String) invocations

jwren:


jwren:
Resolved with a change that just landed. Thanks @pq!

PR #8322, opened 2025-07-09
[logging] Add logging to project initialization

helin24:
This logs any issues from project setup in `FlutterInitializer` to our dedicated log file.

The log variable from `protected val log` is internal, and I'm not sure why (despite asking Gemini multiple questions about it). That's why I'm accessing the log from a function instead.

### DIFF
--- a/flutter-idea/src/io/flutter/FlutterInitializer.java
+++ b/flutter-idea/src/io/flutter/FlutterInitializer.java
@@ -126,9 +126,6 @@ public class FlutterInitializer extends FlutterProjectActivity {
       }
     }
 
-    if (true) {
-      throw new RuntimeException("just for fun");
-    }
     // Lambdas need final vars.
     boolean finalHasFlutterModule = hasFlutterModule;
     ReadAction.nonBlocking(() -> {


flutter-idea/src/io/flutter/FlutterProjectActivity.kt
```diff
@@ -22,4 +23,8 @@ abstract class FlutterProjectActivity : ProjectActivity {
       log.error(it)
     }
   }
+
+  protected fun log(): Logger {
```
pq, line 27: Out of curiosity: why did you decide to use a function rather than just accessing the field directly?

flutter-idea/src/io/flutter/logging/PluginLogger.java
```diff
@@ -30,7 +30,7 @@ public class PluginLogger {
     fileHandler.setFormatter(new SimpleFormatter());
   }
 
-  public static Logger createLogger(@NotNull Class<?> logClass) {
+  public static @NotNull Logger createLogger(@NotNull Class<?> logClass) {
```
pq, line 33: Cool!

flutter-idea/src/io/flutter/FlutterProjectActivity.kt
pq: Ah, hmmm. Just read your note above:

> The log variable from protected val log is internal, and I'm not sure why (despite asking Gemini multiple questions about it). That's why I'm accessing the log from a function instead.

 Seems odd. (Will investigate.)

helin24: I tried, but for some reason it's "internal"? I couldn't figure out why. Gemini suggested this is because the logger class is declared internal at some point, but I couldn't find anything related to that in the source code. 🤷  ???

pq: Are you overriding a `log` variable in `ProjectActivity`?

helin24: `ProjectActivity` is an interface that doesn't seem to have a `log` variable exposed, and `FlutterProjectActivity` doesn't label the `log` var as overridden

PR #8338, opened 2025-07-16
WIP add integration test (not working)

helin24:
From following https://plugins.jetbrains.com/docs/intellij/integration-tests-intro.html and https://blog.jetbrains.com/platform/2025/02/integration-tests-for-plugin-developers-intro-dependencies-and-first-integration-test/ and asking Gemini a lot of questions

The `PluginTest` file doesn't seem to see all of the `com.intellij.ide...` imports it needs. For example, it gets stuck on `TestCase` despite my trying to import all of the classpaths that we have for normal tests and trying to add specific Starter framework libraries.

I added printouts of paths visible at compile time to try to see what's included. There may be other debugging steps like this I'm not aware of when working on gradle issues.

@jwren

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="17" />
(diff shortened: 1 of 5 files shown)


Piinks:

helin24:
Can be closed! Thanks

PR #8339, opened 2025-07-16
[logging] Use separate log for FlutterCommand

helin24:
I added an option (default off) to record full file paths in logs. Originally I was thinking that we shouldn't record any file paths, but it seems reasonable to have them sometimes for debugging setup issues since these are recorded on a user's machine rather than to somewhere centralized that we receive by default.


jwren (approved):

pq:
Cool! I wonder if this is something we should run by @anderdobo or @mit-mit? Do we need any extra messaging to make sure folks know not to have this enabled if filenames are sensitive and they're sharing logs with us?

helin24:
Before this change we were logging these paths in `idea.log`, so this change is moving towards less sharing of filenames. I'm thinking for now we can mention if we're asking for logs, but in the future if we start attaching a log file to issues, we should do a check for the state of the setting.

helin24:
But regardless, if any of you have opinions of what we can/should log in local files, please let me know. I started a doc on this but don't have much content at the moment.

pq:
SGTM. Thanks!

PR #8343, opened 2025-07-17
Move contents of flutter-idea and flutter-studio to top level

helin24:


jwren (approved):

pq:
This is awesome! Is there a tracking bug or doc I can read to get a little more context? (I was going to ask a few questions but realize things may be written up someplace.)

Thanks!

helin24:
There isn't a tracking bug, but there should be - I'll start one so all of these changes can be linked from one place. Feel free to post questions here

PR #8363, opened 2025-07-22
Remove xml templates

helin24:
This removes having the `plugin.dart` tool populate the XML templates with `version` and `ideaVersion` information. We already have configuration in `build.gradle.kts` to do this.

This comes after https://github.com/flutter/flutter-intellij/pull/8360 and should be rebased after that change is merged.


jwren:
woot!

Issue #8369, opened 2025-07-23
Transition to using PluginLogger for logging to a separate file


helin24:
There's an internal doc on this topic, but the goal is to move our logs to a separate file, flutter.log. This should hopefully:

- Make logging easier to see and understand during development
- Make logging more useful for debugging user issues that are reported to us but hard to reproduce

Original change to add file: https://github.com/flutter/flutter-intellij/pull/8253

Migrating files that use logs:
- https://github.com/flutter/flutter-intellij/pull/8320
- https://github.com/flutter/flutter-intellij/pull/8322
- https://github.com/flutter/flutter-intellij/pull/8339, https://github.com/flutter/flutter-intellij/pull/8341
- Further migrations will probably just refer to this issue.

helin24:
As I've been doing more of these and encountering logged exceptions, I've adopted the pattern of checking for the file path enabled setting, then either showing a full exception or recording only the message, i.e.:

```
// pseudocode
if (filePathEnabled) {
  LOG.warn(exception)
} else {
  LOG.warn("descriptive message of what happened, exception is: " + exception.getMessage)
}
```

This is because I don't know what every possible exception could be, but some can contain file paths. If we need more info from exceptions we will ask users to turn on file path logging.

helin24:
All Flutter plugin classes are migrated to using PluginLogger. Going forward I'm going to encourage contributors to add debug logging to new code that may be confusing so that we will be able to have more visibility of how code should work and hopefully have relevant logs that users can send us to debug new issues.

PR #8372, opened 2025-07-24
Change editor directory files to use PluginLogger

helin24:
Related to https://github.com/flutter/flutter-intellij/issues/8369

I also attempted to make the flutter.log file viewable in sandbox mode with a generic file path, but this doesn't work for now - I get an empty file. Oddly this file path will work for `idea.log`, but seems to skip some lines. As the note says, this topic needs further investigation to understand how to optimize this part of setup.


pq (approved):
👍

PR #8373, opened 2025-07-24
[logging] Use PluginLogger for bazel doctor and analytics

helin24:
I'm just looking through the code and checking that paths aren't going to be logged. Bazel path for flutter doctor probably is used very little (if at all) and analytics isn't used at all (will probably be moved to dart plugin?).

Related to https://github.com/flutter/flutter-intellij/issues/8369


jwren:
Bazel support is something we should discuss on VC, it is not used by the majority of our users.  I have left it in the Flutter plugin for IJ as it has been free to bring forward in time, however, I don't believe that we want to leave it in the product going forward.

PR #8387, opened 2025-07-31
Use gradle build directly for dev build

helin24:
This changes over to using gradle for writing the JX Browser key and setting the dev version number, which allows us to call `./gradlew build` for building the dev release instead of using the plugin tool. Note that the plugin tool is still being used to upload the dev build to jetbrains.

I think the next step is removing the gradle build command from the plugin tool and auditing the rest of the commands in the plugin tool to see if we can remove more. I ran a test of building dev from this branch, but I'll wait to make sure this works over the weekend before proceeding.


build.gradle.kts
```diff
@@ -267,3 +274,38 @@ tasks.register("printCompileClasspath") {
     println("--- End Compile Classpath ---")
   }
 }
+
+// This finds the JX Browser license key from the environment and writes it to a file.
```
pq: Nit: => JxBrowser?

(Seems like that's their preferred capitalization? https://teamdev.com/jxbrowser/)

jwren (commented):
LGTM

PR #8390, opened 2025-08-01
Fix the New Flutter Project action from hanging

jwren:
This resolves https://github.com/flutter/flutter-intellij/issues/8310


jwren:
I tried to no-avail here to get the Flutter option on the left hand side to always be selected, but the APIs are not working as expected, see image where Flutter is not selected:

Screenshot 2025-08-01 at 1 09 41 PM

Additional manual testing is needed.

pq:
Oddly, I just tried to reproduce the original issue (https://github.com/flutter/flutter-intellij/issues/8310) and can't. 

(That is, creating a flutter project and opening it in  "This Window".)

image

Are you able to reproduce that locally before this change?

jwren:
It is specifically the async call here which is a kotlin co-routine which is the preferred method. 

The issue I think is a race condition, that this should either completely solve or improve.

jwren:
It is specifically the async call here which is a kotlin co-routine which is the preferred method. 

The issue I think is a race condition, that this should either completely solve or improve.

Issue #8401, opened 2025-08-05
Investigate whether we need the provision tool


helin24:
Search "Provision tool" in our CONTRIBUTING.md - this uses `bin/plugin test` but I'm not sure what it's used for and whether we still need it.

PR #8407, opened 2025-08-06
Change to using gradle for test and verify

helin24:
Test is still being used for provision tool (which I'm not sure what this is about - https://github.com/flutter/flutter-intellij/issues/8401), so I kept the command code for now.

Verify command code is being removed in this PR.


pq (approved):
Awesome! 🎉

Issue #8261, opened 2025-06-12
Call the new `ConnectedAppService` DTD methods when VM Service instances are started / stopped


kenzieschmoll:
This issue tracks the IntelliJ work required for https://github.com/dart-lang/sdk/issues/60540. 

**What needs to be done?**
- Call `ConnectedApp.registerVmService` when a vm service instance is started
- Call `ConnectedApp.unregisterVmService` when a vm service instance is shut down 

As a reference the parallel VS code work was addressed in https://github.com/Dart-Code/Dart-Code/pull/5527. 

**Why?**
This will give DTD clients the ability to fetch all the active VM services (active debug sessions). Without the `ConnectedApp` service, this was only possible if DTD was started from VS Code, where the `Editor.getDebugSessions` method is registered.

PR #8436, opened 2025-08-08
Register VM service with DTD

helin24:
Addresses https://github.com/flutter/flutter-intellij/issues/8261

### DIFF
--- a/src/io/flutter/run/daemon/FlutterApp.java
+++ b/src/io/flutter/run/daemon/FlutterApp.java
@@ -7,6 +7,7 @@ package io.flutter.run.daemon;
 
 import com.google.common.base.Stopwatch;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.process.ProcessAdapter;
@@ -27,13 +28,17 @@ import com.intellij.openapi.util.Key;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.util.EventDispatcher;
 import com.intellij.util.concurrency.AppExecutorUtil;
+import com.jetbrains.lang.dart.ide.toolingDaemon.DartToolingDaemonService;
+import de.roderick.weberknecht.WebSocketException;
 import io.flutter.FlutterMessages;
 import io.flutter.FlutterUtils;
 import io.flutter.ObservatoryConnector;
 import io.flutter.bazel.Workspace;
 import io.flutter.bazel.WorkspaceCache;
+import io.flutter.dart.DtdUtils;
 import io.flutter.dart.FlutterDartAnalysisServer;
 import io.flutter.logging.FlutterConsoleLogManager;
+import io.flutter.logging.PluginLogger;
 import io.flutter.run.FlutterDebugProcess;
 import io.flutter.run.FlutterDevice;
 import io.flutter.run.FlutterLaunchMode;
@@ -677,10 +682,13 @@ public class FlutterApp implements Disposable {
  * Listens for events while running or debugging an app.
  */
 class FlutterAppDaemonEventListener implements DaemonEvent.Listener {
-  private static final @NotNull Logger LOG = Logger.getInstance(FlutterAppDaemonEventListener.class);
+  private static final @NotNull Logger LOG = PluginLogger.createLogger(FlutterAppDaemonEventListener.class);
 
   private final @NotNull FlutterApp app;
   private final @NotNull ProgressHelper progress;
+  private String appVmServiceUri;
+  private final DtdUtils dtdUtils = new DtdUtils();
+
 
   FlutterAppDaemonEventListener(@NotNull FlutterApp app, @NotNull Project project) {
     this.app = app;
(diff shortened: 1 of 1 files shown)


helin24:
CC @jakemac53

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -738,6 +746,12 @@ public void onAppStarting(DaemonEvent.AppStarting event) {
   @Override
   public void onAppDebugPort(@NotNull DaemonEvent.AppDebugPort debugInfo) {
     app.setWsUrl(debugInfo.wsUri);
+    this.appVmServiceUri = debugInfo.wsUri;
```
jakemac53, line 749: Do we get any sort of identifier that we can use as a map key that we can look up when we get the app stopped event for this particular app?

While the MCP server today only supports one connected app, we do want to have it support multiple in the future, so ideally the IDE plugins wouldn't be assuming just one.

helin24: IJ can register multiple apps to DTD. There's an app name that's a unique identifier, or the VM service URI is unique too.

This is an example of the params we send with two apps running at the same time:
```
2025-08-08 09:28:04 io.flutter.run.daemon.FlutterAppDaemonEventListener [INFO   ] Successful request ConnectedApp.registerVmService to DTD with params: {"uri":"ws://127.0.0.1:50489/m1aDg3zyHq4=/ws","name":"752c8ff1-86dd-4c91-abab-825747059057"}  
2025-08-08 09:29:33 io.flutter.run.daemon.FlutterAppDaemonEventListener [INFO   ] Successful request ConnectedApp.registerVmService to DTD with params: {"uri":"ws://127.0.0.1:51041/iUbSRQCLBfg=/ws","name":"65328744-2368-46bf-9e23-70066b219342"}  
```

The `unregisterVmService` takes the URI as a param also, so I imagine you would use that as the key?

jwren (approved):

pq:
/fyi @DanTup who did the analogous work for VS Code.

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -677,10 +682,13 @@ public void dispose() {
  * Listens for events while running or debugging an app.
  */
 class FlutterAppDaemonEventListener implements DaemonEvent.Listener {
-  private static final @NotNull Logger LOG = Logger.getInstance(FlutterAppDaemonEventListener.class);
+  private static final @NotNull Logger LOG = PluginLogger.createLogger(FlutterAppDaemonEventListener.class);
 
   private final @NotNull FlutterApp app;
   private final @NotNull ProgressHelper progress;
+  private String appVmServiceUri;
+  private final DtdUtils dtdUtils = new DtdUtils();
```
pq, line 690: Can this be static or is the util object stateful?

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -677,10 +682,13 @@ public void dispose() {
  * Listens for events while running or debugging an app.
  */
 class FlutterAppDaemonEventListener implements DaemonEvent.Listener {
-  private static final @NotNull Logger LOG = Logger.getInstance(FlutterAppDaemonEventListener.class);
+  private static final @NotNull Logger LOG = PluginLogger.createLogger(FlutterAppDaemonEventListener.class);
 
   private final @NotNull FlutterApp app;
   private final @NotNull ProgressHelper progress;
+  private String appVmServiceUri;
```
pq, line 689: Mark this `@Nullable`?

src/io/flutter/run/daemon/FlutterApp.java
pq: And mark it `@NotNull`?

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -738,6 +746,12 @@ public void onAppStarting(DaemonEvent.AppStarting event) {
   @Override
   public void onAppDebugPort(@NotNull DaemonEvent.AppDebugPort debugInfo) {
     app.setWsUrl(debugInfo.wsUri);
+    this.appVmServiceUri = debugInfo.wsUri;
+
+    final JsonObject params = new JsonObject();
+    params.addProperty("uri", debugInfo.wsUri);
+    params.addProperty("name", debugInfo.appId);
+    sendDtdRequest("ConnectedApp.registerVmService", params);
```
pq, line 754: I'm not sure how we want to test this (And I'd love to get your thoughts), but in anticipation, maybe we should introduce some constants somewhere?

src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -804,7 +818,36 @@ public void onAppStopped(@NotNull DaemonEvent.AppStopped stopped) {
...
     app.getProcessHandler().destroyProcess();
   }
+
+  private void sendDtdRequest(@NotNull String requestName, @NotNull JsonObject params) {
+    try {
+      final DartToolingDaemonService dtdService = dtdUtils.readyDtdService(app.getProject()).get();
+      if (dtdService == null) {
+        return;
+      }
+      // This removes secret from params when we print out after receiving response.
```
pq, line 836: Missing some context but maybe this could be clearer for the next reader.

Removes "secret" _what_?

Maybe another sentence?

pq (commented):
Awesome! 🎉

src/io/flutter/run/daemon/FlutterApp.java
helin24: Yes, for now I can add these methods as constants. Maybe a separate file that links to DTD documentation would be helpful.

I haven't thought much about testing; I suppose one part would be testing `dtdService.sendRequest` in the Dart plugin. And ultimately integration testing in the Dart plugin.

helin24:
Some clean up in #8442

pq:
Awesome. Thanks for following up!

helin24:
I added some notes above on how to test this

PR #8452, opened 2025-08-12
Convert LaunchState, SdkFields to separate plugin logger

helin24:


src/io/flutter/run/LaunchState.java
```diff
@@ -180,7 +182,11 @@ protected RunContentDescriptor launch(@NotNull ExecutionEnvironment env) throws
       }
     }
     catch (IllegalAccessException | InvocationTargetException | NoSuchFieldException | NoSuchMethodException e) {
-      LOG.info(e);
+      if (FlutterSettings.getInstance().isFilePathLoggingEnabled()) {
```
pq: I wonder if we couldn't push this check into the `info` implementation. Would that be limiting?

Maybe we could have an optional `sanitizePaths` argument version? So like:

    LOG.info(e, true);

with an implementation like:


    void info(Exception e, boolean sanitizePaths) {
      if (sanitizePaths && FlutterSettings.getInstance().isFilePathLoggingEnabled()) {
        LOG.info(e);
      } else {
        LOG.info(e.toString());
      }
    }   

Where there's a one arg version that passes `false`.

    void info(Exception e) {
      info(e, false);
    }   

(I do wish there were named params...)

pq: Meh. Dunno.

helin24: Yeah, I think it would be reasonable. There's some old code in `FlutterUtils` that basically does this and I'll try doing that. I don't know if changing the info implementation would be best since we'd have to make a new logger class.

PR #8455, opened 2025-08-13
Add verify and test steps to dev build

helin24:
I can't test that these work without pointing kokoro to a different branch, so I figure I'll just run a test after this is committed to main branch.

### DIFF
--- a/tool/kokoro/deploy.sh
+++ b/tool/kokoro/deploy.sh
@@ -3,6 +3,21 @@
 source ./tool/kokoro/setup.sh
 setup
 
+echo "kokoro verify start"
+
+./gradlew verifyPluginProjectConfiguration
+./gradlew verifyPluginStructure
+./gradlew verifyPluginSignature
+./gradlew verifyPlugin
+
+echo "kokoro verify finished"
+
+echo "kokoro test start"
+
+./gradlew test
+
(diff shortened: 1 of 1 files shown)


tool/kokoro/deploy.sh
```diff
@@ -3,6 +3,21 @@
 source ./tool/kokoro/setup.sh
 setup
 
+echo "kokoro verify start"
+
+./gradlew verifyPluginProjectConfiguration
+./gradlew verifyPluginStructure
+./gradlew verifyPluginSignature
+./gradlew verifyPlugin
```
pq, line 11: Out of curiosity, would it make sense to add a kind of macro `verifyAll` command?

(Obviously not required here but I'm curious what happens when we add a new verification; would we have to remember to update this script and others to include it?)

jwren: In the gradle build file, yes I believe that we could add a new task to call of of these with some, `verifyAll` alias if we want to.

helin24: We could add something like that to gradle. I don't even know what each of these contain.

helin24:
Presubmit failing should be fixed once https://github.com/flutter/flutter/pull/173731 lands

Issue #8448, opened 2025-08-11
Slow operations are prohibited on EDT - FlutterSdk.java:138


jwren:
New Flutter Project > default settings > Open in new Window, if given option.

This was on AS Narwhal.

this sometimes appears when/after the Flutter SDK is automatically configured:

```
java.lang.Throwable: Slow operations are prohibited on EDT. See SlowOperations.assertSlowOperationsAreAllowed javadoc.
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:375)
	at com.intellij.util.SlowOperations.assertSlowOperationsAreAllowed(SlowOperations.java:113)
	at com.intellij.openapi.vfs.newvfs.persistent.FSRecordsImpl.update(FSRecordsImpl.java:756)
	at com.intellij.openapi.vfs.newvfs.persistent.PersistentFSImpl.findChildInfo(PersistentFSImpl.java:620)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findInPersistence(VirtualDirectoryImpl.java:155)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.doFindChild(VirtualDirectoryImpl.java:138)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findChild(VirtualDirectoryImpl.java:84)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findChild(VirtualDirectoryImpl.java:570)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findChild(VirtualDirectoryImpl.java:50)
	at com.intellij.openapi.vfs.newvfs.VfsImplUtil.findFileByPath(VfsImplUtil.java:56)
	at com.intellij.openapi.vfs.impl.local.LocalFileSystemBase.findFileByPath(LocalFileSystemBase.java:57)
	at io.flutter.sdk.FlutterSdk.forPath(FlutterSdk.java:138)
	at io.flutter.sdk.FlutterSdkUtil.addKnownSDKPathsToCombo(FlutterSdkUtil.java:108)
	at io.flutter.module.FlutterGeneratorPeer.init(FlutterGeneratorPeer.java:67)
	at io.flutter.module.FlutterGeneratorPeer.<init>(FlutterGeneratorPeer.java:62)
[...]

Issue #8242, opened 2025-06-04
Loading the Flutter Plugin failed: Too many non-blocking read actions submitted at once.


kfman:
## What happened

(please describe what you were doing when this exception occurred)

## Version information

IntelliJ IDEA `2025.1.1.1` • Flutter plugin `io.flutter 85.3.2` • Dart plugin `251.25410.28`

Flutter 3.27.1 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 17025dd882 (6 months ago) • 2024-12-17 03:23:09 +0900
Engine • revision cb4b5fff73
Tools • Dart 3.6.0 • DevTools 2.40.2

## Exception

null

```
java.lang.Throwable: Too many non-blocking read actions submitted at once. Please use coalesceBy, BoundedTaskExecutor or another way of limiting the number of concurrently running threads.: 11 with similar stack traces are currently active
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:375)
	at com.intellij.openapi.application.impl.SubmissionTracker.preventTooManySubmissions(SubmissionTracker.java:47)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl$Submission.<init>(NonBlockingReadActionImpl.java:278)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl.submit(NonBlockingReadActionImpl.java:233)
	at io.flutter.sdk.FlutterPluginsLibraryManager.updateFlutterPluginsImpl(FlutterPluginsLibraryManager.java:135)
	at io.flutter.sdk.FlutterPluginsLibraryManager.updateFlutterPlugins(FlutterPluginsLibraryManager.java:115)
	at com.intellij.openapi.project.DumbServiceImpl.smartInvokeLater$lambda$21(DumbServiceImpl.kt:613)
	at com.intellij.openapi.application.TransactionGuardImpl.runWithWritingAllowed(TransactionGuardImpl.java:240)
	at com.intellij.openapi.application.TransactionGuardImpl.access$100(TransactionGuardImpl.java:25)
	at com.intellij.openapi.application.TransactionGuardImpl$1.run(TransactionGuardImpl.java:202)
	at com.intellij.openapi.application.impl.AnyThreadWriteThreadingSupport.runIntendedWriteActionOnCurrentThread$lambda$7(AnyThreadWriteThreadingSupport.kt:319)
[...]

PR #8457, opened 2025-08-13
Resolves a "Slow operations are prohibited on EDT" exception

jwren:
This resolves https://github.com/flutter/flutter-intellij/issues/8448

Additional work may be needed to limit the number of read actions by the plugin, see https://github.com/flutter/flutter-intellij/issues/8242

### DIFF
--- a/src/io/flutter/sdk/FlutterSdkUtil.java
+++ b/src/io/flutter/sdk/FlutterSdkUtil.java
@@ -96,26 +96,39 @@ public class FlutterSdkUtil {
    * Adds the current path and other known paths to the combo, most recently used first.
    */
   public static void addKnownSDKPathsToCombo(@NotNull JComboBox combo) {
-    final Set<String> pathsToShow = new LinkedHashSet<>();
-
+    // First, get the current path from the combo box on the EDT.
     final String currentPath = combo.getEditor().getItem().toString().trim();
+    final Set<String> pathsToShow = new LinkedHashSet<>();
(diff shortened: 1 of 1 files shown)


src/io/flutter/sdk/FlutterSdkUtil.java
```diff
@@ -96,26 +96,39 @@ private static void updateKnownPaths(@SuppressWarnings("SameParameterValue") @No
...
     if (!currentPath.isEmpty()) {
       pathsToShow.add(currentPath);
     }
 
-    final String[] knownPaths = getKnownFlutterSdkPaths();
-    for (String path : knownPaths) {
-      if (FlutterSdk.forPath(path) != null) {
-        pathsToShow.add(FileUtil.toSystemDependentName(path));
+    // Now, run the slow operation (finding valid SDK paths) on a background thread.
+    ApplicationManager.getApplication().executeOnPooledThread(() -> {
```
pq, line 107: Neat! I'll have to read up on how this thread pool is managed and what kinds of operations are best run on it. Maybe this is the approach we should take elsewhere?

pq (approved):
Nice! 🎉

src/io/flutter/sdk/FlutterSdkUtil.java
jwren: It is part of it, I was reading up on `coalesceBy`, see the message in https://github.com/flutter/flutter-intellij/issues/8242, which I think will be the right approach.

Issue #8439, opened 2025-08-08
UI glitch: rectangular shadow outside device selector button when window is unfocused


khusantoy:
When another window is in focus, the device selector UI in the top toolbar shows a visual glitch — a rectangular shadow appears outside the button’s border (see attached screenshot). This shadow looks sharp-edged and does not match the rest of the UI design, making it visually unappealing.

<img width="232" height="46" alt="Image" src="https://github.com/user-attachments/assets/0d1c2db0-30ef-4f95-aa8f-b7f48cc4da3c" />

Issue #7972, opened 2025-03-17
Device selector does not support default Light theme


lukef:
The default Light theme in intellij/android studio has a dark header bar. There is a Light + Light header variant as well. If you select the Light theme you will get:

![Image](https://github.com/user-attachments/assets/7365b6b2-c172-4996-80e1-f53fbcba69bb)

```
Platform: Linux
Flutter plugin version: 83.0.4

Android Studio information:
Android Studio Meerkat | 2024.3.1
Build #AI-243.22562.218.2431.13114758, built on February 24, 2025
Runtime version: 21.0.5+-12932927-b750.29 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
Toolkit: sun.awt.X11.XToolkit
Linux 6.13.5-200.fc41.x86_64
GC: G1 Young Generation, G1 Concurrent GC, G1 Old Generation
Memory: 3924M
Cores: 12
Registry:
  ide.experimental.ui=true
  i18n.locale=
Non-Bundled Plugins:
  Dart (243.23177)
  io.flutter (83.0.4)
Current Desktop: GNOME
```

AlexV525:
This looks very different with the latest UI. Here is the selector in my editor: 

![Image](https://github.com/user-attachments/assets/826b2f35-dd68-4e86-b3c5-9e31d372447e)

Could you try to uninstall and install the plugin again and see if it works?

lukef:
I think you're using dark mode.The issue isn't for Dark vs Light. The issue is for the default Light mode that uses a Dark header.

AlexV525:
My result when using **Light** is still different from yours.

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/6811e805-a21b-45b6-ad17-69dcbb5716d8" />

lukef:
What version and IDE is this? I'm using the latest stable releases and mine doesn't look anything like that.

AlexV525:
> What version and IDE is this? I'm using the latest stable releases and mine doesn't look anything like that.

IntelliJ IDEA 2024.3.5 (Ultimate Edition)
Build #IU-243.26053.27, built on March 16, 2025

mahmuttaskiran:
I've the same issue appearing in both IntelliJ IDEA (2024.3.5 (Ultimate Edition)) and Android Studio (Ladybug Feature Drop | 2024.2.2.)

AlexV525:
[...]

PR #8471, opened 2025-08-18
🐛 Set the device selector component to opaque during its creation

AlexV525:
Fixes https://github.com/flutter/flutter-intellij/issues/8439
Fixes https://github.com/flutter/flutter-intellij/issues/7972

This is inspired by [StackOverflow](https://stackoverflow.com/questions/13939176/making-jcombobox-transparent), Claude Sonnet 4, and Gemini 2.5 Pro.

Before:
image

After:
image


AlexV525:
The ultimate solution would be `setSmallVariant(true)` and modify the styles to be the same as the run configuration or the Android Device Selector. However, I didn't find a proper way to do so. Hope someone is familiar with related stuff.

Issue #8478, opened 2025-08-20
🚀 Performance freeze in EDT caused by ExecutorService blocking in Flutter plugin


AlexV525:
*(Generated by GPT-5)*

#### Summary
When changing the editor theme in IntelliJ with the Flutter plugin enabled, the IDE experiences a significant UI freeze (21 seconds). The stack trace indicates that the main thread (EDT) is blocked due to improper handling of ExecutorService lifecycle in `FlutterInitializer.sendThemeChangedEvent`.

#### Details
- **Stack trace snippet:**
```
com.intellij.diagnostic.Freeze
    at java.base@21.0.7/jdk.internal.misc.Unsafe.park(Native Method)
    at java.base@21.0.7/java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:269)
    at java.base@21.0.7/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.awaitNanos(AbstractQueuedSynchronizer.java:1763)
    at java.base@21.0.7/java.util.concurrent.ThreadPoolExecutor.awaitTermination(ThreadPoolExecutor.java:1475)
    at java.base@21.0.7/java.util.concurrent.Executors$DelegatedExecutorService.awaitTermination(Executors.java:780)
    at java.base@21.0.7/java.util.concurrent.ExecutorService.close(ExecutorService.java:417)
    at io.flutter.FlutterInitializer.sendThemeChangedEvent(FlutterInitializer.java:306)
    at io.flutter.FlutterInitializer.lambda$setUpThemeChangeNotifications$3(FlutterInitializer.java:240)
    at io.flutter.FlutterInitializer$$Lambda/0x0000026c765eaf30.uiSettingsChanged(Unknown Source)
```
- **Root cause:**
  - A new ScheduledExecutorService is created and closed within the EDT every time the theme changes. Since JDK 21's `ExecutorService.close()` calls `awaitTermination`, it blocks the EDT if the scheduled task hasn't finished.

#### Suggested Fix
- Use a shared, long-lived ScheduledExecutorService (e.g. IntelliJ's `AppExecutorUtil.getAppScheduledExecutorService`) instead of creating/closing one on each event.
- Avoid any blocking calls on the EDT.
- Remove try-with-resources around executor creation in `sendThemeChangedEvent`.

#### Impact
This freeze severely affects developer experience in IntelliJ-based IDEs when using the Flutter plugin.

#### References
[...]

Issue #7794, opened 2024-11-18
DTD theme changed event exception


jwren:
Flutter plugin version `82.0.3`, IntelliJ IDEA 2024.2 (Ultimate Edition)

Unable to send theme changed event

java.util.concurrent.ExecutionException: java.lang.Exception: Timed out waiting for DTD websocket to start
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
	at io.flutter.FlutterInitializer.lambda$sendThemeChangedEvent$6(FlutterInitializer.java:335)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.Exception: Timed out waiting for DTD websocket to start
	at io.flutter.dart.DtdUtils.readyDtdService(DtdUtils.java:35)
	... 7 more

AlexV525:
Run into this in a relatively high frequency with my production projects:

[idea-dump.log](https://github.com/user-attachments/files/21764030/idea-dump.log)

[flutter-dump.log](https://github.com/user-attachments/files/21764053/flutter-dump.log)

helin24:
Thanks for the logs - it looks like DTD is succeeding eventually but not in the timeout we have set up.

It's odd in the first place that is sent so soon after project startup. Potentially we could add a check somewhere on whether theme info has changed before requesting DTD action.

PR #8479, opened 2025-08-20
Fix DTD freezes when opening projects, and EDT freezes when the theme changed and opening embedded DevTools

AlexV525:
*(originally with https://github.com/flutter/flutter-intellij/pull/8477)*

Summary
- Fix frequent “Timed out waiting for DTD websocket to start” errors when opening projects.
- Fix EDT freezes caused by:
  - Synchronous JxBrowser Engine initialization when opening embedded DevTools.
  - Blocking ExecutorService lifecycle on theme change (JDK 21 ExecutorService.close() awaits termination).

Changes
- DtdUtils
  - Refactor readyDtdService to be truly asynchronous using a shared ScheduledExecutorService for periodic polling (no blocking).
  - De-duplicate and cache the in-flight Future per Project to avoid redundant polling under concurrency.
  - Preserve an overall timeout (~20s by default). On timeout or error, complete exceptionally and clear the cache to allow subsequent retries.
- EmbeddedJxBrowser.openEmbeddedTab:
  - Avoid initializing JxBrowser Engine on the EDT; show “installation in progress” UI until async init completes.
  - This removes the fallback compareAndSet(null, getEngine()) path that could call Engine.newInstance(...) on EDT and freeze it.
- FlutterInitializer.sendThemeChangedEvent:
  - Send themeChanged asynchronously: remove .get() blocking; use a shared scheduler with debounce.
  - If DTD becomes ready later, the event is sent then; if it times out, gracefully skip with debug-level logging (no startup errors).
  - Use shared AppScheduledExecutor (AppExecutorUtil) instead of creating/closing a new executor per event.
  - Remove try-with-resources that triggers ExecutorService.close() on EDT with JDK 21.

Notes
- Added comments explaining why Engine initialization must not occur on EDT and how JDK 21 changed ExecutorService.close().

Fixes
- Fixes https://github.com/flutter/flutter-intellij/issues/7794
- Fixes https://github.com/flutter/flutter-intellij/issues/8394
- Fixes https://github.com/flutter/flutter-intellij/issues/8478

Testing
- Manual:
  - Open Flutter projects; no error logs about DTD timeout.
  - Switch IDE themes repeatedly; no UI freeze.
  - Open DevTools (embedded) after clean startup and after JxBrowser installation completes; no EDT freeze, proper “installing...” UX during setup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 - Made dev release daily instead of weekly
 - Set the device selector component to opaque during its creation to avoid an unexpected background color (#8471)
 - Refactored `DeviceSelectorAction` and add rich icons to different platform devices (#8475)
-- Fix DTD & EDT freezes when the theme changed and opening embedded DevTools (#8477)
+- Fix DTD freezes when opening projects, and EDT freezes when the theme is changed and opening embedded DevTools (#8477)
 
 ## 87.1.0
 


AlexV525:
This saves my day. I've been feeling freezing using the IDEA with Flutter on my Windows machine for months. The performance of the plugin is significantly improved with the PR.

AlexV525:
> @AlexV525 thank you, code and testing looks good. Do validate that with the daily released that the issues are resolved and you don't see any other issues.

Because it's not merged so I'm using my local compiled version and see no issues regarding DTD & EDT so far.

jwren:
> > @AlexV525 thank you, code and testing looks good. Do validate that with the daily released that the issues are resolved and you don't see any other issues.
> 
> Because it's not merged so I'm using my local compiled version and see no issues regarding DTD & EDT so far.

Thank you again!

helin24:
Thanks for the change! I did not see the original issue, so I can't verify myself, but it sounds promising if this is working well for you.

I'm wondering if adding some logs here might help us debug if there are problems? e.g. logging the successful path to make sure we're getting through the steps.

AlexV525:
> I'm wondering if adding some logs here might help us debug if there are problems? e.g. logging the successful path to make sure we're getting through the steps.

I'm not familiar with the logging system of the plugin. Are you indicating we can add logs like *DTD established after x seconds*? (I'd love to investigate this if we have a separate detailed issue!)

helin24:
I've been transitioning classes to use `PluginLogger` (see for example https://github.com/flutter/flutter-intellij/pull/8452) - more background here: https://github.com/flutter/flutter-intellij/issues/8369

And yes, your example of logging is what I was thinking. Do you think that could help for this situation? We have a complex setup for DTD and DevTools so I feel like more info here would be helpful for our own understanding of what happens here and for potential debugging.

Issue #8480, opened 2025-08-20
Bug when I create new flutter project


lvtdel:
## What happened

I created a new Flutter project. After that, IntelliJ showed a notification and threw an exception.

## Version information

IntelliJ IDEA `2025.2` • Flutter plugin `io.flutter 87.1` • Dart plugin `252.24322.5`

Flutter 3.32.8

## Exception

null

```
java.lang.IllegalStateException: Background write action is not permitted on this thread. Consider using `backgroundWriteAction`, or switch to EDT
	at com.intellij.openapi.application.impl.ApplicationImpl.checkWriteActionAllowedOnCurrentThread(ApplicationImpl.java:1092)
	at com.intellij.openapi.application.impl.ApplicationImpl.runWriteAction(ApplicationImpl.java:1099)
	at io.flutter.utils.OpenApiUtils.safeRunWriteAction(OpenApiUtils.java:94)
	at io.flutter.utils.FlutterModuleUtils.enableDartSDK(FlutterModuleUtils.java:418)
	at io.flutter.FlutterInitializer.executeProjectStartup(FlutterInitializer.java:110)
	at io.flutter.FlutterProjectActivity.execute$suspendImpl(FlutterProjectActivity.kt:21)
	at io.flutter.FlutterProjectActivity.execute(FlutterProjectActivity.kt)
	at com.intellij.ide.startup.impl.StartupManagerImplKt$launchActivity$1.invokeSuspend(StartupManagerImpl.kt:511)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:613)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:1189)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:778)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:765)
```

helin24:
Maybe related to one of our recent thread changes? https://github.com/flutter/flutter-intellij/pull/8458? @jwren

helin24:
[...]

PR #8490, opened 2025-08-27
Write action fix in FlutterModuleUtils

jwren:
This resolves https://github.com/flutter/flutter-intellij/issues/8480

### DIFF
--- a/src/io/flutter/utils/FlutterModuleUtils.java
+++ b/src/io/flutter/utils/FlutterModuleUtils.java
@@ -9,6 +9,7 @@ import com.intellij.execution.RunManager;
 import com.intellij.execution.RunnerAndConfigurationSettings;
 import com.intellij.facet.Facet;
 import com.intellij.facet.FacetManager;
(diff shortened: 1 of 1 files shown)


src/io/flutter/utils/FlutterModuleUtils.java
```diff
@@ -418,10 +419,18 @@ public static void enableDartSDK(@NotNull Module module) {
...
-      OpenApiUtils.safeRunWriteAction(() -> {
-        DartPlugin.ensureDartSdkConfigured(module.getProject(), dartSdkPath);
-        DartPlugin.enableDartSdk(module);
-      });
+
+      // Wrap the write action in a thread-safe way
+      // See https://github.com/flutter/flutter-intellij/issues/8480
+      final Application application = ApplicationManager.getApplication();
+      if (application != null) {
+        application.invokeLater(() -> {
```
pq, line 427: Interesting! Is this something we should be doing in general for write actions that don't require synchronicity?

PR #8526, opened 2025-09-12
Convert run files to use PluginLogger

helin24:
Part of #8369

### DIFF
--- a/src/io/flutter/run/FlutterReloadManager.java
+++ b/src/io/flutter/run/FlutterReloadManager.java
@@ -143,11 +143,7 @@ public class FlutterReloadManager {
         }
         catch (Throwable t) {
           // A catch-all, so any exceptions don't bubble through to the users.
-          if (FlutterSettings.getInstance().isFilePathLoggingEnabled()) {
-            LOG.warn("Exception from FlutterReloadManager", t);
-          } else {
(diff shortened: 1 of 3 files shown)


src/io/flutter/run/FlutterReloadManager.java
```diff
@@ -161,7 +166,11 @@ public void afterActionPerformed(@NotNull AnAction action, @NotNull AnActionEven
           handleSaveAllNotification(eventEditor);
         }
         catch (Throwable t) {
-          FlutterUtils.warn(LOG, "Exception from hot reload on save", t);
+          if (FlutterSettings.getInstance().isFilePathLoggingEnabled()) {
```
pq: Where did we land on pushing this conditional logic into the `warn` (and `error`) call(s)?

helin24: 🤦 I completely forgot I added a method for this. `FlutterUtils.warn` with a `sanitizePaths` option, but looking at it again, it may not be safe and could be improved.

pq: Cool. Thanks for looking!

helin24: See https://github.com/flutter/flutter-intellij/pull/8534 instead

PR #8563, opened 2025-09-30
Update Android Studio version

helin24:
I removed code that seems to be related to making gradle files editable within a Flutter project window as the Android Studio code has changed. I will record some additional details in my internal notes.

I also removed nullability annotations that caused errors during `./gradlew test`. These are mostly complaints that we are adding an annotation like `String @NotNull []` or `List<@NotNull Runnable>`. I'm not sure why these are considered invalid with the new platform version - AI suggests it is due to default JDK difference, but I'm also not sure how to update my IDE settings to reflect this supposed difference. In my view, all of these instances with errors are still relevant annotations (e.g. you want to know that the container is not null as well as all of the items within it). 🤷

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,8 +4,8 @@
 # found in the LICENSE file.
 #
 
-ideaVersion=2025.1.3.4
-dartPluginVersion= 251.27623.5
+ideaVersion=2025.2.1.3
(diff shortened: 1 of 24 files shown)


helin24:
Closing in favor of https://github.com/flutter/flutter-intellij/pull/8565

PR #8565, opened 2025-10-01
Update Android Studio version

helin24:
I removed code that seems to be related to making gradle files editable within a Flutter project window as the Android Studio code has changed. I will record some additional details in my internal notes.

There were also test compilation failures that suggested the annotations library wasn't required correctly. (see https://stackoverflow.com/questions/62066166/string-notnull-syntax-causes-notnull-not-applicable-to-type-use-in-in - thanks @pq). When I asked gemini about this, it suggested that we were compiling production source code twice for tests, because the test source sets included `src` as well. I guess it turns out that this isn't needed and the dependency on the production source code is implied; additionally, compiling the production source during test is bad because it can be compiled in the wrong environment. I think this is documentation that includes this info if we want to investigate further at a later time: https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_source_sets

### DIFF
--- a/src/io/flutter/utils/AndroidLocationProvider.java
+++ b/src/io/flutter/utils/AndroidLocationProvider.java
@@ -23,10 +23,14 @@ public class AndroidLocationProvider implements BuildModelContext.ResolvedConfig
   @Nullable
   @Override
   public VirtualFile getGradleBuildFile(@NotNull Module module) {
-    GradleModuleModel moduleModel = GradleProjectSystemUtil.getGradleModuleModel(module);
-    if (moduleModel != null) {
-      return moduleModel.getBuildFile();
-    }
+    // TODO(helin24): Delete this code (and potentially related code) if commenting out has no negative impact on Android editing.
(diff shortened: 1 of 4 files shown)


src/io/flutter/utils/AndroidLocationProvider.java
```diff
@@ -23,10 +23,14 @@ public class AndroidLocationProvider implements BuildModelContext.ResolvedConfig
   @Nullable
   @Override
   public VirtualFile getGradleBuildFile(@NotNull Module module) {
-    GradleModuleModel moduleModel = GradleProjectSystemUtil.getGradleModuleModel(module);
-    if (moduleModel != null) {
-      return moduleModel.getBuildFile();
-    }
+    // TODO(helin24): Delete this code (and potentially related code) if commenting out has no negative impact on Android editing.
+    // I believe this is to make gradle files show up nicely when a flutter project is opened, but this functionality already does not work
+    // and is not needed if we are recommending users edit Android files in a separate project window.
```
pq, line 32: Thanks for inlining this context!

Issue #8556, opened 2025-09-26
When opening a new Flutter project in Android Studio, open top level directory


helin24:
Currently the android subdirectory gets opened instead

PR #8573, opened 2025-10-09
Open project view for flutter projects

helin24:
This shows the project view for flutter projects on first open:
Screenshot 2025-10-09 at 9 45 30 AM

There's a problem with this fix though, in this scenario:
1. Open the flutter project (this will open to the project view correctly)
2. Close the project
3. Open the project's Android subdirectory (this will open to the android view correctly)
4. Open the top-level flutter project again - this will open to the android view, which isn't what we want

So I don't even know if I want to commit this code, because it feels like we are trying to manipulate a mechanism that's already pretty complicated. Any opinions are welcome.

(mostly) Fixes [8556](https://github.com/flutter/flutter-intellij/issues/8556)


helin24:
Thanks for the suggestions! I agree, it should be helpful for the majority of cases. When we have analytics, it would be nice to know how often users are opening the android subdirectory and using that workflow for editing native code.

PR #8576, opened 2025-10-12
💄 Set proper foreground & hover colors for the device selector action

### LINKED ISSUE #8572: Name of the selected device is not visible in toolbar
IntelliJ IDEA 2025.2.3 (Community Edition) for Windows + Flutter Plugin 88.0.0 (the latest)

Name of the selected device is not visible in toolbar if using IDE's default light theme

Image

AlexV525:
Fixes #8572

The request sets the foreground color and hover color consistently across all circumstances.

|     | Before | After |
|----|----|----|
| Dark | image | image |
| Light | image | image |
| Light with Light Header | image | image |
| High Contrast | image | image |

---

<details>

</details>

### DIFF
--- a/src/io/flutter/actions/DeviceSelectorAction.java
+++ b/src/io/flutter/actions/DeviceSelectorAction.java
@@ -91,7 +91,7 @@ public class DeviceSelectorAction extends AnAction implements CustomComponentAct
    * @return A {@link Color} suitable for toolbar text that adapts to the current theme,
    *         including configurations like light themes with dark headers.
    */
-  private static @NotNull Color getToolbarForegroundColor() {
+  @NotNull Color getToolbarForegroundColor() {
     return JBColor.namedColor(TOOLBAR_FOREGROUND_KEY, UIUtil.getLabelForeground());
   }
 
@@ -107,7 +107,7 @@ public class DeviceSelectorAction extends AnAction implements CustomComponentAct
    * @return A {@link Color} suitable for toolbar icon button hover states that adapts to the
    *         current theme, ensuring consistency with other toolbar actions.
    */
-  private static @NotNull Color getToolbarHoverBackgroundColor() {
+  @NotNull Color getToolbarHoverBackgroundColor() {
     return JBColor.namedColor(TOOLBAR_ICON_HOVER_BACKGROUND_KEY, JBUI.CurrentTheme.ActionButton.hoverBackground());
   }
 
--- a/testSrc/unit/io/flutter/actions/DeviceSelectorActionTest.java
+++ b/testSrc/unit/io/flutter/actions/DeviceSelectorActionTest.java
@@ -7,11 +7,10 @@ package io.flutter.actions;
 
 import com.intellij.util.ui.JBUI;
 import com.intellij.util.ui.UIUtil;
-import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.NotNull;
 import org.junit.Test;
 
 import java.awt.*;
-import java.lang.reflect.Method;
 
 import static org.junit.Assert.*;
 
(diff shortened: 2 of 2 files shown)


src/io/flutter/actions/DeviceSelectorAction.java
```diff
@@ -97,14 +99,26 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
     final JButton button = new JButton() {
       @Override
       protected void paintComponent(@NotNull Graphics g) {
         if (getModel() instanceof ButtonModel m && m.isRollover()) {
           final @NotNull Graphics2D g2 = (Graphics2D)Objects.requireNonNull(g.create());
           g2.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
-          g2.setColor(JBUI.CurrentTheme.ActionButton.hoverBackground());
+          // Use MainToolbar hover background to adapt to the toolbar theme (e.g., dark header with light theme)
+          final JBColor hoverColor = JBColor.namedColor(
+            "MainToolbar.Icon.hoverBackground",
```
helin24: I'm nervous about having plain strings to identify colors in these spots - do you (or AI) know how we can find these key names in case they change in the future? What happens if the key is not findable? I wonder if we could instead pull out smaller functions like `getHoverColor` and add unit tests to make sure we're able to get a reasonable color?

Separately, if there are IJ platform constants that we could use here instead, that would be ideal. If not, then let's at least pull out as our own constants in this class.

AlexV525: > I'm nervous about having plain strings to identify colors in these spots - do you (or AI) know how we can find these key names in case they change in the future?

It seems to find keys from https://github.com/JetBrains/intellij-community/blob/idea/252.26830.84/platform/platform-resources/src/themes/HighContrast.theme.json. I didn't see a clear documentation that declares color properties. But I've added some descriptions to explain the behavior clearly.

> What happens if the key is not findable?

It will fall back to a default color value.

> I wonder if we could instead pull out smaller functions like `getHoverColor` and add unit tests to make sure we're able to get a reasonable color?

I've added some unit tests.

> Separately, if there are IJ platform constants that we could use here instead, that would be ideal. If not, then let's at least pull out as our own constants in this class.

Unfortunately, from my understanding and during my code, I found no accurate constant members for these colors.

src/io/flutter/actions/DeviceSelectorAction.java
```diff
@@ -54,6 +56,22 @@ public class DeviceSelectorAction extends AnAction implements CustomComponentAct
   private static final @NotNull Icon DEFAULT_DEVICE_ICON = FlutterIcons.Mobile;
   private static final @NotNull Icon DEFAULT_ARROW_ICON = IconUtil.scale(AllIcons.General.ChevronDown, null, 1.2f);
 
+  /**
+   * Theme property key for the main toolbar foreground color.
+   * This key is used to retrieve the appropriate text color for toolbar components,
+   * ensuring proper visibility in all theme configurations (e.g., light theme with dark header).
+   * If this key is not found in the current theme, falls back to the standard label foreground color.
```
helin24: Let's take out this last line about the fallback as this is mentioned in the method that uses the key and isn't relevant to the key delcaration

src/io/flutter/actions/DeviceSelectorAction.java
```diff
@@ -54,6 +56,22 @@ public class DeviceSelectorAction extends AnAction implements CustomComponentAct
...
+   */
+  private static final String TOOLBAR_FOREGROUND_KEY = "MainToolbar.foreground";
+
+  /**
+   * Theme property key for the main toolbar icon hover background color.
+   * This key is used to retrieve the appropriate hover background color for toolbar icon buttons,
+   * ensuring consistency with other toolbar actions in all theme configurations.
+   * If this key is not found in the current theme, falls back to the standard action button hover background.
```
helin24: Take out this fallback comment as well

testSrc/unit/io/flutter/actions/DeviceSelectorActionTest.java
```diff
@@ -0,0 +1,199 @@
...
+   */
+  @Nullable
+  private Color invokeGetToolbarForegroundColor() throws Exception {
+    final Method method = DeviceSelectorAction.class.getDeclaredMethod("getToolbarForegroundColor");
+    method.setAccessible(true);
+    return (Color)method.invoke(null);
+  }
+
+  /**
+   * Invokes the private static getToolbarHoverBackgroundColor method via reflection.
```
helin24: Is it necessary to make these static methods? Could this test file be simpler if we made them non-static methods?

AlexV525: Agreed!

helin24 (approved):
LGTM! If tests pass I'll land later.

PR #8579, opened 2025-10-15
Fix dev build by using version 2025.2.1.5

helin24:
The dev build was not able to pull down the previous version of Android Studio (2025.2.1.3) and there are newer versions now at https://plugins.jetbrains.com/docs/intellij/android-studio-releases-list.html. I checked that this works from a local invocation of the dev build.


pq:
Nice!

PR #8580, opened 2025-10-15
Add forgotten changelog line for project view

helin24:
Follow up to https://github.com/flutter/flutter-intellij/pull/8573

PR #8581, opened 2025-10-15
Make changelog more consistent

helin24:
Follow up to https://github.com/flutter/flutter-intellij/pull/8580 to make periods and tense consistent.


pq (approved):
A thousand thank yous!

PR #8582, opened 2025-10-16
Add a gradle timeout property

helin24:
I'm trying to fix the dev build again by seeing if I can increase reliability by specifying a longer timeout.


helin24:
Closing because the additional changes in here do not seem to be necessary at the moment.

Issue #7599, opened 2024-08-08
"Open Android module in Android Studio" & "Open iOS module in Xcode" missing after upgrading 80.x


keybow007:
## What happened

"Open Android module in Android Studio" and "Open ios module in XCode" have gone after upgrading Flutter plugin to 80.x, which caused significant inconvenience. 

As shown in the attached images, VSCode offers similar functionality through "Open in Android" and "Open in XCode." The absence of such features in Android Studio is quite surprising and has led me to seriously consider switching to VSCode. I kindly request the prompt restoration of these essential features in Android Studio.

![ScreenShot 2024-07-18 23 14 49](https://github.com/user-attachments/assets/301900d7-9232-43c9-bf6d-fc5d8e127c7d)
![ScreenShot 2024-07-18 23 14 56](https://github.com/user-attachments/assets/139995b6-1a7e-469b-96a0-2c4b1abcce8e)

## Version info

Android Studio Koala | 2024.1.1 Patch 1 • Flutter plugin io.flutter 81.0.2 • Dart plugin 241.18808
Error getting Flutter plugin update to 80.x.

```
[✓] Flutter (Channel stable, 3.24.0, on macOS 14.5 23F79 darwin-x64, locale
    ja-JP)
    • Flutter version 3.24.0 on channel stable at /Users/telulullc/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 80c2e84975 (9 days ago), 2024-07-30 23:06:49 +0700
    • Engine revision b8800d88be
    • Dart version 3.5.0
    • DevTools version 2.37.2

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/telulullc/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • Java binary at: /Applications/Android
      Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build
      17.0.11+0-17.0.11b1207.24-11852314)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 15.4)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15F31d
    • CocoaPods version 1.15.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2023.3)
[...]

PR #8584, opened 2025-10-17
Use new API for opening android module

helin24:
I think this fixes https://github.com/flutter/flutter-intellij/issues/7599.

I checked that I can now use Tools > Flutter > Open Android module in Android Studio, and made sure this works with a new project and an existing project. There may be more testing needed, e.g. is this supposed to work in IntelliJ also?

If we want to use this again, we should probably also re-enable a menu when a user right clicks the android subdirectory, and update comments related to https://github.com/flutter/flutter-intellij/commit/dd5a521382bf20f6cfb1121851f38c398fc9dbaa

### DIFF
--- a/src/io/flutter/actions/OpenAndroidModule.java
+++ b/src/io/flutter/actions/OpenAndroidModule.java
@@ -5,7 +5,6 @@
  */
 package io.flutter.actions;
 
-import com.android.tools.idea.gradle.project.importing.GradleProjectImporter;
 import com.intellij.ide.actions.OpenFileAction;
 import com.intellij.ide.impl.OpenProjectTask;
 import com.intellij.openapi.actionSystem.ActionPlaces;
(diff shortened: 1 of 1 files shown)


src/io/flutter/actions/OpenAndroidModule.java
```diff
@@ -61,8 +61,8 @@ private static void openOrImportProject(@NotNull VirtualFile projectFile,
     if (canImportAsGradleProject(projectFile)) {
       VirtualFile target = findGradleTarget(projectFile);
       if (target != null) {
-        GradleProjectImporter gradleImporter = GradleProjectImporter.getInstance();
-        gradleImporter.importAndOpenProjectCore(null, true, projectFile);
+        com.intellij.ide.impl.ProjectUtil.openOrImport(projectFile.toNioPath(), OpenProjectTask.build().withForceOpenInNewFrame(forceOpenInNewFrame));
```
pq, line 64: I'd be inclined to replace this qualified name with an `import` statement? Or maybe there's some reason not to?

pq:
I haven't dug in but the new build failure seems surprising?

src/io/flutter/actions/OpenAndroidModule.java
helin24: After looking into this more, I think my solution is wrong because this method `openOrImport` is already being imported statically and is used below. There's some additional complexity here about whether a directory should be opened as though it's a project in itself or as a sub-project, but I'm not sure what all of the consequences are. I think I'll mark this PR a draft for now.

Piinks:

helin24:
I'm going to close this for now. This needs more investigation.

Issue #8583, opened 2025-10-17
Command "Open Flutter DevTools in browser" does not open DevTools with VM service connection


elliette:
Steps to reproduce:

1. Run an app and then open one of the DevTools side panels
2. Double-shift to open the universal search dialog and search for command "Open Flutter DevTools in browser"
3. The Flutter DevTools which opens does not have a VM service connection. 

Expected behavior: Open DevTools with a VM service connection using the same URI that is embedded in the side panel

PR #8585, opened 2025-10-17
Fix "Open Flutter DevTools" action opening a instance DevTools disconnected from the VM service

elliette:
Work towards https://github.com/flutter/flutter-intellij/issues/8583
Work towards https://github.com/flutter/devtools/issues/9282

Currently, if a user triggers the "Open Flutter DevTools" command from IntelliJ, it will open up a DevTools instance that is not connected to the VM service, regardless of whether or not the user is already running their app and has a connected DevTools instance open in their IDE. 

This is because it triggers the `OpenDevToolsAction` registered in plugin.xml, which uses the default constructor for the action (which does not have a param for the connector). Therefore, when the action is triggered, we check if the connector is `null`, and if so we first check if there is a running app, and therefore connector, that we can use. 

I'm not sure how to determine the **correct** Flutter app if there are more than one running. Any suggestions here would be appreciated. 

Note: This is part of the work to add better Network panel support to DevTools. We would like to either fix the issues of copy-paste not working when DevTools is embedded in IntelliJ, or as a not-ideal but last case scenario show a banner in DevTools that tells them copy-paste does not work in IntelliJ, and that they should open DevTools in their browser instead (that is what this change is addressing). 

FYI @srawlins

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -40,7 +40,7 @@ plugins {
   // https://plugins.gradle.org/plugin/org.jetbrains.intellij.platform
   // https://plugins.gradle.org/plugin/org.jetbrains.kotlin.jvm
   id("java") // Java support
-  id("org.jetbrains.intellij.platform") version "2.7.2" // IntelliJ Platform Gradle Plugin
+  id("org.jetbrains.intellij.platform") version "2.10.2" // IntelliJ Platform Gradle Plugin
   id("org.jetbrains.kotlin.jvm") version "2.2.0" // Kotlin support
   id("org.jetbrains.changelog") version "2.2.0" // Gradle Changelog Plugin
   idea // IntelliJ IDEA support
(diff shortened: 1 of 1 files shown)


helin24:
Choosing the first one for now seems fine!

I believe we have an open issue of auditing all of the DevTools windows and deciding which VM instance to show. I think @AlexV525 added some functionality to the inspector to show a different browser tab for each VM instance if multiple are running. Would it potentially make sense for DevTools / or a meta-DevTools to manage multiple VM instances at once?

Another option - are we thinking users would choose to open DevTools in browser if something doesn't work? Perhaps we could save the DevTools URLs that are currently trying to be opened in embedded windows and show those as options of URLs to open in the normal browser.

src/io/flutter/run/OpenDevToolsAction.java
```diff
@@ -72,6 +73,17 @@ public void actionPerformed(@NotNull final AnActionEvent event) {
...
     }
 
+    // This action is registered in plugin.xml with the default constructor.
+    // Therefore, if a user triggers this from the IDE, even if there is a
+    // running Flutter app myConnector will be null. In that case, check for a
+    // Flutter app first and use its connector instead.
+    if (myConnector == null) {
+      final List<FlutterApp> apps = FlutterApp.allFromProjectProcess(project);
+      if (!apps.isEmpty()) {
+        myConnector = apps.get(0).getConnector();
```
AlexV525, line 86: Is it possible to find the currently activated app instead of always using the first?

elliette: I couldn't figure out a way to do that unfortunately 😮‍💨 I added a TODO though to capture that.

PR #8590, opened 2025-10-23
Update intellij gradle plugin version and allow compatibility problems temporarily

helin24:


helin24:
The problematic compatibility issue is this:

```
    #Package 'de.roderick' is not found
        Package 'de.roderick' is not found along with its 4 classes.
        Probably the package 'de.roderick' belongs to a library or dependency that is not resolved by the checker.
        It is also possible, however, that this package was actually removed from a dependency causing the detected problems. Access to unresolved classes at runtime may lead to **NoSuchClassError**.
        The following classes of 'de.roderick' are not resolved:
          Class de.roderick.weberknecht.WebSocketException is referenced in
            io.flutter.FlutterInitializer.lambda$sendThemeChangedEvent$6(JsonObject params, DartToolingDaemonService dtdService) : void
            io.flutter.analytics.UnifiedAnalytics.makeUnifiedAnalyticsRequest(String requestName, DartToolingDaemonService service, JsonObject params) : CompletableFuture
            io.flutter.run.daemon.FlutterAppDaemonEventListener.sendDtdRequest(DtdRequest dtdRequest, JsonObject params) : void
            org.dartlang.vm.service.VmServiceBase.connect(String url) : VmService
            org.dartlang.vm.service.internal.WebSocketRequestSink.add(JsonObject json) : void
            org.dartlang.vm.service.internal.WebSocketRequestSink.close() : void
          Class de.roderick.weberknecht.WebSocket is referenced in
            org.dartlang.vm.service.VmServiceBase.connect(String url) : VmService
            org.dartlang.vm.service.internal.WebSocketRequestSink.<init>(WebSocket webSocket)
            org.dartlang.vm.service.internal.WebSocketRequestSink.add(JsonObject json) : void
            org.dartlang.vm.service.internal.WebSocketRequestSink.close() : void
            org.dartlang.vm.service.internal.WebSocketRequestSink.webSocket : WebSocket
          Class de.roderick.weberknecht.WebSocketEventHandler is referenced in
            org.dartlang.vm.service.VmServiceBase$1
          Class de.roderick.weberknecht.WebSocketMessage is referenced in
            org.dartlang.vm.service.VmServiceBase$1.onMessage(WebSocketMessage message) : void
```

This is also being used in the Dart plugin, e.g. the usage in `FlutterInitializer` sending a DTD request is catching an exception from DTD service in Dart. So we can't really remove this dependency without changing the Dart plugin too.

This is only occurring for Android Studio 2025.2, not the earlier versions. I don't understand why that would be, since this library is in the Dart plugin as a jar file.

I've tried adding a ignore file for this specific error so we can prevent any further compatibility problems, but that isn't working currently or I'm doing it incorrectly. (https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/2046)

Potentially we could also try adding the weberknecht depedency as a jar file into the Flutter plugin too, but it's suspicious that this isn't available on maven any more - we should consider changing to using an up to date websocket connectivity package.

At the moment, there's also potentially timeouts resulting from the following:
- Android Studio download just taking a long time. There's a way we can change the timeout on this.
- `verifyPlugin` also generally taking a long time because we have verify run against `recommended()` IDEs, so it's checking against 3 of them at the moment. We could reduce this or start making branches if we want to be more specific.

Apologies for all the text. I'm hoping someone else has an elegant idea of how to fix this aside from turning off the compatibility problems check.

PR #8593, opened 2025-10-27
Add an ignore problems file for webernecht compatibility problem

helin24:
Follow up to https://github.com/flutter/flutter-intellij/pull/8590

This seems at least better than ignoring compatibility problems entirely.


helin24:
> would it be possible to enable all of the FailureLevel enums with appropriate annotations in the verify-ignore-problems.txt file?

I don't think so, see https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/2046#issuecomment-3448971926. Since most of our warnings are API-related, it won't help us much.

Issue #8587, opened 2025-04-02
Support hosting the Widget Previewer in Android Studio / Intellij


bkonyi:
Initial support has landed in VSCode and looks like this:

https://github.com/user-attachments/assets/50a4d280-c15e-44e9-b063-e1b313ad7cb0

To match this, we need to be able to:

- Launch and manage the lifecycle of `flutter widget-preview start --web-server`.
- Open an iframe or browser within Intellij to view the preview application.
- Send IDE navigation events via the `Editor` DTD service to update which previews are visible based on the currently selected file.

helin24:
@bkonyi

1. What is the min flutter SDK version that supports this?
2. Is there an icon that would be appropriate for our tool menu bar?

<img width="81" height="402" alt="Image" src="https://github.com/user-attachments/assets/e9e7cd7b-6a01-47b9-bfe1-89b2539cc2f7" />

PR #8595, opened 2025-10-29
Add widget previewer tool window

helin24:
Changes include:
- Logic for starting the widget previewer server (slightly done by AI)
- A new tool window (mostly done by AI)
- Refactoring browsers to generalize showing a URL (previously we are always showing DevTools URLs)

Screenshot 2025-10-29 at 11 30 44 AM

Things that may change further:
- Specific version of Flutter SDK for gating
- Switching between dark mode/light mode

Addresses https://github.com/flutter/flutter-intellij/issues/8587


jwren (approved):

src/io/flutter/FlutterBundle.properties
```diff
@@ -187,3 +187,10 @@ icon.preview.disallow.flutter_vector_icons=Package "flutter_vector_icons" cannot
 icon.preview.disallow.material_design_icons_flutter=Package "material_design_icons_flutter" always displays blank icon previews.
 icon.preview.disallow.package.title=Unsupported icon package
 icon.preview.analysis=Checking icons...
+
+widget.preview.initializing=Initializing Flutter Widget Preview...
+widget.preview.starting=Starting Flutter widget preview web server...
```
pq: Should it be "Widget Preview"? Or the previous uncapitalized?

src/io/flutter/sdk/FlutterSdkVersion.java
```diff
@@ -60,6 +60,10 @@ public final class FlutterSdkVersion implements Comparable<FlutterSdkVersion> {
   @NotNull
   public static final FlutterSdkVersion MIN_SUPPORTS_PROPERTY_EDITOR = new FlutterSdkVersion("3.32.0-0.1.pre");
 
+  @NotNull
+  public static final FlutterSdkVersion MIN_SUPPORTS_WIDGET_PREVIEWER = new FlutterSdkVersion("3.37.0");
+  //public static final FlutterSdkVersion MIN_SUPPORTS_WIDGET_PREVIEWER = new FlutterSdkVersion("3.39.0");
```
pq: Remove commented out line?

src/io/flutter/view/BrowserUrlProvider.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package io.flutter.view;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Saves a browser URL that may be editable depending on additional settings from the IDE or running applications.
+ */
+public interface BrowserUrlProvider {
+  String widgetId = "";
```
pq: I guess this could be `@NotNull` too?

src/io/flutter/sdk/FlutterSdkVersion.java
```diff
@@ -60,6 +60,10 @@ public final class FlutterSdkVersion implements Comparable<FlutterSdkVersion> {
   @NotNull
   public static final FlutterSdkVersion MIN_SUPPORTS_PROPERTY_EDITOR = new FlutterSdkVersion("3.32.0-0.1.pre");
 
+  @NotNull
+  public static final FlutterSdkVersion MIN_SUPPORTS_WIDGET_PREVIEWER = new FlutterSdkVersion("3.37.0");
```
bkonyi: This should match VSCode's minimum version requirement of the 3.38.0 beta: https://github.com/Dart-Code/Dart-Code/blob/master/src/shared/capabilities/flutter.ts#L18

src/io/flutter/FlutterBundle.properties
helin24: @bkonyi do you have an opinion here? I'll capitalize for now.

src/io/flutter/sdk/FlutterSdkVersion.java
helin24: Done, this was while deciding what version to use.

src/io/flutter/view/BrowserUrlProvider.java
helin24: I think I actually don't need this.

src/io/flutter/sdk/FlutterSdkVersion.java
helin24: Thanks!

src/io/flutter/FlutterBundle.properties
bkonyi: Capitalized, I think!

src/io/flutter/sdk/FlutterSdk.java
```diff
@@ -415,6 +415,19 @@ public FlutterCommand flutterTest(@NotNull PubRoot root, @NotNull VirtualFile fi
     return new FlutterCommand(this, root.getRoot(), FlutterCommand.Type.TEST, args.toArray(new String[]{ }));
   }
 
+  @NotNull
+  public FlutterCommand widgetPreview(@NotNull PubRoot root, boolean isVerboseMode) {
+    final List<String> args = new ArrayList<>();
+    args.add("start");
+    args.add("--web-server");
+    args.add("--machine");
```
bkonyi, line 423: We should also pass `--dtd-url=<url>` here and, if the plugin hosts a DevTools instance, `--devtools-server-address=<url>`.

Does the Intellij plugin provide the DTD `Editor` service?

src/io/flutter/view/WidgetPreviewUrlProvider.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @Override
+  public void setWidgetId(@NotNull String widgetId) {
+
+  }
+
+  @Override
+  public @NotNull String getBrowserUrl() {
+    return url + "?theme=" + (isBackgroundBright ? "light" : "dark");
```
bkonyi, line 27: Do we support background and foreground colors for theming DevTools? If so, we should also pass those values here.

helin24:
Thanks for the reviews! I'm going to land this for now for the dev build over the weekend, but I'll address the outstanding comments from Ben next week.

src/io/flutter/sdk/FlutterSdk.java
helin24: Ah okay, both should be available. How does the the widget preview interact with DevTools?

IntelliJ is intended to serve as an Editor, at least if we're thinking of the same thing (https://github.com/dart-lang/sdk/blob/bc7e4d9fb583c9edeb90393f65989767e865e56f/pkg/dtd_impl/dtd_common_services_editor.md?plain=1#L4), but looking at the code now I'm not sure if this was implemented.

src/io/flutter/view/WidgetPreviewUrlProvider.java
helin24: We only use the params `backgroundColor` (hex code) and `theme` ('light' or 'dark') from the IJ plugin. I don't know if DevTools has other params available that we're not using though. @elliette do you know?

src/io/flutter/sdk/FlutterSdk.java
bkonyi: > How does the the widget preview interact with DevTools?

For the current beta, it doesn't directly interact with DevTools and passing this flag just prevents Flutter Tools from serving another DevTools instance. In the near future, this DevTools instance will be embedded in the widget previewer in an iframe which displays the widget inspector.

> IntelliJ is intended to serve as an Editor, at least if we're thinking of the same thing (https://github.com/dart-lang/sdk/blob/bc7e4d9fb583c9edeb90393f65989767e865e56f/pkg/dtd_impl/dtd_common_services_editor.md?plain=1#L4), but looking at the code now I'm not sure if this was implemented.

Ah, it'd be great if this was implemented. The widget previewer can provide additional functionality if this service is available, including filtering previews based on the currently selected file and jumping to source from file URIs in stack traces displayed for previews which throw exceptions.

helin24: @bkonyi could you file an issue to this repo about implementing the editor services and what functionality will be available for each service that's relevant, maybe with examples from VS Code if you know of them?

PR #8599, opened 2025-11-03
Send DTD and DevTools URI to widget preview

helin24:
Follow up to https://github.com/flutter/flutter-intellij/pull/8595

At the moment I don't see consequences to having or not having the DTD connection, but I expect that in the future, the widget preview will respond to theme change updates over DTD. (and potentially other things?)

Based on the conversation in https://github.com/flutter/flutter-intellij/pull/8595#discussion_r2482336086, the DevTools URI also does not have a visible consequence, but it should prevent flutter tools from starting its own DevTools instance when widget-preview is run.

### DIFF
--- a/src/io/flutter/widgetpreview/WidgetPreviewPanel.java
+++ b/src/io/flutter/widgetpreview/WidgetPreviewPanel.java
@@ -187,6 +187,8 @@ public class WidgetPreviewPanel extends SimpleToolWindowPanel implements Disposa
     });
   }
 
+  // TODO(https://github.com/flutter/flutter/issues/177945): Ideally widget preview would change colors based on theme changes events,
+  //  which we already send for the DevTools panels. If this is implemented then we can remove this listening code.
   private void listenForReload() {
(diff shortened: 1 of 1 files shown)


jwren (approved):

bkonyi:
> At the moment I don't see consequences to having or not having the DTD connection, but I expect that in the future, the widget preview will respond to theme change updates over DTD. (and potentially other things?)

If not given a DTD URI for the DTD instance hosted by the IDE, the widget previewer won't be able to respond to editor events (e.g., changing the set of previews displayed based on the currently selected file, responding to theme change events, etc), which would significantly reduce the UX when using the previewer embedded in the IDE.

helin24:
@bkonyi and I discussed a little further and decided I'll try to implement editor service RPCs in the plugin:
- `navigateToCode`
- `getActiveLocation`

And the events:
- activeLocationChanged
- themeChanged (this is lower priority)

They'll be in upcoming PRs if easy to do. Otherwise we'll decide where to prioritize.

Issue #8612, opened 2025-11-17
Show only last changelog item in what's new


helin24:
See https://github.com/flutter/dart-intellij-third-party/pull/81

PR #8622, opened 2025-11-26
Show only latest version change notes in what's new screen

helin24:
Addresses https://github.com/flutter/flutter-intellij/issues/8612


helin24:
We recently made it this way in the Dart plugin; that's what caused me to make this issue! 😆

helin24:
Uh oh verify has been failing. I guess that's a different task that I'll investigate next..

PR #8637, opened 2025-12-08
Reduce memory use of verify bot by deleting IDEs after verify

helin24:


helin24:
I'm not sure this is worth merging; I mainly wanted to look at the memory usage on this branch.

helin24:
Okay, I think we can land this, though it's not optimal.

Some background:
- We have `gradle verifyPlugin` running against `recommended` IDEs. For the current situation of building a plugin that is for 2024.3 to 2025.3, the intellij gradle extension will select a 243 build, a 251 build, and a 252 build, and verify against those.
- This task was causing an out of memory error because the GitHub actions bots don't have much memory. e.g. from checking memory during this PR:
  ```
  Check on space before verifyPluginProjectConfiguration\n
  Filesystem      Size  Used Avail Use% Mounted on
  /dev/root        72G   55G   17G  77% /
  tmpfs           7.9G   84K  7.9G   1% /dev/shm
  tmpfs           3.2G  1.1M  3.2G   1% /run
  tmpfs           5.0M     0  5.0M   0% /run/lock
  /dev/sdb16      881M   62M  758M   8% /boot
  /dev/sdb15      105M  6.2M   99M   6% /boot/efi
  /dev/sda1        74G  4.1G   66G   6% /mnt
  tmpfs           1.6G   12K  1.6G   1% /run/user/1001
  ```
  By the time we start `verifyPluginProjectConfiguration` there's already 55/72 G used. I'm not sure why that is - is it java? flutter?

The solution for now is to add a way to specify the versions to verify against in the github script and deleting the IDE downloads after verification is run for a specific version. This leaves the normal `gradle verifyPlugin` task the same, so it's easy for us to run manually from our own machines that aren't space-constrained. However, this isn't optimal for a few reasons:
- This adds a second place where we have to specify what versions to verify (in the github.sh script). Ideally I'd like gradle to recommend the versions and then delete after each set of verifications, but this doesn't seem possible.
- The deletion doesn't completely work (see the comment before `ide.deleteRecursively()`). Running twice in a row causes errors. (That doesn't really matter for the GitHub action because it will run different versions each time, but still...)
- Each verification against a version still adds on to memory usage, just less. Before this change, I think verifying for each IDE version was adding maybe 5G of memory; after this change, verifying for each IDE version is adding 2G of memory. I don't know what the other memory usage is coming from. So while this should enable us to verify more IDEs, it's not infinite space. For that, we would probably need to kill the gradle daemon and start it over again? I'm not sure.

helin24:
@jwren  would you mind re-reviewing?

Issue #8631, opened 2025-12-04
JmixLoggerWrapper should override '#setLevel(LogLevel)' - 2025.3.x platform issue


jwren:
New 2025.3 platform issue:

```
Please fix this issue: java.lang.Throwable: class com.haulmont.jmixstudio.logger.JmixLoggerWrapper should override '#setLevel(LogLevel)'
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:375)
	at com.intellij.openapi.diagnostic.Logger.setLevel(Logger.java:491)
	at io.flutter.logging.PluginLogger.updateLogLevel(PluginLogger.java:49)
	at io.flutter.FlutterInitializer.executeProjectStartup(FlutterInitializer.java:87)
	at io.flutter.FlutterProjectActivity.execute$suspendImpl(FlutterProjectActivity.kt:21)
	at io.flutter.FlutterProjectActivity.execute(FlutterProjectActivity.kt)
	at com.intellij.ide.startup.impl.StartupManagerImplKt$launchActivity$1.invokeSuspend(StartupManagerImpl.kt:511)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
```

helin24:
Is this from a gradle verify task or elsewhere? Trying to test this going forward.

PR #8639, opened 2025-12-09
Suppress IDE stack trace for setLevel failure

jwren:
The stack trace is effectively just the log message being excessively loud because of how IntelliJ handles exceptions passed to the logger.

This is a follow up on https://github.com/flutter/flutter-intellij/issues/8631

### DIFF
--- a/src/io/flutter/logging/PluginLogger.java
+++ b/src/io/flutter/logging/PluginLogger.java
@@ -45,13 +45,18 @@ public class PluginLogger {
 
   public static void updateLogLevel() {
     final Logger rootLoggerInstance = Logger.getInstance("io.flutter");
+    // Workaround for https://github.com/flutter/flutter-intellij/issues/8631
+    if (rootLoggerInstance.getClass().getName().equals("com.haulmont.jmixstudio.logger.JmixLoggerWrapper")) {
+      return;
+    }
     try {
       rootLoggerInstance.setLevel(FlutterSettings.getInstance().isVerboseLogging() ? LogLevel.ALL : LogLevel.INFO);
     }
     catch (Throwable e) {
-      // This can happen if the logger is wrapped by a 3rd party plugin that doesn't correctly implement setLevel.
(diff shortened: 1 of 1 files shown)


src/io/flutter/logging/PluginLogger.java
```diff
@@ -51,7 +51,7 @@ public static void updateLogLevel() {
     catch (Throwable e) {
       // This can happen if the logger is wrapped by a 3rd party plugin that doesn't correctly implement setLevel.
       // See https://github.com/flutter/flutter-intellij/issues/8631
-      Logger.getInstance(PluginLogger.class).warn("Failed to set log level", e);
+      // Use an empty catch block to suppress the exception.
```
helin24: can we warn with just the message and no exception?

jwren: I haven't tried that yet.  The goal here is to simply catch and log it without reporting it in the (!) on startup.

src/io/flutter/logging/PluginLogger.java
```diff
@@ -46,15 +46,32 @@ public class PluginLogger {
   public static void updateLogLevel() {
     final Logger rootLoggerInstance = Logger.getInstance("io.flutter");
     try {
-      rootLoggerInstance.setLevel(FlutterSettings.getInstance().isVerboseLogging() ? LogLevel.ALL : LogLevel.INFO);
+      if (isSetLevelOverridden(rootLoggerInstance)) {
+        rootLoggerInstance.setLevel(FlutterSettings.getInstance().isVerboseLogging() ? LogLevel.ALL : LogLevel.INFO);
```
pq: Could you try this and get rid of the reflection?

```java
try {
  rootLoggerInstance.setLevel(FlutterSettings.getInstance().isVerboseLogging() ? LogLevel.ALL : LogLevel.INFO);
} catch (Throwable _) {
  // Ignore
}
```

jwren: Reflection is removed.

PR #8645, opened 2025-12-10
Improve error messages in the build.gradle.kts

jwren:
Improve error messages for network-related failures during Android Studio download and IDEA path resolution in build.gradle.kts


helin24 (approved):
Thanks!

Issue #8657, opened 2025-12-14
io.flutter.dart.DtdUtils static field retains disposed Project


viuginick:
```
(root): java.lang.Class(io.flutter.dart.DtdUtils)
(static): java.util.concurrent.ConcurrentHashMap
table: java.util.concurrent.ConcurrentHashMap$Node[]
[]: java.util.concurrent.ConcurrentHashMap$Node
key: com.intellij.openapi.project.impl.ProjectImpl (disposed)
```

helin24:
Note that there is more info on an internal issue.

jwren:
Fix landed on `main`, closing.

PR #8658, opened 2025-12-16
Fix DtdUtils static field retaining disposed Project

jwren:
https://github.com/flutter/flutter-intellij/issues/8657

### DIFF
--- a/src/io/flutter/dart/DtdUtils.java
+++ b/src/io/flutter/dart/DtdUtils.java
@@ -14,17 +14,17 @@ import java.util.Map;
 import java.util.concurrent.*;
 
 public class DtdUtils {
-  private static final Map<Project, CompletableFuture<DartToolingDaemonService>> WAITERS = new ConcurrentHashMap<>();
+  // Visible for testing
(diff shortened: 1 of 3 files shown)


helin24:
I don't think this is a new issue, although is it newly surfacing in 2025.3? Are you thinking we need another release for this?

jwren:
> I don't think this is a new issue, although is it newly surfacing in 2025.3? Are you thinking we need another release for this?

I don't think that a release is needed to patch this in the short term.

I added a line in the changelog before landing.

src/io/flutter/dart/DtdUtils.java
```diff
@@ -42,9 +42,17 @@ public class DtdUtils {
       readyService.whenComplete((s, t) -> {
         poll.cancel(false);
         timeout.cancel(false);
-        if (t != null) {
```
pq, line 45: Is `t` now sure not to be null or was this test always unnecessary?

jwren: It is my belief that the conditional check here is what led to the leak.

pq: Ah. Interesting! Thanks for clarifying.

PR #8670, opened 2026-01-02
Fix ProcessAdapter deprecation warnings

jwren:
Replaces deprecated ProcessAdapter with ProcessListener. Reduces deprecated API usages by 33 in 2025.2 (future-proofing for when check is enabled).

https://github.com/JetBrains/intellij-community/blob/master/platform/util/src/com/intellij/execution/process/ProcessAdapter.java

### DIFF
--- a/src/io/flutter/actions/OpenInAppCodeAction.java
+++ b/src/io/flutter/actions/OpenInAppCodeAction.java
@@ -8,7 +8,7 @@ package io.flutter.actions;
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.process.ColoredProcessHandler;
-import com.intellij.execution.process.ProcessAdapter;
+import com.intellij.execution.process.ProcessListener;
 import com.intellij.execution.process.ProcessEvent;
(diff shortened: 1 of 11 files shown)


pq:
Cool!

Here's a thought... `ProcessAdapter` was nice in that it avoided a bunch of boilerplate which this PR adds in. How about introducing our own `ProcessAdapter` so we can reap the benefits (without depending on deprecated API)? Maybe in `io.flutter.utils`?

jwren:
I like the adapter pattern as well, but expectations are shifting under us here.  This PR is _almost_ free and moves our repo in a direction that will have it more accepted and green in the JB Marketplace.

Lets chat on VC, I can completely see both sides of the argument here, lets defer to the larger team.

pq:
> This PR is almost free 

I think that's debatable given our time talking about it 😄 -- but it's all good!  It's more than worth the time since it's a nice way to frame a larger and more important question about evolving coding standards.

My 2 cents is that while it may be free to add the boilerplate, we'll continue to pay the price in the ways we always have with unnecessary code (in the way of maintenance and understandability). I don't really see a shift here but I think it's an awesome conversation to have. Looking forward to it!

PR #8671, opened 2026-01-02
Add unit tests for JsonUtils

jwren:
Added JsonUtilsTest.java to test JsonUtils utility class. Covered getStringMember, getIntMember, and getKeySet methods.

### DIFF
--- a/testSrc/unit/io/flutter/utils/JsonUtilsTest.java
+++ b/testSrc/unit/io/flutter/utils/JsonUtilsTest.java
@@ -1,0 +1,41 @@
+package io.flutter.utils;
+
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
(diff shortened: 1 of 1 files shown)


testSrc/unit/io/flutter/utils/JsonUtilsTest.java
```diff
@@ -0,0 +1,41 @@
+package io.flutter.utils;
```
pq, line 1: Copyright?

pq (changes requested):

testSrc/unit/io/flutter/utils/JsonUtilsTest.java
jwren: 100%, this can and needs to be automated.

PR #8680, opened 2026-01-05
Refactor DiagnosticsNode to use SmartList usage patterns

jwren:

This change updates DiagnosticsNode to use List interface and SmartList implementation for properties and children, optimizing memory usage for tree nodes.


jwren:

PR #8682, opened 2026-01-06
Refactor EmbeddedJxBrowser to use SmartList

jwren:
Replaces ArrayList with SmartList for the inputs list in handleJxBrowserInstallationFailed. This list typically contains 0-2 elements, making it a good candidate for SmartList optimization.

### DIFF
--- a/src/io/flutter/jxbrowser/EmbeddedJxBrowser.java
+++ b/src/io/flutter/jxbrowser/EmbeddedJxBrowser.java
@@ -41,6 +41,7 @@ import com.intellij.ide.ui.UISettingsUtils;
 import javax.swing.*;
 import java.awt.*;
 import java.io.File;
+import com.intellij.util.SmartList;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


src/io/flutter/jxbrowser/EmbeddedJxBrowser.java
```diff
@@ -41,6 +41,7 @@
 import javax.swing.*;
 import java.awt.*;
 import java.io.File;
+import com.intellij.util.SmartList;
 import java.util.ArrayList;
```
pq, line 45: Can this import be removed?

pq (approved):
Cool!

src/io/flutter/jxbrowser/EmbeddedJxBrowser.java
jwren: No, it is a requirement

pq: Interesting! I don't see any `ArrayList` references after this change but maybe I'm missing something?

PR #8684, opened 2026-01-07
Refactor EmbeddedBrowser to use SmartList usage patterns

jwren:
Updates showMessageWithUrlLink and showMessage to use SmartList for labels. These lists typically contain only 1 element, making SmartList an ideal optimization.


jwren:

PR #8685, opened 2026-01-07
Refactor PubRootCache to use SmartList usage patterns

jwren:
Updates getRoots methods to use SmartList for collecting results. PubRoots are typically found in small numbers per module/project, making this memory optimization appropriate.


jwren:

PR #8686, opened 2026-01-07
Refactor FlutterRetargetAppAction to use SmartList usage patterns

jwren:
Updates myPlaces to use SmartList for storing action places. The list of places is typically small and fixed size, suitable for SmartList.


jwren:

PR #8689, opened 2026-01-07
Refactor StdoutJsonParser to use SmartList

jwren:
Replaces ArrayList with SmartList to optimize memory usage for lists that typically contain few elements.


jwren:

PR #8693, opened 2026-01-07
Refactor io.flutter.sdk to use SmartList

jwren:


jwren:

Issue #7336, opened 2024-04-09
Breakpoints don't work on a mounted dev drive


mvarendorff2:
## Steps to Reproduce

1. On Windows 11, create a Dev Drive (let's call it D: here)
2. Mount D: to a subfolder of C: (for example C:\Projects)
3. Create a project in D:\my-great-flutter-project
4. Open it in Android Studio using the path C:\Projects\my-great-flutter-project
5. Try to debug the app with breakpoints -> doesn't work and breakpoints are shown as 🚫 :( 
6. Open it in Android Studio using the path D:\my-great-flutter-project
7. Try to debug the app with breakpoints -> works :)

## Version info

<details>

<summary> Flutter Doctor output </summary>

```
[✓] Flutter (Channel stable, 3.19.1, on Microsoft Windows [Version 10.0.22631.3374], locale
    en-GB)
    • Flutter version 3.19.1 on channel stable at C:\Users\miche\fvm\versions\3.19.1
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision abb292a07e (7 weeks ago), 2024-02-20 14:35:05 -0800
    • Engine revision 04817c99c9
    • Dart version 3.3.0
    • DevTools version 2.31.1

[✓] Windows Version (Installed version of Windows is version 10 or higher)

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at C:\development\Android\Sdk
    • Platform android-34, build-tools 34.0.0
    • ANDROID_HOME = C:\development\Android\Sdk
    • Java binary at: C:\Users\miche\AppData\Local\Programs\Android Studio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 17.0.9+0--11185874)
    • All Android licenses accepted.

[✗] Chrome - develop for the web (Cannot find Chrome executable at
    .\Google\Chrome\Application\chrome.exe)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.

[!] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.9.4)
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community
    • Visual Studio Community 2022 version 17.9.34714.143
    • Windows 10 SDK version 10.0.22621.0
    ✗ Visual Studio is missing necessary components. Please re-run the Visual Studio installer
      for the "Desktop development with C++" workload, and include these components:
[...]

PR #8706, opened 2026-01-13
Add logs of breakpoint URI setup

helin24:
This is relevant for debugging https://github.com/flutter/flutter-intellij/issues/7336. I'm trying to log some points in breakpoint handling to see how paths are working.

### DIFF
--- a/src/io/flutter/run/FlutterPositionMapper.java
+++ b/src/io/flutter/run/FlutterPositionMapper.java
@@ -22,6 +22,7 @@ import com.jetbrains.lang.dart.util.DartResolveUtil;
 import com.jetbrains.lang.dart.util.DartUrlResolver;
 import io.flutter.dart.DartPlugin;
 import io.flutter.logging.PluginLogger;
+import io.flutter.settings.FlutterSettings;
 import io.flutter.utils.OpenApiUtils;
 import io.flutter.vmService.DartVmServiceDebugProcess;
 import org.dartlang.vm.service.element.LibraryRef;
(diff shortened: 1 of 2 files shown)


src/io/flutter/run/FlutterPositionMapper.java
```diff
@@ -140,7 +141,15 @@ public void onLibrariesDownloaded(@NotNull final Iterable<LibraryRef> libraries)
       if (remoteUri.startsWith(DartUrlResolver.DART_PREFIX)) continue;
       if (remoteUri.startsWith(DartUrlResolver.PACKAGE_PREFIX)) continue;
       remoteSourceRoot = findRemoteSourceRoot(remoteUri);
-      if (remoteSourceRoot != null) return;
+      if (remoteSourceRoot != null) {
+        if (FlutterSettings.getInstance().isFilePathLoggingEnabled()) {
```
pq, line 145: Aside: did we talk about ways to push this check into the logger? These logging calls introduce a lot of boilerplate...

pq: If the issue is that we only want to conditionally evaluate the code that assembles the message, we could consider passing in a lambda?

(Not for fixing now obviously; just food for thought!)

helin24: I added an issue for this: https://github.com/flutter/flutter-intellij/issues/8707

I think the difficulty is that logs that could involve file paths don't seem that predictable, in terms of what might need to change? And then if we passed into a logging function something like `log(messageWithFilePaths, messageWithoutFilePaths)` then that may be more confusing to follow at the spot where we're logging? But I suspect if we thought about it for a bit these issues could be solved very sensibly.

pq: Thanks!

PR #8712, opened 2026-01-17
Introduce a code formatting script, as well as inspections from IntelliJ.

jwren:
Formatting includes Dart / Java / Kotlin / XML / Form / Properties / JSON.

The inspections are currently the default settings from IntelliJ, but this could be iterated on if we want the repo to have the inspections enforced.

If we agree on having the repo formatted / inspected, a follow up here would be documentation of this tool and the addition of it to the presubmit to prevent regressions.

With the follow-through of formatting, and adding this to the bots, this will resolve https://github.com/flutter/flutter-intellij/issues/8098

Formatting output visible in this PR: https://github.com/flutter/flutter-intellij/pull/8715

### DIFF
--- a/tool/format.sh
+++ b/tool/format.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+# Copyright 2026 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This script formats the codebase.
+# - Dart: Uses 'dart format'
+# - Java/Kotlin: Uses IntelliJ's format.sh (if configured)
+
+set -e
+
+echo "Formatting Dart files..."
+dart format .
+
+# Check for IDEA_HOME or common locations
+IDEA_FORMAT_SCRIPT=""
+
+if [ -n "$IDEA_HOME" ]; then
+  IDEA_FORMAT_SCRIPT="$IDEA_HOME/bin/format.sh"
+elif [ -f "/Applications/IntelliJ IDEA CE.app/Contents/bin/format.sh" ]; then
+  IDEA_FORMAT_SCRIPT="/Applications/IntelliJ IDEA CE.app/Contents/bin/format.sh"
+elif [ -f "/Applications/IntelliJ IDEA.app/Contents/bin/format.sh" ]; then
+  IDEA_FORMAT_SCRIPT="/Applications/IntelliJ IDEA.app/Contents/bin/format.sh"
+fi
+
+if [ -n "$IDEA_FORMAT_SCRIPT" ] && [ -f "$IDEA_FORMAT_SCRIPT" ]; then
+  echo "Formatting Java/Kotlin files using $IDEA_FORMAT_SCRIPT..."
+  "$IDEA_FORMAT_SCRIPT" -r -m *.java,*.kt,*.xml,*.form,*.properties,*.json \
+      src testSrc resources tool/plugin/lib tool/plugin/test
+
+else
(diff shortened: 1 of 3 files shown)


tool/format.sh
```diff
@@ -0,0 +1,36 @@
...
+set -e
+
+echo "Formatting Dart files..."
+dart format .
+
+# Check for IDEA_HOME or common locations
+IDEA_FORMAT_SCRIPT=""
+
+if [ -n "$IDEA_HOME" ]; then
+  IDEA_FORMAT_SCRIPT="$IDEA_HOME/bin/format.sh"
```
helin24, line 19: Is this IDEA `format.sh` shared for all of us? Or is it a local file that we can modify? I'm wondering if we could have a situation where we each have separate settings and end up with conflicting corrections.

jwren: It is where the IntelliJ is launched from, yes. Versioning is a question for this, as well as which dart SDK we would use. 

I suspect a on the latest stable platform version would be more than sufficient, formatting and verifications also don't change that much from one version to the next.

jwren: It is where the IntelliJ is launched from, yes. Versioning is a question for this, as well as which dart SDK we would use. 

I suspect a on the latest stable platform version would be more than sufficient, formatting and verifications also don't change that much from one version to the next.

helin24: Ohh I think I understand more after I read this:

> The format.sh script is an executable located in the bin directory of your IntelliJ IDEA installation, used for applying the IDE's configured code style formatting to specified files or directories from the command line.

So this doesn't contain the information of what code styles we want for java, it just uses the code styles we set. But where are those from? I guess my question still remains of whether we have a spot where we set the code styles for our project, rather than potentially each of us having different settings depending on our individual configuration of IntelliJ options.

jwren: That's a valid concern, but I don't think it will be an issue as long as we're all on the latest stable version of the platform (currently 2025.3). The Java formatter doesn't tend to change dramatically from quarter to quarter.

If we do run into inconsistencies, we could change the script to download a specific version of IntelliJ locally—outside of the repo—and hardcode our scripts to use that one, ensuring we're all aligned.

jwren:
@pq & @helin24  I went through and added documentation, and stricter assertions in the `verify_inspections.sh`.  Assuming that we agree that code should be formatted and we want to track inspections.  I have follow up PRs to do the formatting, and the (most importantly) have these rules enforced by our presubmits.

tool/format.sh
helin24: We could potentially commit `.idea/codeStyles/Project.xml`

pq: From my experience this kind of skew is worth working hard to avoid and I don't think we can rely on every contributor being on the same plugin version.

pq:
This is awesome! Let's discuss before it lands so we make sure to have churn only once.

Thanks!

pq:
Was this AI-Assisted? Tag if so?

jwren:
> Hi! Asking for a hold until we can all discuss.
> 
> Thanks!

100%.  One of the first things I will say here is that we should land this so you can experiment some here with the idea of having the repo formatted.  This PR doesn't format any code, nor does it enforce it in any way.  It is just 3 simple shell scripts that we should have had in the repo from day 1.

tool/format.sh
jwren: Maybe, and then we'd have more information and we could add complexity if we need to.

The Dart language changes more frequently than Java and there hasn't been the need for super strict versioning between contributors for the formatting.

Issue #8723, opened 2026-01-21
[cq] remove unused `product-matrix.json` and references


pq:
We're not using the product matrix anymore so it'd be good to remove. (Note there are some script references so those will need some attention too.)

PR #8726, opened 2026-01-23
Remove product-matrix.json and related code

helin24:
This deletes the `product-matrix.json` file and places that use it. I checked that `bin/plugin generate` still works (though we probably don't need this either) to make sure the plugin tool still compiles.

Related to https://github.com/flutter/flutter-intellij/issues/8723

### DIFF
--- a/tool/plugin/lib/plugin.dart
+++ b/tool/plugin/lib/plugin.dart
@@ -62,10 +62,10 @@ List<String> findJavaFiles(String path) {
       .toList();
 }
 
-bool isTravisFileValid() {
-  var travisPath = p.join(rootPath, '.github/workflows/presubmit.yaml');
-  var travisFile = File(travisPath);
-  return travisFile.existsSync();
+bool isPresubmitFileValid() {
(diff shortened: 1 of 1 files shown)


tool/plugin/lib/plugin.dart
```diff
@@ -77,15 +65,7 @@ List<String> findJavaFiles(String path) {
 bool isTravisFileValid() {
```
pq: Since this isn't really about travis anymore, consider changing to something like `isPresubmitFileValid`?

tool/plugin/lib/plugin.dart
```diff
@@ -77,15 +65,7 @@ List<String> findJavaFiles(String path) {
...
   var travisPath = p.join(rootPath, '.github/workflows/presubmit.yaml');
```
pq: Rename to `presubmitPath`?

tool/plugin/lib/plugin.dart
```diff
@@ -77,15 +65,7 @@ List<String> findJavaFiles(String path) {
...
   var travisFile = File(travisPath);
```
pq: And `presubmitFile`?

Issue #8725, opened 2026-01-22
Plugin exception


Claytonious:
Please file this bug report at https://github.com/flutter/flutter-intellij/issues/new.

---

## What happened

Loading project

## Version information

JetBrains Rider `2025.3.1` • Flutter plugin `io.flutter 88.2.0` • Dart plugin `502.0.0`

Error getting Flutter sdk information.

## Exception

null

```
java.lang.NoClassDefFoundError: com/jetbrains/lang/dart/util/DotPackagesFileUtil
	at io.flutter.pub.PubRoot.getPackagesMap(PubRoot.java:321)
	at io.flutter.sdk.FlutterPluginsLibraryManager.getFlutterPluginPaths(FlutterPluginsLibraryManager.java:141)
	at io.flutter.sdk.FlutterPluginsLibraryManager.lambda$updateFlutterPluginsImpl$0(FlutterPluginsLibraryManager.java:125)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl$OTelMonitor.callWrapped(NonBlockingReadActionImpl.java:929)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl$OTelMonitor$MonitoredComputation.call(NonBlockingReadActionImpl.java:961)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl$Submission.insideReadAction(NonBlockingReadActionImpl.java:685)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl$Submission.lambda$attemptComputation$6(NonBlockingReadActionImpl.java:648)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.tryRunReadAction(NestedLocksThreadingSupport.kt:902)
	at com.intellij.openapi.application.impl.ApplicationImpl.tryRunReadAction(ApplicationImpl.java:1237)
	at com.intellij.openapi.progress.util.ProgressIndicatorUtils.lambda$runInReadActionWithWriteActionPriority$0(ProgressIndicatorUtils.java:94)
	at com.intellij.openapi.progress.util.ProgressIndicatorUtilService.runActionAndCancelBeforeWrite(ProgressIndicatorUtilService.java:73)
	at com.intellij.openapi.progress.util.ProgressIndicatorUtils.runActionAndCancelBeforeWrite(ProgressIndicatorUtils.java:152)
[...]

Issue #8598, opened 2025-11-03
☂️ [analytics] instrument all contributed actions


pq:
Tracking issue for adding analytics gathering to our contributed actions.

- [x] `AttachDebuggerAction`  (https://github.com/flutter/flutter-intellij/pull/8597)
- [x] `DeviceSelectorAction` (https://github.com/flutter/flutter-intellij/pull/8606)
- [x] `DeviceSelectorRefresherAction` (https://github.com/flutter/flutter-intellij/pull/8606)
- [x] `ExtractWidgetAction` (https://github.com/flutter/flutter-intellij/pull/8735)
- [x] `FlutterBuildAction` (https://github.com/flutter/flutter-intellij/pull/8735)
- [x] `FlutterCleanAction`  (https://github.com/flutter/flutter-intellij/pull/8597)
- [x] `FlutterDoctorAction`  (https://github.com/flutter/flutter-intellij/pull/8597)
- [x] `FlutterGettingStartedAction` (https://github.com/flutter/flutter-intellij/pull/8604)
- [x] `FlutterNewProjectAction` (https://github.com/flutter/flutter-intellij/pull/8735)
- [x] `FlutterPackagesAddAction` (https://github.com/flutter/flutter-intellij/pull/8597)
- [x] `FlutterPackagesGetAction` (https://github.com/flutter/flutter-intellij/pull/8597)
- [x] `FlutterPackagesUpgradeAction` (https://github.com/flutter/flutter-intellij/pull/8597)
- [x] `FlutterRetargetAppAction` (https://github.com/flutter/flutter-intellij/pull/8733)
- [x] `FlutterSubmitFeedback` (https://github.com/flutter/flutter-intellij/pull/8604)
- [x] `FlutterUpgradeAction` (https://github.com/flutter/flutter-intellij/pull/8597)
- [x] `OpenAndroidModule` (https://github.com/flutter/flutter-intellij/pull/8734)
- [x] `OpenEmulatorAction` (https://github.com/flutter/flutter-intellij/pull/8734)
- [x] `OpenInAndroidStudioAction`  (https://github.com/flutter/flutter-intellij/pull/8734)
- [x] `OpenInAppCodeAction` (https://github.com/flutter/flutter-intellij/pull/8734)
- [x] `OpenInXcodeAction` (https://github.com/flutter/flutter-intellij/pull/8734)
- [x] `OpenSimulatorAction` (https://github.com/flutter/flutter-intellij/pull/8737)
[...]

PR #8730, opened 2026-01-26
[analytics] migrate to core Dart Plugin reporting

pq:
Dart 502 introduces the core analytics implementation and this migrates our stub implementations to use it. We remove the now defunct `UnifiedAnalytics` implementation to keep things tidy.

Note that this does introduce more version coupling. Once published we'll have more places that will break in the presence of a pre-502 Dart plugin. (Since this was already happening with the `PubRoot` cleanup, e.g., https://github.com/flutter/flutter-intellij/issues/8725, I didn't consider doing version testing and any defensive programming; if folks feel strongly we can revisit though!)

Follow-up from #8598 

---

<details>

- See our [contributor guide]([https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md](https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md) for general expectations for PRs.
</details>


helin24:
Great! I agree we probably don't need to worry about the old Dart version issue.

pq:
Cool, thanks! I'm a little surprised that IDEA doesn't support explicit version dependencies but since they don't I think occasional issues with version skew are probably tolerable (if not expected). Worth keeping an eye out though!

PR #8773, opened 2026-02-09
Update platform version for 261

helin24:
A problem is that I may need a way for users to download a dev Dart plugin as well.

Another problem is there are many new verify issues for 2026.1 (these can be seen on the jetbrains site already, even though our current plugins aren't technically compatible with 2026.1). Maybe it would make sense to focus on fixing those first, and make this an issue instead so we know that compatibility for 2026.1 is on our agenda.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@
 #
 
 ideaVersion=2025.2.3.9
-dartPluginVersion= 503.0.0
+dartPluginVersion= 502.0.0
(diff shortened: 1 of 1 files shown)


pq:
Great points.  Let's walk through them in person!

helin24:
Update to initial comments: I think the compatibility problems are okay, as they reflect Dart plugin not yet being available for 2026.1 also. We will release updated Dart plugin before Flutter this time (last release this was harder because the new Dart version broke old Flutter versions)

PR #8787, opened 2026-02-11
[verifier] add baseline checking support to plugin verification

pq:
## Plugin Verifier Baseline support

Adds baseline support to the verifier check that compares verification results against established baselines. This ensures that no new verification errors are introduced and go undetected.

Note that the plugin verifier’s mechanism for ignoring warnings is not sufficient since it only supports a subset of warnings and we need the whole gamut and so we have some bespoke scripting. IntelliJ verifier docs are [here](https://github.com/JetBrains/intellij-plugin-verifier).

More complete docs to come, but the basic idea is:
* baselines are extracted from verifier reports and stored for each verfied IDE (`tool/baseline/$version`)
* verifier runs extract results from fresh reports and compare them against the baselines
* if there are unaccounted for errors, they are printed and the build is failed
* in all cases, complete reports are still uploaded as build artifacts for manual inspection

Some care is taken to make the results "pop" in the action runs with terminal styling. For example:

image

with a successful check looking like:

image

and a failed one like:

image

---

### Important

❗ Note that with this change, builds will **BREAK** if source changes stray from our 251 and 252 baselines.

---

### Implementation Notes

I elected not to weigh the script down with a lot of implementation notes but there are a few interesting points that might help in reviewing.

#### Gemini Summary

This PR introduces baseline checking support for plugin verification in tool/github.sh, allowing the build to pass even if known issues exist, while failing if any new regressions are introduced.

* **Version-Specific Baselines**: Baseline files are maintained individually for each IDE version in `tool/baseline/$version/verifier-baseline.txt`.

* **Dynamic Report Discovery**: The script automatically locates the correct `report.md` for each version using a path-specific search: `find build/reports/pluginVerifier -path "*-$version.*/report.md"`.

* **Automated Issue Comparison:**

  * It extracts current issues (bulleted items) from the generated report.

  * It uses `comm -13` to compare these against the sorted baseline to identify only the new errors.

* **Non-Blocking Failure Logic:** An `EXIT_STATUS` variable tracks failures across multiple IDE versions (e.g., 251 and 252). This allows all verifications to complete before the script exits with an error code if any version failed.

* **Improved Log Readability:** The script uses ANSI escape codes for **bold**, **green** (success), and **red** (error) output to make verification results easier to parse in GitHub Actions logs.

---

<details>

- See our [contributor guide]([https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md](https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md) for general expectations for PRs.
</details>

### DIFF
--- a/tool/github.sh
+++ b/tool/github.sh
@@ -80,9 +80,25 @@ elif [ "UNIT_TEST_BOT" = "$BOT" ] ; then
 
 elif [ "VERIFY_BOT" = "$BOT" ] ; then
   EXIT_STATUS=0
+  
+  echo "Check on space before verifyPluginProjectConfiguration\n"
+  df -h
+  ./gradlew verifyPluginProjectConfiguration
+  
+  echo "Check on space before verifyPluginStructure\n"
+  df -h
+  ./gradlew verifyPluginStructure
+  
+  echo "Check on space before verifyPluginSignature\n"
+  df -h
+  ./gradlew verifyPluginSignature
 
   for version in 251 252; do
     echo -e "${BOLD}Running verifyPlugin for $version...${NC}"
+    
+    echo "Check on space before run\n"
+    df -h
+    
     ./gradlew verifyPlugin -PsingleIdeVersion=$version || true
 
     BASELINE="$GITHUB_WORKSPACE/tool/baseline/$version/verifier-baseline.txt"


tool/github.sh
```diff
@@ -72,26 +79,49 @@ elif [ "UNIT_TEST_BOT" = "$BOT" ] ; then
   ./gradlew test
 
 elif [ "VERIFY_BOT" = "$BOT" ] ; then
-  # Run the verifier
-  echo "Check on space before verifyPluginProjectConfiguration\n"
```
helin24, line 76: Are these other verify tasks no longer being done at all? Are they important?

pq: Oh gosh. Good catch. That was an accident! Will add back. Thank you!

Issue #8795, opened 2026-02-15
Device labels missing from run tabs on IntelliJ 2025.3 / Android Studio Panda


lukemmtt:
## Description

Device labels (e.g., "main.dart (macOS)", "main.dart (iPhone 16 Pro)") no longer appear in Run/Debug tab titles starting with IntelliJ 2025.3 (Android Studio Panda). All tabs show only the run configuration name (e.g., "main.dart"), making it impossible to distinguish which device a build is running on when targeting multiple platforms simultaneously.

## Root Cause

The reflection code in `LaunchState.java` (lines 173-185) that sets the display name is silently failing on every launch.

The code gets the private `myDisplayNameView` field from `RunContentDescriptor`, then calls `getMethod("setValue", Object.class)` on the field's runtime class, then calls `invoke()`. This worked when the field type was `MutableReactiveProperty` (a public IntelliJ-internal class), but broke in IntelliJ 2025.3 when the field type was changed to `MutableStateFlow` from kotlinx-coroutines ([JetBrains/intellij-community@`aaa88849ba4c`](https://github.com/JetBrains/intellij-community/commit/aaa88849ba4ceec53a0e65279e9ee79071f27451)).

The runtime implementation class `kotlinx.coroutines.flow.StateFlowImpl` is **package-private**. Java's access control prevents `Method.invoke()` on a public method when the declaring class is not accessible from the caller, throwing `IllegalAccessException`. The exception is caught silently (logged at INFO without stack trace), so the device name is never applied.

## Verification

The Android Studio `idea.log` shows `"Error setting display name"` (from the catch block in `LaunchState.java:184`) on every Flutter launch in Panda.

Reproduced independently by calling the same reflection against the actual `kotlinx-coroutines-core-jvm.jar` shipped with Android Studio Panda:

```java
MutableStateFlow<String> flow = StateFlowKt.MutableStateFlow(null);
Method m = flow.getClass().getMethod("setValue", Object.class);
m.invoke(flow, "test");
// IllegalAccessException: cannot access a member of class
// kotlinx.coroutines.flow.StateFlowImpl with modifiers "public"
```

## Proposed Fix

Use `RunContentDescriptor.setDisplayName()` (an existing protected method on `RunContentDescriptor`) via reflection instead of reaching into the private field's internal type:

```java
Method setDisplayName = RunContentDescriptor.class.getDeclaredMethod("setDisplayName", String.class);
setDisplayName.setAccessible(true);
[...]

Issue #8794, opened 2026-02-14
Regression for linux desktop in Panda 1 in starting/stopping app.


hapasa:
Hi,

I submitted ssue 483732396: Regression in Android Studio Panda 1, and got a reply that I should report it here?

Anyway, prior to upgrading to Panda 1, I was a happy camper,  developing and testing my flutter app using linux desktop, even if eventual platform is mobile.

Since the upgrade, the Run, Stop buttons in IDE are no longer reliable.   Either Run does not always work, nothing happens when I click on the button, or Stop does not work, the click on button does nothing.  Debug button always seems to work, launching a the debugged app.

The IDE is still aware of the app running, even when Stop button is not working, if I quite the app by closing it's window, the Stop button disappears and run button looses the Green background.

It seems some race/async condition, because sometimes it requires only 1 click on Run to start my app, but occasionally I need to click multiple times, often then ending up triggering hot restarts due to my frustrated attempts to start my app.

It was working fine before Panda 1.

Best regards,
Harri

pq:
Hi! I'm pretty sure this is fixed w/ @lukemmtt's change in https://github.com/flutter/flutter-intellij/pull/8796.

If you're interested, follow the instructions in https://github.com/flutter/flutter-intellij/issues/8242 to get a dev channel build to try it out. Otherwise the fix will be available in the next release.

Thanks for the report. And sorry for the hassle!

hapasa:
Hi,

Sorry, but no cigar. I switch to dev version and it did not improve, running 90.0.0-dev.20260221-9fef30b

[image.png]

Click on arrow still does not always "take"
Teaching myself to hit Shift-F10 instead.
Best regards,
Harri

PR #8796, opened 2026-02-15
Fix device labels missing from run tabs on IntelliJ 2025.3+

lukemmtt:
## Summary

- Fix `IllegalAccessException` that silently prevents device names from appearing in Run/Debug tab titles on IntelliJ 2025.3 (Android Studio Panda)
- Use `RunContentDescriptor.setDisplayName()` API instead of reaching into a private field's internal type

Fixes https://github.com/flutter/flutter-intellij/issues/8795

## Problem

In IntelliJ 2025.3, the `myDisplayNameView` field on `RunContentDescriptor` changed from `MutableReactiveProperty` (a public class) to `MutableStateFlow` (whose runtime class `StateFlowImpl` is package-private). The existing reflection code calls `Method.invoke()` on the `setValue` method found via `getMethod()`, which throws `IllegalAccessException` when the declaring class is not accessible — even though the method itself is public. The exception is caught silently, so run tabs show "main.dart" instead of "main.dart (macOS)".

Regression introduced by [JetBrains/intellij-community@`aaa88849ba4c`](https://github.com/JetBrains/intellij-community/commit/aaa88849ba4ceec53a0e65279e9ee79071f27451) (Jul 14, 2025).

## Fix

Replace the field-level reflection (`getDeclaredField("myDisplayNameView")` → `getMethod("setValue")` → `invoke()`) with a call to `RunContentDescriptor.setDisplayName()`, an existing protected method on `RunContentDescriptor`. This calls a stable API method on a public class rather than depending on the internal type of a private field.

An alternative minimal fix would be to add `setValueMethod.setAccessible(true)` before the existing `invoke()` call, but calling a stable API method on a public class is more resilient than depending on the internal type of a private field.

### Before fix
Screenshot 2026-02-15 at 2 53 56 PM

### After fix:
Screenshot 2026-02-15 at 2 52 38 PM

---

- [x] I've reviewed the contributor guide and applied the relevant portions to this PR.

<details>

- See our [contributor guide](https://github.com/flutter/flutter/blob/main/docs/contributing/Tree-hygiene.md#overview) for general expectations for PRs.
</details>

### DIFF
--- a/src/io/flutter/run/LaunchState.java
+++ b/src/io/flutter/run/LaunchState.java
@@ -55,7 +55,6 @@ import io.flutter.toolwindow.ToolWindowBadgeUpdater;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
@@ -171,16 +170,12 @@ public class LaunchState extends CommandLineState {
     final String nameWithDeviceName = descriptor.getDisplayName() + " (" + device.deviceName() + ")";
 
     try {
-      // There is no public way to set display name so we resort to reflection.
-      final Field f = descriptor.getClass().getDeclaredField("myDisplayNameView");
-      f.setAccessible(true);
-      Object viewInstance = f.get(descriptor);
-      if (viewInstance != null) {
-        final Method setValueMethod = viewInstance.getClass().getMethod("setValue", Object.class);
-        setValueMethod.invoke(viewInstance, nameWithDeviceName);
-      }
+      // RunContentDescriptor.setDisplayName() is protected, so we use reflection to call it.
+      final Method setDisplayName = RunContentDescriptor.class.getDeclaredMethod("setDisplayName", String.class);
+      setDisplayName.setAccessible(true);
+      setDisplayName.invoke(descriptor, nameWithDeviceName);
     }
-    catch (IllegalAccessException | InvocationTargetException | NoSuchFieldException | NoSuchMethodException e) {
+    catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
       FlutterUtils.info(LOG, "Error setting display name", e, true);
     }
 


pq:
Hey wow. This is really fantastic. Thanks especially for the thoughtful write-up.

I do *so* wish there was a way around this that didn't lean on reflection but this is a lot safer than what was there so it feels like a win.

I wonder if this could be the root of https://github.com/flutter/flutter-intellij/issues/8794 too?

lukemmtt:
Glad to help! This feature is pretty helpful and important for my workflow, so when I found those labels to be removed, my first thought was "why on earth did they remove the labels!"; I was dying to know who was behind the decision and how / why that choice was made, which started me on this wild goose chase, scouring the IntelliJ codebase before finally realizing (1) it's not an Android studio feature, and (2) it wasn't an intentional removal, but rather just a regression 😂 

As for the reflection approach, yeah not my favorite pattern, & I'm open to others, seems robust enough for me though, now that I understand it.

pq:
Awesome. This LGTM, so let's land it. We may have missed today's dev-channel build, but it should be in tomorrow's and you can grab it and enjoy the fruits of your labor!

Instructions here if you're game to try that: https://github.com/flutter/flutter-intellij/blob/main/docs/Dev-Channel.md

pq (approved):
💎

pq:
One thing that occurs to me...

We should confirm (if you haven't already) that `setDisplayName` is available in `251.0`; if not, we should add back the old bits as a fallback.

For extra credit, we might consider adding a test that asserts that we can access something that can set a display name. Maybe introducing a function like:

```java
@visibleForTesting
static @Nullable BiConsumer<RunContentDescriptor, String> getDisplaySetter() { ... }
```

where the test could be something like

```java
assertNotNull(LaunchState.getDisplaySetter());
```

@lukemmtt: WDYT?

lukemmtt:
Good thinking; I checked the [251 branch source](https://github.com/JetBrains/intellij-community/blob/idea/251.23774.435/platform/execution/src/com/intellij/execution/ui/RunContentDescriptor.java) and can confirm `setDisplayName` is there; it was added back in February 2024 (~243 ish), well before build 251. No fallback needed. 

As for the test, good idea, sorry that slipped my mind! I went ahead and implemented your suggestion with `getDisplaySetter()`; PR here: https://github.com/flutter/flutter-intellij/pull/8809

Thank you @pq!

pq:
That's awesome. Thanks so much! Looking at your PR now...

blasko:
The issue still persists, on both the stable and dev channel. 

image

lukemmtt:
@blasko I looked into this; what you've reported is a new IntelliJ-specific regression. I filed a new issue here: https://github.com/flutter/flutter-intellij/issues/8907

Issue #8806, opened 2026-02-18
☂️ [CQ] Manage our dependence on reflection in the plugins


pq:
Use of reflection to access hidden APIs is prone to runtime failures (see e.g., the recent #8795) and circumvents our verification baselining.

The regression reported in #8795 could have been avoided.

1. If we hadn't used reflection, we'd have seen a static compilation failure.
2. If we *had* used reflection but had it under-test, we would have detected when the API broke.

To protect against future issues, I propose a few next steps.

1. Review all the places we're using reflection and remove them where possible and add tests where not.
2. Explore a verification mechanism whereby new uses of reflection are detected (akin to our baseline verifier).

(This is an umbrella issue for the Flutter AND Dart plugins.)

PR #8809, opened 2026-02-18
Add test asserting display name setter is accessible on current IntelliJ build

lukemmtt:
Follow-up to #8796.

Extracts the reflection lookup into a `@VisibleForTesting getDisplaySetter()` method and adds a test asserting it returns non-null. This serves as a regression guard: if JetBrains removes or renames `setDisplayName` in a future build, the test will fail loudly rather than the feature silently breaking for users (which was the original failure mode in #8795).

Verified locally by running the test against the real SDK (passes), then temporarily corrupting the method name to `setDisplayName_doesNotExist` (fails with the expected assertion message).

 I've reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/src/io/flutter/run/LaunchState.java
+++ b/src/io/flutter/run/LaunchState.java
@@ -60,6 +60,8 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


pq:
This is fantastic and a great reference point for #8806 too. Thanks for following up and moving this along!

PR #8827, opened 2026-02-25
Fix: Flutter SDK path in settings is not updated

### LINKED ISSUE #8641: Flutter SDK path setting issue
UI does not remember the actual saved value for the "Flutter SDK path" setting.

In the attached video the behavior is visible:

1. Open Android Studio => Settings
2. Flutter SDK path: /Users/johannesroest/fvm/versions/3.27.1, the button "Apply" is greyed out
3. Choose Flutter 3.38.4
4. Notice that the UI reflects the choice, but the button "Apply" is still greyed out
5. Choose OK to save the usage of Flutter 3.38.4
6. Open Android Studio => Settings
7. Notice the Flutter SDK version still shows 3.27.1
8. Run Flutter doctor, you can see that it uses Flutter 3.38.4
9. Open Android Studio => Settings
10. Notice the Flutter SDK version still shows 3.27.1

Flutter 3.38.4
Flutter plugin 88.1.0
Dart plugin 500.0.0

https://github.com/user-attachments/assets/fbccc47a-00c2-4aad-8beb-6ded1786ef87

cj-radcliff:
The settings page UI for the Flutter SDK path was not updating after being changed. When a user updated the path, applied the change, and re-opened settings, the old path would still be displayed.

This was caused by the settings page reading the SDK path from a stale cache provided by the Dart plugin (FlutterSdk.getFlutterSdk()).

The fix is to use FlutterSdk.getIncomplete() instead, which bypasses the cache and reads the path directly from the authoritative source: the project's Library Table configuration. This ensures the UI always displays the currently configured value.

Fixes: https://github.com/flutter/flutter-intellij/issues/8641

Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.


src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -110,7 +110,7 @@ public class FlutterSettingsConfigurable implements SearchableConfigurable {
   }
 
   private void init() {
-    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(myProject);
+    final FlutterSdk sdk = com.intellij.openapi.application.ReadAction.compute(() -> FlutterSdk.getIncomplete(myProject));
```
helin24: Is there a reason to use qualified name instead of importing?

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -185,7 +185,7 @@ public JComponent createComponent() {
   @Override
   public boolean isModified() {
 
-    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(myProject);
+    final FlutterSdk sdk = FlutterSdk.getIncomplete(myProject);
```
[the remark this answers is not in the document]
helin24: This seems like a reasonable change if the other spots are also using a `ReadAction`.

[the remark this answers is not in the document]
helin24: Maybe we can pull this "live SDK access" into another method and comment about how it works?

helin24 (approved):

src/io/flutter/sdk/FlutterSettingsConfigurable.java
pq: +1. I don't think there's a good one. Could you fix?

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -312,7 +312,7 @@ public void apply() throws ConfigurationException {
 
   @Override
   public void reset() {
-    final FlutterSdk sdk = FlutterSdk.getFlutterSdk(myProject);
+    final FlutterSdk sdk = com.intellij.openapi.application.ReadAction.compute(() -> FlutterSdk.getIncomplete(myProject));
```
pq: Let's fix this qualified reference here too.

pq (approved):
If this works, I'm OK to land it but I have to wonder why the cache is stale and if there aren't other places where `getFlutterSdk(...)` isn't working as expected  -- looks like it's used 40+ times. 🧐 

I'm also curious if this is a newly broken bit or if it's been misbehaving for a while and noone has reported it? (More just asking out loud... I don't expect an answer!)

pq:
More thoughts: it'd be interesting to ask gemini some follow up questions:

* Is it likely that this is a regression or a long-standing bug? Is there anything in git history that could explain a breakage? Any changes in platform libraries that could be to blame?
* If the cache is stale in this case, are there other places where it may also be stale to negative effect?
* Let's say we don't want to go this route and want to fix the stale cache issue at the root, how would we do that? Might there be any unexpected consequences?

Including the results with the review would be great as a kind of "alternatives considered" addendum.

pq:
Another question:

* How would I validate that this change was working as expected? (This might give us a hint for how well gemini "understands" the fix.)

@cj-radcliff: it would be great to include your verification steps too. (That is: how you tested the fix.)

cj-radcliff:
I asked a few questions of gemini, adding here. @pq I will also ask your questions and follow up further:



# Explanation of Cache Fix for Reviewers

## Q1: Who introduced the cache and why is it used instead of reading from disk?

**A:** The commit to fix the issue by bypassing the cache was made by **Cory R** on **Wed Feb 25, 2026**. The commit message explains that the original cached method, `FlutterSdk.getFlutterSdk()`, was provided by the Dart plugin.

The reason a cache is used is for performance. Reading from disk is thousands of times slower than reading from memory. Since many parts of the IDE need the SDK path frequently, caching the value after the first read prevents redundant, slow disk operations and keeps the UI responsive.

The bug occurred because the settings page, which needs the authoritative value from the disk, was incorrectly using the potentially stale cached value. The fix was to use `getIncomplete()` in that specific location to read the authoritative value.

** NOTE! This answer is partially incorrect because its not what I asked. I include it because the second half seems to be a good answer.

## Q2: Does this fix cause normal access methods to read from disk more often?

**A:** No, it does not. The fix was carefully targeted *only* to the settings page code (`FlutterSettingsConfigurable.java`).

Other parts of the plugin that use the cached `FlutterSdk.getFlutterSdk()` method for normal, frequent operations (like running an app) are completely unaffected. They continue to get the performance benefit of the cache. The direct disk read only happens in the infrequent case of a user opening the settings page, where correctness is more important than a minor performance optimization.

## Q3: Is there a way to make the fix by refreshing the cache properly before reading it, instead of bypassing it?

**A:** While theoretically possible, bypassing the cache is the better and safer design choice in this specific context for several reasons:

1.  **Ownership:** The cache belongs to the Dart plugin. Forcing a refresh on another component's internal state is risky and can have unintended side effects.
2.  **API Availability:** There's no guarantee the Dart plugin exposes a public, stable API for forcing a refresh. `getIncomplete()`, however, is clearly designed for this purpose.
3.  **Correctness:** A settings page's primary job is to reflect the **authoritative source of truth** (the configuration on disk). Bypassing a cache to read this ground truth is the most direct and correct design pattern for a UI that edits that truth.

Piinks:

pq:
@cj-radcliff: do you want to follow-up on the feedback or close for now?

cj-radcliff:
> @cj-radcliff: do you want to follow-up on the feedback or close for now?

I'll pull this into my working repo and close on the main repo. Thank you.

src/io/flutter/sdk/FlutterSettingsConfigurable.java
cj-radcliff: Fixed. Thank you!

src/io/flutter/sdk/FlutterSettingsConfigurable.java
cj-radcliff: Fixed. Thank you!

helin24:
@cj-radcliff I'm still seeing the SDK not persisting (I am testing in Android Studio 2025.3) - could you test the dev build or the potential release build I uploaded to the google drive to see if you can reproduce?

PR #8854, opened 2026-03-16
Fix existing integration test

helin24:
The `isVisible` function was not found; I expect maybe it was from a previous version of dependencies or something. I also specified ms between text input because it feels very slow. I don't see an option to just paste with the keyboard but maybe that exists elsewhere.


helin24:
@gemini-code-assist is there a way to copy and paste without being platform-specific?

pq:
(FWIW: I think direct text input will be more robust than copy-paste. It is super slow though!)

Issue #8831, opened 2026-02-25
Error during debug application for IDEA 2026.1


helin24:
## What happened

Debug hello flutter app targeting macos device.

## Version information

IntelliJ IDEA `2026.1 EAP` • Flutter plugin `io.flutter 90.0.0` • Dart plugin `503.0.0`

## Exception

null

```
java.lang.Throwable: [Split debugger] RunContentDescriptor should not be used in split mode from XDebugSession. XDebugSession.getRunContentDescriptor is deprecated, see the javadoc for details
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:375)
	at com.intellij.xdebugger.impl.XDebugSessionImpl.getRunContentDescriptor(XDebugSessionImpl.kt:218)
	at io.flutter.run.LaunchState.launch(LaunchState.java:163)
	at io.flutter.run.LaunchState$Runner.doExecute(LaunchState.java:443)
	at com.intellij.execution.runners.GenericProgramRunner.execute$lambda$0(GenericProgramRunner.kt:20)
	at com.intellij.execution.impl.ExecutionManagerImpl.startRunProfile$lambda$0(ExecutionManagerImpl.kt:239)
	at com.intellij.execution.impl.ExecutionManagerImpl.doStartRunProfile$lambda$1(ExecutionManagerImpl.kt:306)
	at com.intellij.openapi.application.TransactionGuardImpl.runWithWritingAllowed(TransactionGuardImpl.java:236)
	at com.intellij.openapi.application.TransactionGuardImpl.access$100(TransactionGuardImpl.java:22)
	at com.intellij.openapi.application.TransactionGuardImpl$1.run(TransactionGuardImpl.java:198)
	at com.intellij.openapi.application.impl.AppImplKt$runnableUnitFunction$1.invoke(appImpl.kt:123)
	at com.intellij.openapi.application.impl.AppImplKt$runnableUnitFunction$1.invoke(appImpl.kt:123)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.runWriteIntentReadAction(NestedLocksThreadingSupport.kt:686)
	at com.intellij.openapi.application.impl.ApplicationImpl.runIntendedWriteActionOnCurrentThread(ApplicationImpl.java:1045)
	at com.intellij.openapi.application.impl.ApplicationImpl$7.lambda$run$0(ApplicationImpl.java:599)
[...]

PR #8877, opened 2026-03-31
Fix split debugging error in 2026.1

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/8831

This includes a small set of changes to use new APIs while debugging, and changes to our build files to contain these changes to 2026.1 only.

I'm inclined to make a new branch for platform version 2026.1 for this change:
- Even though the deprecated API (`XDebugSessionImpl.getRunContentDescriptor`) throws an error, the alternative recommended is marked experimental (`XSessionStartedResult`), so I don't think it makes sense to commit this to main.
- The build changes are temporary - since Android Studio hasn't released 2026.1 yet, I had to remove some Android Studio-specific files from src so that we can compile the plugin against IntelliJ IDEA instead. But once Android Studio 2026.1 is released, then we should  revert most of the build changes and go back to compiling against that.

Alternatively, we could use reflection to get rid of the error. I will try to push up a draft for that as well.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -155,6 +155,8 @@ sourceSets {
         "resources"
       )
     )
+
+    // TODO(helin24): Include these files when we go back to building against Android Studio.
     java.exclude("**/AndroidStudioGradleSyncProvider.java")
     java.exclude("**/GradleUtils.java")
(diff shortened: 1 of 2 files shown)


helin24:
Closing in favor of https://github.com/flutter/flutter-intellij/pull/8878

PR #8905, opened 2026-04-15
Expose Flutter debug toolbar extension group

mdddj:
## What changed
- exposed a new public action group id: `Flutter.DebugProcess.TopToolbar`
- appended that group after Flutter's built-in debug toolbar actions in `FlutterDebugProcess`
- applied the same hook to Flutter test and Bazel test debug toolbars
- kept the group hidden when no downstream plugin contributes actions

## Why
`FlutterDebugProcess.registerAdditionalActions()` currently builds the top run/debug toolbar directly from code by mutating the runtime `DefaultActionGroup`. That makes the toolbar effectively closed to dependent plugins: there is no public action group id and no extension point they can target from `plugin.xml`.

Exposing a dedicated inline `ActionGroup` is the smallest platform-native way to make this extensible. Downstream plugins can now contribute with standard IntelliJ action registration, for example:

```xml
<action id="..." class="...">
  <add-to-group group-id="Flutter.DebugProcess.TopToolbar" anchor="last"/>
</action>
```

This avoids adding a new custom EP just to append toolbar actions, keeps the existing toolbar construction intact, and preserves current behavior when nobody contributes anything.

## Validation
- `./gradlew buildPlugin`

### DIFF
--- a/src/io/flutter/run/FlutterDebugProcessActions.java
+++ b/src/io/flutter/run/FlutterDebugProcessActions.java
@@ -24,14 +24,21 @@ public final class FlutterDebugProcessActions {
   private FlutterDebugProcessActions() {
   }
 
+  /**
+   * Appends toolbar actions contributed by downstream dependent plugins.
+   *
+   * <p>{@link ActionManager#getAction(String)} returns {@link AnAction} because action groups are
+   * actions too. The extension id resolves to an inline {@link ActionGroup}, which can contain
+   * multiple contributed child actions.
+   */
   public static void addTopToolbarExtensionActions(@NotNull DefaultActionGroup topToolbar) {
-    final AnAction action = ActionManager.getInstance().getAction(TOP_TOOLBAR_EXTENSION_GROUP_ID);
-    if (!(action instanceof ActionGroup actionGroup) || isEmpty(actionGroup)) {
+    final AnAction extensionGroupAction = ActionManager.getInstance().getAction(TOP_TOOLBAR_EXTENSION_GROUP_ID);
+    if (!(extensionGroupAction instanceof ActionGroup actionGroup) || isEmpty(actionGroup)) {
       return;
     }
 
     topToolbar.addSeparator();
-    topToolbar.addAction(action);
+    topToolbar.addAction(extensionGroupAction);
   }
 
   private static boolean isEmpty(@NotNull ActionGroup actionGroup) {
--- a/src/io/flutter/run/FlutterDebugProcessActions.java
+++ b/src/io/flutter/run/FlutterDebugProcessActions.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2026 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+package io.flutter.run;
+
+import com.intellij.openapi.actionSystem.ActionGroup;
+import com.intellij.openapi.actionSystem.ActionManager;
+import com.intellij.openapi.actionSystem.AnAction;
+import com.intellij.openapi.actionSystem.DefaultActionGroup;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Shared action groups for Flutter-specific run and debug toolbars.
+ */
+public final class FlutterDebugProcessActions {
+  /**
+   * Third-party plugins can contribute inline toolbar actions with:
+   * {@code <add-to-group group-id="Flutter.DebugProcess.TopToolbar" anchor="last" />}.
+   */
+  public static final @NotNull String TOP_TOOLBAR_EXTENSION_GROUP_ID = "Flutter.DebugProcess.TopToolbar";
+
+  private FlutterDebugProcessActions() {
+  }
+
+  public static void addTopToolbarExtensionActions(@NotNull DefaultActionGroup topToolbar) {
+    final AnAction action = ActionManager.getInstance().getAction(TOP_TOOLBAR_EXTENSION_GROUP_ID);
+    if (!(action instanceof ActionGroup actionGroup) || isEmpty(actionGroup)) {
+      return;
+    }
+
+    topToolbar.addSeparator();
+    topToolbar.addAction(action);
+  }
+
+  private static boolean isEmpty(@NotNull ActionGroup actionGroup) {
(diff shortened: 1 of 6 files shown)


helin24:
Thanks for the contribution! This isn't something we've thought about at all, so can you provide some more context of what a dependent plugin will be able to do if we provide this extension point? e.g. would this likely be to move the Flutter actions around more easily? Or an example of how you would accomplish an equivalent UI with and without this code?

mdddj:
I’m developing a Flutter plugin called [FlutterX](https://github.com/mdddj/dd_flutter_idea_plugin), and I’d like to add some custom Actions to this toolbar.

I tried several approaches, but after digging into the source code, I found that the toolbar is created and populated directly in code.

Would you consider exposing some extension points for third-party plugins that depend on the Flutter plugin, so they can contribute their own actions to this toolbar?

src/io/flutter/run/FlutterDebugProcessActions.java
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Third-party plugins can contribute inline toolbar actions with:
+   * {@code <add-to-group group-id="Flutter.DebugProcess.TopToolbar" anchor="last" />}.
+   */
+  public static final @NotNull String TOP_TOOLBAR_EXTENSION_GROUP_ID = "Flutter.DebugProcess.TopToolbar";
+
+  private FlutterDebugProcessActions() {
+  }
+
+  public static void addTopToolbarExtensionActions(@NotNull DefaultActionGroup topToolbar) {
```
helin24, line 33: I'm confused about how this function works - it's called by the debug process setup, but the action can be added by a third-party plugin in plugin.xml, and found here? Does this mean there can only be one action added to the toolbar?

helin24 (commented):
Thanks for explaining more @mdddj ! Exposing this extension sounds good; I'm not totally following how the code works, so if you can explain a bit more or show me an example of how the plugin.xml will look for your plugin, that may help.

mdddj:
Thanks, this is a good question.

The goal here is not to let downstream plugins reorder or replace Flutter’s existing toolbar actions. It is only to provide an extension point after the built-in Flutter debug/test actions, so plugins that depend on the Flutter plugin can contribute additional inline actions there.

`Flutter.DebugProcess.TopToolbar` is an `ActionGroup`, not a single action. A dependent plugin contributes actions to that group from its own `plugin.xml`, and this code looks up that group via `ActionManager` and appends it to the toolbar. Since the group is declared with `popup="false"`, its children are rendered inline on the toolbar, so this can expose multiple actions rather than just one.

For example, a dependent plugin could contribute:

```xml
<actions>
  <action id="FlutterX.AttachFoo"
          class="com.example.flutterx.AttachFooAction"
          text="Foo">
    <add-to-group group-id="Flutter.DebugProcess.TopToolbar" anchor="last"/>
  </action>

  <action id="FlutterX.AttachBar"
          class="com.example.flutterx.AttachBarAction"
          text="Bar">
    <add-to-group group-id="Flutter.DebugProcess.TopToolbar" anchor="last"/>
  </action>
</actions>

**Comment by helin24** [comment]
@mdddj thanks for the example! This makes sense, though I suppose the name `addAction` is a bit of a misnomer given that an action group can be added there.

File: src/io/flutter/run/FlutterDebugProcessActions.java [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+    final AnAction action = ActionManager.getInstance().getAction(TOP_TOOLBAR_EXTENSION_GROUP_ID);
```
helin24: Could you add a comment to clarify that `AnAction` and `getAction` can refer to groups of actions and not necessarily a single action?

src/io/flutter/run/FlutterDebugProcessActions.java
helin24, line 33: Can you add a comment to this method noting that this appends actions from downstream dependent plugins?

helin24 (approved):
Sorry I'm asking for so many comments! I'm hoping they will reduce confusion for anyone looking at the code in the future.

src/io/flutter/run/FlutterDebugProcessActions.java
mdddj: Added a method comment explaining that this id resolves to an inline ActionGroup, so dependent plugins can contribute multiple child actions instead of only one toolbar action.

src/io/flutter/run/FlutterDebugProcessActions.java
mdddj: Done. The method comment now notes that ActionManager#getAction returns AnAction because action groups are actions too.

src/io/flutter/run/FlutterDebugProcessActions.java
mdddj: Done. The method comment now states that this appends toolbar actions contributed by downstream dependent plugins.

helin24 (approved):
Thanks for addressing all of the questions! I'll let @pq review and merge.

pq (approved):
Thanks for the thoughtful explanation. LGTM!

pq (changes requested):
Hey! It looks like we're hitting an issue during plugin verification.


`* Override-only method com.intellij.openapi.actionSystem.ActionGroup.getChildren(AnActionEvent) is invoked in io.flutter.run.FlutterDebugProcessActions.isEmpty(ActionGroup) : boolean. This method is marked with @org.jetbrains.annotations.ApiStatus.OverrideOnly annotation, which indicates that the method must be only overridden but not invoked by client code. See documentation of the @ApiStatus.OverrideOnly for more info.`

https://github.com/flutter/flutter-intellij/actions/runs/25895252883/job/76564968851?pr=8905

Could you investigate? Is there a way to accomplish this with supported APIs?

mdddj:
Thanks for catching this. I replaced the manual `getChildren(null)` expansion with a `DefaultActionGroup` check using
  `getChildActionsOrStubs()`, since this group is declared statically in plugin.xml. Pushed the fix in 56f2e136b.

pq:
Awesome. Thanks a million, @mdddj!

pq:
Sorry. Last thing, I promise.

Could you add yourself to `AUTHORS`?

Thanks!

mdddj:
Done, added myself to AUTHORS. Thanks!

pq (approved):
Thanks!

PR #8946, opened 2026-05-04
Upgrade version of Android Studio

helin24:
### The motivations for this change:
- The latest Dart plugin (v505) has been released, so we should start depending on it.
- Android Studio Panda (2025.3) is now the stable release, so we should build against it.

### Some other changes followed from those updates:
- I had to add `lsp4ij` as a dependency since Dart depends on it now. `bytebuddy` was also added as a new dependency (for mockito, I think?)
- `intellij.platform.langInjection"` is now a module replacing `org.intellij.intelliLang` (old plugin). See notes on https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/2026
- Verification ran into an issue that I couldn't fix by running `update_baselines.sh` - more info below.

### More on the verification issue:
- When I generate baselines for 2025.3, it produces a line that keeps parameter names, e.g. `* Deprecated constructor com.jetbrains.lang.dart.ide.template.DartTemplateContextType.<init>(java.lang.String id, java.lang.String presentableName, java.lang.Class baseContextType) is invoked in io.flutter.template.DartToplevelTemplateContextType.<init>()`
- On CI, verify does not preserve parameter names, so it finds `* Deprecated constructor com.jetbrains.lang.dart.ide.template.DartTemplateContextType.<init>(java.lang.String arg0, java.lang.String arg1, java.lang.Class arg2) is invoked in io.flutter.template.DartToplevelTemplateContextType.<init>()`
- I could potentially ask CI to always generate verification reports with parameters, but this actually requires a change to Dart plugin, not Flutter plugin. So while I could do that, it would require releasing Dart again and updating the version dependency here again, which seems like a lot...
- Instead I'm asking `github.sh` to instead match the class name and ignore the params. Hopefully that works.


gradle/libs.versions.toml
```diff
@@ -27,6 +29,7 @@ kodein-di = { group = "org.kodein.di", name = "kodein-di-jvm", version.ref = "ko
 kotlinx-coroutines = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-core", version.ref = "kotlinx-coroutines" }
 junit-jupiter = { group = "org.junit.jupiter", name = "junit-jupiter", version.ref = "junit-jupiter" }
 junit-platform-launcher = { group = "org.junit.platform", name = "junit-platform-launcher", version.ref = "junit-platform" }
+bytebuddy = { group = "net.bytebuddy", name = "byte-buddy", version.ref = "bytebuddy" }
```
pq, line 32: Curious about this... Another LSP dep?

Odd too that Dart deps flow down to the Flutter plugin. Seems a little surprising that you'd need to enumerate transitive dependencies.

**EDIT** gemini seems to think if Dart is in our `pluginDependencies`, gradle will handle transitive dependencies for us. But maybe that's not right?

tool/github.sh
```diff
@@ -109,10 +109,12 @@ elif [ "VERIFY_BOT" = "$BOT" ] ; then
 
     if [ -f "$REPORT" ]; then
       echo "Comparing baseline against report in $REPORT"
-      grep "^*" "$REPORT" | sort > current_issues.tmp
+      # Normalize the specific constructor signature to ignore parameter names.
+      # This bridges the gap between local builds (with names) and CI builds (without names).
+      grep "^*" "$REPORT" | sed -E 's/DartTemplateContextType\.<init>\([^)]+\)/DartTemplateContextType.<init>(args)/' | sort > current_issues.tmp
```
pq, line 114: Ick. (But you know that.)

I'm confused by this skew.  If we're running `./gradlew verifyPlugin -PsingleIdeVersion=$version` locally and on the CI it seems like we should be verifying against the same binaries so I'm puzzled about this incompatibility. I'm pretty sure we want to keep them in sync too since that seems like a place that could become a headache debugging down the road.

If you don't want to address this now, could you open an issue to track the concern?

gradle.properties
```diff
@@ -4,8 +4,8 @@
 # found in the LICENSE file.
 #
 
-ideaVersion=2025.2.3.9
-dartPluginVersion= 504.0.0
+ideaVersion=2025.3.4.6
```
pq: Dart is on `2026.1.1`. Is there any reason not to jump ahead to that? (Or any good reason that they don't agree?)

gradle/libs.versions.toml
helin24: I think the LSP dependency is needed for testing, not for building. Perhaps the transitive part is handled for building only? I don't really understand this.

tool/github.sh
helin24: Gemini suggested that one way to fix this is to request is to add this to the Dart plugin (NOT this plugin) build file:

```
tasks.withType<JavaCompile> {
    options.compilerArgs.add("-parameters")
}
```

And that should make parameters always keep their initial names. It seems annoying, but if this works, I can start making that change in the Dart plugin and then remove this hack soon. I mainly just don't want to have to do another plugin release to fix this.

gradle.properties
helin24: Yeah, it's confusing... Most flutter plugin users are on Android Studio, and 2025.3 is stable for Android Studio, while 2026.1 is stable for IntelliJ IDEA. We could consider building against Android Studio for Dart too, but I don't think that makes as much sense. Or we could switch to IntelliJ IDEA for both, but then it seems more likely we'd miss bugs that affect large numbers of users.

tool/github.sh
pq: If you want to open a tracking issue that has some of this context I'm fine punting for now. Thanks!

gradle.properties
pq: Got it. Let's stick w/ 2025 here then. (And perhaps consider documenting somewhere?)

Thanks!

gradle/libs.versions.toml
helin24: More info on this:
- Compiling code during `./gradlew build` only requires code be available that the Flutter code uses directly, so it doesn't require lsp4ij.
- Testing spins up a minimal instance of the IDE and tries to load the plugins to run the tests, so it needs all of the dependencies. Otherwise the Dart plugin can't be loaded, and so tests that depend on the Dart plugin fail.

tool/github.sh
helin24: https://github.com/flutter/dart-intellij-third-party/issues/373

gradle/libs.versions.toml
pq: Is there any way to add this dependency resolution only to the test run flow?

settings.gradle.kts
```diff
@@ -4,9 +4,12 @@
...
-    gradlePluginPortal()
-  }
+//pluginManagement {
+//  repositories {
+//    maven("https://oss.sonatype.org/content/repositories/snapshots/")
+//    gradlePluginPortal()
+//  }
+//}
+plugins {
+  id("org.gradle.toolchains.foojay-resolver-convention") version "1.0.0"
```
pq, line 7: I think this may be causing the Kokoro build failures internally.

gradle/libs.versions.toml
helin24: Well, it turns out we actually need it for `runIde` too, since that's also creating the IDE environment. I think technically we can also call `runIde` and then install it manually, then run again (and the workbench IDE will remember the plugin is installed), but that's also not great.

settings.gradle.kts
helin24: I actually added this to resolve the kokoro build failures. We were getting the same error as we had on the Dart plugin, where there is a "toolchain dependency" requiring a specific version of Java, which isn't available in CI unless we also use this plugin.

pq: Aha. Interesting. Thanks!

build.gradle.kts
```diff
@@ -225,17 +225,16 @@ dependencies {
...
-        "org.intellij.intelliLang",
         "org.jetbrains.android",
         "com.android.tools.idea.smali"
       )
     )
     plugin("Dart:$dartPluginVersion")
+    plugin("com.redhat.devtools.lsp4ij:${libs.versions.lsp4ij.get()}")
+    bundledModule("intellij.platform.langInjection")
 
-    if (sinceBuildInput == "243" || sinceBuildInput == "251") {
```
pq, line 235: 🧹  Thanks for cleaning up!

pq (approved):
Ok. I'll stop resisting 🙃 

Thanks for all the follow-up!

helin24:
Haha we can also discuss later, along with why our build takes an hour 😢 

It's not a very pleasant set of changes.

pq:
Oof. An hour? Let's talk.

PR #8995, opened 2026-06-04
Update name of file

helin24:
Sorry, I neglected to actually run the dev build to make sure it worked.

### DIFF
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="21" />
+    <bytecodeTargetLevel target="17" />
   </component>
 </project>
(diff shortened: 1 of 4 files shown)


tool/plugin/lib/plugin.dart
```diff
@@ -393,7 +393,7 @@ abstract class ProductCommand extends Command<int> {
       'releases',
       subDir,
       spec.version,
-      'flutter-intellij.zip',
+      'Flutter.zip',
```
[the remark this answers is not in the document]
helin24: I believe this is no longer used, and we should actually remove this code in a following PR.

pq (approved):

.idea/compiler.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="CompilerConfiguration">
-    <bytecodeTargetLevel target="17" />
+    <bytecodeTargetLevel target="21" />
```
pq: I wonder if you want to commit these changes here?

helin24: Definitely not! I missed them while I was scanning the diff.

PR #9005, opened 2026-06-17
Add analytics for Flutter Run and Debug sessions

helin24:
See https://github.com/flutter/dart-intellij-third-party/pull/478 for context

### DIFF
--- a/src/io/flutter/analytics/Analytics.kt
+++ b/src/io/flutter/analytics/Analytics.kt
@@ -6,6 +6,11 @@
 
 package io.flutter.analytics
 
+import com.intellij.execution.Executor
+import com.intellij.execution.executors.DefaultDebugExecutor
+import com.intellij.openapi.project.Project
+import com.jetbrains.lang.dart.analytics.Analytics as DartAnalytics
+import com.jetbrains.lang.dart.analytics.AnalyticsData
 import com.jetbrains.lang.dart.analytics.BooleanValue
 
 /**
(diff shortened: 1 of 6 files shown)


src/io/flutter/run/test/TestConfig.java
```diff
@@ -89,6 +90,7 @@ public FlutterSdk getSdk() {
   @NotNull
   @Override
   public CommandLineState getState(@NotNull Executor executor, @NotNull ExecutionEnvironment env) throws ExecutionException {
+    Analytics.recordRunOrDebugSession("flutter_tests", executor, getProject());
```
pq: Could you extract this and the following into constants? (See for example what we did in `DartProblemsViewPanel`.)

pq (approved):

src/io/flutter/run/test/TestConfig.java
pq: Ah, and above too!

src/io/flutter/analytics/Analytics.kt
```diff
@@ -39,3 +44,27 @@ object AnalyticsConstants {
...
+      DartAnalytics.report(data)
+    } catch (t: Throwable) {
+      // Fail-safe: analytics should never crash the IDE or run/debug sessions.
+    }
+  }
+
+  @JvmStatic
+  fun recordRunOrDebugSession(mechanism: String, executor: Executor, project: Project?) {
+    val type = if (executor.id == DefaultDebugExecutor.EXECUTOR_ID) {
+      "debug_session"
```
pq: Can you use `AnalyticsConstants` as you did in the other PR?

PR #9022, opened 2026-06-29
Update versions to include 262, add baseline for 261

### LINKED ISSUE #9019: Update versioning to include 2026.2

helin24:
Fixes https://github.com/flutter/flutter-intellij/issues/9019

IntelliJ 2026.2 is now available, so we want our full versioning to include that. However, there is no Android Studio 2026.2 yet, so I can only expand verification versioning to 2026.1. We could always add to our verify checks to do an IntelliJ IDEA 2026.2 check, but I don't think this is a priority at the moment.

Similar to the Dart change, I built the plugin with these settings and verified I can install to IntelliJ IDEA 2026.2. I haven't checked the other versions yet.


pq (approved):
Sweet!

PR #9038, opened 2026-07-15
Remove Bazel settings and analytics telemetry

helin24:
To manually test this:
- Check flutter settings panel - bazel-related checkbox should be removed, but other things should still be changeable. SDK should be changeable (though there are some bugs here that are unrelated to this change, mostly from clicking around quickly during SDK changes)
- Hot reload and hot restart should still work
- Pub commands should still work, e.g. try running flutter doctor

### DIFF
--- a/src/io/flutter/actions/FlutterSdkAction.java
+++ b/src/io/flutter/actions/FlutterSdkAction.java
@@ -43,7 +43,6 @@ public abstract class FlutterSdkAction extends DumbAwareAction {
       if (workspace != null) {
         FileDocumentManager.getInstance().saveAllDocuments();
         startCommandInBazelContext(project, workspace, event);
-        analyticsData.add(AnalyticsConstants.IN_BAZEL_CONTEXT, true);
         Analytics.report(analyticsData);
         return;
       }
--- a/src/io/flutter/sdk/FlutterSettingsConfigurable.java
+++ b/src/io/flutter/sdk/FlutterSettingsConfigurable.java
@@ -38,8 +38,6 @@ import icons.FlutterIcons;
 import io.flutter.FlutterBundle;
 import io.flutter.FlutterConstants;
 import io.flutter.FlutterMessages;
-import io.flutter.bazel.Workspace;
-import io.flutter.bazel.WorkspaceCache;
 import io.flutter.font.FontPreviewProcessor;
 import io.flutter.pub.PubRoot;
 import io.flutter.pub.PubRoots;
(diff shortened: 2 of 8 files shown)


src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -255,36 +241,31 @@ public boolean isModified() {
...
+        FlutterSdkUtil.enableDartSdk(myProject);
+
+        ApplicationManager.getApplication().executeOnPooledThread(() -> {
+          final FlutterSdk sdk = FlutterSdk.forPath(sdkHomePath);
+          if (sdk != null) {
+            try {
+              lock.acquire();
+              sdk.queryFlutterChannel(false);
+              lock.release();
+            }
```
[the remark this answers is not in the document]
helin24: This is part of https://github.com/flutter/flutter-intellij/issues/8641, but I don't want to address here.

src/io/flutter/actions/FlutterSdkAction.java
```diff
@@ -43,7 +43,6 @@ public void actionPerformed(@NotNull AnActionEvent event) {
       if (workspace != null) {
         FileDocumentManager.getInstance().saveAllDocuments();
         startCommandInBazelContext(project, workspace, event);
```
pq, line 45: Should this be renamed? Or will it ever be called? (Actually: *is* it currently being called?)

src/io/flutter/sdk/FlutterSettingsConfigurable.java
```diff
@@ -255,36 +241,31 @@ public boolean isModified() {
 
   @Override
   public void apply() throws ConfigurationException {
-    // Bazel workspaces do not specify a sdk path so we do not need to update the sdk path if using
```
pq, line 258: Removing this feels a little different than removing the run configs. This *could* be google3-impacting?

pq: As per our conversation, I think we're making ourselves with chancing it. 🤞

pq (approved):
Open question about a rename but no bigge.

src/io/flutter/actions/FlutterSdkAction.java
helin24: This command will be removed in the next PR (related to workspace management), but it's not being touched in this one since I'm limiting this one to settings and analytics.

PR #9060, opened 2026-07-29
Remove Weberknecht from the project

JavadAsadi:
## Summary

This pull request copies the Jdk's WebSocket into `third_party/vmServiceDrivers/org/dartlang/vm/service/internal`.
And replaces it with the Weberknecht.

Note about FlutterApp.java:
Since the new version of dart plugin is no longer throwing weberknecht.WebSocketException, we need to migrate the exception handling to the new Exception as well. But because we are only logging the exception here generalizing is a good idea to keep the compatibiity with both old and new dart plugin.

### DIFF
--- a/src/io/flutter/run/daemon/FlutterApp.java
+++ b/src/io/flutter/run/daemon/FlutterApp.java
@@ -28,7 +28,6 @@ import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.util.EventDispatcher;
 import com.intellij.util.concurrency.AppExecutorUtil;
 import com.jetbrains.lang.dart.ide.toolingDaemon.DartToolingDaemonService;
-import de.roderick.weberknecht.WebSocketException;
 import io.flutter.FlutterUtils;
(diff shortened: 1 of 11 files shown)


src/io/flutter/run/daemon/FlutterApp.java
```diff
@@ -841,7 +840,10 @@ private void sendDtdRequest(@NotNull DtdRequest dtdRequest, @NotNull JsonObject
         }
       });
     }
-    catch (InterruptedException | java.util.concurrent.ExecutionException | WebSocketException e) {
+    // The Dart plugin changed its checked WebSocket exception type when it replaced
+    // Weberknecht. Catching the common base type keeps this call compatible with both APIs.
+    // We can potnetitaly catch WebSocketException in the future from Dart plugin, but since it's only used for logging, the general exception is ok
```
helin24, line 845: potentially (typo)

helin24 (approved):
This LGTM - fix typo and we'll merge if tests pass

